PR #6, opened 2020-04-06
Batch writes form multiple WriteReqs

JLockerman:
Prometheus batches writes on a per-series basis, while we store data on
a per-metric basis. This means that a single write batch from
Prometheus may contain only a few writes to a large number of database
tables. To improve performance, this commit batches writes at the IO
level, so samples from multiple `WriteReq`s can be stored using a single
database write.

We do this by maintaining a goroutine per-metric, which is responsible
for getting input from one or more WriteReqs, and forwarding them to
the database in `CopyFrom`s. Each goroutine listens on a channel, and
will attempt to pull more data from that channel until the channel runs
dry, or a threshold is reached. The goroutine will then issue a
CopyFrom, and, once it completes, go back to reading from the channel.

The goroutines responsible for handling `WriteReq`s wait until all of
their writes have been completed before returning, ensuring that we
have the same synchronous API as before.

On certain workloads, this leads to a significant speedup, with an 8x
improvement having been seen.


pkg/pgmodel/ingestor.go
```diff
@@ -107,7 +108,7 @@ func (i *DBIngestor) Ingest(tts []prompb.TimeSeries) (uint64, error) {
 
 	rowsInserted, err := i.db.InsertNewData(newSeries, data)
 	if err == nil && int(rowsInserted) != totalRows {
-		return rowsInserted, fmt.Errorf("Failed to insert all the data! Expected: %d, Got: %d", totalRows, rowsInserted)
+		// return rowsInserted, fmt.Errorf("Failed to insert all the data! Expected: %d, Got: %d", totalRows, rowsInserted)
```
cevian, line 111: why is this commented out? Can't we support this check now?

pkg/pgmodel/pgx.go
```diff
@@ -323,6 +353,78 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
 	return tableName, err
 }
 
+func (p *pgxInserter) getMetricTableInserter(metric string) chan insertDataRequest {
+	var inserterI interface{}
+	var ok bool
+	var inserter chan insertDataRequest
+	if inserterI, ok = p.metricTableInserters.Load(metric); !ok {
+		ch := make(chan insertDataRequest, 1000)
+		ii, loaded := p.metricTableInserters.LoadOrStore(metric, ch)
```
cevian, line 362: What happens if the cache evicts an item? Is the value channel closed somehow? Does BigCache have facility for this?

pkg/pgmodel/ingestor.go
JLockerman: nope! since we batch rows from different requests together, when we get a number of rows-inserted back, we have no way to no which request they belong to.

pkg/pgmodel/pgx.go
JLockerman: This is not a cache, it's a map, my expectation is that we'll ≈never have enough dead metrics that it's worthwhile to garbage collect them.

pkg/pgmodel/pgx.go
```diff
@@ -323,6 +355,78 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
 	return tableName, err
 }
 
+func (p *pgxInserter) getMetricTableInserter(metric string) chan insertDataRequest {
+	var inserterI interface{}
+	var ok bool
+	var inserter chan insertDataRequest
+	if inserterI, ok = p.metricTableInserters.Load(metric); !ok {
+		ch := make(chan insertDataRequest, 1000)
```
blagojts, line 363: i feel like this 1000 should be configurable

PR #7, opened 2020-04-06
REFACTOR: only pass SampleInfoIterator when needed

JLockerman:
We don't need to create a SampleInfoIterator until we're actually
passing data to pgx, until that point a regular slice will do.

### DIFF
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -19,6 +19,7 @@ import (
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/prompb"
 )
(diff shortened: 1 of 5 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -138,6 +139,54 @@ func (p *pgxConnImpl) SendBatch(ctx context.Context, b pgxBatch) (pgx.BatchResul
...
+// SampleInfoIterator is an iterator over a collection of sampleInfos that returns
+// data in the format expected for the data table row.
+type SampleInfoIterator struct {
+	sampleInfos     []*samplesInfo
+	sampleInfoIndex int
+	sampleIndex     int
+}
+
+// NewSampleInfoIterator is the constructor
+func NewSampleInfoIterator() *SampleInfoIterator {
```
cevian, line 151: Never Used. Should be called by  line 431

pkg/pgmodel/pgx.go
```diff
@@ -391,14 +440,14 @@ func (p *pgxInserter) metricTableInsert(metric string, input chan insertDataRequ
 			return
 		}
 		needsResponse = append(needsResponse, insertDataTask{finished: req.finished, errChan: req.errChan})
-		batch.sampleInfos = append(batch.sampleInfos, req.data.sampleInfos...)
+		batch.sampleInfos = append(batch.sampleInfos, req.data...)
```
cevian, line 443: This should call batch.AppendSlice (new function) or better yet batch.Append should be changed to work with this.

pkg/pgmodel/pgx.go
```diff
@@ -138,6 +139,54 @@ func (p *pgxConnImpl) SendBatch(ctx context.Context, b pgxBatch) (pgx.BatchResul
 	return conn.SendBatch(ctx, b.(*pgx.Batch)), nil
 }
 
+// SampleInfoIterator is an iterator over a collection of sampleInfos that returns
+// data in the format expected for the data table row.
+type SampleInfoIterator struct {
```
cevian, line 144: I think this should also have a Reset function()

cevian:
Included in PR #11

PR #9, opened 2020-04-07
Add github action for tests

cevian:
Add go test checks and also make sure that go generate
was run (doesn't create a diff).


.github/workflows/go.yml
```diff
@@ -31,4 +31,12 @@ jobs:
...
\ No newline at end of file
+      run: go build -v ./...
+
+    - name: Test
+      run: go test -race ./...
+
+    - name: Generated
+      run: |
+        go generate ./...
+        git diff --exit-code
```
antekresic, line 42:
```suggestion
        diff <(git diff -G "modTime:") <(git diff)
```

PR #11, opened 2020-04-08
Improve performance by reducing goroutines

JLockerman:
Prof numbers indicate we're spending an inordinate amount of time in
selectgo trying to determine which channels to access; mostly in a
single select in CopyFrom. This commit improves performance by reducing
the number of goroutines in play at once, and the number of times that
select can be accessed.

Instead of having a single goroutine per-metric, we now have a fixed
number of them, equal to the GOMAXPROCS, and we hash metric table names
to determine which goroutine is responsible for each insert. The
inserters maintain a set of pending inserts, and they flush the values
to the database if:
  a. the amount of pending writes for that metric becomes large enough
  b. a sufficient amount of time has elapsed since we received the
     first sample in a metric's batch
  c. if the inserter runs out of things to do, it flushes the first
     batch it received

Overall, this commit improves one benchmark I've been using from around
90K writes/s to around 140K writes/s.

### DIFF
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"regexp"
+	"runtime"
 	"time"
 
 	"github.com/allegro/bigcache"
@@ -52,7 +53,14 @@ type Client struct {
 func NewClient(cfg *Config) (*Client, error) {
 	connectionStr := cfg.GetConnectionStr()
 
-	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr)
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxProcs, maxProcs))
 
 	log.Info("msg", regexp.MustCompile("password='(.+?)'").ReplaceAllLiteralString(connectionStr, "password='****'"))
 
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -19,6 +19,7 @@ import (
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/prompb"
 )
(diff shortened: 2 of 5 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -404,76 +435,208 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
...
+	h.handleReq(req)
+
+	return true
 }
 
-func (p *pgxInserter) metricTableInsert(metric string, input chan insertDataRequest) {
-	tableName, err := p.getMetricTableName(metric)
-	if err != nil {
+func (h *InsertHandler) handleMoreReqs() bool {
+	select {
```
cevian, line 507: I wonder if this should actually be a for {select{}} with one of the cases being a timer that checks for the livelock timeout

pkg/pgclient/client.go
```diff
@@ -52,7 +53,14 @@ type Client struct {
...
 
-	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr)
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxProcs, maxProcs))
```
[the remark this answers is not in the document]
JLockerman: at some point. I hesitate to do so until we're sure this is the strategy we want to take

pkg/pgmodel/pgx.go
```diff
@@ -218,17 +221,40 @@ func NewPgxIngestor(c *pgxpool.Pool) *DBIngestor {
...
 
+func newPgxInserter(conn pgxConn, cache MetricCache) *pgxInserter {
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+	inserters := make([]chan insertDataRequest, maxProcs)
```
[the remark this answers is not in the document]
JLockerman: same, don't think it's worth for this PR

pkg/pgmodel/pgx.go
JLockerman: select performance is wonky, and this will ~never be taken; I think we should optimize for performance here

pkg/pgmodel/pgx.go
```diff
@@ -404,76 +435,230 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
...
+	needsFlush, pending := h.pending.addReq(req)
+	if needsFlush {
+		h.flushPending(pending)
+		return true
+	}
+	return false
+}
 
-	var needsResponse []insertDataTask
+func (p *PendingBuffer) addReq(req insertDataRequest) bool {
```
cevian, line 543: nit: super confusing to have this sandwitched between InsertHandler functions

pkg/pgmodel/pgx.go
```diff
@@ -404,76 +435,230 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
...
+		pending.needsResponse[i] = insertDataTask{}
+	}
+	pending.needsResponse = pending.needsResponse[:0]
+
+	for i := 0; i < len(pending.batch.sampleInfos); i++ {
+		pending.batch.sampleInfos[i] = nil
+	}
+	pending.batch = SampleInfoIterator{sampleInfos: pending.batch.sampleInfos[:0], sampleIndex: -1, sampleInfoIndex: 0}
+
+	h.pending.giveOldBuffer(pending)
```
cevian, line 602: why don't we just call h.pending.Remove here and have that call giveOldBuffer

pkg/pgmodel/pgx.go
```diff
@@ -404,76 +435,230 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
...
+		earliest := h.pending.Front()
+		if earliest == nil {
 			return
 		}
-		needsResponse = append(needsResponse, insertDataTask{finished: req.finished, errChan: req.errChan})
-		batch.sampleInfos = append(batch.sampleInfos, req.data...)
 
-	TryReadMore:
-		for len(batch.sampleInfos) < batchSize {
+		earliestPending := earliest.Value.(*PendingBuffer)
```
cevian, line 556: nit can we avoid these casts in non-orderedMap code. can Front() return both a *PendingBuffer and a List.Element (which I would call a ref). So the code would be `earliest, earliestRef := h.pending.Front()`

pkg/pgmodel/pgx.go
```diff
@@ -404,76 +435,230 @@ func (p *pgxInserter) getMetricTableName(metric string) (string, error) {
...
-	return inserter
+func (h *InsertHandler) hasPendingReqs() bool {
+	return !h.pending.IsEmpty()
 }
 
-func (p *pgxInserter) metricTableInsert(metric string, input chan insertDataRequest) {
-	tableName, err := p.getMetricTableName(metric)
-	if err != nil {
-		return
+func (h *InsertHandler) waitForReq() bool {
```
cevian, line 513: nit maybe call `getReqBlocking` and below `getReqNonBlocking`

pkg/pgmodel/pgx.go
JLockerman: I don't want anything in `pending` that's being inserted.

PR #14, opened 2020-04-09
Test handling of staleness marker

cevian:
Prometheus injects a staleness marker into the data for a series if
it has not seen the series in a certain amount of time (default 5 min).
The staleness marker is represented as a float8 NaN with some additional
bit twiddling.

This PR tests that a staleness marker is correctly propagated to and from
the database with the bit twiddling intact. It also adds sql functions
to differentiate a normal NaN from a stale marker.

The alternative is to represent the stale marker as NULL inside the DB,
but the semantics there aren't clear either and it's further removed
from the prometheus implementation.


pkg/pgmodel/nan_test.go
```diff
@@ -0,0 +1,175 @@
...
+			{
+				isNaN: true,
+				rrq: prompb.ReadRequest{
+					Queries: []*prompb.Query{
+						{
+							Matchers: []*prompb.LabelMatcher{
+								{
+									Type:  prompb.LabelMatcher_EQ,
+									Name:  metricNameLabelName,
+									Value: metricName,
```
sgichohi, line 108: nit: consider making this a variable since it doesn't change between tests.

pkg/pgmodel/nan_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+			if err != nil {
+				t.Fatalf("unexpected error while ingesting test dataset: %s", err)
+			}
+			answer := getSingleSampleValue(t, resp)
+			isStaleNaN := getBooleanSQLResult(t, db,
+				fmt.Sprintf(
+					`SELECT prom.is_stale_marker(value)
+					 FROM prom."StaleMetric"
+					 WHERE %s
```
sgichohi, line 135: nit: consider using variable interpolation if it works, for sql best practice ` time >= 'epoch'::timestamptz +interval '$1 ms' AND "+
 				"time <= 'epoch'::timestamptz +interval $2 ms' `

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -744,4 +744,18 @@ BEGIN
...
     END LOOP;
 END;
-$$;
\ No newline at end of file
+$$;
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.is_stale_marker(value double precision)
+RETURNS BOOLEAN
+AS $func$
+    SELECT float8send(value) = '\x7ff0000000000002'
```
sgichohi, line 752: this is cool

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -744,4 +744,18 @@ BEGIN
         COMMIT;
     END LOOP;
 END;
-$$;
\ No newline at end of file
+$$;
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.is_stale_marker(value double precision)
```
sgichohi, line 749: This would be great to document for users as well, and is great UX. The bit twiddling could be non trivial for some users.

cevian: yeah will document everything in schema_prom

PR #20, opened 2020-04-10
Fix lint issues

cevian:


pkg/pgclient/client.go
```diff
@@ -79,14 +79,18 @@ func NewClient(cfg *Config) (*Client, error) {
...
 	return fmt.Sprintf("host=%v port=%v user=%v dbname=%v password='%v' sslmode=%v connect_timeout=10",
 		cfg.host, cfg.port, cfg.user, cfg.database, cfg.password, cfg.sslMode)
 }
+
+// Close closes the client and performs cleanup
 func (c *Client) Close() {
 	c.ingestor.Close()
 }
 
+// Ingest write the timeseries object into the DB
```
blagojts, line 93:
```suggestion
// Ingest writes the timeseries object into the DB
```

pkg/util/election.go
```diff
@@ -86,10 +93,12 @@ func (se *ScheduledElector) resumeScheduledElection() {
 	se.pausedScheduledElection = false
 }
 
-func (se *ScheduledElector) IsPausedScheduledElection() bool {
+func (se *ScheduledElector) isPausedScheduledElection() bool {
```
blagojts, line 96: maybe this would sound better if it were isScheduledElectionPaused?

pkg/util/election.go
```diff
@@ -223,12 +235,14 @@ func (r *RestElection) BecomeLeader() (bool, error) {
 	return r.leader, nil
 }
 
+// IsLeader return whether or not you are the leader
```
blagojts, line 238:
```suggestion
// IsLeader returns whether or not you are the leader
```

PR #22, opened 2020-04-13
Switch to Array based interchange instead of JSONB

JLockerman:
This commit switches the interchange format we use for labels to be two
`TEXT[]`s instead of a `JSONB`. This should be a bit more efficient, as we
no longer have to marshal the labels to JSON on the database side, and,
more significantly, have to parse JSON at the adapter.


JLockerman:
The actual change to the result-parsing logic can be found [here in the second commit](https://github.com/timescale/timescale-prometheus/pull/22/commits/17214b91b951b20df388d0674b0430ff2bf95212#diff-047c81ab212d92a340724a7fa6617c1e).

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -429,10 +429,10 @@ LANGUAGE SQL VOLATILE PARALLEL SAFE;
 
 --Returns the jsonb for a series defined by a label_array
 --Note that this function should be optimized for performance
-CREATE OR REPLACE FUNCTION SCHEMA_PROM.label_array_to_jsonb(labels int[])
-RETURNS jsonb AS $$
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.label_array_to_kv(labels int[], OUT keys text[], OUT vals text[])
```
cevian, line 432: This shouldn't replace the function but be in addition to. People querying from SQL may very well want to get jsonb out. Also note the inverse is called `key_value_array_to_label_array` we should be. consistent either way (don't feel strongly which)

pkg/pgmodel/migrate_test.go
```diff
@@ -598,19 +600,24 @@ func TestSQLJsonLabelArray(t *testing.T) {
 						t.Fatalf("Expected label arrays to be equal: %v != %v", labelArray, labelArrayKV)
 					}
 
-					var jsonres []byte
-					err = db.QueryRow(context.Background(), "SELECT * FROM prom.label_array_to_jsonb($1)", labelArray).Scan(&jsonres)
+					var (
```
cevian, line 603: we should have both `prom.label_array_to_jsonb` and `prom.label_array_to_kv` so we should have tests for both.

JLockerman: It's a simple enough wrapper around the array version that I'm not sure it's worth it; it'd require us to re-add the label-set logic for one...

cevian: it'd require us to re-add the label-set logic for one. isn't that just re-adding lines 630-642 back in?

cevian:
@antekresic please also review. Also for context this really mattered on the insert side for performance so switching will both help performance and make the insert and select paths more similar

PR #26, opened 2020-04-15
Add go report card and fix some suggestions from it

blagojts:
There were several functions with cyclomatic complexities > 15 but other than one, all of them were in the tests, so I skipped fixing those.


blagojts:
@cevian added action to request report update

PR #30, opened 2020-04-15
SQL Function Audit

davidkohn88:
Modify SQL/PLPGSQL functions to have same naming style, proper volatility
and parallel safety markings, and follow formatting styles. Also updates
a few comments for clarity.


cevian:
The build is failing because you have to run `go generate ./...` and commit the change to the binary string

PR #32, opened 2020-04-16
Update the flag names and default value of the dbname flag

blagojts:
The flag names are changed to signify that the connector is no longer 
compatible with Postgres, just TimescaleDB.

Also the deployment in the helm-chart is simplified to use
the TS_PROM environment variables to set the connection
flags.

*Warning* Upon merging, the helm chart will require a new docker build to be uploaded to the docker hub because it assumes the new flag names

### DIFF
--- a/helm-chart/templates/deployment-connector.yaml
+++ b/helm-chart/templates/deployment-connector.yaml
@@ -29,25 +29,21 @@ spec:
             - containerPort: 9201
               name: connector-port
           env:
-            - name: PGPORT
+            - name: TS_PROM_DB_PORT
               value: {{ .Values.connection.port | quote }}
-            - name: PGUSER
+            - name: TS_PROM_DB_USER
               value: {{ .Values.connection.user }}
-            - name: PGPASSWORD
+            - name: TS_PROM_DB_PASSWORD
               valueFrom:
                 secretKeyRef:
                   name: {{ tpl .Values.connection.password.secretTemplate . }}
                   key: {{ .Values.connection.user }}
-            - name: PGHOST
-              value: {{ tpl .Values.connection.host.nameTemplate . }} 
-          args:
-            [
-              -pg-ssl-mode=require,
-              -pg-host=$(PGHOST),
-              -pg-user=$(PGUSER),
-              -pg-password=$(PGPASSWORD),
-              -pg-port=$(PGPORT)
-            ]
+            - name: TS_PROM_DB_HOST
+              value: {{ tpl .Values.connection.host.nameTemplate . }}
+            - name: TS_PROM_DB_NAME
+              value: {{ .Values.connection.dbName }}
+            - name: TS_PROM_DB_SSL_MODE
+              value: require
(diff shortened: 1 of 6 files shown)


pkg/pgclient/client.go
```diff
@@ -30,17 +30,17 @@ type Config struct {
...
 // ParseFlags parses the configuration flags specific to PostgreSQL and TimescaleDB
 func ParseFlags(cfg *Config) *Config {
-	flag.StringVar(&cfg.host, "pg-host", "localhost", "The PostgreSQL host")
-	flag.IntVar(&cfg.port, "pg-port", 5432, "The PostgreSQL port")
-	flag.StringVar(&cfg.user, "pg-user", "postgres", "The PostgreSQL user")
-	flag.StringVar(&cfg.password, "pg-password", "", "The PostgreSQL password")
-	flag.StringVar(&cfg.database, "pg-database", "postgres", "The PostgreSQL database")
-	flag.StringVar(&cfg.sslMode, "pg-ssl-mode", "disable", "The PostgreSQL connection ssl mode")
-	flag.IntVar(&cfg.dbConnectRetries, "pg-db-connect-retries", 0, "How many times to retry connecting to the database")
+	flag.StringVar(&cfg.host, "db-host", "localhost", "The TimescaleDB host")
```
blagojts, line 33: i'm not sure about this change, I wanted to signify that the connector is no longer compatible with PostgreSQL, but it is still a Postgres cluster with users, databases etc created in Postgres

helm-chart/templates/deployment-connector.yaml
```diff
@@ -29,25 +29,21 @@ spec:
             - containerPort: 9201
               name: connector-port
           env:
-            - name: PGPORT
+            - name: TS_PROM_DB_PORT
```
cevian, line 32: how are these env variables read by the connector if they aren't in the args?

pkg/pgclient/client.go
cevian: I like this change

helm-chart/templates/deployment-connector.yaml
blagojts: we use a package that is set up to search ENV for vars starting with TS_PROM

helm-chart/templates/deployment-connector.yaml
```diff
@@ -29,25 +29,21 @@ spec:
...
-              -pg-ssl-mode=require,
-              -pg-host=$(PGHOST),
-              -pg-user=$(PGUSER),
-              -pg-password=$(PGPASSWORD),
-              -pg-port=$(PGPORT)
-            ]
+            - name: TS_PROM_DB_HOST
+              value: {{ tpl .Values.connection.host.nameTemplate . }}
+            - name: TS_PROM_DB_NAME
+              value: {{ .Values.connection.dbName }}
```
cevian, line 44: can we add the dbname to the cronjob too?

PR #35, opened 2020-04-17
Sql changes

cevian:


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -42,17 +44,17 @@ CREATE TABLE SCHEMA_CATALOG.label_key(
 );
 
 CREATE TABLE SCHEMA_CATALOG.label_key_position (
-    metric text,
+    metric_name text, --references metric.metric_name NOT metric.id for performance reasons
     key TEXT, --NOT label_key.id for performance reasons.
     pos int,
```
JLockerman, line 49: I'm a bit concerned with running out of label positions, especially since I don't know a good way to GC them. Should we switch these to a `BIGINT`? It's won't really help now, but we can switch to a custom array-type than can handle them...

cevian: I think if we have an array that big we are in real trouble. fwiw we see sizes of mostly < 20 now. I'd rather push this off

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -8,14 +8,16 @@ CREATE SCHEMA IF NOT EXISTS SCHEMA_METRIC;
...
 -----------------------
 
 CREATE TABLE SCHEMA_CATALOG.series (
     id bigserial PRIMARY KEY,
     metric_id int,
-    labels int[],
+    labels SCHEMA_PROM.label_ids,
     UNIQUE(labels) INCLUDE (id)
 );
 CREATE INDEX series_labels_id ON SCHEMA_CATALOG.series USING GIN (labels);
```
davidkohn88, line 23: Does this need to be on `labels::int[]`?

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
davidkohn88: Agreed, if we have arrays the size of int we're screwed.

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
cevian: I don't think so, in any case it seems to work and doesn't complain.

PR #37, opened 2020-04-19
Place the data tables in prom_data

cevian:
Previously, data tables and public api were both in the prom
schema. But it makes sense to separate those out because
the public api but not the data tables should be in the search
path. So, move the data tables out into a new schema - prom_data.


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -5,11 +5,23 @@ CREATE SCHEMA IF NOT EXISTS SCHEMA_CATALOG; -- catalog tables + internal functio
...
 CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;
 
 CREATE DOMAIN SCHEMA_PROM.label_ids AS int[] NOT NULL;
+
+DO $$
+DECLARE
+   new_path text;
+BEGIN
+   new_path := current_setting('search_path') || format(',%L,%L', 'SCHEMA_PROM', 'SCHEMA_METRIC');
+   execute format('ALTER DATABASE %I SET search_path = %s', current_database(), new_path);
```
davidkohn88, line 20: I'm not sure how this is going to interact with cloud and whether this will be allowed through Aiven...but we'll deal with that when the time comes.

PR #38, opened 2020-04-20
Add Go benchmarking for certain SQL functions

antekresic:
Some of the SQL functions are suspected to be very performance
intensive. Adding the benchmark allows us to setup a baseline for future
performance testing.

Note: `BenchmarkKeyValueArrayToLabelArrayExistingLabels` fails to run, we need to take a look at it. I might have hit a bug with the benchmark.

### DIFF
--- a/pkg/pgmodel/sql_bench_test.go
+++ b/pkg/pgmodel/sql_bench_test.go
@@ -1,0 +1,266 @@
+package pgmodel
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+)
+
+const (
+	labelCount      = 5
+	metricName      = "benchmark_metric"
+	otherMetricName = "other_benchmark_metric"
+)
+
+func BenchmarkGetSeriesIDForKeyValueArrayExistingSeries(b *testing.B) {
+	b.StopTimer()
+	withDB(b, "bench_1", func(db *pgxpool.Pool, t testing.TB) {
+		err := createMetricTableName(db, metricName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		keys, values := generateKeysAndValues(b.N+labelCount, "label")
+		for n := 0; n < b.N; n++ {
+			err = getSeriesIDForKeyValueArray(db, metricName, keys[n:n+labelCount], values[n:n+labelCount])
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		var bench *testing.B
+		var ok bool
+		if bench, ok = t.(*testing.B); !ok {
+			t.Fatal("Not a benchmarking instance, stopping benchmark")
+		}
+
+		bench.ResetTimer()
+		bench.StartTimer()
+		for n := 0; n < b.N; n++ {
+			err = getSeriesIDForKeyValueArray(db, metricName, keys[n:n+labelCount], values[n:n+labelCount])
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+		bench.StopTimer()
+	})
+}
+
+func BenchmarkGetSeriesIDForKeyValueArrayNewSeriesExistingLabels(b *testing.B) {
+	b.StopTimer()
+	withDB(b, "bench_2", func(db *pgxpool.Pool, t testing.TB) {
+		err := createMetricTableName(db, metricName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		keys, values := generateKeysAndValues(b.N+labelCount, "label")
+		for n := 0; n < b.N; n++ {
+			err = getSeriesIDForKeyValueArray(db, otherMetricName, keys[n:n+labelCount], values[n:n+labelCount])
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		var bench *testing.B
+		var ok bool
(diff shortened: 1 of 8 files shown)


cevian:
can we include the benchmark results in a comment on this pr?

antekresic:
Benchmarks updated with @cevian suggestions. Here are the results from my machine:
```goos: linux
goarch: amd64
pkg: github.com/timescale/timescale-prometheus/pkg/pgmodel
BenchmarkGetSeriesIDForKeyValueArrayExistingSeries-12                       1374            770349 ns/op
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesExistingLabels-12               360           3878838 ns/op
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesNewLabels-12                    236           4535138 ns/op
BenchmarkKeyValueArrayToLabelArrayCreateNewLabels-12                         319           4002234 ns/op
BenchmarkKeyValueArrayToLabelArrayExistingLabels-12                         2163            570341 ns/op
BenchmarkKeyValueArrayToLabelArrayCreateNewLabelKeys-12                      346           3894518 ns/op
BenchmarkGetOrCreateMetricTableName-12                                      4569            237746 ns/op
```

pkg/pgmodel/sql_bench_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+func BenchmarkGetSeriesIDForKeyValueArrayExistingSeries(b *testing.B) {
+	b.StopTimer()
+	withDB(b, "bench_1", func(db *pgxpool.Pool, t testing.TB) {
+		err := createMetricTableName(db, metricName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		keys, values := generateKeysAndValues(labelCount, "label")
```
cevian, line 25: shouldn't this be b.N+labelCount?

pkg/pgmodel/sql_bench_test.go
```diff
@@ -0,0 +1,262 @@
...
+		var bench *testing.B
+		var ok bool
+		if bench, ok = t.(*testing.B); !ok {
+			t.Fatal("Not a benchmarking instance, stopping benchmark")
+		}
+
+		bench.ResetTimer()
+		bench.StartTimer()
+		for n := 0; n < b.N; n++ {
+			err = getSeriesIDForKeyValueArray(db, metricName, keys, values)
```
cevian, line 40: don't we want this going through N different series?

pkg/pgmodel/sql_bench_test.go
```diff
@@ -0,0 +1,262 @@
...
+func BenchmarkGetSeriesIDForKeyValueArrayNewSeriesExistingLabels(b *testing.B) {
+	b.StopTimer()
+	withDB(b, "bench_2", func(db *pgxpool.Pool, t testing.TB) {
+		err := createMetricTableName(db, metricName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		keys, values := generateKeysAndValues(b.N+labelCount, "label")
+		err = getSeriesIDForKeyValueArray(db, metricName, keys, values)
```
cevian, line 58: This calls get_series_id_for_key_value_array() with really big keys and values slices. This function is meant to be called for keys and values for one series. also don't we want this going to otherMetricName. So I think this should be 
```
		for n := 0; n < b.N; n++ {
			err = getSeriesIDForKeyValueArray(db, otherMetricName, keys[n:n+labelCount], values[n:n+labelCount])
			if err != nil {
				t.Fatal(err)
			}
		}
```

pkg/pgmodel/sql_bench_test.go
antekresic: I thought this test was running the function for an existing series. Hence it only uses one series with 5 labels. Why would it need to be more than that?

pkg/pgmodel/sql_bench_test.go
antekresic: Again, why? Benchmark is about using existing series. Do you think it would give us better insight into the performance if we run the function on multiple series twice and measuring the second run? If so, I'll update it.

pkg/pgmodel/sql_bench_test.go
antekresic: This part isn't in the benchmark timing. I just do it like so to insert all the labels at once. Theoretically, doing it this way or your way, shouldn't matter for the sql performance?

antekresic:
Updated the benchmarks with @cevian s latest comments. Here are the new results: 
```
goos: linux
goarch: amd64
pkg: github.com/timescale/timescale-prometheus/pkg/pgmodel
BenchmarkGetSeriesIDForKeyValueArrayExistingSeries-12                       1454            753370 ns/op
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesExistingLabels-12              1545            820569 ns/op
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesNewLabels-12                    322           3998221 ns/op
BenchmarkKeyValueArrayToLabelArrayCreateNewLabels-12                         289           3475210 ns/op
BenchmarkKeyValueArrayToLabelArrayExistingLabels-12                         1963            583342 ns/op
BenchmarkKeyValueArrayToLabelArrayCreateNewLabelKeys-12                      355           3716721 ns/op
BenchmarkGetOrCreateMetricTableName-12                                      4980            233311 ns/op

```

PR #40, opened 2020-04-20
Add functions for selecting labels based on filters

cevian:
Added functions and operators for working with selectors
and filtering on labels. More documentation in sql_schema.md
changes in this commit.


pkg/docs/sql_schema.md
```diff
@@ -95,4 +97,78 @@ Example query to look at all the series:
...
+- `!=~` match tag values that are not equal to the pattern
+
+These four matchers correspond to each of the four selectors in PromQL but with slightly
+different names (to avoid clashing with other PostgreSQL operators). They can
+be combined together using any boolean logic with any arbitrary where clauses.
+
+For those coming from PromQL there are a few differences to keep in mind:
+- Regexes are not anchored for you. Although, you can of course add anchors (`^$`) yourself.
+- The logic for whether series that are missing the tag key pass the qualifier is slightly different:
+  The `!==` and `!=~` do always match series with missing keys while `==` and `==~` do not.
```
JLockerman, line 147:
```suggestion
  If the key on the left-hand side is not found `!==` and `!=~` always match, while `==` and `==~` never match.
```

pkg/docs/sql_schema.md
```diff
@@ -95,4 +97,78 @@ Example query to look at all the series:
...
+
+
+### Equivalence
+
+The `eq` function tests equivalence between labels without comparing the metric name (`__name__`) label key.
```
JLockerman, line 152:
```suggestion
The `eq` function tests exact equivalence between labels, without comparing the metric name (`__name__`) label key.
For instance if the labels `a` is `{"__name__":"metric", "foo":"bar", "baz":"frob"}`
then `SELECT eq(a, jsonb {"__name__":"something else", "foo":"bar", "baz":"frob"})` will evaluate to `true`, however, unlike `@>` if `SELECT eq(a, {"__name__":"metric", "foo":"bar"})` will evaluate to `false`.
```

pkg/docs/sql_schema.md
```diff
@@ -95,4 +97,78 @@ Example query to look at all the series:
...
+
+```SQL
+SELECT *
+FROM cpu_usage
+WHERE eq(labels , jsonb '{“namespace”:”prod”,”node”:”pinky”,"zone":"us-a1-east"}')
+```
+
+Note the above json equivalence requires the entire json object
+to be specified. If you want a partial match see the Containment
+section above.
```
JLockerman, line 173: I think we should cut this, or expand it to be clear that exactness is required for `eq(label_ids, label_ids)` as well

pkg/docs/sql_schema.md
```diff
@@ -95,4 +97,78 @@ Example query to look at all the series:
...
+To test whether a series contains a json fragment you can use the `@>` containment operator.
+
+For example to find all metrics on in the dev namespace and on node pinky, run:
+
+```SQL
+SELECT *
+FROM cpu_usage  u
+WHERE labels @> jsonb '{"namespace":"dev", "node": "pinky"'}
+```
+
```
JLockerman, line 117: perhaps a sample output?

> Sample output
> ``` 
>  series_id | labels  | namespace  | node  | region
>-----------+---------+------------+-------+--------
>          4 | {3,4,5} | dev        | pinky | East
>          5 | {3,4,5} | dev        | pinky | West
> ```
To demonstrate that unspecified matchers can be different

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -974,3 +974,179 @@ BEGIN
...
+--------------------- op ? ------------------------
+
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_contains(labels SCHEMA_PROM.label_ids, json_labels jsonb)
+RETURNS BOOLEAN
+AS $func$
+    SELECT labels @> SCHEMA_PROM.get_matcher_from_json(json_labels)
+$func$
+LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
+
+CREATE OPERATOR SCHEMA_PROM.@> (
```
JLockerman, line 1057: Should this be in a separate section `--- op @> ---`?

pkg/pgmodel/view_test.go
```diff
@@ -167,8 +167,164 @@ func TestSQLView(t *testing.T) {
...
+
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		queries := []struct {
+			where string
+			rows  int
+		}{
+			{
```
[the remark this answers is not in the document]
davidkohn88: And especially joins with filters from the other bits. Just want to make sure everything comes together properly in that case.

pkg/pgmodel/view_test.go
```diff
@@ -167,8 +167,164 @@ func TestSQLView(t *testing.T) {
...
+				where: "WHERE labels ? ('namespace' == 'dev')",
+				rows:  3,
+			},
+			{
+				where: "WHERE labels ? ('namespace' !== 'dev')",
+				rows:  2,
+			},
+			{
+				where: "WHERE labels ? ('namespace' ==~ 'de.*')",
```
davidkohn88, line 261: Should we do some tests with explain on and see if we can get things to inline and get initplans?

davidkohn88: I think a lateral join from the metric table with eq to the series table would also be a very useful explain test / I'm just curious how that one looks. The explain stuff can probably just be in json format and the comparison should be relatively easy then? Not sure how PGX handles that output though.

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -974,3 +974,179 @@ BEGIN
...
+
+
+
+---------------- eq functions ------------------
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.eq(labels1 SCHEMA_PROM.label_ids, labels2 SCHEMA_PROM.label_ids)
+RETURNS BOOLEAN
+AS $func$
+    --assumes labels have metric name in position 1 and have no duplicate entries
+    SELECT array_length(labels1, 1) = array_length(labels2, 1) AND labels1 @> labels2[2:]
```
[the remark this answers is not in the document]
cevian: I don't think so, most things will have the same array_length. It's not selective at all. But `labels1 @> labels2[2:]` should still use that index

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -974,3 +974,179 @@ BEGIN
...
+LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
+COMMENT ON FUNCTION SCHEMA_PROM.eq(SCHEMA_PROM.label_ids, SCHEMA_PROM.label_ids)
+IS 'returns true if two label arrays are equal, ignoring the metric name';
+
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.eq(labels1 SCHEMA_PROM.label_ids, matchers SCHEMA_PROM.matcher_positive)
+RETURNS BOOLEAN
+AS $func$
+    --assumes no duplicate entries
+     SELECT (array_length(labels1, 1) - 1) = array_length(matchers, 1)
```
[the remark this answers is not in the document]
cevian: I'll change that

PR #43, opened 2020-04-21
Fix docker compose example for 0.1.0-alpha.2

farrajota:
This PR fixes the environment variable names in docker-compose.yml example setup for the v0.1.0-alpha.2.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     environment:
       POSTGRES_PASSWORD: postgres
       POSTGRES_USER: postgres
-    image: timescale/pg_prometheus:latest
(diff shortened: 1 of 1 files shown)


antekresic:
Thank you for the contribution! 

Would you mind squashing the commits? We can merge this immediately after that.

farrajota:
Since the latest docker image of timescale is working with this setup, would you rather have the current image `timescale/pg_prometheus:latest-pg11` replaced by  `timescale/timescaledb:latest-pg12` ?

antekresic:
@farrajota that sounds good! We definitely don't tepend on pg_prometheus for the connector.

antekresic:
Looking good! Thanks again @farrajota

PR #44, opened 2020-04-22
Add links to Go builds and reviewdog-golangci badges

antekresic:


antekresic:
@blagojts could you add the badge URL from codeclimate here? I'll add it to this PR. Thanks!

PR #45, opened 2020-04-22
Set query integration tests to read-only mode

antekresic:
Setting the read-only mode on the test database ensures we are not
writing to the database when executing queries.

### DIFF
--- a/pkg/internal/testhelpers/containers.go
+++ b/pkg/internal/testhelpers/containers.go
@@ -18,6 +18,11 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
+const (
+	defaultDB       = "postgres"
+	connectTemplate = "postgres://postgres:password@%s:%d/%s"
+)
+
 var (
 	PromHost          = "localhost"
 	PromPort nat.Port = "9090/tcp"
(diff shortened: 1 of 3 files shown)


pkg/internal/testhelpers/containers.go
```diff
@@ -26,24 +31,28 @@ var (
...
 	db := dbSetup(t, DBName)
 	defer func() {
 		db.Close()
 	}()
-	f(db, t, pgConnectURL(t, DBName))
+	f(db, t, pgConnectURL(DBName))
+}
+
+func SetDatabaseToReadOnly(t testing.TB, db *pgxpool.Pool, DBName string) {
+	_, err := db.Exec(context.Background(), fmt.Sprintf("ALTER DATABASE %s SET default_transaction_read_only = true", DBName))
```
cevian, line 48: I think you want to close all the pool connections after this to make sure new connections are started with this mode?

antekresic: Updated to use a read-only session for query integration tests.

Could you take another look?

pkg/pgmodel/query_integration_test.go
```diff
@@ -527,6 +529,7 @@ func TestSQLQuery(t *testing.T) {
 
 			})
 		}
+		readOnly.Close()
```
cevian, line 532: can't this be in a defer?

pkg/pgmodel/query_integration_test.go
```diff
@@ -908,6 +913,7 @@ func TestPromQL(t *testing.T) {
 				}
 			})
 		}
+		readOnly.Close()
```
cevian, line 916: same here

PR #49, opened 2020-04-24
Add main package unit tests

antekresic:
These tests cover most of the basic declaritive logic of the main package.
Some of the functions require a DB to be fully tested, to test them would
require major refactorings of the project so they are out-of-scope for this
commit.

### DIFF
--- a/cmd/timescale-prometheus/main_test.go
+++ b/cmd/timescale-prometheus/main_test.go
@@ -1,0 +1,577 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+	"testing/iotest"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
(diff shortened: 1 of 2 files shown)


cmd/timescale-prometheus/main_test.go
```diff
@@ -0,0 +1,577 @@
...
+				t.Errorf("leader gauge metric not set correctly: got %f when is leader", mockGauge.value)
+			case !c.isLeader && mockGauge.value != 0:
+				t.Errorf("leader gauge metric not set correctly: got %f when is not leader", mockGauge.value)
+			}
+		})
+	}
+}
+
+func TestInitElector(t *testing.T) {
+	// TODO: refactor the function to be fully testable without using a DB.
```
blagojts, line 446: can we set this up as a integration test for now?

antekresic: The reason we can't use DB for testing is because main package can't access internal packages. Otherwise, I would have added it here. I'm unsure how to resolve it right now so I'm leaving it for the future. Ideally, we should refactor the code to be more testable vs. just using integration tests and spinning up docker containers for every test.

PR #58, opened 2020-04-27
Add a postgres extension with some optimizations

JLockerman:
There are some optimizations we would like to perform (adding a support
function, and declaring some functions in a way that allows for
inlining) which can only be performed by a superusers. To avoid
requiring superuser permissions for everyone who wants to use the
adapter we will instead package this functionality in a postgres
extension, for which they can be given permission separately.


extension/sql/timescale_prometheus.sql
```diff
@@ -0,0 +1,62 @@
+
+CREATE OR REPLACE FUNCTION _prom_catalog.const_support(internal) RETURNS INTERNAL
```
cevian, line 2: Let's not specify a schema for the objects in any of these. Rather the schema will be installed in a particular schema with `CREATE EXTENSION ... WITH SCHEMA '_prom_catalog'`. That way the user of the extension can put it anywhere it wants and there is less coupling between the connector and extension.

extension/Makefile
```diff
@@ -0,0 +1,62 @@
+PG_CONFIG=pg_config
+
+EXTENSION=timescale_prometheus_support
```
cevian, line 3: why don't we just call this timescale-prometheus?

extension/sql/timescale_prometheus.sql
cevian, line 2: also maybe let's now give this function a better name. `const_arguments_initplan_support` ?

extension/sql/timescale_prometheus.sql
```diff
@@ -0,0 +1,62 @@
...
+--wrapper around unnest to give better row estimate (10 not 100)
+CREATE OR REPLACE FUNCTION _prom_catalog.label_unnest(label_array anyarray)
+ RETURNS SETOF anyelement
+ LANGUAGE internal
+ IMMUTABLE PARALLEL SAFE STRICT ROWS 10
+AS $function$array_unnest$function$;
+
+---------------------  comparison functions ---------------------
+
+CREATE OR REPLACE FUNCTION _prom_catalog.label_find_key_equal(label_key SCHEMA_PROM.label_key, pattern SCHEMA_PROM.pattern)
```
cevian, line 23: instead of overwritting all these functions can we just have a noop stub for const_support in the connector sql and only overwrite that?

extension/Readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+This [Postgres extension](https://www.postgresql.org/docs/12/extend-extensions.html)
+contains support functions to improve the performance of Timescale-Prometheus.
+While Timescale-Prometheus will run without it, adding this extension will
+cause it perform better.
+
+## Installation ##
+
+The extension is installed by default on the timescale-prometheus docker image,
+and will be installed automatically by timescale-prometheus, if it has
```
cevian, line 11: I wouldn't put the ```CREATE EXTENSION timescale_prometheus_support;``` instruction here because the connector and not the use should call that. I'd just put the make && make install here and say that the prometheus-connector will call CREATE EXTENSION

extension/sql/timescale_prometheus.sql
JLockerman: unfortunately we can't; you're not allowed to use the `SUPPORT` option unless you're superuser

extension/Readme.md
JLockerman: oh, whoops, I thought I'd deleted that line 😅

extension/sql/timescale_prometheus.sql
JLockerman: While I like the idea, I think the has issues with the operator functions: we need to create the operator functions in the extension, because they need superuser, and the connector needs to know their names to be able to define the operators (I don't really want to rely on what happens to postgres when you create multiple of the same operator...)

extension/Makefile
JLockerman: The name is ambiguous, it's difficult to tell if `timescale-prometheus` refers to the connector or the extension. Since the only time the name will ever come up is during debugging, I think that disambiguation is more important than prettiness

cevian: let's call it `timescale-prometheus-extra` then note also - instead of _

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -1127,6 +1127,14 @@ AS $func$
 $func$
 LANGUAGE SQL STABLE PARALLEL SAFE;
 
+DO $$
```
cevian, line 1130: I think we want to separate out stub-of-functions in extension and extension-install into a separate migrate script just for clarity. can we do that?

JLockerman: I not sure we should, we'll end up four different scripts, with weird dependencies:

JLockerman:
due to
```SQL
create extension timescale-prometheus-extra;
ERROR:  syntax error at or near "-"
LINE 1: create extension timescale-prometheus-extra;
```
we're naming the extension
```SQL
timescale_prometheus_extra
```

JLockerman:
Note that the image with the extension is not yet pushed, so the tests cannot be run without building the image locally.

JLockerman:
Image is now up

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -16,7 +16,7 @@ DO $$
 DECLARE
    new_path text;
 BEGIN
-   new_path := current_setting('search_path') || format(',%L,%L', 'SCHEMA_PROM', 'SCHEMA_METRIC');
+   new_path := current_setting('search_path') || format(',%L,%L,%L,%L', 'SCHEMA_EXT', 'SCHEMA_PROM', 'SCHEMA_METRIC', 'SCHEMA_CATALOG');
```
cevian, line 19: can we add a comment about why the order is the way it is and how this interacts with extensions.

extension/src/support.c
```diff
@@ -0,0 +1,64 @@
+/* This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE for a copy of the license
+ */
+
+#include <postgres.h>
+#include <fmgr.h>
+
+#include <nodes/pathnodes.h>
+#include <nodes/supportnodes.h>
```
davidkohn88, line 10: I think this should be documented significantly more, in terms of what it's doing, what its limitations are etc. It's very difficult to review this way and also will mean that it's much more difficult to maintain.

extension/sql/timescale-prometheus.sql
```diff
@@ -0,0 +1,85 @@
...
+CREATE OR REPLACE FUNCTION label_unnest(label_array anyarray)
+ RETURNS SETOF anyelement
+ LANGUAGE internal
+ IMMUTABLE PARALLEL SAFE STRICT ROWS 10
+AS $function$array_unnest$function$;
+
+---------------------  comparison functions ---------------------
+
+CREATE OR REPLACE FUNCTION label_find_key_equal(label_key prom.label_key, pattern prom.pattern)
+RETURNS prom.matcher_positive
```
davidkohn88, line 24: I find it confusing where we use the templating thing with the different schemas and where we don't. Is it that we can't run the matching thingy over these files? Or is it something else, it just strikes me that the mixture of both strategies is odd and could easily lead to errors in the future.

pkg/pgmodel/migrate.go
```diff
@@ -82,6 +96,10 @@ func Migrate(db *sql.DB) error {
 	}
 
 	err = m.Up()
+	if err == nil {
+		_, err = db.Exec(fmt.Sprintf(extensionInstall, extSchema))
```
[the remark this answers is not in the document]
JLockerman: if we do that, I think we should just log the error at the connector, and not try to parse it; not sure we'll get all the error cases, and I think it's better to not kill the connector if this fails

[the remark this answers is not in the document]
cevian: yep, we catch, log, and ignore the error. It's ok if the extension isnt installed.

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -12,6 +12,9 @@ CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;
 
 CREATE DOMAIN SCHEMA_PROM.label_array AS int[] NOT NULL;
 
+-- the timescale_prometheus_extra extension contains optimized version of some
+-- of our. To ensure the correct version of the functions are used, SCHEMA_EXT
```
cevian, line 16:
```suggestion
-- of our functions and operators. To ensure the correct version of the functions are used, SCHEMA_EXT
```

extension/sql/timescale-prometheus.sql
cevian: This is a good point. this shouldn't be schema qualified.

JLockerman:
@cevian IIRC, `label_find_key_equal` will exist in whatever schema the extension is put it

extension/sql/timescale-prometheus.sql
JLockerman: ok, should work now that they're in the path

JLockerman: Nope, doesn't work. I think the way that the search path is set on `CREATE EXTENSION` prevents it.

JLockerman: Ok, I got it to work, but the extension will only be relocatable at install

pkg/pgmodel/migrate_test.go
```diff
@@ -1061,11 +1062,56 @@ func TestSQLDropChunk(t *testing.T) {
...
+		functions := []string{
+			"label_jsonb_each_text",
+			"label_unnest",
+			"label_find_key_equal",
+			"label_find_key_not_equal",
+			"label_find_key_regex",
+			"label_find_key_not_regex",
+		}
+		for _, fn := range functions {
+			const query = "SELECT nspname FROM pg_proc LEFT JOIN pg_namespace ON pronamespace = pg_namespace.oid WHERE pg_proc.oid = $1::regproc;"
```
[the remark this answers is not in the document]
davidkohn88: Might want to add a comment that that's how it works, we'll recognize but if someone else is looking at this code they might not get it.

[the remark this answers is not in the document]
davidkohn88: @JLockerman

extension/src/support.c
```diff
@@ -0,0 +1,64 @@
...
+
+#ifdef PG_MODULE_MAGIC
+PG_MODULE_MAGIC;
+#endif
+
+PG_FUNCTION_INFO_V1(const_support);
+Datum
```
cevian, line 17:
```suggestion
/* This is a support function that optimizes calls to the supported function if it's called with constant arguments.
     Such calls are transformed into a subselect of the function call. This allows the planner to make this call
     an InitPlan which is evaluated once per query instead of multiple times (e.g. on every tuple when the function is 
     used in a WHERE clause). This should be used on any stable function that is often called with constant
     arguments. 
*/
Datum
```

extension/src/support.c
```diff
@@ -0,0 +1,64 @@
...
+		Query *query;
+		SupportRequestSimplify *req = (SupportRequestSimplify *) rawreq;
+		FuncExpr   *expr = req->fcall;
+		ListCell *lc;
+		SubLink    *sublink;
+		if (req->root == NULL)
+		{
+			PG_RETURN_POINTER(NULL);
+		}
+		if (req->root->query_level > 1)
```
cevian, line 35:
```suggestion
                 /* This prevents recursion of this optimization when the subselect is planned */
		if (req->root->query_level > 1)
```

PR #59, opened 2020-04-27
Add informational views

cevian:
Add views about metrics and labels.


docs/sql_schema.md
```diff
@@ -112,6 +112,48 @@ Example query to look at all the series:
...
+the following columns
+
+- id
+- metric_name
+- table_name - name of the tables and views in the `prom_metric` and `prom_series`
+  schemas
+- retention_period - the length of time data in this table will be kept
+- label_keys - an array of the label keys used with this metric
+- size - disk space used for storing the metric
+- compression_ratio - the compression ratio achieved for this metric
```
JLockerman, line 127: is this `compressed_size / uncompressed_size`? if so we should say it

cevian: 1 - (compressed_size / uncompressed_size)

davidkohn88: Maybe just give them compressed bytes and uncompressed bytes and let them figure out how they want to deal with that themselves?

davidkohn88: It's also very useful in that case to be able to sum sizes etc, because otherwise, with the text repr, it gets really hard

cevian: I added the description but don't think that giving raw numbers is necessary. Most people only care about ratio and people can always get more info from hypertable_information view.

PR #61, opened 2020-04-29
Mask password data before outputting to log

antekresic:


pkg/util/util_test.go
```diff
@@ -55,3 +55,25 @@ func TestThroughputCalc(t *testing.T) {
 		}
 	}
 }
+
+func TestMaskPassword(t *testing.T) {
+	testData := map[string]string{
+		"password='foobar' host='localhost'":   "password='****' host='localhost'",
```
blagojts, line 61: Maybe add an example where password is not at the start of the string

antekresic: Thanks for that, tweaked the regex a bit, take a look.

pkg/util/util.go
```diff
@@ -55,3 +56,9 @@ func (dt *ThroughputCalc) Start() {
 		}()
 	}
 }
+
+// MaskPassword is used to mask sensitive password data before outputing to persistent stream like logs.
+func MaskPassword(s string) string {
+	s = regexp.MustCompile(`password=(\s*?)'([^']+?)'`).ReplaceAllString(s, "password=$1'****'")
```
cevian, line 62: should we save these Regexp objects (returned by MustCompile) as global vars or const so that we don't need to recompile on every call to the function?

pkg/util/util.go
```diff
@@ -55,3 +56,9 @@ func (dt *ThroughputCalc) Start() {
 		}()
 	}
 }
+
+// MaskPassword is used to mask sensitive password data before outputing to persistent stream like logs.
+func MaskPassword(s string) string {
```
cevian, line 61: Should this be called in our Logger on every log or is that too expensive? Are we sure we caught all places that passwords can leak?

antekresic: I went through the code and added masking to all places that could leak. We could add this to all logs, not sure about the perf penalty. I could do some benchmarks to confirm. Might not be worth the trouble.

What do you think? @cevian

cevian: Ok, yeah lets leave as-is then.

PR #62, opened 2020-04-29
Add custom logger to caching layer

### LINKED ISSUE #52: Add custom logger to caching config
Currently used caching lib, bigcache, outputs its logs to standard out. 

We should set a custom logger to it that conforms the format of the global logging lib and only outputs the logs to debug level.

antekresic:
Default caching logger would output logs to standard output which did not conform
to our logging library output standards. This commit adds a custom logger which
transforms that output to match our logger.

Closes #52

### DIFF
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"regexp"
 	"runtime"
-	"time"
 
 	"github.com/allegro/bigcache"
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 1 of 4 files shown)


pkg/pgclient/client.go
```diff
@@ -69,8 +68,7 @@ func NewClient(cfg *Config) (*Client, error) {
 		return nil, err
 	}
 
-	config := bigcache.DefaultConfig(10 * time.Minute)
-	metrics, _ := bigcache.NewBigCache(config)
+	metrics, _ := bigcache.NewBigCache(pgmodel.DefaultCacheConfig())
```
blagojts, line 71: kind-of unrelated to this pr, but do we want to have diff retention periods for series and metrics

antekresic: I would leave that decision for when we start doing some benchmarks and monitoring later on.

PR #68, opened 2020-04-30
Add infrastructure for psql-based golden tests

cevian:
This adds the infrastructure for simple golden file tests.
Also add a simple golden file test for views.


cevian:
Having talked to Josh, I mentioned that these tests will always run on latest PG version. If we need tests that check older PG versions, we'll use a separate modality. I think that addresses a lot of the problems we had with these sorts of tests in TimescaleDB. We'll also work on other modalities and libraries in the future.

PR #72, opened 2020-05-01
Add Prerequisite check for timescale-prometheus-extra

cevian:
The extension requires that the schema for timescale-prometheus
connector installed and the migration is run. We add a check so
that the extension let's people know if something is wrong.


JLockerman:
(you'll need to run `make docker-image` `make docker-push`)

PR #73, opened 2020-05-01
Whitelist Timescale sxtensions so they can be installed by non-superusers

JLockerman:
This PR adds `pgextwlist` to our docker image, and sets it so that the `timescaledb` and `timescale_prometheus_extra` extensions can be added by non-superusers.


pkg/pgmodel/migrate_test.go
```diff
@@ -1157,6 +1157,15 @@ func TestExtensionFunctions(t *testing.T) {
...
 	}
 	withDB(t, *database, func(db *pgxpool.Pool, t testing.TB) {
+
+		searchPath := ""
+		expected := fmt.Sprintf(`"$user", public, %s, %s, %s, %s`, extSchema, promSchema, metricViewSchema, catalogSchema)
+		db.QueryRow(context.Background(), "SHOW search_path;").Scan(&searchPath)
+		if searchPath != expected {
+			t.Errorf("incorrect search path\nexpected\n\t%s\nfound\n\t%s", expected, searchPath)
+		}
+		t.Logf("search path: %s", searchPath)
```
cevian, line 1167: remove log

extension/sql/timescale-prometheus.sql
```diff
@@ -1,5 +1,27 @@
 
-SET LOCAL search_path TO DEFAULT;
+-- Set the search path to one that will find all the defitions provided by the
```
cevian, line 2:
```suggestion
-- Set the search path to one that will find all the definitions provided by the
```

extension/sql/timescale-prometheus.sql
```diff
@@ -1,5 +1,27 @@
...
+-- connector. Since the connector can change the schemas it stores things
+-- in we cannot just hardcode the searchpath, instead we switch the search path
+-- to one procided by the connector. If there was no search path provided, we
```
cevian, line 5:
```suggestion
-- to one provided by the connector. If there was no search path provided, we
```

extension/sql/timescale-prometheus.sql
```diff
@@ -1,5 +1,27 @@
...
+-- be the search path the connector set, though this does not work if the
+-- extension is installed by pgextwlist. We use timescale_prometheus.search_path
+-- as the guc as it is unlikely to be used by anything else, we cannot use a guc
+-- in the timescale_prometheus_extra namespace, as this will cause the
+-- CREATE EXTENSION to fail.
+DO $$
+    DECLARE
+        val text;
+    BEGIN
+        val := current_setting('timescale_prometheus.search_path');
```
cevian, line 16: I actually think we want this recorded in the _prom_catalog.default table (which already exists) during connector install and read here instead of passed in as a guc. That way the install only requires the connect installed its sql stuff and removing the extra set-this-guc appropriately during extension install step.

JLockerman: that would mean that the extension needs to know the `_prom_catalog` schema, and the user couldn't rename it

extension/sql/timescale-prometheus.sql
```diff
@@ -1,5 +1,35 @@
 
-SET LOCAL search_path TO DEFAULT;
+-- Set the search path to one that will find all the definitions provided by the
```
cevian, line 2: This comment need to be rewrittten.

pkg/internal/testhelpers/containers.go
```diff
@@ -22,7 +22,13 @@ import (
 
 const (
 	defaultDB       = "postgres"
-	connectTemplate = "postgres://postgres:password@%s:%d/%s"
+	connectTemplate = "postgres://%s:password@%s:%d/%s"
+
+	postgresUser = "postgres"
+	promUser     = "prom"
+
+	SUPERUSER   = true
```
cevian, line 30: I think it is idiomatic for this to be `Superuser` and `NonSuperuser` all caps not Idiomatic

pkg/internal/testhelpers/containers.go
```diff
@@ -33,25 +39,29 @@ var (
...
+	if !superuser {
+		user = promUser
+	}
+	return fmt.Sprintf(connectTemplate, user, pgHost, pgPort.Int(), dbName)
 }
 
 // WithDB establishes a database for testing and calls the callback
-func WithDB(t testing.TB, DBName string, f func(db *pgxpool.Pool, t testing.TB, connectString string)) {
-	db, err := dbSetup(DBName)
+func WithDB(t testing.TB, DBName string, superuser bool, f func(db *pgxpool.Pool, t testing.TB, connectString string)) {
```
cevian, line 51: if we are gonna use a IsSuperuser type, should be used here as well, no?

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -9,8 +9,6 @@ CREATE SCHEMA IF NOT EXISTS SCHEMA_METRIC;
 CREATE SCHEMA IF NOT EXISTS SCHEMA_DATA;
 CREATE SCHEMA IF NOT EXISTS SCHEMA_INFO;
 
-CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;
```
cevian, line 12: should this be moved out of here? Why was this moved?

pkg/internal/testhelpers/containers.go
```diff
@@ -64,8 +74,10 @@ func GetReadOnlyConnection(t testing.TB, DBName string) *pgxpool.Pool {
 	return dbPool
 }
 
-func dbSetup(DBName string) (*pgxpool.Pool, error) {
-	db, err := pgx.Connect(context.Background(), pgConnectURL(defaultDB))
+type IsSuperuser = bool
```
cevian, line 77: Maybe SuperuserStatus a better name?

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
JLockerman: if we leave it hear the migrate fails with
```SQL
ERROR: parameter "extwlist.extensions" must be a list of extension names
```
I suspect the migrate library is doing something non-standard with connections, which is causing the guc to not load correctly.

PR #76, opened 2020-05-05
Refactor docker tests into their own subpackage

JLockerman:
This commit refactors out all the tests that rely on a running instance
of the database into their own package end_to_end_tests, and breaks up
migrate_test.go into multiple files, whose names are more descriptive
of their roles. Creating a separate end_to_end_tests package has a
number of advantages:

  1. The other tests run faster as they no longer need to start a
     docker container.

  2. The end-to-end tests are more restricted as to what they can rely
     on, ensuring the correct encapsulation.

  3. It is clearer which tests need docker, and which can run without.

  4. There are a significant number of end-to-end tests, and this
     allows us to keep them together in a single subdirectory.

Performing this refactor found two things that are part of the public
API of pgmodel that were mistakenly made private: metricLabelName and
ErrNoMetricName. The former is defacto public as it is written to the
database, and users can easily query based on it. The latter should be
public as that is the only users can check for that specific error.


pkg/internal/testhelpers/containers_test.go
```diff
@@ -46,7 +46,7 @@ func TestWithDB(t *testing.T) {
 	if testing.Short() {
```
antekresic, line 46: With this test setup, there is no need for checking the short flag, since the integration tests are in a separate package.

pkg/pgmodel/end_to_end_tests/concurrent_sql_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Fatal(err)
+	}
+	if id == nil {
+		t.Fatalf("NULL found")
+	}
+	return *id
+}
+
+func TestConcurrentSQL(t *testing.T) {
+	if testing.Short() {
```
antekresic, line 56: No need for short check.

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -0,0 +1,420 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/prometheus/prompb"
+
+	_ "github.com/jackc/pgx/v4/stdlib"
+
+	. "github.com/timescale/timescale-prometheus/pkg/pgmodel"
+)
+
+func TestSQLGetOrCreateMetricTableName(t *testing.T) {
+	if testing.Short() {
```
antekresic, line 21: No need for short check.

JLockerman:
I'm leaving the `Short` decision for a followup PR

PR #78, opened 2020-05-06
Create Docker images in CI

JLockerman:
This commits edits our CI to always create its own docker images, this should
catch any bugs edits to the extension could introduce.

### DIFF
--- a/extension/DockerMakefile
+++ b/extension/DockerMakefile
@@ -1,0 +1,25 @@
+# the regular Makefile requires a valid postgres install as it uses the postgres
+# makefiles. Since when building docker images we don't care what the local
+# install is, we've seperated it out into this file.
+
+include dependencies.makefile
+
+EXTENSION=timescale_prometheus_extra
+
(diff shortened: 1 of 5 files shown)


extension/Makefile
```diff
@@ -64,7 +64,7 @@ package: clean $(EXT_SQL_FILE)
 	$(install_sh) -m 644 $(EXTENSION).control 'package/extension/'
 	$(install_sh) -m 644 $(EXT_SQL_FILE) 'package/extension/'
 
-docker-image: Dockerfile
+docker-image: Dockerfile $(SQL_FILES)
```
cevian, line 67: we should delete this section

extension/DockerMakefile
```diff
@@ -0,0 +1,25 @@
+PG_CONFIG?=pg_config
```
cevian, line 1: please add a comment about why we need a separate makefile

extension/Makefile
cevian: But add a comment about the other makefile

PR #81, opened 2020-05-08
Fix series deadlock

cevian:
Previoussly there was a deadlock in get_series_id functions
because the get_or_create_metric_table_name call may take
an exclusive lock on the series table but the startup of
the sql function took an access lock on that same table because
SQL functions parse/plan all statements in the function before
executing any. Thus, even though get_or_create_metric_table_name
was executed before any series table access, the lock was already taken.

This lock escalation caused a deadlock because the insert on the
metric table has an ON CONFLICT DO NOTHING and thus a waiting
process may wait on a winning process to complete while holding
an access lock on the series table, but the winning
process needs an exclusive lock on the series table.

Switching to a PLPGSQL function helps because locks are
taken in order of execution and so there is no lock escalation.

We also added complete ingest concurrency tests to try
to catch this in the future.

Finally, we ran benchmark tests and it seems that the switch to
PLPGSQL does not significantly degrade performance.


pkg/pgmodel/end_to_end_tests/concurrent_sql_test.go
```diff
@@ -108,3 +166,110 @@ func TestConcurrentSQL(t *testing.T) {
...
+}
+
+func TestConcurrentInsert(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	withDB(t, *testDatabase, func(db *pgxpool.Pool, t testing.TB) {
+		var wg sync.WaitGroup
+		wg.Add(2)
```
antekresic, line 253: Very minor nit, usually you add inside the goroutine so its more understandable who you are waiting for :)

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -693,8 +699,11 @@ GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_series(int, name, SCHEMA_PROM.la
 
 CREATE OR REPLACE  FUNCTION SCHEMA_PROM.series_id(label jsonb)
```
davidkohn88, line 700: Given all of these locking issues, I'd want a get only version of this function I think that didn't try and create any of this infrastructure and wouldn't accidentally end up locking (or not working on a replica?)

davidkohn88: Also, I thought we were saying that all of the volatile functioins weren't going to be in the public api essentially? Because they shouldn't be run by users?

pkg/pgmodel/end_to_end_tests/concurrent_sql_test.go
cevian: I'd have to change it all over, so separate pr :)

PR #82, opened 2020-05-08
Fix the cronjob call

jpigree:
The `prom` schema does not exist anymore. It is replaced by `prom_api`.
cf. Issue #83  for more context.


cevian:
@thanks a lot @jpigree

PR #84, opened 2020-05-08
Add version const and print on boot

blagojts:
When debugging issues that users are facing we need to know
which version they are using. Especially in container environments 
because of stale :latest images.

### DIFF
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -142,17 +142,19 @@ func main() {
 	cfg := parseFlags()
 	err := log.Init(cfg.logLevel)
 	if err != nil {
+		fmt.Println("Version: ", Version, "Commit Hash: ", CommitHash)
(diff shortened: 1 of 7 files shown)


cmd/timescale-prometheus/main.go
```diff
@@ -47,6 +47,7 @@ type config struct {
 }
 
 const (
+	version           = "0.1.0-dev"
```
blagojts, line 50: Maybe we need to make this a placeholder string so it can be replaced when releasing by an arg, or when the CI is building

antekresic: Maybe we could include the commit? Using `go generate` since we are already using that.

cevian: I like the idea of a git commit a lot

blagojts:
Added the git hash to be applied if go generate is run on ./cmd/timescale-prometheus

PR #85, opened 2020-05-08
Take into account ErrNoChange returned from pkg migrate

blagojts:
The migrate package uses ErrNoChange to signal that
no changes were migrations were applied because
the database was up to date. We didn't check this
case and it made it impossible to restart the connector.

Branch was made from 0.1.0-alpha.3 release, we should issue a patch to the release before rebasing this on the current master

### DIFF
--- a/pkg/pgmodel/migrate_test.go
+++ b/pkg/pgmodel/migrate_test.go
@@ -47,20 +47,41 @@ func TestMigrate(t *testing.T) {
 		t.Skip("skipping integration test")
 	}
 	withDB(t, *database, func(db *pgxpool.Pool, t testing.TB) {
-		var version int64
-		var dirty bool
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/migrate_test.go
```diff
@@ -47,22 +47,43 @@ func TestMigrate(t *testing.T) {
 		t.Skip("skipping integration test")
 	}
 	withDB(t, *database, func(db *pgxpool.Pool, t testing.TB) {
-		var version int64
-		var dirty bool
-		err := db.QueryRow(context.Background(), "SELECT version, dirty FROM prom_schema_migrations").Scan(&version, &dirty)
+		checkMigrationsState(db, t)
+	})
+}
```
blagojts, line 52: Maybe we should delete one of these since they test the same thing

PR #89, opened 2020-05-11
Optimize series inserts

cevian:
Previously, series inserts were slower because they were done to the
series parent. Since we are now using a PLPGSQL function, we can
optimize this to go to the partition.

Also prevent the series table lock from going to the children,
this is unnecessary and can cause deadlocks because access to the
children also takes a lock on the parent.

On the BenchmarkGetSeriesIDForKeyValueArrayNewMetric benchmark
this improves performance from 250ms/op to 50ms/op.


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -197,7 +197,8 @@ BEGIN
     -- stupidly create table ... partition of X first takes an access share lock
     -- and then an access exclusive lock. Such an upgrade can, and does, cause
     -- deadlocks. Taking the stronger lock beforehand prevents such deadlocks.
-    LOCK TABLE _prom_catalog.series in ACCESS EXCLUSIVE mode;
+    -- We only need the lock on the parent.
+    LOCK TABLE ONLY SCHEMA_CATALOG.series in ACCESS EXCLUSIVE mode;
```
davidkohn88, line 201: Got it, we're probably still access share locking hte other children to prevent their constraints from being changed at the same time. But that's fine.

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -727,34 +732,38 @@ IS 'returns the series id that exactly matches a JSONB of labels';
 GRANT EXECUTE ON FUNCTION SCHEMA_PROM.series_id(jsonb) TO prom_writer;
 
 CREATE OR REPLACE  FUNCTION SCHEMA_CATALOG.get_series_id_for_key_value_array(metric_name TEXT, label_keys text[], label_values text[])
```
davidkohn88, line 734: This is also improperly named according to our naming convention as it also creates. I would also be okay with renaming all the get or create functions to goc_ or gc_ or something if we think that that would be better for name length etc.

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -701,21 +702,25 @@ CREATE OR REPLACE  FUNCTION SCHEMA_PROM.series_id(label jsonb)
...
+        SELECT SCHEMA_PROM.label_array($1)
+    )
+    SELECT id
+    FROM SCHEMA_DATA_SERIES.%1$I as series
+    WHERE labels = (SELECT * FROM cte)
+    UNION ALL
+    SELECT SCHEMA_CATALOG.create_series(%2$L, %1$L, (SELECT * FROM cte))
+    LIMIT 1
+   $query$, table_name, metric_id)
+   USING label
```
davidkohn88, line 723: I still think this should be renamed to get or create and then we should have a get only one or something like that that is exposed. Especially given that it takes an exclusive lock on the table.

PR #94, opened 2020-05-20
Switch to per-metric inserter

cevian:
With or recent path to pgx's pgconn, CopyFrom is much less CPU intensive
than it was, making a goroutine per-metric look practical again. This
commit does that

We also move the seriesID inserts into the per-metric goroutine logic
This creates more parallelism and should help with locking issues
since the series table is partitioned by metric. Also makes the series
caches not have to be thread-safe.

This commit also  ensures that table-creation happens with the same
batching as everything else.


pkg/pgmodel/pgx.go
```diff
@@ -187,140 +184,41 @@ func NewPgxIngestor(c *pgxpool.Pool) *DBIngestor {
...
+	inserters        sync.Map
 }
 
 func (p *pgxInserter) Close() {
-	for i := 0; i < len(p.inserters); i++ {
-		close(p.inserters[i])
-	}
+	// for i := 0; i < len(p.inserters); i++ {
+	// 	close(p.inserters[i])
+	// }
```
JLockerman, line 202:
```suggestion
```

pkg/pgmodel/pgx.go
```diff
@@ -455,27 +342,19 @@ func runInserterRoutine(conn pgxConn, input chan insertDataRequest) {
...
-				if !receivingNewReqs {
-					break hotReceive
-				}
-			}
-			if time.Since(startHandling) > flushTimeout {
-				handler.flushTimedOutReqs()
-				startHandling = time.Now()
+		for handler.nonblockingHandleReq() {
+			if len(handler.pending.batch.sampleInfos) >= flushSize {
+				break hotReceive
```
JLockerman, line 348: We should benchmark sleeping when we have data, instead of relying on the hot-loop; I suspect that the individual workers won't have enough work for the loop to be justified now that each one only handles a single metric. (this can go in the next PR)

pkg/pgmodel/pgx_test.go
```diff
@@ -432,18 +415,15 @@ func TestPGXInserterInsertSeries(t *testing.T) {
...
-					return
 				default:
 					t.Errorf("unexpected error: %v", err)
 				}
 			}
 
-			if calls != len(c.series) {
-				t.Errorf("Callback called wrong number of times: got %v expected %d", calls, len(c.series))
+			for _, si := range lsi {
+				if si.seriesID == 0 {
```
JLockerman, line 424: we should also check `si.seriesID == -1` since that's the sentinel we use for not-yet-set

pkg/pgmodel/ingestor.go
```diff
@@ -77,43 +78,22 @@ func (i *DBIngestor) parseData(tts []prompb.TimeSeries) ([]seriesWithCallback, m
...
-				return nil, nil, rows, err
-			}
-			newSeries = true
-		}
-
-		sample := samplesInfo{
-			seriesID,
+		sample := labeledSamplesInfo{
+			seriesLabels,
+			-1,
```
JLockerman, line 88:
```suggestion
			-1, // sentinel marking the seriesId as unset
```

pkg/pgmodel/ingestor.go
```diff
@@ -52,22 +53,22 @@ type DBIngestor struct {
...
 		return rowsInserted, fmt.Errorf("Failed to insert all the data! Expected: %d, Got: %d", totalRows, rowsInserted)
 	}
 	return rowsInserted, err
 }
 
-func (i *DBIngestor) parseData(tts []prompb.TimeSeries) ([]seriesWithCallback, map[string][]*samplesInfo, int, error) {
-	var seriesToInsert []seriesWithCallback
-	dataSamples := make(map[string][]*samplesInfo)
+func (i *DBIngestor) parseData(tts []prompb.TimeSeries) (map[string][]samplesInfo, int, error) {
+	// var seriesToInsert []seriesWithCallback
```
JLockerman, line 70:
```suggestion
```

PR #95, opened 2020-05-20
Split metric creation into sync and async parts

cevian:
Metric creation can happen on the insert path and
the insert path is latency-sensitive because it
needs to ack the request within a reasonable
amount of time. To prevent problems when first
starting the connector, we make the metric creation
happen in two phases. Only the first phase happens
during insert. Phase 2 happens async.

Phase 1) creates just enough structures to insert
data. Because that does not require heavy locks
it's low-latency.
Phase 2) Finishes all the other setup. Most
importantly it attaches the series partition
to the series table. This requires a heavy lock
held for a short time. So this is a low latency
operation but doesn't happen on the hot path.

We also do the following:
- Add not null constraints on data table
- Set up time index explicitly and turn off default indexes
- Document and fix up canonical lock ordering between tables


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -171,47 +182,91 @@ $func$
 LANGUAGE SQL STABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_default_retention_period() TO prom_reader;
 
+--lock-order: metric table, data_table, series parent, series partition
+CREATE PROCEDURE SCHEMA_CATALOG.finalize_metric_creation()
```
davidkohn88, line 186: Add a comment explaining why we're splitting these steps out
Also, explain how this is run in the comment, but that we build it to allow concurrency

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -171,47 +182,91 @@ $func$
...
+AS $proc$
+DECLARE
+    r RECORD;
+    created boolean;
+BEGIN
+    FOR r IN
+        SELECT *
+        FROM SCHEMA_CATALOG.metric
+        WHERE NOT creation_completed
```
davidkohn88, line 195: Should we order by random() to allow for better concurrency in the case where that happens?

pkg/pgmodel/pgx.go
```diff
@@ -322,12 +364,68 @@ const (
...
+}
+
+func runInserterRoutine(conn pgxConn, input chan insertDataRequest, metricName string, completeMetricCreationSignal chan bool, errChan chan error, metricTableNames MetricCache) {
+	tableName, err := metricTableNames.Get(metricName)
+	if err == ErrEntryNotFound {
+		var possiblyNew bool
+		tableName, possiblyNew, err = getMetricTableName(conn, metricName)
+		if err != nil {
+			select {
+			case errChan <- err:
```
JLockerman, line 399: In this case, when we we don't have a `tableName` we should exit the inserter routine; we won't be able to insert anyway.

pkg/pgmodel/pgx.go
```diff
@@ -173,30 +179,66 @@ func NewPgxIngestorWithMetricCache(c *pgxpool.Pool, cache MetricCache) *DBIngest
...
 }
 
 type pgxInserter struct {
-	conn             pgxConn
-	metricTableNames MetricCache
-	inserters        sync.Map
+	conn                   pgxConn
+	metricTableNames       MetricCache
+	inserters              sync.Map
+	completeMetricCreation chan bool
```
JLockerman, line 216:
```suggestion
	completeMetricCreation chan struct{}
```
since we never use the value, we might as well. It indicates that it's just a signaling channel, and saves a bit of space.

Issue #96, opened 2020-05-21
Using the regex label matcher ==~ doesn't check right hand operand when using the extra extension


blagojts:
When using TimescaleDB with the Timescale-Prometheus Extra extension, if I write a query with a regex label matcher
```
SELECT *
FROM cpu_usage  u
WHERE labels ? ('namepace' ==~ 'dev')
```

It returns all the rows whose label value matches the 'dev'  regex, the right-hand side label-key ('namespace') is ignored

cevian:
Fixed by #100

PR #100, opened 2020-05-26
Fix bug in extension selector

cevian:
The extension SQL code had a bug in the selector logic where the
function parameter was named `key` but the code in SQL that said
`l.key = key` interpreted that as `l.key=l.key` not `l.key=
parameter.key`. Changed the parameter to `key_to_match` to avoid
this type of confusion and made the parameter names consistent
in the extension and non-extension versions.


cevian:
This fixes #96

PR #101, opened 2020-05-27
Decompress automatically if trying to insert to a compressed chunk

cevian:
This may happen if prometheus is replaying old data.


pkg/pgmodel/pgx.go
```diff
@@ -529,6 +546,47 @@ func (h *insertHandler) flushPending(pending *pendingBuffer) {
 			copyColumns,
 			&pending.batch,
 		)
+		if err != nil {
+			if pgErr, ok := err.(*pgconn.PgError); ok && strings.Contains(pgErr.Message, "insert/update/delete not permitted") {
```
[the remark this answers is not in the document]
cevian: Error code may change so leaving out for now

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -1471,3 +1471,38 @@ CREATE VIEW SCHEMA_INFO.label AS
     ARRAY(SELECT value FROM SCHEMA_CATALOG.label l WHERE l.key = lk.key ORDER BY value)
       AS values
   FROM SCHEMA_CATALOG.label_key lk;
+
+
+
+CREATE PROCEDURE SCHEMA_CATALOG.decompress_chunks_after(metric_table NAME, min_time TIMESTAMPTZ)
```
davidkohn88, line 1477: Maybe add a comment about why we're doing this in a procedure rather than just calling the function? Note the locking behavior?

pkg/pgmodel/pgx.go
```diff
@@ -573,6 +590,49 @@ func (h *insertHandler) flushPending(pending *pendingBuffer) {
 			copyColumns,
 			&pending.batch,
 		)
+		if err != nil {
```
JLockerman, line 593: I think this block should be moved into its own function

cevian: moved part of it

PR #102, opened 2020-05-28
Reduce memory usage on ingest

JLockerman:
This commit reduces memory usage by ensuring WriteRequests and Labels
are freed as soon as they are no longer needed. For WriteRequests, this
is immediately after parseData is called, as at that point we are only
going to used the parsed version. For Labels this is as soon as we have
the SeriesId for the Labels.

This commit also adds a command line flag for async-acks, and the
ability ro report the actual throughput of said acks.

### DIFF
--- a/pkg/pgmodel/bigcache.go
+++ b/pkg/pgmodel/bigcache.go
@@ -7,6 +7,7 @@ package pgmodel
 import (
 	"encoding/binary"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/allegro/bigcache"
@@ -62,7 +63,13 @@ func (m *MetricNameCache) Get(metric string) (string, error) {
 
 // Set stores table name for specified metric.
 func (m *MetricNameCache) Set(metric string, tableName string) error {
-	return m.Metrics.Set(metric, []byte(tableName))
+	// deep copy the strings so the original memory doesn't need to stick around
+	metricBuilder := strings.Builder{}
+	metricBuilder.Grow(len(metric))
+	metricBuilder.WriteString(metric)
+	table := make([]byte, len(tableName))
+	copy(table, tableName)
+	return m.Metrics.Set(metricBuilder.String(), table)
 }
 
 func DefaultCacheConfig() bigcache.Config {
--- a/pkg/pgmodel/insert_ctx.go
+++ b/pkg/pgmodel/insert_ctx.go
@@ -6,51 +6,50 @@ import (
 	"github.com/timescale/timescale-prometheus/pkg/prompb"
 )
 
-var pool = sync.Pool{
+var wrPool = sync.Pool{
 	New: func() interface{} {
-		return new(InsertCtx)
+		return new(prompb.WriteRequest)
 	},
 }
 
-type InsertCtx struct {
-	WriteRequest prompb.WriteRequest
-	Labels       []Labels
+var lPool = sync.Pool{
+	New: func() interface{} {
+		return new(Labels)
+	},
 }
 
-func NewInsertCtx() *InsertCtx {
-	return pool.Get().(*InsertCtx)
+func NewWriteRequest() *prompb.WriteRequest {
+	return wrPool.Get().(*prompb.WriteRequest)
 }
 
-func (t *InsertCtx) clear() {
-	for _, ts := range t.WriteRequest.Timeseries {
+func FinishWriteRequest(wr *prompb.WriteRequest) {
+	for i := range wr.Timeseries {
+		ts := &wr.Timeseries[i]
+		for j := range ts.Labels {
+			ts.Labels[j] = prompb.Label{}
+		}
 		ts.Labels = ts.Labels[:0]
 		ts.Samples = ts.Samples[:0]
+		ts.XXX_unrecognized = nil
 	}
-	t.WriteRequest.Timeseries = t.WriteRequest.Timeseries[:0]
-	t.Labels = t.Labels[:0]
+	wr.Timeseries = wr.Timeseries[:0]
+	wr.XXX_unrecognized = nil
+	wrPool.Put(wr)
 }
 
-func (t *InsertCtx) resetLabels(l *Labels) {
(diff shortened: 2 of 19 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -282,8 +301,16 @@ func (p *pgxInserter) InsertData(rows map[string][]samplesInfo, ctx *InsertCtx)
...
 			workFinished.Wait()
-			ctx.Close()
+			select {
+			case err = <-errChan:
+			default:
+			}
 			close(errChan)
+			if err != nil {
+				log.Error("msg", "error on send", "error", err)
+			} else if p.insertedDatapoints != nil {
```
cevian, line 311: maybe make it clear in this message that this is in the async path and data is being lost

pkg/prompb/remote.pb.go
```diff
@@ -881,38 +881,10 @@ func sozRemote(x uint64) (n int) {
...
-		if fieldNum <= 0 {
-			return fmt.Errorf("proto: WriteRequest: illegal tag %d (wire type %d)", fieldNum, wire)
-		}
-		switch fieldNum {
-		case 1:
-			if wireType != 2 {
-				return fmt.Errorf("proto: wrong wireType = %d for field Timeseries", wireType)
-			}
-			var msglen int
+		iNdEx := 0
```
cevian: changes should be removed

pkg/pgmodel/labels.go
```diff
@@ -23,53 +24,56 @@ type Labels struct {
...
 }
 
 // initLabels intializes labels
-func initLabels(l *Labels) error {
-
-	if !sort.IsSorted(l) {
-		sort.Sort(l)
+func getStr(labels []prompb.Label) (string, error) {
+
+	if !sort.SliceIsSorted(labels, func(i, j int) bool {
```
cevian, line 60: nit: can we format this  in a saner way. Also these functions are identical, just make a function variable?

pkg/pgmodel/insert_ctx.go
```diff
@@ -23,7 +23,7 @@ func FinishWriteRequest(wr *prompb.WriteRequest) {
 			ts.Labels[j] = prompb.Label{}
 		}
 		ts.Labels = ts.Labels[:0]
-		ts.Samples = ts.Samples[:0]
+		ts.Samples = nil
```
cevian, line 26: wonder if this log should be in parseData instead?

pkg/pgmodel/bigcache.go
```diff
@@ -62,7 +63,13 @@ func (m *MetricNameCache) Get(metric string) (string, error) {
 
 // Set stores table name for specified metric.
 func (m *MetricNameCache) Set(metric string, tableName string) error {
-	return m.Metrics.Set(metric, []byte(tableName))
+	// deep copy the strings so the original memory doesn't need to stick around
+	metricBuilder := strings.Builder{}
```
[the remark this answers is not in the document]
JLockerman: The first one has to be a `string` when we're done, the second a `[]byte`. While `copy([]byte, string)` is a thing, `string.Builder` seems to be the best way to get a deep copy of a `string`.

pkg/pgmodel/insert_ctx.go
JLockerman: mmm, both

JLockerman: ehhh, fine, just in `parseData`

pkg/pgmodel/pgx.go
```diff
@@ -186,16 +192,29 @@ func NewPgxIngestorWithMetricCache(c *pgxpool.Pool, cache MetricCache) (*DBInges
...
 		conn:                   conn,
 		metricTableNames:       cache,
 		completeMetricCreation: cmc,
+		asyncAcks:              cfg.AsyncAcks,
+	}
+	if cfg.AsyncAcks && cfg.ReportInterval > 0 {
+		inserter.insertedDatapoints = new(int64)
+		reportInterval := int64(cfg.ReportInterval)
+		go func() {
+			log.Info("msg", fmt.Sprintf("outputting throughpput info once every %ds", reportInterval))
```
antekresic, line 211:
```suggestion
			log.Info("msg", fmt.Sprintf("outputting throughput info once every %ds", reportInterval))
```

pkg/pgmodel/pgx.go
```diff
@@ -509,6 +537,22 @@ func (h *insertHandler) handleReq(req insertDataRequest) bool {
 	return false
 }
 
+func (h *insertHandler) fillKnowSeriesIds(sampleInfos []samplesInfo) (numMissingSeries int) {
```
antekresic, line 540:
```suggestion
func (h *insertHandler) fillKnownSeriesIDs(sampleInfos []samplesInfo) (numMissingSeries int) {
```

pkg/pgmodel/insert_ctx.go
```diff
@@ -6,51 +6,27 @@ import (
...
-	for _, ts := range t.WriteRequest.Timeseries {
+func FinishWriteRequest(wr *prompb.WriteRequest) {
+	for i := range wr.Timeseries {
+		ts := &wr.Timeseries[i]
+		for j := range ts.Labels {
+			ts.Labels[j] = prompb.Label{}
+		}
 		ts.Labels = ts.Labels[:0]
 		ts.Samples = ts.Samples[:0]
+		ts.XXX_unrecognized = nil
```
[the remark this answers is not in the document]
JLockerman: see [here](https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order); slices must contain valid pointers throughout their capacity, so even if the length is truncated, the GC won't collect the pointees, the actual strings, unless the pointers are overwritten.

JLockerman:
@antekresic I'll have to read more of github.com/gogo/protobuf but from what I've seen so far it doesn't seem to support the kind of buffer-reuse we want, so we may have to alter generated files anyway.

PR #106, opened 2020-06-01
Add option to output postgres log, and clean docker image when done

JLockerman:
Just like it says in the title. Add
```
-print-logs true
```
to print the postgres logs in the `end_to_end` tests

### DIFF
--- a/pkg/internal/testhelpers/containers.go
+++ b/pkg/internal/testhelpers/containers.go
@@ -106,8 +106,20 @@ func dbSetup(DBName string, superuser SuperuserStatus) (*pgxpool.Pool, error) {
 	return dbPool, nil
 }
 
+type stdoutLogConsumer struct {
+}
+
+func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
+	if l.LogType == testcontainers.StderrLog {
+		fmt.Print(l.LogType, " ", string(l.Content))
+	} else {
+		fmt.Print(string(l.Content))
+	}
+
+}
+
 // StartPGContainer starts a postgreSQL container for use in testing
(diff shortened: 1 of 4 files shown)


pkg/pgmodel/end_to_end_tests/main_test.go
```diff
@@ -91,6 +100,12 @@ func withDB(t testing.TB, DBName string, f func(db *pgxpool.Pool, t testing.TB))
 	})
 }
 
+func withSuperuser(t testing.TB, DBName string, f func(db *pgxpool.Pool, t testing.TB)) {
```
cevian, line 103: doesn't seem to be used?

pkg/internal/testhelpers/containers.go
```diff
@@ -106,8 +106,20 @@ func dbSetup(DBName string, superuser SuperuserStatus) (*pgxpool.Pool, error) {
 	return dbPool, nil
 }
 
+type stdoutLogConsumer struct {
+}
+
+func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
+	if l.LogType == testcontainers.StderrLog {
+		fmt.Print(l.LogType, " ", string(l.Content))
```
cevian, line 114: shouldn't this be Println (i.e. don't you need to add line breaks)

JLockerman: @cevian `l.Content` always has a trailing `\n`, if we use `Println` we get double-whitespace between the lines

pkg/pgmodel/end_to_end_tests/main_test.go
JLockerman: must've removed that test. Ah well, it was too brittle anyway

PR #107, opened 2020-06-01
Switch to dedicated CopyFrom workers

JLockerman:
This commit switches the way we do inserts into the database. Instead
of each InsertHandler running CopyFrom directly when it finishes a
batch of data, it instead forwards the values to a pool of CopyFrom
workers, who perform the actual insertion. This is intended to provide
a few advantages:
  1. If one metric is very hot, it will be able to continue
     processing new data at the same time as earlier values are
     inserted. Before this commit, individual metrics would experience
     head-of-line blocking in such cases.
  2. This may allow a single worker to handle all inserts in the event
     that the insert-rate is low enough.
  3. This should ease experimentation with batching across metrics.

### DIFF
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -210,14 +211,32 @@ func NewPgxIngestor(c *pgxpool.Pool) (*DBIngestor, error) {
 	return NewPgxIngestorWithMetricCache(c, cache, &Cfg{})
 }
 
+var ConnectionsPerProc = 5
+
 func newPgxInserter(conn pgxConn, cache MetricCache, cfg *Cfg) (*pgxInserter, error) {
 	cmc := make(chan struct{}, 1)
+
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+
+	// we leave one connection per-core for other usages
+	numCopiers := maxProcs*ConnectionsPerProc - maxProcs
+	toCopiers := make(chan copyRequest, numCopiers)
+	for i := 0; i < numCopiers; i++ {
+		go runCopyFrom(conn, toCopiers)
+	}
 
 	inserter := &pgxInserter{
 		conn:                   conn,
 		metricTableNames:       cache,
 		completeMetricCreation: cmc,
 		asyncAcks:              cfg.AsyncAcks,
+		toCopiers:              toCopiers,
 	}
 	if cfg.AsyncAcks && cfg.ReportInterval > 0 {
 		inserter.insertedDatapoints = new(int64)
@@ -250,6 +269,7 @@ type pgxInserter struct {
 	completeMetricCreation chan struct{}
 	asyncAcks              bool
 	insertedDatapoints     *int64
+	toCopiers              chan copyRequest
 }
 
 func (p *pgxInserter) CompleteMetricCreation() error {
@@ -275,6 +295,7 @@ func (p *pgxInserter) Close() {
 		close(value.(chan insertDataRequest))
 		return true
 	})
+	close(p.toCopiers)
 }
 
 func (p *pgxInserter) InsertNewData(rows map[string][]samplesInfo) (uint64, error) {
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -198,11 +199,27 @@ func NewPgxIngestor(c *pgxpool.Pool) (*DBIngestor, error) {
...
 
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+
+	// we leavel one connection per-core for other usages
```
cevian, line 210: what does this mean? It would have been 5 otherwise? The comment makes little sense

pkg/pgmodel/pgx.go
```diff
@@ -557,25 +591,38 @@ func (h *insertHandler) flush() {
...
+	}
+
+	h.toCopiers <- copyRequest{h.pending, h.metricTableName}
+	h.pending = pendingBuffers.Get().(*pendingBuffer)
+}
 
-		_, err = h.conn.CopyFrom(
+func runCopyFrom(conn pgxConn, in chan copyRequest) {
+	for {
+		req, ok := <-in
```
cevian, line 610: where is this channel closed?

pkg/pgmodel/pgx.go
```diff
@@ -557,25 +591,38 @@ func (h *insertHandler) flush() {
...
 }
 
-func (h *insertHandler) flushPending(pending *pendingBuffer) {
-	err := func() error {
-		_, err := h.setSeriesIds(pending.batch.sampleInfos)
-		if err != nil {
-			return err
-		}
+func (h *insertHandler) flushPending() {
+	_, err := h.setSeriesIds(h.pending.batch.sampleInfos)
```
cevian, line 598: hmm, what's the reason this is done here and not by the copiers?

pkg/pgmodel/pgx.go
```diff
@@ -477,9 +510,10 @@ func runInserterRoutine(conn pgxConn, input chan insertDataRequest, metricName s
 	handler := insertHandler{
 		conn:            conn,
 		input:           input,
-		pending:         &pendingBuffer{make([]insertDataTask, 0), NewSampleInfoIterator()},
+		pending:         pendingBuffers.Get().(*pendingBuffer),
```
cevian, line 513: ` pendingBuffers.Get().(*pendingBuffer)` should be in it's own function

pkg/pgmodel/pgx.go
```diff
@@ -557,25 +591,38 @@ func (h *insertHandler) flush() {
...
-			pgx.Identifier{dataSchema, h.metricTableName},
+			pgx.Identifier{dataSchema, req.table},
 			copyColumns,
-			&pending.batch,
+			&req.data.batch,
 		)
-		return err
-	}()
+		req.data.finish(err)
+		pendingBuffers.Put(req.data)
```
cevian, line 621: let's make a proper function for this line using our FinishXXX naming convention

pkg/pgmodel/pgx.go
JLockerman: @cevian yup, otherwise we'd have 5 workers, one per-connection. I'm concerned that if we do that new metrics would starve when trying to create tables. (alternatively, we could create `5*maxprocs - 1` workers, and only leave one for other tasks)

pkg/pgmodel/pgx.go
JLockerman: @cevian the series cache needs to be per-metric-inserter, while the copiers are multiplexed across all inserters.

pkg/pgmodel/pgx.go
cevian: I see that fine. But lets make 5 a variable somewhere that is used in both setting pool size and this. Right now there is nothing keeping the two values in sync.

pkg/pgmodel/pgx.go
JLockerman: @cevian I really don't think it should be: it should never be used outside this file, and it's just adding a gratuitous layer of indirection. If we did more work than just `Put`ing in `pendingBuffers` that would be one thing, but `data.finish` is _not_ that (and, incidentally, is now renamed to `reportResults`).

pkg/pgmodel/pgx.go
```diff
@@ -210,14 +211,32 @@ func NewPgxIngestor(c *pgxpool.Pool) (*DBIngestor, error) {
...
+	maxProcs := runtime.GOMAXPROCS(-1)
+	if maxProcs <= 0 {
+		maxProcs = runtime.NumCPU()
+	}
+	if maxProcs <= 0 {
+		maxProcs = 1
+	}
+
+	// we leave one connection per-core for other usages
+	numCopiers := maxProcs*ConnectionsPerProc - maxProcs
```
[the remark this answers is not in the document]
JLockerman: please ignore the fact that I cannot do basic arithmetic...

PR #110, opened 2020-06-03
Prevent concurrent table creation errors for long metric name

cevian:
Previously, when two metrics with really long names were created
concurrently, postgres could throw errors due to name collision
violations when it created constraint names as well as type names
for the type corresponding to the table.

Constraint names was fixed by explicitly assigning a unique constraint
name instead of having PG autogenerate one.

Types are a bit trickier. Turns out for every table X, PG assigns
an array type '_X' (makeArrayTypeName() in PG source). That can
also collide and cause erroors. The fix was for us to change the
max length of the table names we geneate from 63 chars to 62 chars,
thus reserving one char for the underscore.


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -307,38 +307,53 @@ CREATE TRIGGER make_metric_table_trigger
 
 -- Return a table name built from a full_name and a suffix.
 -- The full name is truncated so that the suffix could fit in full.
--- name size will always be exactly 63 chars.
+-- name size will always be exactly final_length chars.
 CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.pg_name_with_suffix(
-        full_name text, suffix text)
+        full_name text, suffix text, final_length smallint)
```
davidkohn88, line 312: Not a huge fan of this being a variable. I think it might be better to do the macro like thing we're doing with schemas to replace that?

PR #113, opened 2020-06-08
Add read only version to get key position, clarify comments

davidkohn88:


pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -797,7 +817,7 @@ BEGIN
 
    EXECUTE format($query$
     WITH CTE AS (
-        SELECT SCHEMA_PROM.label_array($1)
+        SELECT SCHEMA_PROM.get_or_create_label_array($1)
```
cevian, line 820: should the parent function here `SCHEMA_PROM.series_id(` be renamed and go into SCHEMA_CATALOG?

cevian: it's a GoC right?

pkg/pgmodel/migrations/sql/1_base_schema.up.sql
```diff
@@ -834,7 +854,7 @@ BEGIN
 
    EXECUTE format($query$
     WITH CTE AS (
-        SELECT SCHEMA_PROM.label_array($1, $2, $3)
+        SELECT SCHEMA_PROM.get_or_create_label_array($1, $2, $3)
```
cevian, line 857: should parent be renamed `get_or_create_series_id_for_kv_array`

PR #117, opened 2020-06-10
Validate Request headers

JLockerman:
Validate some HTTP headers when receiving a remote request to prevent unmarshalling of irrelevant data.

### DIFF
--- a/pkg/api/read.go
+++ b/pkg/api/read.go
@@ -1,21 +1,29 @@
 package api
 
 import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
 	"github.com/timescale/timescale-prometheus/pkg/log"
 	"github.com/timescale/timescale-prometheus/pkg/pgmodel"
 	"github.com/timescale/timescale-prometheus/pkg/prompb"
-	"io/ioutil"
-	"net/http"
-	"time"
 )
 
 func Read(reader pgmodel.Reader, metrics *Metrics) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !validateReadHeaders(w, r) {
+			metrics.InvalidReadReqs.Inc()
+			return
+		}
+
 		compressed, err := ioutil.ReadAll(r.Body)
 		if err != nil {
-			log.Error("msg", "Read error", "err", err.Error())
+			log.Error("msg", "Read header validation error", "err", err.Error())
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
@@ -67,4 +75,38 @@ func Read(reader pgmodel.Reader, metrics *Metrics) http.Handler {
 			return
 		}
 	})
+}
+
+func validateReadHeaders(w http.ResponseWriter, r *http.Request) bool {
+	// validate headers from https://github.com/prometheus/prometheus/blob/2bd077ed9724548b6a631b6ddba48928704b5c34/storage/remote/client.go
+	if r.Method != "POST" {
+		buildReadError(w, fmt.Sprintf("HTTP Method %s instead of POST", r.Method))
+		return false
+	}
+
+	if !strings.Contains(r.Header.Get("Content-Encoding"), "snappy") {
+		buildReadError(w, fmt.Sprintf("non-snappy compressed data got: %s", r.Header.Get("Content-Encoding")))
+		return false
+	}
+
+	if r.Header.Get("Content-Type") != "application/x-protobuf" {
+		buildReadError(w, "non-protobuf data")
+		return false
+	}
+
+	remoteReadVersion := r.Header.Get("X-Prometheus-Remote-Read-Version")
+	if remoteReadVersion == "" {
+		err := "missing X-Prometheus-Remote-Read-Version"
+		log.Warn("msg", "Read header validation error", "err", err)
+	} else if !strings.HasPrefix(remoteReadVersion, "0.1.") {
+		buildReadError(w, fmt.Sprintf("unexpected Remote-Read-Version %s, expected 0.1.X", remoteReadVersion))
+		return false
+	}
+
+	return true
+}
+
+func buildReadError(w http.ResponseWriter, err string) {
+	log.Error("msg", "Read header validation error", "err", err)
+	http.Error(w, err, http.StatusBadRequest)
 }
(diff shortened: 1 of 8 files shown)


pkg/api/read.go
```diff
@@ -68,3 +74,40 @@ func Read(reader pgmodel.Reader, metrics *Metrics) http.Handler {
 		}
 	})
 }
+
+func validateReadHeaders(w http.ResponseWriter, r *http.Request) bool {
+	// validate headers from https://github.com/prometheus/prometheus/blob/2bd077ed9724548b6a631b6ddba48928704b5c34/storage/remote/client.go
+	if r.Method != "POST" {
```
cevian, line 80: According to our Router this also supports Get, which semantically makes sense. Lets add that as a valid option

pkg/api/read.go
```diff
@@ -68,3 +74,40 @@ func Read(reader pgmodel.Reader, metrics *Metrics) http.Handler {
...
+		err := fmt.Sprintf("HTTP Method %s instead of POST", r.Method)
+		log.Error("msg", "Read error", "err", err)
```
cevian, line 82: let's change all the msgs to `Read header validation error`

pkg/api/write.go
```diff
@@ -104,3 +110,44 @@ func getCounterValue(counter prometheus.Counter) float64 {
 	}
 	return dtoMetric.GetCounter().GetValue()
 }
+
+func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
+	// validate headers from https://github.com/prometheus/prometheus/blob/2bd077ed9724548b6a631b6ddba48928704b5c34/storage/remote/client.go
+	if r.Method != "POST" {
+		err := fmt.Sprintf("HTTP Method %s instead of POST", r.Method)
+		log.Error("msg", "Write error", "err", err)
```
cevian, line 118: change all the msg to `Write header validation error`

pkg/api/write.go
```diff
@@ -104,3 +110,44 @@ func getCounterValue(counter prometheus.Counter) float64 {
...
+	if r.Header.Get("Content-Type") != "application/x-protobuf" {
+		err := "non-protobuf data"
+		log.Error("msg", "Write error", "err", err)
+		http.Error(w, err, http.StatusBadRequest)
+		return false
+	}
+
+	remoteWriteVersion := r.Header.Get("X-Prometheus-Remote-Write-Version")
+	if remoteWriteVersion == "" {
+		err := "non-Remote-Write"
```
cevian, line 139:
```suggestion
		err := "Missing X-Prometheus-Remote-Write-Version header"
```

pkg/api/write.go
```diff
@@ -30,6 +31,11 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
 			return
 		}
 
+		if !validateWriteHeaders(w, r) {
+			metrics.LeaderGauge.Set(0)
```
cevian, line 35: This is wrong. Validate headers do not change leader status

pkg/api/read.go
JLockerman: GET request bodies are supposed to be ignored; I think our current routing is incorrect...

pkg/api/write.go
```diff
@@ -18,6 +19,12 @@ import (
 
 func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		if !validateWriteHeaders(w, r, metrics) {
```
cevian, line 23: error here

pkg/api/read.go
```diff
@@ -68,3 +76,40 @@ func Read(reader pgmodel.Reader, metrics *Metrics) http.Handler {
...
+		log.Error("msg", "Read header validation error", "err", err)
+		http.Error(w, err, http.StatusBadRequest)
+		return false
+	}
+
+	remoteReadVersion := r.Header.Get("X-Prometheus-Remote-Read-Version")
+	if remoteReadVersion == "" {
+		err := "missing X-Prometheus-Remote-Read-Version"
+		log.Warn("msg", "Read header validation error", "err", err)
+	} else if !strings.HasPrefix(remoteReadVersion, "0.1") {
```
cevian, line 107:
```suggestion
	} else if !strings.HasPrefix(remoteReadVersion, "0.1.") {
```

pkg/api/write.go
```diff
@@ -104,3 +111,44 @@ func getCounterValue(counter prometheus.Counter) float64 {
...
+
+	remoteWriteVersion := r.Header.Get("X-Prometheus-Remote-Write-Version")
+	if remoteWriteVersion == "" {
+		err := "Missing X-Prometheus-Remote-Write-Version header"
+		log.Error("msg", "Write header validation error", "err", err)
+		http.Error(w, err, http.StatusBadRequest)
+		return false
+	}
+
+	if !strings.HasPrefix(remoteWriteVersion, "0.1") {
```
cevian, line 146:
```suggestion
	if !strings.HasPrefix(remoteWriteVersion, "0.1.") {
```

pkg/api/write.go
cevian, line 23: please add a comment as to why we validate before checking leadership

pkg/api/read_test.go
```diff
@@ -75,9 +78,9 @@ func TestRead(t *testing.T) {
 			}
 			handler := Read(mockReader, metrics)
 
-			test := GenerateHandleTester(t, handler)
+			test := GenerateReadHandleTester(t, handler)
```
antekresic, line 81: Should we add test cases to check for invalid headers and method?

pkg/api/write.go
```diff
@@ -104,3 +113,44 @@ func getCounterValue(counter prometheus.Counter) float64 {
 	}
 	return dtoMetric.GetCounter().GetValue()
 }
+
+func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
+	// validate headers from https://github.com/prometheus/prometheus/blob/2bd077ed9724548b6a631b6ddba48928704b5c34/storage/remote/client.go
+	if r.Method != "POST" {
+		err := fmt.Sprintf("HTTP Method %s instead of POST", r.Method)
+		log.Error("msg", "Write header validation error", "err", err)
+		http.Error(w, err, http.StatusBadRequest)
```
antekresic, line 122: Seems like these two things are done a lot of times, maybe we could refactor it into a separate function and cleanup the code.

antekresic (changes requested):
Looks good, I think we just need to add test cases that verify we are handling invalid headers.

antekresic (approved):

PR #119, opened 2020-06-10
Add promql endpoint pushdown

cevian:
This PR Pushes some query processing logic from the PromQL engine into the database.

We modify the interaction between the query engine and the storage layer so that the storage layer now takes in the node-branch it's evaluating, and computes as much of the branch as it's capable of, returning the top-most node it computed. The engine then walks up the execution path, starting at that node, and performs the remainder of the computation.

This PR puts in the framework to accomplish this and tests it with one pushdown function `delta`, which serves as a proof of concept. Further PRs will be used to push down other functions and processing nodes.

The approach used here is limited in certain ways:
1) It is not meant to push down past branches that split. This may or may not be fundamental but is out of scope since it's rarely used.
2) We don't push down queries with subqueries. Again this is rarely used. Unsure how difficult this would be to implement.

### DIFF
--- a/pkg/pgmodel/end_to_end_tests/query_integration_test.go
+++ b/pkg/pgmodel/end_to_end_tests/query_integration_test.go
@@ -2,6 +2,7 @@ package end_to_end_tests
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -14,10 +15,15 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
+	"github.com/prometheus/prometheus/util/testutil"
 	"github.com/timescale/timescale-prometheus/pkg/internal/testhelpers"
+	"github.com/timescale/timescale-prometheus/pkg/log"
 	"github.com/timescale/timescale-prometheus/pkg/prompb"
+	"github.com/timescale/timescale-prometheus/pkg/promql"
+	"github.com/timescale/timescale-prometheus/pkg/query"
 
 	. "github.com/timescale/timescale-prometheus/pkg/pgmodel"
 )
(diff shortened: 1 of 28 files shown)


pkg/pgmodel/end_to_end_tests/query_integration_test.go
```diff
@@ -1124,3 +1130,82 @@ func generateLargeTimeseries() []prompb.TimeSeries {
...
+		// Getting a read-only connection to ensure read path is idempotent.
+		readOnly := testhelpers.GetReadOnlyConnection(t, *testDatabase)
+		defer readOnly.Close()
+
+		var tester *testing.T
+		var ok bool
+		if tester, ok = t.(*testing.T); !ok {
+			t.Fatalf("Cannot run test, not an instance of testing.T")
+			return
+		}
```
JLockerman, line 1185: is it really worth passing the `testing.TB` out of `withDB` if we're just going to have to do stuff like this? Realistically it's a bug if `withDB` ever swaps out the `testing.<T/B>` for anything else, so we can just capture the `testing.<T/B>` from the environment

pkg/pgmodel/pgx.go
```diff
@@ -904,10 +904,10 @@ func (q *pgxQuerier) Query(query *prompb.Query) ([]*prompb.TimeSeries, error) {
 	return results, nil
 }
 
-func (q *pgxQuerier) getResultRows(startTimestamp int64, endTimestamp int64, matchers []*labels.Matcher) ([]pgx.Rows, error) {
+func (q *pgxQuerier) getResultRows(startTimestamp int64, endTimestamp int64, hints *storage.SelectHints, path []parser.Node, matchers []*labels.Matcher) ([]pgx.Rows, parser.Node, error) {
 	metric, cases, values, err := buildSubQueries(matchers)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
```
JLockerman, line 910: I wonder if it would be better to start using named-`return` so we can just assign the return values and use `return` without parameters; it might make these diffs cleaner

pkg/pgmodel/query_builder.go
```diff
@@ -245,6 +245,101 @@ func buildTimeseriesBySeriesIDQuery(filter metricTimeRangeFilter, series []Serie
...
+	restOfQueryParams []interface{}
+}
+
+func (t *queryFinalizer) Finalize() (string, []interface{}, error) {
+	fullQuery := `SELECT (key_value_array(s.labels)).*, ` + t.timeClause + `, ` + t.valueClause + t.restOfQuery
+	newParams := append(t.timeParams, t.valueParams...)
+	return fillInParameters(fullQuery, t.restOfQueryParams, newParams...)
+}
+
+/* The path is the list if ancestors (direct parent last) returned node is the last node processed by the pushdown */
```
JLockerman, line 297:
```suggestion
/* The path is the list of ancestors (direct parent last) returned node is the most-ancestral node processed by the pushdown */
```
?

pkg/pgmodel/query_builder.go
```diff
@@ -296,7 +296,7 @@ func (t *queryFinalizer) Finalize() (string, []interface{}, error) {
 
 /* The path is the list if ancestors (direct parent last) returned node is the last node processed by the pushdown */
 func getQueryFinalizer(otherClauses string, values []interface{}, hints *storage.SelectHints, path []parser.Node) (*queryFinalizer, parser.Node, error) {
-	if path != nil && hints != nil && len(path) >= 2 && !partOfSubquery(path) {
+	if ExtensionIsInstalled && path != nil && hints != nil && len(path) >= 2 && !partOfSubquery(path) {
```
JLockerman, line 299: I don't love this...
though it should be correct

pkg/pgmodel/query_builder.go
```diff
@@ -245,6 +245,101 @@ func buildTimeseriesBySeriesIDQuery(filter metricTimeRangeFilter, series []Serie
...
+	}
+
+	query, newValues, err := qf.Finalize()
+	if err != nil {
+		return "", nil, nil, err
+	}
+	return query, newValues, node, nil
+}
+
+// cumulativeSubqueryOffset returns the sum of range and offset of all subqueries in the path.
```
JLockerman, line 271: What does this comment mean?

pkg/pgmodel/pgx_test.go
```diff
@@ -855,7 +855,7 @@ func TestPGXQuerierQuery(t *testing.T) {
 				},
 			},
 			sqlQueries: []string{`SELECT table_name FROM _prom_catalog.get_metric_table_name_if_exists($1)`,
-				`SELECT (key_value_array(s.labels)).*, array_agg(m.time ORDER BY time), array_agg(m.value ORDER BY time)
+				`SELECT (key_value_array(s.labels)).*, array_agg(m.time ORDER BY time) as time_array, array_agg(m.value ORDER BY time)
```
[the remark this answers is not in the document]
cevian: this is an alias not a type

pkg/pgmodel/pgx.go
cevian: I don't think for this PR

cevian:
Statements are trees. The leaves fetch data bringing it up to the leaves. Not sure if this would work for pushdowns higher than a node split, but I believe that's out of scope for now.

cevian:
@JLockerman updated the PR description

davidkohn88:
THe only question I had was whether we needed to check for the extension's presence before doing this? But that might have been in a different commit...

PR #120, opened 2020-06-10
Add prometheus `rate` functions

JLockerman:
This commit adds implementations of the prometheus `rate`, `increase` and `delta`
functions to the timescale_prometheus_extra extension. These features are added
using a rust addition to the extension, based on our [timescale-extension-utils](https://github.com/timescale/timescale-extension-utils-rs)
library.

The functions are a fairly direct post of the relevant prometheus code

Note that while these functions "work" in the sense that the produce output and
don't crash, they have not been extensively tested against prometheus yet, and
may still have behavioral differences


extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,67 @@ AS $func$
...
+LANGUAGE SQL VOLATILE SECURITY DEFINER;
+
+----------------------------  aggregation functions ----------------------------
+
+CREATE FUNCTION @extschema@.gp_d_t(state internal, lowest_time timestamptz,
+    greatest_time timestamptz, step_size bigint, window_size bigint,
+    timev timestamptz, val double precision
+    )
+RETURNS internal AS '$libdir/timescale_prometheus_extra', 'gapfill_delta_transition'
+LANGUAGE C STABLE PARALLEL SAFE;
```
cevian, line 171: I think this and all the other functions are IMMUTABLE not STABLE. They don't read the state of the db other than the inputs

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,67 @@ AS $func$
...
+CREATE FUNCTION @extschema@.gp_i_t(state internal, lowest_time timestamptz,
+    greatest_time timestamptz, step_size bigint, window_size bigint,
+    timev timestamptz, val double precision
+    )
+RETURNS internal AS '$libdir/timescale_prometheus_extra', 'gapfill_increase_transition'
+LANGUAGE C STABLE PARALLEL SAFE;
+
+CREATE FUNCTION @extschema@.gp_d_f(state internal)
+RETURNS DOUBLE PRECISION[]
+AS '$libdir/timescale_prometheus_extra', 'gapfill_delta_final' LANGUAGE C;
```
cevian, line 189: also immutable and parallel safe

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,67 @@ AS $func$
...
+
+CREATE AGGREGATE @extschema@.gapfill_delta(lowest_time timestamptz,
+    greatest_time TIMESTAMPTZ,
+    step_size BIGINT,
+    window_size BIGINT,
```
cevian, line 194: in this public api may want to call it range_size to be compatible with prom terminology

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,67 @@ AS $func$
...
+    timev TIMESTAMPTZ,
+    val DOUBLE PRECISION)
+(
+    sfunc=@extschema@.gp_d_t,
+    stype=internal,
+    finalfunc=@extschema@.gp_d_f
+);
+
+CREATE AGGREGATE @extschema@.gapfill_rate(lowest_time timestamptz,
```
cevian, line 203: let's discuss these argument name. prom is start, end, step, range, time ,value

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+pg_agg!{
+    // prom divides time into no-sliding windows of fixed size, e.g.
+    // |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |
+    // we take the first and last values in that bucket and uses `last-first` as the
+    // value for that bucket.
+    //  | a b c d e | f g h i | j   k |   m    |
+    //  |   e - a   |  i - f  | k - j | <null> |
+    pub fn gapfill_delta_transition(
+        state: Option<Pox<GapfillDeltaTransition>>,
+        lowest_time: TimestampTz,
```
cevian, line 34: so we have to discuss the api we want. In prom start is before the the - step and here after. I know we discussed this before but let's reconsider after talking more broadly about the sql api

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        let mut state = state.unwrap_or_else(|| {
+            let state: Pox<_> = GapfillDeltaTransition::new(
+                lowest_time,
+                greatest_time,
+                window_size,
+                step_size,
+                false,
+                false,
+            ).into();
```
cevian, line 53: what's into again?

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,67 @@ AS $func$
...
+    greatest_time timestamptz, step_size bigint, window_size bigint,
+    timev timestamptz, val double precision
+    )
+RETURNS internal AS '$libdir/timescale_prometheus_extra', 'gapfill_increase_transition'
+LANGUAGE C STABLE PARALLEL SAFE;
+
+CREATE FUNCTION @extschema@.gp_d_f(state internal)
+RETURNS DOUBLE PRECISION[]
+AS '$libdir/timescale_prometheus_extra', 'gapfill_delta_final' LANGUAGE C;
+
```
cevian, line 190: should document ordering assumptions about each of these aggregates in a comment

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            step_size: step_size*USECS_PER_MS,
+            greatest_time,
+            is_counter,
+            is_rate,
+        }
+    }
+
+    fn add_data_point(&mut self, time: TimestampTz, val: f64) {
+        // skip stale NaNs
+        const STALE_NAN: u64 = 0x7ff0000000000002;
```
cevian, line 168: I think this should be a global const somewhere, no?

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        let current_window_min = self.current_window_min;
+        while self.window.front().map(|(time, _)| *time < current_window_min).unwrap_or(false) {
+            self.window.pop_front();
+        }
+    }
+
+    //based on extrapolatedRate
+    // https://github.com/prometheus/prometheus/blob/e5ffa8c9a08a5ee4185271c8c26051ddc1388b7a/promql/functions.go#L59
+    fn add_delta_for_current_window(&mut self) {
```
cevian, line 198: this function name makes less sense when it's used for delta, rate, increase etc. Maybe call it add_extrapolated_rate

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            current_window_max: lowest_time + window_size*USECS_PER_MS,
+            current_window_min: lowest_time,
+            step_size: step_size*USECS_PER_MS,
+            greatest_time,
+            is_counter,
+            is_rate,
+        }
+    }
+
+    fn add_data_point(&mut self, time: TimestampTz, val: f64) {
```
cevian, line 166: do we check somewhere that data is coming in in the correct order wrt time?

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    fn flush_current_window(&mut self) {
+        self.add_delta_for_current_window();
+
+        self.current_window_min += self.step_size;
+        self.current_window_max += self.step_size;
+
+        let current_window_min = self.current_window_min;
+        while self.window.front().map(|(time, _)| *time < current_window_min).unwrap_or(false) {
```
cevian, line 191: not sure I follow this ... can discuss tomorrow during walkthru

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            construct_md_array(
+                self.deltas.as_mut_ptr(),
+                self.nulls.as_mut_ptr(),
+                1,
+                &mut (self.nulls.len() as _),
+                &mut 1,
+                FLOAT8OID,
+                size_of::<f64>() as _,
+                FLOAT8PASSBYVAL != 0,
+                'd' as u8 as _,
```
cevian, line 290: what does this `as _` do

extension/src/support.c
```diff
@@ -16,6 +16,21 @@ PG_MODULE_MAGIC;
 #endif
 
 PG_FUNCTION_INFO_V1(make_call_subquery_support);
+PG_FUNCTION_INFO_V1(gapfill_delta_transition);
+PG_FUNCTION_INFO_V1(gapfill_increase_transition);
+PG_FUNCTION_INFO_V1(gapfill_rate_transition);
+PG_FUNCTION_INFO_V1(gapfill_delta_final);
+
+/* linker hack to make sure the rust code is actually linked in */
```
cevian, line 24: why do we need this?

cevian: also probably in wrong file

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    pub fn gapfill_delta_final(state: Option<Pox<GapfillDeltaTransition>>) -> Option<*mut ArrayType> {
+        state.map(|mut s| s.to_pg_array())
+    }
+}
+
+#[derive(Debug)]
+struct GapfillDeltaTransition {
+    window: VecDeque<(TimestampTz, f64)>,
+    deltas: Vec<Datum>,
```
cevian, line 130: lets leave comment that the delta and null vec are not same size. Rather the delta vec is of size = # items not null.

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            extrapolate_to_interval += avg_duration_between_samples / 2.0
+        }
+
+        result_val = result_val * (extrapolate_to_interval / sampled_interval);
+
+        if self.is_rate {
+            //TODO is this right?
+            result_val = result_val
+                / ((self.current_window_max - self.current_window_min) /USECS_PER_SEC) as f64
```
cevian, line 269: I think this is right, but why not just store window_size/range in the transition state to make this more explicit and more like the go code?

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+#[derive(Debug)]
+struct GapfillDeltaTransition {
+    window: VecDeque<(TimestampTz, f64)>,
+    deltas: Vec<Datum>,
+    nulls: Vec<bool>,
+    current_window_max: TimestampTz,
+    current_window_min: TimestampTz,
+    step_size: TimestampTz,
```
cevian, line 134: this cannot be a timestamptz. It's an interval

cevian: maybe have a Nanosecond type?

extension/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+                }
+                last_value = *sample
+            }
+        }
+
+        let (latest_time, latest_val)= self.window.back().cloned().unwrap();
+        let (earliest_time, earliest_val) = self.window.front().cloned().unwrap();
+        let mut result_val = latest_val - earliest_val + counter_correction;
+
+        let mut duration_to_start = (earliest_time - self.current_window_min) as f64
```
cevian, line 220: can we indicate this is seconds either in comments or through naming convention

extension/src/lib.rs
JLockerman: arbitrary lossless type-conversion, in this case `Pox`ing the value

extension/src/lib.rs
JLockerman: it is 😁

extension/src/lib.rs
JLockerman: I think technically they're all kinds of deltas, or alternatively all kinds of rates, so why not

extension/src/lib.rs
JLockerman: nope, do want me to?

extension/src/lib.rs
JLockerman: if's basically
```python
while window.front() exists and window.front() < current_window_min
```

extension/src/lib.rs
JLockerman: converts it to whatever type is expected in that context, in this case the first one casts from `usize` to postgres's `Size` and the second one casts from `u8` to `c_char`

extension/src/support.c
JLockerman: postgres passes a flag to the linker asking it to garbage collect all unused functions which sees that our rust functions are never used by this library and deletes them. This makes the functions actually used so they stick around. There may be some linker-hackery I can do to mark the entire archive we pass in as used, but I want to be careful that we don't accidentally include too much.

extension/src/lib.rs
JLockerman: wouldn't it be Microseconds>

extension/src/lib.rs
JLockerman: comment I think

extension/src/lib.rs
cevian: yes pls.

extension/src/lib.rs
JLockerman: or we can use newtypes if you really want

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,78 @@ AS $func$
...
-LANGUAGE SQL VOLATILE SECURITY DEFINER;
\ No newline at end of file
+LANGUAGE SQL VOLATILE SECURITY DEFINER;
+
+----------------------------  aggregation functions ----------------------------
+
+CREATE FUNCTION @extschema@.prom_delta_transition(state internal, lowest_time timestamptz,
+    greatest_time timestamptz, step bigint, range bigint,
+    timev timestamptz, val double precision
+    )
```
cevian, line 169: nit: no newline before `)`. Sorry my OCD was kicking in.

extension/sql/timescale-prometheus.sql
```diff
@@ -159,4 +159,78 @@ AS $func$
...
+
+
+-- implementation of prometheus delta function
+-- for proper behavior the input must be ORDER BY timev
+CREATE AGGREGATE @extschema@.prom_delta(
+    lowest_time timestamptz,
+    greatest_time TIMESTAMPTZ,
+    step BIGINT,
+    range BIGINT,
+    timev TIMESTAMPTZ,
```
cevian, line 200: nit maybe data_time and data_value as arg parameters. `timev` is just weird and this is public-facing api

extension/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    // we take the first and last values in that bucket and uses `last-first` as the
+    // value for that bucket.
+    //  | a b c d e | f g h i | j   k |   m    |
+    //  |   e - a   |  i - f  | k - j | <null> |
+    pub fn gapfill_delta_transition(
+        state: Option<Pox<GapfillDeltaTransition>>,
+        lowest_time: TimestampTz,
+        greatest_time: TimestampTz,
+        step_size: Milliseconds, // `prev_now - step` is where the next window starts
+        window_size: Milliseconds, // the size of a window to delta over
```
cevian, line 40: nit: worth renaming to range for consistency?

extension/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        Some(state)
+    }
+
+    pub fn gapfill_rate_transition(
+        state: Option<Pox<GapfillDeltaTransition>>,
+        lowest_time: TimestampTz,
+        greatest_time: TimestampTz,
+        step_size: Milliseconds, // `prev_now - step` is where the next window starts
+        window_size: Milliseconds, // the size of a window to delta over
```
cevian, line 70:
```suggestion
        window_size: Milliseconds, // the size of a window to rate over
```

extension/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    // |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |
+    // we take the first and last values in that bucket and uses `last-first` as the
+    // value for that bucket.
+    //  | a b c d e | f g h i | j   k |   m    |
+    //  |   e - a   |  i - f  | k - j | <null> |
+    pub fn gapfill_delta_transition(
+        state: Option<Pox<GapfillDeltaTransition>>,
+        lowest_time: TimestampTz,
+        greatest_time: TimestampTz,
+        step_size: Milliseconds, // `prev_now - step` is where the next window starts
```
cevian, line 39: not sure what that comment means

extension/src/lib.rs
cevian: I still don't think this location makes sense

extension/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        if duration_to_end < extrapolation_threshold {
+            extrapolate_to_interval += duration_to_end
+        } else {
+            extrapolate_to_interval += avg_duration_between_samples / 2.0
+        }
+
+        result_val = result_val * (extrapolate_to_interval / sampled_interval);
+
+        if self.is_rate {
+            let range = (self.current_window_max - self.current_window_min) as f64
```
cevian, line 276: I think this is right, but why not just store range in the transition state to make this more explicit and more like the go code?

extension/src/support.c
```diff
@@ -16,6 +16,21 @@ PG_MODULE_MAGIC;
 #endif
 
 PG_FUNCTION_INFO_V1(make_call_subquery_support);
+PG_FUNCTION_INFO_V1(gapfill_delta_transition);
```
cevian, line 19: This stuff is in the wrong .c file. It has nothing to do with support.

cevian: I'd try to fix this not to need this too...

extension/sql/timescale-prometheus.sql
JLockerman: I philosophically disagree 😉 
I think with multi-line arguments `( )` should be formatted as you would `{ }`

extension/sql/timescale-prometheus.sql
JLockerman: bleh, don't love that either, do you prefer `data_time` or `"time"`?

extension/src/lib.rs
JLockerman: It's only used and relevant in exactly this location; if we use it more places we should move it to the LUB of those locations, but that's not relevant right now.

extension/src/support.c
JLockerman: hmm, I'm a bit worried that DLL export won't be set correctly if I move it...

JLockerman: also, I don't think I can generate the function right now, `concat_idents!` is unstable

extension/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+            extrapolate_to_interval += duration_to_end
+        } else {
+            extrapolate_to_interval += avg_duration_between_samples / 2.0
+        }
+
+        result_val = result_val * (extrapolate_to_interval / sampled_interval);
+
+        if self.is_rate {
+            result_val = result_val / self.range as f64
+        }
```
antekresic, line 279: I'm used to writing code like this but the compiler suggests to use an if expression. I _assume_ this is to enable possible compiler optimizations.

JLockerman: @antekresic nah, they're [completely equivalent](https://godbolt.org/z/-h2744), it's just a stylistic thing. Clippy (which is what I assume is firing here) has a bunch of style lints, some of which aren't even normative. Since this is a direct port of the go code, I think it's best to maintain similarly with the original over style

PR #122, opened 2020-06-11
Add promql endpoint unit tests

blagojts:


pkg/api/query_range_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestRangedQuery(t *testing.T) {
+	_ = log.Init("debug")
+	testCases := []struct {
+		desc        string
```
antekresic, line 22: I would rename this to `name` to keep consistent with the rest of the test cases, since it is used as name param in `t.Run`.

PR #124, opened 2020-06-15
Add PromQL query and query_range support

blagojts:
By using the Prometheus QueryEngine we enable
Timescale-Prometheus to return results for query and query_range
PromQL requests.


pkg/pgmodel/pgx.go
```diff
@@ -64,7 +68,8 @@ type MetricCache interface {
 }
 
 type pgxConnImpl struct {
-	conn *pgxpool.Pool
+	conn     *pgxpool.Pool
+	readHist prometheus.ObserverVec
```
JLockerman, line 72: Since this only seems to be needed on the Query side, can we add it to some reader-only state instead?

pkg/pgmodel/pgx.go
```diff
@@ -677,19 +690,64 @@ func (q *pgxQuerier) HealthCheck() error {
 	return nil
 }
 
+func (q *pgxQuerier) Select(mint int64, maxt int64, sortSeries bool, hints *storage.SelectHints, ms ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
```
JLockerman, line 693: Sinve all of this stuff seems to be non-pgx specific, should it go in `querier.go`, or another file, instead?

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,238 @@
...
+			found = true
+			break
+		}
+		p.cur++
+	}
+
+	return found
+}
+
+func (p *pgxSeriesIterator) getCurrTs() int64 {
```
JLockerman, line 208: should this return a `model.Time`?
(btw, I have a version with what I believe to be correct overflow handling [here](https://github.com/JLockerman/timescale-prometheus/blob/6aa57869548259701bc601ff856ef0850a1e9244/pkg/pgmodel/pgx.go#L92-L115))

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,238 @@
...
+	rp := p.cur*12 + 4
+	return (microsecFromUnixEpochToY2K + int64(binary.BigEndian.Uint64(p.samples[2][rp:]))) / 1000
+}
+
+func (p *pgxSeriesIterator) getCurrVal() float64 {
+	rp := p.cur*12 + 4
+	return math.Float64frombits(binary.BigEndian.Uint64(p.samples[3][rp:]))
```
JLockerman, line 215: I think we will need to handle NULLs correctly, which makes me want to use a custom type with `DecodeBinary` more...

JLockerman:
I also don't like how the decoding code is using arbitrary-seeming array-indexes everywhere; at minimum I'd like documentation for expected layout, and ideally I'd like pgx to error if the postgres types don't match the type we expect

cevian:
This also needs a rebase

pkg/pgmodel/series_set.go
blagojts: implements a prometheus defined interface

antekresic:
@cevian @JLockerman I've updated the PR with your suggestion. I didn't include the overflow checks as suggested but we really care, can add those as well. 

Can you take another look? Thanks!

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,196 @@
...
+			found = true
+			break
+		}
+		p.cur++
+	}
+
+	return found
+}
+
+func (p *pgxSeriesIterator) getTs() int64 {
```
cevian, line 158: Nit: Please add comment that this returns millisecond or return model.Time

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,196 @@
...
+		return math.MinInt64
+	case pgtype.Infinity:
+		return math.MaxInt64
+	default:
+		return v.Time.UnixNano() / 1e6
+	}
+}
+
+func (p *pgxSeriesIterator) getVal() float64 {
+	return p.values.Elements[p.cur].Float
```
cevian, line 176: This is actually what can be null so the return here should be (float64, bool) and we should fix NULL handling here

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,196 @@
...
+// At implements storage.SeriesIterator.
+func (p *pgxSeriesIterator) At() (t int64, v float64) {
+	if p.cur >= p.totalSamples {
+		return 0, 0
+	}
+	return p.getTs(), p.getVal()
+}
+
+// Next implements storage.SeriesIterator.
+func (p *pgxSeriesIterator) Next() bool {
```
cevian, line 188: The right thing to do here is probably iterate until both the Time and  Value is not null.

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return found
+}
+
+func (p *pgxSeriesIterator) getTs() int64 {
+	v := p.times.Elements[p.cur]
+
+	if v.Status == pgtype.Null {
+		return 0
```
cevian, line 162: should probably be a two-value return of (model.Time, bool). 0 is a valid time value so we shouldn't use this to indicate NULL.

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,196 @@
...
+		return v.Time.UnixNano() / 1e6
+	}
+}
+
+func (p *pgxSeriesIterator) getVal() float64 {
+	return p.values.Elements[p.cur].Float
+}
+
+// At implements storage.SeriesIterator.
+func (p *pgxSeriesIterator) At() (t int64, v float64) {
```
cevian, line 180: This can stay as-is if Next() iterates util both time and value are Not NULL

pkg/prompb/query.go
```diff
@@ -0,0 +1,66 @@
+package prompb
```
cevian, line 1: I don't think this file is ever used. We should delete it.

pkg/prompb/labels.go
```diff
@@ -0,0 +1,32 @@
+package prompb
+
+import (
+	"errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// FromLabelMatchers parses protobuf label matchers to Prometheus label matchers.
+func FromLabelMatchers(matchers []*LabelMatcher) ([]*labels.Matcher, error) {
```
cevian, line 9: Nit would prefer this in another package. This package is only for protobuf pseudo-generated code.

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil
+	}
+
+	if err := ps.times.DecodeBinary(nil, arrays[2]); err != nil {
+		return nil
+	}
+
+	if err := ps.values.DecodeBinary(nil, arrays[3]); err != nil {
+		return nil
+	}
```
JLockerman, line 82: Why not just `p.rows[p.rowIdx].Scan(&ps.labelNames, &ps.LabelValues, &ps.times, &ps.values)`?

JLockerman:
further I think [this code](https://github.com/timescale/timescale-prometheus/pull/124/files#diff-25015b1d7c86d8c5fab194419dee6c51R930-R942) may be a problem; if I'm reading it right it'll send the queries one-at-a-time to the database, and incur latency proportional to the number of metrics, if we batched them we would only have 1 round-trip-time. Additionally, if we batched them that function could return a `BatchResults`, and we wouldn't have to store `[]Rows` everywhere.

pkg/pgmodel/series_set.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// At returns the current storage.Series. It expects to get rows to contain
+// four arrays in binary format which it attempts to deserialize into specific types.
+// It also expects that the first two and second two arrays are the same length.
+func (p *pgxSeriesSet) At() storage.Series {
+	if p.rowIdx >= len(p.rows) {
+		return nil
+	}
+	arrays := p.rows[p.rowIdx].RawValues()
```
cevian, line 54: This should be p.rows[p.rowIdx].Scan(&ps.labelNames, &ps.labelValues, &ps.times, &ps.values)

JLockerman:
(that last bit should probably wait for the next PR)

antekresic:
@cevian @JLockerman updated the PR with the latest comments and created a new issue for Joshes comment: https://github.com/timescale/timescale-prometheus/issues/128

pkg/pgmodel/labels.go
```diff
@@ -168,3 +169,29 @@ func (l *Labels) Swap(i, j int) {
 	l.values[j] = l.values[i]
 	l.values[i] = tmp
 }
+
+// FromLabelMatchers parses protobuf label matchers to Prometheus label matchers.
+func FromLabelMatchers(matchers []*prompb.LabelMatcher) ([]*labels.Matcher, error) {
```
JLockerman, line 174: Doesn't really belong here either, this is for our own `Labels` datatype 🤔

PR #135, opened 2020-06-25
Add integration tests for PromQL query and query_range endpoints

antekresic:
Integration tests are performed by checking the results of the new endpoints
against the results from a Prometheus instance that has ingested the same
dataset. It tests most of the PromQL queries, its aggregators and functions.

### DIFF
--- a/pkg/pgmodel/end_to_end_tests/query_integration_test.go
+++ b/pkg/pgmodel/end_to_end_tests/query_integration_test.go
@@ -22,6 +22,11 @@ import (
 	. "github.com/timescale/timescale-prometheus/pkg/pgmodel"
 )
 
+var (
+	startTime int64 = 1577836800000
+	endTime   int64 = 1577886800000
+)
+
 // PromClient is a wrapper around http.Client for sending read requests.
 type PromClient struct {
 	url        *url.URL
@@ -941,14 +946,16 @@ func generateSamples(index int) []prompb.Sample {
 		timeDelta int64   = 30000
 	)
 	samples := make([]prompb.Sample, 0, 3)
-	i := 1
+	i := 0
+	time := startTime + (timeDelta * int64(i))
 
-	for i <= 1000 {
+	for time < endTime {
 		samples = append(samples, prompb.Sample{
-			Timestamp: timeDelta * int64(i),
+			Timestamp: time,
 			Value:     delta * float64(i),
 		})
 		i++
+		time = startTime + (timeDelta * int64(i))
 	}
 
 	return samples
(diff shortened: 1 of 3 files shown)


pkg/pgmodel/end_to_end_tests/query_integration_test.go
```diff
@@ -570,6 +654,296 @@ func ingestQueryTestDataset(db *pgxpool.Pool, t testing.TB, metrics []prompb.Tim
...
+	}
+
+	promClient, err := NewPromClient(fmt.Sprintf("http://%s:%d/api/v1", testhelpers.PromHost, testhelpers.PromPort.Int()), 10*time.Second)
+	if err != nil {
+		t.Fatalf("unable to create PromQL query client for Prometheus: %s", err)
+	}
+
+	startTime := time.Unix(1577836800, 0)
+	endTime := time.Unix(1577866800, 0)
+	step := 10 * time.Second
```
blagojts, line 669: does it make sense to run the test with diff steps?

antekresic: Sure... Let me add 3-4 step sizes and update this to run for each of them.

pkg/pgmodel/end_to_end_tests/query_integration_test.go
```diff
@@ -939,13 +1316,16 @@ func generateSamples(index int) []prompb.Sample {
 	var (
 		delta     float64 = float64(index * 2)
 		timeDelta int64   = 30000
+
+		// Jan 1st, 2020
+		startTime int64 = 1577836800000
```
cevian, line 1321: nit: let's make this a constant used here and in startTime on line 667 too? Maybe end time as well (and use here)

pkg/pgmodel/end_to_end_tests/query_integration_test.go
```diff
@@ -38,6 +68,60 @@ func NewPromClient(urlStr string, timeout time.Duration) (*PromClient, error) {
 	return &PromClient{url: url, httpClient: httpClient}, nil
 }
 
+func (c *PromClient) genInstantRequest(query string, start time.Time) (*http.Request, error) {
```
cevian, line 71: nit: a bit weird to have this function be part of PromClient since the returned request is also used to send to the connector. Not a big deal but may be cleaner as a standalone function

pkg/pgmodel/end_to_end_tests/query_integration_test.go
```diff
@@ -2,26 +2,56 @@ package end_to_end_tests
```
cevian, line 2: Hmm one more nit. Maybe these added tests are better in a separate file? This file is already long and checking promql queries vs remote_read stuff seems like a good delineation to split things out along.

PR #138, opened 2020-06-30
Implement promql meta endpoints

blagojts:
The LabelNames and LabelValues endpoints are required
so timescale-prometheus can be directly queried by Grafana.

Not all metadata endpoints are implemented as a
prometheus server has (the series  and metadata endpoints
are missing) but they are never queried by Grafana.

An http router is added to handle path params and scale
better than the default multiplexer under load

### DIFF
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -217,25 +217,18 @@ func main() {
 	queryable := client.GetQueryable()
 	queryEngine := query.NewEngine(log.GetLogger(), time.Minute)
 	queryHandler := timeHandler(httpRequestDuration, "query", api.Query(queryEngine, queryable))
-	router.Get("/query", queryHandler)
-	router.Post("/query", queryHandler)
 	router.Get("/api/v1/query", queryHandler)
 	router.Post("/api/v1/query", queryHandler)
 
 	queryRangeHandler := timeHandler(httpRequestDuration, "query_range", api.QueryRange(queryEngine, queryable))
-	router.Get("/query_range", queryRangeHandler)
-	router.Post("/query_range", queryRangeHandler)
 	router.Get("/api/v1/query_range", queryRangeHandler)
 	router.Post("/api/v1/query_range", queryRangeHandler)
 
 	labelsHandler := timeHandler(httpRequestDuration, "labels", api.Labels(queryable))
-	router.Get("/labels", labelsHandler)
-	router.Post("/labels", labelsHandler)
 	router.Get("/api/v1/labels", labelsHandler)
 	router.Post("/api/v1/labels", labelsHandler)
 
 	labelValuesHandler := timeHandler(httpRequestDuration, "label/:name/values", api.LabelValues(queryable))
-	router.Get("/label/:name/values", labelValuesHandler)
 	router.Get("/api/v1/label/:name/values", labelValuesHandler)
 
 	router.Get("/healthz", api.Health(client))
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -10,6 +10,7 @@ import (
 	"database/sql"
 	"flag"
 	"fmt"
+	"github.com/prometheus/common/route"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -191,10 +192,9 @@ func main() {
 		os.Exit(1)
 	}
 	defer client.Close()
-	queryable := client.GetQueryable()
-	queryEngine := query.NewEngine(log.GetLogger(), time.Minute)
-	queryHandler := timeHandler(httpRequestDuration, "query", api.Query(queryEngine, queryable))
-	queryRangeHandler := timeHandler(httpRequestDuration, "query_range", api.QueryRange(queryEngine, queryable))
+
+	router := route.New()
+
 	promMetrics := api.Metrics{
 		LeaderGauge:         leaderGauge,
 		ReceivedSamples:     receivedSamples,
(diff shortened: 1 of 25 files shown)


cmd/timescale-prometheus/main.go
```diff
@@ -208,18 +208,44 @@ func main() {
...
+	labelsHandler := timeHandler(httpRequestDuration, "labels", api.Labels(queryable))
+	router.Get("/labels", labelsHandler)
+	router.Post("/labels", labelsHandler)
+	router.Get("/api/v1/labels", labelsHandler)
+	router.Post("/api/v1/labels", labelsHandler)
+
+	labelValuesHandler := timeHandler(httpRequestDuration, "label/:name/values", api.LabelValues(queryable))
+	router.Get("/label/:name/values", labelValuesHandler)
+	router.Get("/api/v1/label/:name/values", labelValuesHandler)
+
```
cevian, line 240: I think we are missing a /api/v1/series endpoint https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers

This can be in a separate PR

cmd/timescale-prometheus/main.go
```diff
@@ -208,18 +208,44 @@ func main() {
...
 	writeHandler := timeHandler(httpRequestDuration, "write", api.Write(client, elector, &promMetrics))
-	http.Handle("/write", writeHandler)
-	http.Handle("/read", timeHandler(httpRequestDuration, "read", api.Read(client, &promMetrics)))
-	http.Handle("/query", queryHandler)
-	http.Handle("/query_range", queryRangeHandler)
-	http.Handle("/api/v1/query", queryHandler)
-	http.Handle("/api/v1/query_range", queryRangeHandler)
-	http.Handle("/healthz", api.Health(client))
+	router.Post("/write", writeHandler)
+
```
cevian, line 212: I think we should document all the endpoint that we support. Doesn't have to be detailed, but a list/table of endpoints with links to the prom documentation would be good. Can be in a separate PR.

pkg/api/label_values.go
```diff
@@ -0,0 +1,39 @@
...
+		if err != nil {
+			respondError(w, http.StatusInternalServerError, err, "internal")
+			return
+		}
+		names, warnings, err := querier.LabelValues(name)
+		if err != nil {
+			respondError(w, http.StatusInternalServerError, err, "internal")
+			return
+		}
+
```
cevian, line 34: In prom, I see a code path for `w.WriteHeader(http.StatusNoContent)`  (api.go in the handler function). We don't have such a path. Should we (applies to all handlers)?

blagojts: Added a NoContent path in the respondQuery function, should cover the same path as in prometheus api impl

docs/prometheus_api.md
```diff
@@ -0,0 +1,43 @@
...
+
+We implemented the endpoints for executing PromQL and labels metadata to allow the Timescale-Prometheus connector to be 
+used as a Prometheus compatible data-source in applications like Grafana. 
+
+<img src="./timescale-prometheus-arch.png" alt="Timescale-Prometheus Architecture Diagram" width="800"/>
+
+
+When requesting a query to run on data that is in the long-term storage through Prometheus, the query goes through the 
+following steps. 
+1. Query engine requests the data through the remote_read protocol;
```
cevian, line 15: can we add `1. The user issues a query to the prometheus server `

docs/prometheus_api.md
```diff
@@ -0,0 +1,43 @@
...
+1. The connector translates the request to SQL statement with only a time range and label matchers applied;
+1. The database returns almost raw data;
+1. The connector marshals it into a format compatible with the remote_read protocol;
+1. The Query engine combines the local and remote data and applies any functions or aggregations before returning a 
+result.
+
+By having the Connector implement the PromQL APIs, the connector can:
+1. Parse the PromQL and translate it to a SQL statement that with a time range, label matchers, calculations and
```
cevian, line 23: please add `1. the user issues a query directly to the connector`

docs/prometheus_api.md
```diff
@@ -0,0 +1,43 @@
...
+1. A modified Prometheus Query Engine finalizes the calculation and marshals the data to a proper response.
+
+By using the Connector for PromQL queries directly a network trip is avoided, and TimescaleDB is better utilized to 
+actually perform some calculations. 
+
+## Implemented Endpoints
+
+|                 Name               |              Endpoint                   |                          Description                      | 
+|------------------------------------|-----------------------------------------|-----------------------------------------------------------|
+|[Instant Queries][instant-queries]  |`GET /api/v1/query``POST /api/v1/query`  |Evaluate an instant query at a single point in time        |
```
cevian, line 35:
```suggestion
|[Instant Queries][instant-queries]  |`GET,POST  /api/v1/query`  |Evaluate an instant query at a single point in time        |
```
and similar below

pkg/api/labels.go
```diff
@@ -0,0 +1,56 @@
...
+	return parser.ValueTypeNone
+}
+
+func (l labelsValue) String() string {
+	return strings.Join(l, "\n")
+}
+
+func Labels(queriable *query.Queryable) http.Handler {
+	hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
```
cevian, line 28: This is controlled by a command line flag in Prom 
```
	a.Flag("web.cors.origin", `Regex for CORS origin. It is fully anchored. Example: 'https?://(domain1|domain2)\.com'`).
		Default(".*").StringVar(&cfg.corsRegexString)
```

Should we implement this in the future too (if so please file an issue)

cevian: Also why isn't this in `setHeaders`?

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -358,24 +510,9 @@ func testRequest(t *testing.T, req *http.Request, handler http.Handler, client *
 		}
 		defer tsResp.Body.Close()
 
-		var got, wanted response
-
-		err = json.Unmarshal(tsContent, &got)
+		err = comparator(promContent, tsContent)
```
cevian, line 513: should we also be comparing headers in this function? Not just the body? (ok to leave to another PR but let's file an issue if we do)

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -320,20 +368,124 @@ func TestPromQLEndpoint(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("could not get label names from querier")
+		}
+		for _, label := range labelNames {
+			req, err = getLabelValuesRequest(apiURL, label)
+			if err != nil {
+				t.Fatalf("unable to create PromQL label values request: %v", err)
 			}
+			testMethod := testRequest(req, router, client, labelsResultComparator)
+			tester.Run(fmt.Sprintf("get label values for %s", label), testMethod)
```
cevian, line 437: let's also add a test for a request for a label name that does not exist

pkg/api/labels.go
blagojts: Moved to setHeaders, and created a card in project board

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
blagojts: added to project board

PR #139, opened 2020-06-30
Fix pushdown array construction

JLockerman:
Looks like we misunderstood construct_md_array, it needs to have an elem
for every slot in the array regardless of rather it's NULL (we had
thought it only needed non-NULL elements). This commit fixes this
oversight.

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -130,7 +130,7 @@ pg_agg!{
 #[derive(Debug)]
 struct GapfillDeltaTransition {
     window: VecDeque<(TimestampTz, f64)>,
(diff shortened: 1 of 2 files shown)


extension/src/lib.rs
```diff
@@ -208,6 +208,7 @@ impl GapfillDeltaTransition {
     fn add_delta_for_current_window(&mut self) {
         if self.window.len() < 2 {
             // if there are 1 or fewer values in the window, store NULL
+            self.deltas.push(0);
```
cevian, line 211: need to change comment on line 133

PR #141, opened 2020-07-01
Handwritten json encoder

JLockerman:


pkg/api/marshal_test.go
```diff
@@ -0,0 +1,334 @@
...
+	}{
+		{
+			name:     "nil",
+			value:    promql.Vector{},
+			expected: `{"status":"success","data":{"resultType":"vector","result":[]}}`,
+		},
+		{
+			name:     "empty",
+			value:    promql.Vector{},
+			expected: `{"status":"success","data":{"resultType":"vector","result":[]}}`,
```
cevian, line 26: hmm do we need these expected strings? Can't we just use the standard json encoder for the source of truth?

JLockerman: the standard json encoder vs string is a bit of a tradeoff, on the one hand the standard encoder guarantees that we’ll always match the output prom expects, on the other hand it’s nice having the strings for documentation purposes…

PR #146, opened 2020-07-06
Add label cache, and only fetch missing labels at query time 

JLockerman:
In simple queries, approximately 1/3rd our queries runtime is spent in
key_value_array. To reduce the time spent, this commit adds a cache of
ids-to-labels in the connector, so we can only fetch those labels we
have never seen before.

### DIFF
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -25,6 +25,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
+	"github.com/timescale/timescale-prometheus/pkg/clockcache"
 	"github.com/timescale/timescale-prometheus/pkg/log"
 	"github.com/timescale/timescale-prometheus/pkg/prompb"
 )
@@ -883,12 +884,13 @@ func (p *pendingBuffer) addReq(req insertDataRequest) bool {
 
 // NewPgxReaderWithMetricCache returns a new DBReader that reads from PostgreSQL using PGX
 // and caches metric table names using the supplied cacher.
-func NewPgxReaderWithMetricCache(c *pgxpool.Pool, cache MetricCache) *DBReader {
+func NewPgxReaderWithMetricCache(c *pgxpool.Pool, cache MetricCache, labelsCacheSize uint64) *DBReader {
 	pi := &pgxQuerier{
 		conn: &pgxConnImpl{
 			conn: c,
 		},
 		metricTableNames: cache,
+		labels:           clockcache.WithMax(labelsCacheSize),
 	}
 
 	return &DBReader{
@@ -897,10 +899,10 @@ func NewPgxReaderWithMetricCache(c *pgxpool.Pool, cache MetricCache) *DBReader {
 }
 
 // NewPgxReader returns a new DBReader that reads that from PostgreSQL using PGX.
-func NewPgxReader(c *pgxpool.Pool, readHist prometheus.ObserverVec) *DBReader {
+func NewPgxReader(c *pgxpool.Pool, readHist prometheus.ObserverVec, labelsCacheSize uint64) *DBReader {
 	metrics, _ := bigcache.NewBigCache(DefaultCacheConfig())
 	cache := &MetricNameCache{metrics}
-	return NewPgxReaderWithMetricCache(c, cache)
+	return NewPgxReaderWithMetricCache(c, cache, labelsCacheSize)
 }
 
 type metricTimeRangeFilter struct {
@@ -912,6 +914,8 @@ type metricTimeRangeFilter struct {
 type pgxQuerier struct {
 	conn             pgxConn
 	metricTableNames MetricCache
+	// contains [int64]labels.Label
+	labels *clockcache.Cache
 }
 
 // HealthCheck implements the healtchecker interface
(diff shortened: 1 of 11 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -1021,6 +1025,81 @@ func (q *pgxQuerier) LabelValues(labelName string) ([]string, error) {
...
+		}
+		if len(keys) != len(vals) {
+			return fmt.Errorf("query returned a mismatch in timestamps and values: %d, %d", len(keys), len(vals))
+		}
+		if len(keys) != len(misses) {
+			return fmt.Errorf("query returned wrong number of labels: %d, %d", len(misses), len(keys))
+		}
+
+		for i := range misses {
+			newLabels[i] = labels.Label{Name: keys[i], Value: vals[i]}
```
cevian, line 1093: I'm pretty sure this has to be sorted. A labels.Labels has to be sorted.

pkg/pgmodel/pgx.go
```diff
@@ -1021,6 +1025,81 @@ func (q *pgxQuerier) LabelValues(labelName string) ([]string, error) {
 	return labelValues, nil
 }
 
+const GetLabelsSQL = "SELECT (key_value_array($1::int[])).*"
```
cevian, line 1028: (don't feel strongly) should this cache stuff be broken out into it's own .go file?

pkg/pgmodel/series_set_test.go
```diff
@@ -450,93 +392,43 @@ func genRows(count int) [][][]byte {
...
-		kIdx += len(v)
-	}
-	for _, v := range labelValues {
-		valLen := len(v)
-		values = append(values, []byte("1234")...)
-		binary.BigEndian.PutUint32(values[vIdx:], uint32(valLen))
-		vIdx += 4
-		values = append(values, v...)
-		vIdx += len(v)
+func pgLabel(id int64) pgtype.Int8 {
```
cevian, line 408: where is this used?

pkg/pgmodel/pgx.go
JLockerman: Not yet, but it should later: we should split out the caching layer from the talks-to-database layer, with only that final layer in pgx and a well-defined go api between them. This would make our tests _much_ nicer as we can tests value-injection from the database using the go api, and only test SQL output from the pgx layer.

pkg/pgmodel/series_set_test.go
JLockerman: it's a leftover from an old rewrite of these tests. Removed.

JLockerman:
comments addressed, sorting code is [here](https://github.com/timescale/timescale-prometheus/pull/146/files#diff-25015b1d7c86d8c5fab194419dee6c51R1092-R1093)

pkg/pgmodel/pgx.go
```diff
@@ -1021,6 +1025,84 @@ func (q *pgxQuerier) LabelValues(labelName string) ([]string, error) {
...
+	}
+
+	return
+}
+
+func (q *pgxQuerier) fetchMissingLabels(misses []interface{}, missedIds []int64, newLabels []interface{}) error {
+	for i := range misses {
+		missedIds[i] = misses[i].(int64)
+	}
+	rows, err := q.conn.Query(context.Background(), GetLabelsSQL, missedIds)
```
cevian, line 1073: doesn't rows have to be Closed?

pkg/pgmodel/pgx.go
```diff
@@ -1021,6 +1025,84 @@ func (q *pgxQuerier) LabelValues(labelName string) ([]string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	for rows.Next() {
```
cevian, line 1078: should make sure we get at most one row?

pkg/pgmodel/pgx.go
cevian, line 1073: cant we do `rows, err := q.conn.Query(context.Background(), GetLabelsSQL, misses)` and not have missedIds at all?

pkg/pgmodel/pgx.go
```diff
@@ -1021,6 +1025,84 @@ func (q *pgxQuerier) LabelValues(labelName string) ([]string, error) {
...
+		if len(keys) != len(misses) {
+			return fmt.Errorf("query returned wrong number of labels: %d, %d", len(misses), len(keys))
+		}
+
+		lls := Labels{names: keys, values: vals}
+		sort.Sort(&lls)
+
+		for i := range misses {
+			newLabels[i] = labels.Label{Name: lls.names[i], Value: lls.values[i]}
+		}
```
cevian, line 1097:
```suggestion
		for i := range keys {
			newLabels[i] = labels.Label{Name: keys[i], Value: vals[i]}
		}
```

pkg/pgmodel/pgx.go
cevian: should we use QueryRow instead?

JLockerman: our shim doesn't expose that function ☹️

pkg/pgmodel/pgx.go
JLockerman: IIRC pgx doesn't handle deserializing to an `interface{}`; it doesn't know which type to write to

pkg/pgmodel/pgx.go
JLockerman: I have mixed feelings: on the one hand the resulting code is smaller, on the other hand once the Labels take ownership of the arrays it seems better to always access it through that

PR #147, opened 2020-07-10
Add CLOCK based approximate LRU

JLockerman:
We have a number of places in the connector where we want to store a
cache of elements. Our expectation is that the set of live elements
will be time-correlated, and therefore, when we want to evict elements
from the cache the least-recently-used element is the one least likely
to be used again in the future. There for this commit adds an LRU
implementation suitable for use in such caches. Traditional LRUs suffer
from large amounts of contention in multi-threaded workloads, so we use
a CLOCK-based approximate LRU; more information can be found in the
README.

### DIFF
--- a/pkg/clockcache/Readme.md
+++ b/pkg/clockcache/Readme.md
@@ -1,0 +1,72 @@
+# clockcache #
+
+This package contains a CLOCK based approximate-LRU cache optimized for
+concurrent usage. It minimizes both the number of locks held, and the number of
+writes to shared cache lines. In the worst case, this implementation requires
+_at most_ 1 shared-cache write on a hit, not counting lock acquisition.
+
+## Issues With Concurrent LRU ##
+
+Typically, LRU is implemented with a map and a doubly-linked list, all elements
+are stored in both the map, and the list. Whenever there's a hit on a
+cache-item, it's moved to the head of the list; evictions are done from the tail
+of the list. This ensures that it's always the least-recently-hit element that's
+evicted.
+
+While conceptually simple, the standard implementation of LRU as significant
+issues in a multi-processor setting, since _every_ read to the cache writes to
+the underlying datastructures. Updating the LRU list requires at least 4 writes
+on shared cache lines; 2 to unlink the element from it's old place in the list,
+and 2 more to add it to the head. When the LRU is being accessed from multiple
+cores, these cache lines will bounce among them, causing a significant slowdown.
+
+## CLOCK-based LRU ##
+
+The CLOCK algorithm approximates LRU. Conceptually, it augments a map with a
+bitset, with each element owning one bit in the bitset. Whenever an element is
+hit, it's bit is set to `1`. The evictor maintains a cursor into the bitset, and
+reads it round-robin.
+
+```
+                    eviction cursor
+                            v
+bitset: 0 1 0 0 1 1 0 1 0 0 1 1 1 1 0
+```
+
+Whenever the evictor sees a `1` it sets the bit to `0`.
+Whenever the evictor see a `0`, it evicts the element.
+This means that eviction always evicts an element that hasn't been read since
+the last time the evictor ran, and, since it reads the element in order, the
+first such element is the element that hasn't been touched the longest since the
+last run.
+
+### Concurrent CLOCK ###
+
+CLOCK two important benefits over regular LRU in a concurrent setting; the state
+that changes on a hit is per-element instead of shared, and the changes made by
+a hit are idempotent.
+
+Regular LRU relies on a shared list, which is manipulated every time an element
+is hit. CLOCK only requires a per-element bit to be set on a hit. This means
+that on a hit, only a single value, and with the proper layout, a single cache
+line, needs to be written to. Already, this reduces the worst-case number of
+writes to 1 from 4. We can further reduce the number of writes by observing that
+for an LRU to be useful, hits must be more common than evictions, and in CLOCK
+any successive hits before an eviction need not change any state; the bit is
+already set to `1`. This means that we can check the bit before writing it, and
+do nothing if it's already set, further reducing the number of writes.
+
+## Other Optimizations ##
+
+The map consists of a hashmap and slice. The map is used to perform key
+lookups, and the slice is used to ensure a consistent order for eviction checks.
+The actual elements and staleness-bits are stored in the slice, there is no
+separate bitset. Slice-elements nodes are padded to take a cache line
+each to prevent false-sharing. Staleness-bits are set using atomic operations,
+and are only written after they have been read to be `0`. The slice and map are
+preallocated to the maximum capacity of the cache, this prevents most
+allocations within the critical section (the map will occasionally allocate
+even with preallocation). We have separate locks for the slice and the cache;
+during eviction we grab a lock on only the slice, and allow gets to proceed
+until we find an element to evict. At that point we lock the map and replace the
+element.
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -1,0 +1,288 @@
+package clockcache
+
+import (
+	"fmt"
+	"sync"
+	"sync/atomic"
+)
+
+// CLOCK based approximate LRU storing designed for concurrent usage.
+// Gets only require a read lock, while Inserts take at least one write lock.
+type Cache struct {
+	// guards elements and all fields except for `used` in Element, must have at
+	// least a read-lock to access, and a write-lock to insert/update/delete.
+	elementsLock sync.RWMutex
+	// stores indexes into storage
+	elements map[interface{}]*element
+	storage  []element
+
+	// guards next, and len(storage) and ensures that at most one eviction
+	// occurs at a time, always grabbed _before_ elementsLock
+	insertLock sync.Mutex
+	// CLOCK sweep state, must have the insertLock
+	next int
+}
+
+type element struct {
+	// The value stored with this element.
+	key   interface{}
+	value interface{}
+
+	// CLOCK marker if this is recently used
+	used uint32
+
+	// pad Elements out to be cache aligned
+	_ [24]byte
+}
+
+func WithMax(max uint64) *Cache {
+	if max < 1 {
+		panic("must have max greater than 0")
+	}
+	return &Cache{
+		elements: make(map[interface{}]*element, max),
+		storage:  make([]element, 0, max),
+	}
+}
+
+// Insert a key/value mapping into the cache if the key is not already present
+// returns the canonical version of the value
+// and if the value is in the map
+func (self *Cache) Insert(key interface{}, value interface{}) (canonicalValue interface{}, in_cache bool) {
+	self.insertLock.Lock()
+	defer self.insertLock.Unlock()
+
+	_, canonicalValue, in_cache = self.insert(key, value)
+	return
+}
+
+// Insert a batch of keys with their corresponding values.
+// This function will _overwrite_ the keys and values slices with their
+// canonical versions.
+// returns the number of elements inserted, is lower than len(keys) if insertion
+// starved
+func (self *Cache) InsertBatch(keys []interface{}, values []interface{}) int {
+	if len(keys) != len(values) {
+		panic(fmt.Sprintf("keys and values are not the same len. %d keys, %d values", len(keys), len(values)))
+	}
+	values = values[:len(keys)]
+	self.insertLock.Lock()
+	defer self.insertLock.Unlock()
+
+	for idx := range keys {
+		var inserted bool
+		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx])
+		if !inserted {
+			return idx
+		}
+	}
+	return len(keys)
+}
+
+func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey interface{}, canonicalValue interface{}, inserted bool) {
+	elem, present := self.elements[key]
+	if present {
+		// we'll count a double-insert as a hit. See the comment in get
+		if atomic.LoadUint32(&elem.used) != 0 {
+			atomic.StoreUint32(&elem.used, 1)
+		}
+		return elem.key, elem.value, true
+	}
+
+	var insertLocation *element
+	if len(self.storage) >= cap(self.storage) {
+		insertLocation = self.evict()
+		if insertLocation == nil {
+			return key, value, false
+		}
+		self.elementsLock.Lock()
+		defer self.elementsLock.Unlock()
+		delete(self.elements, insertLocation.key)
+		*insertLocation = element{key: key, value: value}
+	} else {
+		self.elementsLock.Lock()
+		defer self.elementsLock.Unlock()
+		self.storage = append(self.storage, element{key: key, value: value})
+		insertLocation = &self.storage[len(self.storage)-1]
+	}
+
+	self.elements[key] = insertLocation
+	return key, value, true
+}
+
+func (self *Cache) evict() (insertPtr *element) {
+	// this code goes around storage in a ring searching for the first element
+	// not marked as used, which it will evict. The code has two unusual
+	// features:
+	//  1. it will go through storage at most twice before giving up. Concurrent
+	//     gets can starve out the evictor, in which case the cache is too small
+	//  2. it divides the walk through storage into two loops, one walk through
+	//     all the elements after the last place the evictor stopped, one
+	//     through all elements before that location. This is due to a limitation
+	//     in go's bounds check elimination, where it will only eliminate checks
+	//     based off an induction variable e.g. `next := range slice`,
+	//     if the value is merely guarded by e.g. `if next >= len(slice) { next = 0 }`
+	//     the bounds check will not be elided. Doing the walk like this lowers
+	//     eviction time by about a third
+	startLoc := self.next
+	postStart := self.storage[startLoc:]
+	preStart := self.storage[:startLoc]
+	for i := 0; i < 2; i++ {
+		for next := range postStart {
+			elem := &postStart[next]
+			old := atomic.SwapUint32(&elem.used, 0)
+			if old == 0 {
+				insertPtr = elem
+			}
+
+			if insertPtr != nil {
+				self.next = next + 1
+				return
+			}
+		}
+		for next := range preStart {
+			elem := &preStart[next]
+			old := atomic.SwapUint32(&elem.used, 0)
+			if old == 0 {
+				insertPtr = elem
+			}
+
+			if insertPtr != nil {
+				self.next = next + 1
+				return
+			}
+		}
+	}
+
+	return
+}
+
+// tries to get a batch of keys and store the corresponding values is valuesOut
+// returns the number of keys that were actually found.
+// NOTE: this function does _not_ preserve the order of keys; the first numFound
+//       keys will be the keys whose values are present, while the remainder
+//       will be the keys not present in the cache
+func (self *Cache) GetValues(keys []interface{}, valuesOut []interface{}) (numFound int) {
+	if len(keys) != len(valuesOut) {
+		panic(fmt.Sprintf("keys and values are not the same len. %d keys, %d values", len(keys), len(valuesOut)))
+	}
+	valuesOut = valuesOut[:len(keys)]
+	n := len(keys)
+	idx := 0
+
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+
+	for idx < n {
+		value, found := self.get(keys[idx])
+		if !found {
+			if n == 0 {
+				return 0
+			}
+			// no value found for key, swap the key with the last element, and shrink n
+			n -= 1
+			keys[n], keys[idx] = keys[idx], keys[n]
+			continue
+		}
+		valuesOut[idx] = value
+		idx += 1
+	}
+	return n
+}
+
+func (self *Cache) Get(key interface{}) (interface{}, bool) {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	return self.get(key)
+}
+
+func (self *Cache) get(key interface{}) (interface{}, bool) {
+
+	elem, present := self.elements[key]
+	if !present {
+		return 0, false
+	}
+
+	// While logically this is a CompareAndSwap, this code has an important
+	// advantage: in the common case of the element already being marked as used,
+	// this is a read-only operation, and doesn't trash the cache line that used
+	// is stored on. The lack of atomicity of the update doesn't matter for our
+	// use case.
+	if atomic.LoadUint32(&elem.used) == 0 {
+		atomic.StoreUint32(&elem.used, 1)
+	}
+
+	return elem.value, true
+}
+
+func (self *Cache) unmark(key string) bool {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+
+	elem, present := self.elements[key]
+	if !present {
+		return false
+	}
+
+	// While logically this is a CompareAndSwap, this code has an important
+	// advantage: in the common case of the element already being marked as used,
+	// this is a read-only operation, and doesn't trash the cache line that used
+	// is stored on. The lack of atomicity of the update doesn't matter for our
+	// use case.
+	if atomic.LoadUint32(&elem.used) != 0 {
+		atomic.StoreUint32(&elem.used, 0)
+	}
+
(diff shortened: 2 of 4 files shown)


pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+type Element struct {
+	// The value stored with this element.
+	key   interface{}
+	Value interface{}
+
+	//CLOCK marker if this is recently used
+	used uint32
+
+	// pad Elements out to be cache aligned
+	_padding [24]byte
```
cevian, line 36: `_padding` should be named `_`

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+func (self *Cache) evict() (insertPtr *Element) {
+	for {
+		insertLocation, evicted := self.tryEvict()
+		if evicted {
+			return insertLocation
+		}
+	}
+}
+
+func (self *Cache) tryEvict() (insertPtr *Element, evicted bool) {
```
cevian, line 104: can't we rewrite cache.evict() as the following and avoid the try. The following should guarantee we always return
```
for {
      if self.next >= len(self.storage) {
		self.next = 0
      }
      elem := &self.storage[self.next]
      if elem.used != 0 {
         elem.used = 0
      }
      else {
	       delete(self.elements, elem.key)
               return elem
      }
      self.next += 1
}
```

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if atomic.LoadUint32(&elem.used) == 0 {
+		atomic.StoreUint32(&elem.used, 1)
+	}
+
+	return elem.Value, true
+}
+
+func (self *Cache) Unmark(key string) bool {
```
cevian, line 181: Wouldn't a Remove be more appropriate? What's the use-case for just an Unmark where you wouldn't want a remove?

pkg/clockcache/cache_test.go
```diff
@@ -0,0 +1,173 @@
+package clockcache
```
cevian, line 1: This would be a great use case for some testing benchmarking functions too

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"fmt"
+	"sync"
+	"sync/atomic"
+)
+
+// CLOCK based approximate LRU storing mappings from strings to
+// int64s designed for concurrent usage.
+// It is sharded into 127 blocks, each of which can be
```
cevian, line 11: comment needs updating; no longer sharded.

pkg/clockcache/cache.go
iand: Think this would need to increment self.next before returning elem

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+// locked independently. Gets only require a read lock on the
+// individual block the element can be found in. When the cache is
+// not full, Insert only requires a write lock on the individual
+// block. Eviction locks blocks one at a time looking for a value
+// that's valid to evict/
+type Cache struct {
+	lock sync.RWMutex
```
iand, line 18: This lock protects access to `elements`+`storage` and `next`. I suggest having a separate Mutex for `storage`+`next`. This allows cache reads to take a read lock without having to wait for a pending eviction from a concurrent insert.

RWMutexes have a pathological case where every call to Lock causes all RLocks to be queued until Unlock is called.

pkg/clockcache/Readme.md
```diff
@@ -0,0 +1,72 @@
...
+line, needs to be written to. Already, this reduces the worst-case number of
+writes to 1 from 4. We can further reduce the number of writes by observing that
+for an LRU to be useful, hits must be more common than evictions, and in CLOCK
+any successive hits before an eviction need not change any state; the bit is
+already set to `1`. This means that we can check the bit before writing it, and
+do nothing if it's already set, further reducing the number of writes.
+
+## Other Optimizations ##
+
+The cache is built off a sharded hashmap; values are first hashed to see which
```
iand, line 61: Needs removing now

pkg/clockcache/Readme.md
```diff
@@ -0,0 +1,72 @@
...
+hashmap they should be stored in, then stored in an individual shard. The shards
+are protected by Read/Write locks, with the read side taken for cache lookups,
+and the write side for cache inserts and evictions. Eviction locks a single
+shard at a time, going in round-robin fashion until it finds an element to evict.
+
+Each shard consists of a hashmap and linkedlist. The map is used to perform key
```
iand, line 67: It's not a linked list, just a list.

pkg/clockcache/cache.go
JLockerman: @iand you mean something like
```
1. lock storage+next, search for an element to be evicted
2. lock elements, evict old element and re-insert a new one
```
?
I'll have to think about that a bit more...

JLockerman: one issue is that it can cause the evictor to starve if the other threads are hitting the cache fast enough...

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Insert a bactch of keys with their corresponding values.
+// This function will _overwrite_ the keys and values slices with their
+// canonical versions.
+func (self *Cache) InsertBatch(keys []interface{}, values []interface{}) {
+	if len(keys) != len(values) {
+		panic(fmt.Sprintf("keys and values are not the same len. %d keys, %d values", len(keys), len(values)))
+	}
+	values = values[:len(keys)]
```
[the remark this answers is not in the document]
JLockerman: it's mostly a lint; if you forget a key or a slot for a value it's really annoying to debug, and it's not difficult to reslice before calling `InsertBatch` when you want to store into a longer sice

pkg/clockcache/cache.go
iand: Yes, that occured to me while I was out for my walk. If the cache is hot such that every key is being used between writes then it's probably too small and inserting will just cause unnecessary churn.  However, is there a requirement for insert to always succeed? It could attempt one run through the eviction loop and return if it can't insert. On a hot cache that's not much different from inserting and being immediately evicted.

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+		self.next += 1
+		reachedEnd = self.next >= len(self.storage)
+	}
+
+	return
+}
+
+// tries to get a batch of keys and store the corresponding values is valuesOut
+// returns the number of keys that were actually found.
+// NOTE: this function does _not_ preserve the order of keys; the first numFound
```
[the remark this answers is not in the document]
JLockerman: The real reason is because the API was designed around non-pointer keys. However, the current version has the advantage of constant-time discovery of hits/misses. In our labels cache, the expected use looks something like:
```go
numFound := labelsCache.GetValue(labelIds, labels)
if numFound < len(labelIds) {
    missed := fetchLabels(labelIds[numFound:])
    copy(labels[numFound:], missed)
}
sort.Slice(labels, ...)
return labels
```
which I think lends itself better to the current API

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (self *Cache) get(key interface{}) (interface{}, bool) {
+
+	elem, present := self.elements[key]
+	if !present {
+		return 0, false
+	}
+
+	if atomic.LoadUint32(&elem.used) == 0 {
```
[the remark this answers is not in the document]
JLockerman: unless we do the changes Ian proposes around multiple locks, it shouldn't: the lock will exclude the evictor, and all readers are going to write the same value at this point, so the store is idempotent. An atomic store is potentially a decent bit faster than a `cmpxchg`, so we want to use the simplest instruction we can

pkg/clockcache/cache.go
JLockerman: 1. Testing 😋 
2. incremental unmarking: one of the big downsides to this kind of cache is the ~linear time it takes to evict. One thing that may help with that is incrementally unmarking elements during normal operation so that when it comes time to evict there's less to do. `Remove` is much to aggressive an operation for that.

pkg/clockcache/cache.go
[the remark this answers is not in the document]
JLockerman: I assumed you meant the next line down, but I realize now you may mean the while
```go
if atomicLoad(...)  {
    atomicStore(...)
}
```
The current way actually performs a bit better: this is a read-mostly operation; the majority of access to the cache should be `Get`s, which means most of the time when we reach this code `used` will already be `1`, and we'll never hit the store. If we perform a `CompareAndSwap` we'll store a value every single time we perform a `Get`. Performing a read to a shared CPU cache-line is basically free (if it's already in the cache), while performing a store require gaining exclusive access to the line, both increasing the latency for the store-er, and forcing a cache-miss on all other cores.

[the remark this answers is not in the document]
JLockerman: I'l add a comment summarizing this.

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+func (self *Cache) Unmark(key string) bool {
+	self.lock.RLock()
+	defer self.lock.RUnlock()
+
+	elem, present := self.elements[key]
+	if !present {
+		return false
+	}
+
+	if atomic.LoadUint32(&elem.used) != 0 {
```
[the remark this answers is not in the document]
JLockerman: if you're talking about the entire `if` then see https://github.com/timescale/timescale-prometheus/pull/147/files#r453666623.
If you're worried about races, then there are two cases:

1. A `Get` sets used to `1` after we perform the `Load`. This doesn't matter, since the value remains the same
2. Another `Unmark` sets the value to `0` after the `Load`. Both threads write the same value, so it's idempotent.

Notably the evictor cannot be running because of the lock.

pkg/clockcache/cache.go
JLockerman: That's a good point, and in the current implementation `InsertBatch` can actually "succeed" by evicting its own elements, which is morally the same thing.

JLockerman: @iand @cevian Ok, let;s go through this in more depth.
The evict we use in `Insert` will go through `storage` until we either find an element to evict or we reach the element we started at without having done so. In the second case we’ll give up, as all the elements are too useful (maybe we’ll add a new heuristic later, really this probably means the cache is too small). I’m going to assume it’s ok to be a little sloppy with `used`, and that a sequence like
```
1. evictor sees Element e is unused, decides to evict it
2. getter marks e as used
3. evictor grabs lock on elements and evicts e
```
is ok: it’s morally equivalent to swapping the order of steps `1` and `2`.

The changed functions are
```go
func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey interface{}, canonicalValue interface{}, inserted bool) {
	elem, present := self.elements[key]
	if present {
		return elem.key, elem.Value, false
	}

	var insertLocation *Element
	if len(self.storage) >= cap(self.storage) {
		insertLocation = self.evict()
		if insertLocation == nil {
			return nil, nil, false
		}
		self.elementsLock.Lock()
		defer self.elementsLock.Unlock()
		key := insertLocation.key
		delete(self.elements, key)
		*insertLocation = Element{key: key, Value: value}
	} else {
		self.storage = append(self.storage, Element{key: key, Value: value})
		insertLocation = &self.storage[len(self.storage)-1]
	}

	self.elements[key] = insertLocation
	return key, value, true
}

func (self *Cache) evict() (insertPtr *Element) {
	for {
		if self.next >= len(self.storage) {
			self.next = 0
		}

		insertPtr = self.tryEvictNext()
		if insertPtr != nil {
			return
		}
		if self.next == startLock {
			break
		}
	}
	// we've gone through all the elements, and they're all used
	// we'll try going through them once more, and if we cannot find
	// an unused element, the cache is to small, and we'll give up
	for {
		if self.next >= len(self.storage) {
			self.next = 0
		}

		insertPtr = self.tryEvictNext()
		if insertPtr != nil {
			return
		}
		if self.next == startLock {
			break
		}
	}
	return
}

func (self *Cache) tryEvictNext() (insertPtr *Element) {
	elem := &self.storage[self.next]
	used := atomic.CompareAndSwapUint32(&elem.used, 1, 0)
	if !used {
		insertPtr = elem
	}
	self.next += 1
	return
}

```

JLockerman: so:

1. The long-duration lock only excludes other evictions, which is really nice; the only place we lock out `Get`s is for a `delete;insert` into the `map`
2.Correctness looks pretty trivial: the only state touched by the evictor is `next`, which is guarded by the lock, and `used` which is only accessed via atomics.
3. We can _almost_ forgo locking during the eviction-search entirely, if we use atomic ops to update `next` there're no race conditions, though there is a little weirdness when to evictors decide to evict the same element if one loops the other. I'm still tempted tho, I think that case should be rare enough, and the extra performance would be worth it. (this can be mitigated to a large degree by css-ing `2` into `used` when deciding to evict, it's not perfect because the `Get` update of `used` is non-atomic, but I'm not sure we care)
3. CPU-cache behavior is roughly the same: everyone has an oblige cache-miss to grab the lock, the evictor has an obligate miss per-element searched through, and however much it does when manipulating the map; the only addition is an extra lock-acquisition

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	elements map[interface{}]*Element
+	storage  []Element
+
+	// CLOCK sweep state, must have write lock
+	next int
+}
+
+type Element struct {
+	// The value stored with this element.
+	key   interface{}
```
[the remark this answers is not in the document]
JLockerman: Yeah, I don't love this. The problem we have is that we want to use both `int64`s and `string`s as keys, and I'm not sure @cevian is willing for us to `generate` a unique cache type for each. I considered storing redundant key and fields for different types, but since we'd need to `interface{}` the keys anyway to store in the `map`, I don't think it's worthwhile.
(`interface{}`ing will _also_ increase the cache miss rate normally, because of the pointer-chasing you mentioned, the padding below is to prevent false-sharing and bouncing elements around cores because of new adjacent elements. I'm not sure it's worthwhile in this case, because we should be read-mostly anyway, and I have concerns about the `map`, but ehh, unless we have more stuff to store here, it doesn't really matter)

pkg/clockcache/cache_test.go
JLockerman: hmm, I have some benchmarks I copied of `bigcache` that I've been using, but I don't love them. It's a bit difficult to get benchmarks that are indicative of real-world contention, which is what should really matter for performance. I'll see what I can do...

pkg/clockcache/cache.go
JLockerman: @iand nice find! if you and @cevian can confirm that the above code looks right (modulo the lock changes in `Insert[Batch]` which I elided) I'll update the PR to match

JLockerman: writing this down so I don't forget: non-locking eviction has an actual data-race where the
```go
*insertLocation = Element{key: key, Value: value}
```
races against 
```go
used := atomic.CompareAndSwapUint32(&elem.used, 1, 0)
```
I think this is "benign" in that both are writing the same value, and I don't know of an architecture that would mess it up, but we still shouldn't do it.

cevian: A few questions:
1) what lock protects next? I don't see it in the code above.
2) why is the Get update to Used non-atomic?

cevian: Also what are the locks taken by Get?

JLockerman: @cevian 
> what lock protects next? I don't see it in the code above.
Yeah, that's the one I elided 😊 . There's a mutex that protects next, and is acquired [here](https://github.com/timescale/timescale-prometheus/pull/147/files#diff-666b355b2c2f901c37b3cccc26156de4R52-R53l) and [here](https://github.com/timescale/timescale-prometheus/pull/147/files#diff-666b355b2c2f901c37b3cccc26156de4R67-R68) instead of the current lock.

> why is the Get update to Used non-atomic?
we don't have enough words. It's non-atomic because it uses
```go
if atomic.LoadUint32(&elem.used) == 0 {
    atomic.StoreUint32(&elem.used, 0)
}
```
instead of 
```go
atomic.CompareAndSwap(...)
```
for reasons described [here](https://github.com/timescale/timescale-prometheus/pull/147/files#r453666623), and that means the `insert` update to `used` can occur between the `if` and the `Store`. (It still is atomic in the sense that it uses atomic operations to shared memory, and does not have data-races)

JLockerman: > Also what are the locks taken by Get?

@cevian The exact same locks that are taken now; the `Get` code should not need to change.

iand: This looks better to me

iand:
A stylistic nit: conventionally the receiver in Go is a single character, e.g. `c` for cache. `self` stands out as odd.

JLockerman:
I pushed a commit with @iand's edits to the locking.

JLockerman:
I pushed a new version of the eviction code that eliminates bounds checking.
@iand I pushed a version of your benchmarks. I ended up changing them slightly to ensure that evictions happened.

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,289 @@
+package clockcache
+
+import (
+	"fmt"
+	"sync"
+	"sync/atomic"
+)
+
+// CLOCK based approximate LRU storing mappings from strings to
```
cevian, line 9: the mapping is between interfaces right not strings and int64s

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,289 @@
...
+		var inserted bool
+		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx])
+		if !inserted {
+			return idx
+		}
+	}
+	return len(keys)
+}
+
+func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey interface{}, canonicalValue interface{}, inserted bool) {
```
cevian, line 83:
```suggestion
func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey interface{}, canonicalValue interface{}, in_cache bool) {
```

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,289 @@
...
+	self.storage = newStorage
+}
+
+func (self *Cache) Len() int {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	return len(self.storage)
+}
+
+func (self *Cache) DebugString() string {
```
cevian, line 280: we should either delete this or make debugString

pkg/clockcache/cache.go
```diff
@@ -0,0 +1,289 @@
...
+	defer self.insertLock.Unlock()
+
+	oldMax := cap(self.storage)
+	if newMax <= oldMax {
+		return
+	}
+
+	newStorage := make([]element, 0, newMax)
+
+	// can we use copy here despite the data race
```
cevian, line 251:
```suggestion
	// cannot use copy here because of the data race on element.used
```

JLockerman:
benchmark results on a standard droplet
```
goos: linux
goarch: amd64
pkg: github.com/timescale/timescale-prometheus/pkg/clockcache
BenchmarkIntCache-8              	 4132209	       314 ns/op	       0 B/op	       0 allocs/op
BenchmarkStringCache-8           	 3913479	       341 ns/op	       0 B/op	       0 allocs/op
BenchmarkEviction/500-8          	 1377842	       862 ns/op	       0 B/op	       0 allocs/op
BenchmarkEviction/5000-8         	  165669	      7369 ns/op	       0 B/op	       0 allocs/op
BenchmarkEviction/50000-8        	  191414	     70761 ns/op	       0 B/op	       0 allocs/op
BenchmarkMembership/500-8        	20423635	        58.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkMembership/5000-8       	17888761	        65.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkMembership/50000-8      	12135523	        96.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkMembership/500000-8     	 6010875	       198 ns/op	       0 B/op	       0 allocs/op
BenchmarkNotFound/500-8          	24013395	        51.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkNotFound/5000-8         	20420546	        58.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkNotFound/50000-8        	17699750	        67.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkNotFound/500000-8       	13643972	        83.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkInsertUnderCapacity/500-8         	 7455434	       162 ns/o   7455434 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertUnderCapacity/5000-8        	 7307496	       164 ns/o   7307496 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertUnderCapacity/50000-8       	 6761834	       179 ns/o   6761834 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertUnderCapacity/500000-8      	 4576710	       233 ns/o   4576710 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertOverCapacity/500-8          	 5698213	       211 ns/o   5698213 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertOverCapacity/5000-8         	 6743119	       177 ns/o   6743119 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertOverCapacity/50000-8        	 7207106	       167 ns/o   7207106 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertOverCapacity/500000-8       	 6356412	       175 ns/o   6356412 inserts	         1.00 inserts/ops	       0 B/op	       0 allocs/op
BenchmarkInsertConcurrent/500-8            	 4390713	       280 ns/o       0 B/op	       0 allocs/op
BenchmarkInsertConcurrent/5000-8           	 3873616	       296 ns/o       0 B/op	       0 allocs/op
BenchmarkInsertConcurrent/50000-8          	 4385242	       317 ns/o       0 B/op	       0 allocs/op
BenchmarkInsertConcurrent/500000-8         	 3538814	       373 ns/o       0 B/op	       0 allocs/op
```

iand:
The insert counts on the benchmark look incorrect. 1 per op means every insert was new or evicted an existing entry. 

For comparison, this is what I saw when I originally wrote them in my branch:

```
BenchmarkInsertUnderCapacity/500-8         	12273985	        98.5 ns/op	         0.000030 inserts/op
BenchmarkInsertUnderCapacity/5000-8        	12334939	        97.7 ns/op	         0.000229 inserts/op
BenchmarkInsertUnderCapacity/50000-8       	11569123	       101 ns/op	         0.00194 inserts/op
BenchmarkInsertUnderCapacity/500000-8      	 8644075	       131 ns/op	         0.0201 inserts/op
BenchmarkInsertOverCapacity/500-8          	 9168835	       131 ns/op	         0.783 inserts/op
BenchmarkInsertOverCapacity/5000-8         	 9761977	       128 ns/op	         0.596 inserts/op
BenchmarkInsertOverCapacity/50000-8        	 9785414	       123 ns/op	         0.458 inserts/op
BenchmarkInsertOverCapacity/500000-8       	10211001	       135 ns/op	         0.350 inserts/op
```

The UnderCapacity benchmark shows that repeated iterations of the benchmark are almost always hitting a pre-existing key, which is what you expect if the cache is large enough to cover the test data.

The OverCapacity benchmark shows lots of evictions since the cache is only 25% of the size of the test data set.

JLockerman:
Ah whoops. I changed the meaning of the `bool` returned by `Insert` a never updated the benchmarks 😅 
(for those of you coming in from the future, at one point `Insert` returned `true` if the call to the function inserted the value into the cache and `false` if the value was already there and we're returning the old value. At the last minute I changed it to return `true` if the value is now in the cache and `false` if the `Insert` starved. My logic is that concurrent `Insert`s should be trying to insert the same value so it's not very useful to know which version inserted, you just always use the one returned, on the other hand, it can be very useful to know if the cache is overloaded)

PR #148, opened 2020-07-15
Add sql schema versioning and migration system

antekresic:


pkg/pgmodel/migrate.go
```diff
@@ -58,101 +57,159 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+	if err != nil {
+		return fmt.Errorf("unable to get migration scripts: %w", err)
+	}
+
+	sort.Slice(entries, func(i, j int) bool {
+		first, err := semver.Make(entries[i].Name())
+		if err != nil {
+			return false
+		}
+		second, err := semver.Make(entries[i].Name())
```
cevian, line 132: BUG: second also has index `i`. So both first and second will be the same thing.

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +57,159 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
 	if err != nil {
-		return fmt.Errorf("cannot create driver due to %w", err)
+		return fmt.Errorf("unable to get migration scripts: name %s, err %w", name, err)
 	}
 
-	_, err = db.Exec(timescaleInstall)
+	contents, err := replaceSchemaNames(f)
 	if err != nil {
-		return fmt.Errorf("timescaledb failed to install due to %w", err)
+		return fmt.Errorf("unable to read preinstall scripts: name %s, err %w", name, err)
```
cevian, line 106:
```suggestion
		return fmt.Errorf("unable to read migration scripts: name %s, err %w", name, err)
```

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +57,159 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
 	}
 
-	src, err := httpfs.New(migrations.SqlFiles, "/")
+	_, err = tx.Exec(string(contents))
 	if err != nil {
-		return err
+		return fmt.Errorf("error executing preinstall scripts: name %s, err %w", name, err)
```
cevian, line 111:
```suggestion
		return fmt.Errorf("error executing migration scripts: name %s, err %w", name, err)
```

pkg/pgmodel/migrations/generate.go
```diff
@@ -9,11 +9,19 @@
 package main
```
cevian, line 9: I am not sure I like the approach of merging files in a directory at the generate level. I'd prefer to have execMigrationFiles go through the directory and execute the files one at a time. The reason is debuggability, If we merge files here it's very hard to determine which sql file an error originates from; if. we do it in execMigrationFiles then we can print out the filename along with the error. Also from the point of view of separation of concerns I think keeping the filesystem layer as simple as possible is a good goal.

pkg/pgmodel/migrations/generate.go
```diff
@@ -9,11 +9,19 @@
...
```
cevian, line 10: all these changes need to be undone, correct?

pkg/pgmodel/migrate.go
```diff
@@ -58,99 +69,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
 	}
+	if err = setDBVersion(db, &appVersion, false); err != nil {
+		return fmt.Errorf("error setting clean app version to DB: %w", err)
+	}
+
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("unable to commit migration transaction: %w", err)
+	}
 
 	_, extErr := db.Exec(fmt.Sprintf(extensionInstall, extSchema))
```
cevian, line 312: I think some of the issues in the tests are because ExtensionIsInstalled is no longer set

pkg/pgmodel/migrate.go
```diff
@@ -6,48 +6,61 @@ package pgmodel
...
 )
 
 const (
 	timescaleInstall            = "CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;"
 	extensionInstall            = "CREATE EXTENSION IF NOT EXISTS timescale_prometheus_extra WITH SCHEMA %s;"
-	metadataUpdateWithExtension = "SELECT update_tsprom_metadata($1, $2, $3)"
+	metadataUpdateWithExtension = "SELECT %s.update_tsprom_metadata($1, $2, $3)"
 	metadataUpdateNoExtension   = "INSERT INTO _timescaledb_catalog.metadata(key, value, include_in_telemetry) VALUES ($1, $2, $3) ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value, include_in_telemetry = EXCLUDED.include_in_telemetry"
-)
+	createMigrationsTable       = "CREATE TABLE IF NOT EXISTS %s (version text not null primary key, dirty boolean not null)"
```
cevian, line 32: are we using the dirty flag for anything, if not then let's remove.

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,262 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+	if dirty {
+		return fmt.Errorf("version found in database dirty, cannot migrate")
+	}
+
+	// If already at correct version, nothing to migrate.
+	if dbVersion.Compare(appVersion) == 0 {
+		return nil
+	}
+
+	// If already at correct version, nothing to migrate.
```
cevian, line 277:
```suggestion
	// error if at a greater version
```

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,262 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+		}
+	} else {
+		if err = installVersionScripts(tx, dbVersion, appVersion); err != nil {
+			return err
+		}
+	}
+	if err = execMigrationFiles(tx, idempotentScripts); err != nil {
 		return err
 	}
+	if err = setDBVersion(db, &appVersion, false); err != nil {
```
cevian, line 312: this should use the tx not `db` object. Needs to be transactional.

pkg/pgmodel/migrations/generate.go
```diff
@@ -50,6 +51,8 @@ func (modTimeFileInfo) ModTime() time.Time {
 	return time.Time{}
 }
 
+var idempotentTOC = []string{}
```
cevian, line 54:
```suggestion

```

pkg/pgmodel/migrations/generate.go
```diff
@@ -9,11 +9,12 @@
 package main
 
 import (
-	"github.com/shurcooL/vfsgen"
 	"log"
 	"net/http"
 	"os"
 	"time"
+
+	"github.com/shurcooL/vfsgen"
```
cevian, line 17:
```suggestion
```

antekresic: Why remove this? Its used to generate the virtual filesystem for migration files.

go.mod
```diff
@@ -17,6 +18,7 @@ require (
 	github.com/jackc/pgerrcode v0.0.0-20190803225404-afa3381909a6
 	github.com/jackc/pgproto3/v2 v2.0.1
 	github.com/jackc/pgtype v1.2.0
+	github.com/jackc/pgx v3.2.0+incompatible
 	github.com/jackc/pgx/v4 v4.4.1
```
JLockerman, line 22: can we specify the version so we only have one pgx?

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/OneOfOne/xxhash v1.2.5 // indirect
 	github.com/allegro/bigcache v1.2.1
+	github.com/blang/semver v3.5.1+incompatible
```
JLockerman, line 9: this version looks suspicious

go.sum
```diff
@@ -118,6 +118,9 @@ github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6r
 github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kBD4zp0CCIs=
 github.com/bitly/go-hostpool v0.0.0-20171023180738-a3a6125de932/go.mod h1:NOuUCSz6Q9T7+igc/hlvDOUdtWKryOrtFyIVABv/p7k=
 github.com/bkaradzic/go-lz4 v1.0.0/go.mod h1:0YdlkowM3VswSROI7qDxhRvJ3sLhlFrRRwjwegp5jy4=
+github.com/blang/semver v1.1.0 h1:ol1rO7QQB5uy7umSNV7VAmLugfLRD+17sYJujRNYPhg=
+github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
+github.com/blang/semver v3.5.1+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
```
JLockerman, line 123: We should prune this so we only have one version of semver

pkg/pgmodel/migrate.go
```diff
@@ -6,48 +6,61 @@ package pgmodel
...
 	timescaleInstall            = "CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;"
 	extensionInstall            = "CREATE EXTENSION IF NOT EXISTS timescale_prometheus_extra WITH SCHEMA %s;"
-	metadataUpdateWithExtension = "SELECT update_tsprom_metadata($1, $2, $3)"
+	metadataUpdateWithExtension = "SELECT %s.update_tsprom_metadata($1, $2, $3)"
 	metadataUpdateNoExtension   = "INSERT INTO _timescaledb_catalog.metadata(key, value, include_in_telemetry) VALUES ($1, $2, $3) ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value, include_in_telemetry = EXCLUDED.include_in_telemetry"
-)
+	createMigrationsTable       = "CREATE TABLE IF NOT EXISTS %s (version text not null primary key)"
+	getVersion                  = "SELECT version FROM %s LIMIT 1"
+	setVersion                  = "INSERT INTO %s (version) VALUES ($1)"
+	truncateMigrationsTable     = "TRUNCATE %s"
```
JLockerman, line 35: if these aren't going to be user-settable then I think we should either do
```suggestion
	createMigrationsTable       = "CREATE TABLE IF NOT EXISTS prom_schema_migrations (version text not null primary key)"
	getVersion                  = "SELECT version FROM prom_schema_migrations LIMIT 1"
	setVersion                  = "INSERT INTO prom_schema_migrations (version) VALUES ($1)"
	truncateMigrationsTable     = "TRUNCATE prom_schema_migrations"
```
or at, if we must have the `migrationsTableName` constant, at least
```suggestion
	createMigrationsTable       = "CREATE TABLE IF NOT EXISTS "+ migrationsTableName + " (version text not null primary key)"
	getVersion                  = "SELECT version FROM "+ migrationsTableName + " LIMIT 1"
	setVersion                  = "INSERT INTO "+ migrationsTableName + " (version) VALUES ($1)"
	truncateMigrationsTable     = "TRUNCATE "+ migrationsTableName 
```
when reading through the control flow it's easier if you don't have to jump between the constants and `Sprintf` to figure out what's going on.

pkg/pgmodel/migrate.go
```diff
@@ -6,48 +6,61 @@ package pgmodel
...
+)
 
-type mySrc struct {
-	source.Driver
-}
+var (
+	ExtensionIsInstalled = false
+	toc                  = map[string][]string{
+		"idempotent": []string{},
+	}
```
JLockerman, line 47: this might be more readable on one lie if gofmt will let you, right now it's a bit difficult to determine where the type ends and the value begins

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+	var version semver.Version
+	res, err := db.Query(context.Background(), fmt.Sprintf(getVersion, migrationsTableName))
+
 	if err != nil {
-		return
+		return version, err
+	}
+
+	for res.Next() {
+		err = res.Scan(&version)
```
JLockerman, line 104: note to self: ensure this is tested (kinda surprised it works)

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+			}
+
+			if entry, err = file.Stat(); err != nil {
+				return fmt.Errorf("unable to stat migration script from toc: name %s, err %w", fullName, err)
+			}
+
+			entries = append(entries, entry)
+		}
+	} else {
+		entries, err = f.Readdir(-1)
```
JLockerman, line 156: this branch needs documentation and the function needs a high-level overview for what our strategy for finding migration files is

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+			first, err := strconv.Atoi(strings.Split(entries[i].Name(), "-")[0])
+			if err != nil {
+				return false
+			}
+			second, err := strconv.Atoi(strings.Split(entries[j].Name(), "-")[0])
+			if err != nil {
+				return false
+			}
+
+			return first < second
```
JLockerman, line 171: Is this checking file-order based on the assumption that the filenames are are `<filename>-<order num>`?
I'm betting this would read better split out into it's own function.

I'm also  not sure of the behavior that files without a version are incomparable, that might break things.

(Also, it probably doesn't matter, but it bugs me a little that we split and parse the string every time we call the ordering function, instead of just doing so once at the start)

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
+		if err != nil {
+			return fmt.Errorf("error executing migration script: name %s, err %w", fileName, err)
+		}
 	}
 
-	src, err := httpfs.New(migrations.SqlFiles, "/")
+	return nil
+}
+
+func installVersionScripts(tx pgx.Tx, dbVersion, appVersion semver.Version) error {
```
JLockerman, line 198: is this installing scripts up to the supplied version, greater than the supplied version or between them? Either way, it should be reflected in the naming: either something like
`installScriptsOlderThan`, something like `installScriptsNewerThan` or `installScriptsBetween`

JLockerman, line 198: note to self: ensure this is tested with multiple versions

pkg/pgmodel/migrate.go
```diff
@@ -58,101 +71,251 @@ func (t *mySrc) replaceSchemaNames(r io.ReadCloser) (io.ReadCloser, error) {
...
 var migrateMutex = &sync.Mutex{}
 
-// Migrate performs a database migration to the latest version
-func Migrate(db *sql.DB, versionInfo VersionInfo) (err error) {
-	migrateMutex.Lock()
-	defer migrateMutex.Unlock()
-	// The migration table will be put in the public schema not in any of our schema because we never want to drop it and
-	// our scripts and our last down script drops our shemas
-	driver, err := postgres.WithInstance(db, &postgres.Config{MigrationsTable: "prom_schema_migrations"})
+func execMigrationFiles(tx pgx.Tx, name string) error {
```
JLockerman, line 128: Does this execute _all_ the migration files? if so it should be named such. (If not, it could still use a more meaningful name; I'm still not entirely sure what this function does, and how it differs from `installVersionScripts`)

JLockerman: If, as I'm guessing now this is for non-versioned files, like the pre-init stuff that's only ever done once, or the idempotent things which are done each time, than `execNonVersionedMigrations` might work, though `name` probably can use a better name.

JLockerman:
Things remaining for me to review:

1. all the notes I left for myself above about testing
2. the new go libraries we're using

pkg/pgmodel/end_to_end_tests/main_test.go
```diff
@@ -102,17 +101,14 @@ func withDB(t testing.TB, DBName string, f func(db *pgxpool.Pool, t testing.TB))
...
 
 func performMigrate(t testing.TB, DBName string, connectURL string) {
-	dbStd, err := sql.Open("pgx", connectURL)
-	defer func() {
-		err := dbStd.Close()
-		if err != nil {
-			t.Fatal(err)
-		}
-	}()
+	dbStd, err := pgxpool.Connect(context.Background(), connectURL)
```
cevian, line 104: nit: can we change `dbStd` to `migratePool` or something? dbStd was because it wasn't a pgx connection but a sql.DB connection before

go.mod
```diff
@@ -27,9 +34,10 @@ require (
 	github.com/prometheus/prometheus v1.8.2-0.20200507164740-ecee9c8abfd1
 	github.com/spf13/cobra v0.0.7 // indirect
 	github.com/testcontainers/testcontainers-go v0.5.1
+	github.com/uber/jaeger-client-go v2.23.0+incompatible
```
JLockerman, line 37: as near as I can tell, all of the things here are either the same as in the old version, a removal, or an indirect. Would you try to remove the new lines and see if it still works? It seems odd that all of this should be needed

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
+	if dbVersion.Compare(semver.Version{}) == 0 {
+		if err = execMigrationFiles(tx, preinstallScripts); err != nil {
+			return err
+		}
 	} else {
-		_, _ = db.Exec(metadataUpdateNoExtension, key, value, true)
+		if err = upgradeVersion(tx, dbVersion, appVersion); err != nil {
+			return err
+		}
+	}
```
JLockerman, line 141: how about
```go
} else if err = upgradeVersion(tx, dbVersion, appVersion); err != nil {
	return err
}
```

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
+	for _, entry := range entries {
+		// Ignoring directories.
+		if entry.IsDir() {
+			continue
+		}
+
+		_, err := fmt.Sscanf(entry.Name(), "%d-%s", &prefix, &name)
+		// Ignore malformated file names.
+		if err != nil {
+			continue
```
JLockerman, line 277: I think we should log a warning

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
+		return "", err
 	}
+	s := buf.String()
+	s = strings.ReplaceAll(s, "SCHEMA_CATALOG", catalogSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_EXT", extSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_PROM", promSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_SERIES", seriesViewSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_METRIC", metricViewSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_DATA", dataSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_DATA_SERIES", dataSeriesSchema)
```
JLockerman, line 303: we should switch these two lines; it won't work if [done in the current order](https://play.golang.org/p/h7HeUFqqi6U)

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
+	}
+
+	versions := make(semver.Versions, 0, len(entries))
+
+	for _, e := range entries {
+		version, err := semver.Make(e.Name())
+
+		// Ignoring malformated entries.
+		if err != nil {
+			continue
```
JLockerman, line 328: we should log a warning here too

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
-		return r, err
+		return fmt.Errorf("app version is not semver format, aborting migration")
+	}
+	if err := ensureVersionTable(db); err != nil {
+		return fmt.Errorf("error ensuring version table: %w", err)
+	}
+	dbVersion, err := getDBVersion(db)
+	if err != nil {
+		return fmt.Errorf("failed to get the version from database: %w", err)
+	}
```
JLockerman, line 107: I suspect we'll eventually need a way to handle a malformed db-version, probably by forcing it to assume the db has a specific version, but I'm fine leaving that for a latter date

JLockerman:
Great work on the code cleanup btw, it's very readable now 👍

pkg/pgmodel/migrate.go
JLockerman: we should probably use a [strings.Replacer](https://golang.org/pkg/strings/#Replacer) for performance, though it won't fix the bug

pkg/pgmodel/migrate.go
```diff
@@ -26,133 +28,344 @@ const (
...
-	if withExtension {
-		_, _ = db.Exec(metadataUpdateWithExtension, key, value, true)
+	// No version in DB.
+	if dbVersion.Compare(semver.Version{}) == 0 {
+		if err = execMigrationFiles(tx, preinstallScripts); err != nil {
+			return err
+		}
 	} else {
-		_, _ = db.Exec(metadataUpdateNoExtension, key, value, true)
+		if err = upgradeVersion(tx, dbVersion, appVersion); err != nil {
```
cevian, line 138: @ante let's put an error before this line for now (temporarily), because this isn't tested yet. Otherwise after you put in the WARNINGS @JLockerman suggested plus swap the order of the two replace lines, it's good to merge. Does not need another review before merge

PR #149, opened 2020-07-16
Add info to logging during decompression

cevian:
Adds more logging when decompression happens.

### DIFF
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -675,7 +675,6 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
 }
 
 func decompressChunks(conn pgxConn, pending *pendingBuffer, table string) error {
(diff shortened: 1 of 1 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -686,6 +685,7 @@ func decompressChunks(conn pgxConn, pending *pendingBuffer, table string) error
 	if delayBy > maxDelayBy {
 		delayBy = maxDelayBy
 	}
+	log.Warn("msg", fmt.Sprintf("Table %s was compressed, decompressing", table), "table", table, "min-time", minTime, "age", time.Since(minTime), "delay-job-by", delayBy)
```
sgichohi, line 688: nit:could we add a metric as well so we can trend of Grafana

cevian: let's do that in another PR

PR #155, opened 2020-07-22
Ignore duplicate data points

### LINKED ISSUE #118: Make sure to align de-duplication logic with prometheus.
It seems like prometheus doesn't allow points with same metric/timestamp/series. We should align this as well. 

Also avoiding duplicates may be a worthwhile feature in any case.

JLockerman:
Prometheus does not allow data points with duplicate
(timestamp, series id). To replicate this, this commit adds a `UNIQUE`
index to each of the timestamp data tables, and adds a fallback path so
that COPY FROMs that fail due to duplicate keys are retried as
`INSERT ... ON CONFLICT DO NOTHING`s.

Fixes #118


pkg/pgmodel/pgx.go
```diff
@@ -643,37 +643,68 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+		&req.data.batch,
+	)
+	if err != nil {
+		err = copyFromErrorFallback(conn, req, err)
+	}
+	return err
+}
+
+// certain errors are recoverable
+func copyFromErrorFallback(conn pgxConn, req copyRequest, err error) error {
```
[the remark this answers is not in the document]
JLockerman: that's why the `copyFromErrorFallback` calls `copyData` after decompression; so if there're duplicate keys it'll be caught and handled. I don't believe it can happen the other way since we don't insert compressed data at the moment

pkg/pgmodel/pgx.go
```diff
@@ -643,37 +643,68 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+		}
+
+		req.data.batch.ResetPosition()
+		err = copyData(conn, req)
+		return err
+	}
+
+	// if we have duplicate data points try again using
+	// INSERT ... ON CONFLICT DO NOTHING to insert only the new ones
+	if strings.Contains(pgErr.Message, "duplicate key value violates unique constraint") {
```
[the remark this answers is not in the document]
JLockerman: should work, will fix

pkg/pgmodel/pgx.go
```diff
@@ -643,68 +643,88 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+// certain errors are recoverable, handle those we can
+//   1. if the data contains duplicates, switch to INSERT ... ON CONFLICT DO NOTHING
+//      so the duplicates are dropped and the other data can go in
+//   2. if the table is compressed, decompress and retry the insertion
 func copyFromErrorFallback(conn pgxConn, req copyRequest, err error) error {
+	// we use either COPY FROM or INSERT depending on if the dataset has duplicates
+	// if we do need to use INSERT, we need to keep doing so, otherwise we'll
+	// fail due to duplicates.
+	insertFn := doCopyFrom
+	for err != nil {
```
cevian, line 675: let's also put in a max err count as a fallback safety against infinite loops

pkg/pgmodel/pgx.go
```diff
@@ -643,68 +643,88 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+		insertFn, err = tryRecovery(conn, req, insertFn, err)
+		if insertFn == nil {
```
cevian, line 677:
```suggestion
		if insertFn == nil || err != nil {
```

pkg/pgmodel/pgx.go
```diff
@@ -643,68 +643,88 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+			return err
+		}
+		err = insertFn(conn, req)
+	}
+	return nil
+}
+
+type insertionFunction = func(conn pgxConn, req copyRequest) error
+
+func tryRecovery(conn pgxConn, req copyRequest, insertFn insertionFunction, err error) (insertionFunction, error) {
```
cevian, line 687: nit: let's document this better?

pkg/pgmodel/pgx.go
```diff
@@ -643,68 +643,88 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
-		err = copyData(conn, req)
-		return err
+		return insertFn, nil
 	}
 
 	log.Warn("msg", fmt.Sprintf("unexpected postgres error while inserting to %s", req.table), "error", pgErr.Error())
+	return nil, err
+}
+
+func doInsert(conn pgxConn, req copyRequest) error {
```
cevian, line 716: nit: I think doCopyFrom and doInsert should be next to each other in code.

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -452,6 +456,88 @@ func TestSQLIngest(t *testing.T) {
...
+
+		ts = []prompb.TimeSeries{
+			{
+				Labels: []prompb.Label{
+					{Name: MetricNameLabelName, Value: "test"},
+					{Name: "test", Value: "test"},
+				},
+				Samples: []prompb.Sample{
+					{Timestamp: 1, Value: 0.1},
+					{Timestamp: 10000000, Value: 0.1},
```
cevian, line 528:
```suggestion
					{Timestamp: 10000000, Value: 0.2},
					{Timestamp: 10000001, Value: 0.3},
```

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -452,6 +456,88 @@ func TestSQLIngest(t *testing.T) {
...
+				},
+			},
+		}
+
+		//ingest duplicate after compression
+		_, err = ingestor.Ingest(ts, NewWriteRequest())
+		if err != nil {
+			t.Fatal(err)
+		}
```
cevian, line 537: can we make sure there are 4 points in here (after my suggestions above) and they have the values we expect?

pkg/pgmodel/pgx.go
```diff
@@ -643,37 +643,104 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+	pgErr, ok := err.(*pgconn.PgError)
+	if !ok {
+		log.Warn("msg", fmt.Sprintf("unexpected error while inserting to %s", req.table), "error", err.Error())
+		return nil, err
+	}
+
+	// if we have duplicate data points try again using
+	// INSERT ... ON CONFLICT DO NOTHING to insert only the new ones
+	if pgErr.Code == "23505" {
+		log.Warn("msg", fmt.Sprintf("duplicate data points inserted to %s", req.table), "table", req.table)
```
blagojts, line 698: nit: since duplicate points were not inserted, maybe the error should be "duplicate data points in sample"?

pkg/pgmodel/pgx.go
```diff
@@ -643,37 +643,104 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+		req.data.batch.ResetPosition()
+		return doInsert, nil
```
blagojts, line 700: since this function internally is aware of the `doInsert` function, does it make sense to drop the insertFn argument, and just return `doCopyFrom` (as doInsert is returned)

JLockerman: the issue with that is an sequence of inserts which fails like
```
duplicate key
compressed data
```
after we decompress the data, we still need to `doInsert`, otherwise we'll fail again. (technically retry will catch that, but it'll be yet another round trip) (it's not the best that we have to restart the entire thing as it is, but that's the cost of large batches)

PR #156, opened 2020-07-23
Truncate long pod/service names (timescale/timescale-observability#11)

archen2019:


helm-chart/templates/svc-connector.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ .Release.Name }}-timescale-prometheus-connector
+  name: {{ template "connector.fullname" . }}-connector
```
cevian, line 4: do we need a `| trunc X` here too ?

PR #158, opened 2020-07-27
 Replace remaining uses of bigcache with clockcache

JLockerman:
Bigcache is a TTL cache, making it largely inappropriate for our use
case where the cache is mainly a garbage collection mechanism for dead
metrics and series, both of which we expect to change in a
time-correlated fashion. This commit switches the remaining usages to
our own clockcache, whose eviction policy we expect to be better suited
to our usage.

Right now the cache sizes are constants. It may be worth keeping a
hit-rate for each cache so we can start off with a smaller initial size
and grow the caches if the miss-rate is too large.

This PR depends on PR #146, and contains commits from that PR. After that PR is merged, this one will only contain the final commit


pkg/pgmodel/pgx.go
```diff
@@ -220,8 +219,8 @@ func NewPgxIngestorWithMetricCache(c *pgxpool.Pool, cache MetricCache, cfg *Cfg)
 
 // NewPgxIngestor returns a new Ingestor that write to PostgreSQL using PGX
 func NewPgxIngestor(c *pgxpool.Pool) (*DBIngestor, error) {
-	metrics, _ := bigcache.NewBigCache(DefaultCacheConfig())
-	cache := &MetricNameCache{metrics}
+	//TODO what size should we use?
```
cevian, line 222: I think 10k is ok here too.

pkg/pgmodel/cache_test.go
```diff
@@ -7,20 +7,14 @@ import (
...
 	"strings"
 	"testing"
-	"time"
 
-	"github.com/allegro/bigcache"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/timescale/timescale-prometheus/pkg/clockcache"
 )
 
 func TestBigCache(t *testing.T) {
```
cevian, line 15: change function name

pkg/pgmodel/cache.go
```diff
@@ -0,0 +1,61 @@
...
+	if !ok {
+		return "", ErrEntryNotFound
+	}
+	return result.(string), nil
+}
+
+// Set stores table name for specified metric.
+func (m *MetricNameCache) Set(metric string, tableName string) error {
+	// deep copy the strings so the original memory doesn't need to stick around
+	metricBuilder := strings.Builder{}
```
cevian, line 53: is this really the best way to copy a string?

JLockerman: yup, I think so 😞 
The other way I know is something like
```go
newbytes := make([]byte, len(str))
copy(newbytes, str)
return BytesToString(newbytes)

func BytesToString(bytes []byte) (s string) {
	if len(bytes) == 0 {
		return ""
	}

	str := (*reflect.StringHeader)(unsafe.Pointer(&s))
	str.Data = uintptr(unsafe.Pointer(&bytes[0]))
	str.Len = len(bytes)

	return s
}
```
which has the disadvantage of using `unsafe`, and is basically what `string.Builder` does internally anyway.

pkg/pgmodel/pgx.go
JLockerman: My bigger question for this, and the [series cache](https://github.com/timescale/timescale-prometheus/pull/158/files#diff-25015b1d7c86d8c5fab194419dee6c51R212) is if we should start the cache smaller and grow it to some maximum if the miss-rate is too high. Generally the smallest cache which fits the data should perform better...

ksten-ab: I suspect the performance impact will be negligible for Metric cache (favoring a bit oversized), but Series cache will probably be very dependent on the size of records--- especially if undersized for large datasets.

Whether or not it can be grown automatically, maybe the default size be configured via CLI Args/ENV Vars?

JLockerman: @ksten-ab Yup, we're definitely adding CLI configuration before this PR, _probably_ not dynamic-expansion though

cmd/timescale-prometheus/main.go
```diff
@@ -194,6 +195,61 @@ func main() {
 	}
 	defer client.Close()
 
+	cachedMetricNames := prometheus.NewCounterFunc(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Name:      "cached_metric_names",
```
cevian, line 200:
```suggestion
		Name:      "metric_names_cache_used",
```

cmd/timescale-prometheus/main.go
```diff
@@ -194,6 +195,61 @@ func main() {
...
+		Namespace: promNamespace,
+		Name:      "metric_names_cache_capacity",
+		Help:      "Maximum number of elements in the metric names cache.",
+	}, func() float64 {
+		return float64(client.MetricNamesCacheCapacity())
+	})
+
+	cachedSeriesSets := prometheus.NewCounterFunc(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Name:      "cached_series_sets",
```
cevian, line 216:
```suggestion
		Name:      "series_set_cache_used",
```

cmd/timescale-prometheus/main.go
```diff
@@ -194,6 +195,61 @@ func main() {
...
+		Namespace: promNamespace,
+		Name:      "series_set_cache_capacity",
+		Help:      "Maximum number of elements in the series-set cache.",
+	}, func() float64 {
+		return float64(client.SeriesCacheCapacity())
+	})
+
+	cachedLabels := prometheus.NewCounterFunc(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Name:      "cached_labels",
```
cevian, line 232:
```suggestion
		Name:      "labels_cache_used",
```

pkg/pgmodel/pgx.go
```diff
@@ -220,8 +216,8 @@ func NewPgxIngestorWithMetricCache(c *pgxpool.Pool, cache MetricCache, cfg *Cfg)
...
```
cevian, line 219: remove the TODO

PR #159, opened 2020-07-28
Add proper CORS handling by the HTTP API

blagojts:
A flag can be set by the user to specify the allowed
origins to access the HTTP API. A wrapper handler
sets the proper CORS headers if the flag is enabled.

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/util/httputil"
 	"github.com/timescale/timescale-prometheus/pkg/promql"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/api/common.go
```diff
@@ -22,8 +24,18 @@ var (
 	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
 )
 
-func setHeaders(w http.ResponseWriter, res *promql.Result, warnings storage.Warnings) {
-	w.Header().Set("Access-Control-Allow-Origin", "*")
+type Config struct {
+	AllowedOrigin *regexp.Regexp
+}
+
+func corsWrapper(conf *Config, f http.HandlerFunc) http.HandlerFunc {
```
cevian, line 31: Design question: why do we have this as a separate wrapper? Why isn't this part of `setResponseHeaders`?

blagojts: 1. I wanted to re-use the prometheus httputil.SetCORS function, so we're more future proof against prometheus changing their api code
2. could write better tests where we test if the cors settings are ok once, not part of the tests for each endpoint

pkg/api/label_values.go
```diff
@@ -12,15 +12,23 @@ import (
 	"net/http"
 )
 
-func LabelValues(queriable *query.Queryable) http.Handler {
-	hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+func LabelValues(conf *Config, queryable *query.Queryable) http.Handler {
+	hf := labelValues(queryable)
+	if conf.AllowedOrigin != nil {
+		hf = corsWrapper(conf, hf)
+	}
```
JLockerman, line 19: 1. how come this branch is repeated around each call to `corsWrapper` instead of being within `corsWrapper`?
2. How come each API function calls `corsWrapper` instead of it wrapping the API function like we do for `timeHandler`?

pkg/api/common.go
cevian: Fixed 1. About 2 i'm kinda torn. Your way would be less code but not all api endpoints do/should do this and from a layering point of view perhaps the api endpoints should know this, not the routing code

JLockerman: (I think you accidentally responed to Blagoj instead of me)
fair; I was inclined to think the other way, but I guess I buy that it's part of the API.

PR #166, opened 2020-07-30
Fix bugs with concurrent decompression

cevian:
- Prevent trying to decompress already decompressed chunks
- Handle alter_job_schedule not locking correctly


pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1196,6 +1212,19 @@ BEGIN
...
     LOOP
+
+        --lock the chunk exclusive.
+        EXECUTE format('LOCK %I.%I;', chunk_row.schema_name, chunk_row.table_name);
+        --double check it's still compressed.
+        PERFORM c.*
+        FROM _timescaledb_catalog.chunk c
+        WHERE c.id = chunk_row.id AND c.compressed_chunk_id IS NOT NULL;
+
+        IF NOT FOUND THEN
```
davidkohn88, line 1223: It might be cleaner to just do the notice and decompression in an if block rather than having a commit/continue here, but I don't really mind it either way.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1166,7 +1166,23 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+DECLARE
+  bgw_job_id int;
+BEGIN
+        SELECT job_id INTO bgw_job_id
+		FROM _timescaledb_config.bgw_policy_compress_chunks p
+		INNER JOIN _timescaledb_catalog.hypertable h ON (h.id = p.hypertable_id)
+		WHERE h.schema_name = 'SCHEMA_DATA' and h.table_name = ht_table;
 
+        --alter job schedule is not currently concurrency-safe (timescaledb issue #2165)
+        PERFORM pg_advisory_xact_lock(12377, bgw_job_id);
```
davidkohn88, line 1180: Oh. How'd you get this number? Just a random one?

cevian: yep random

PR #167, opened 2020-07-30
Add concurrent write test

cevian:


cevian:
This test was used to find the bugs in #165 and #166.

pkg/pgmodel/end_to_end_tests/write_test.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return http.NewRequest(
+		"POST",
+		u.String(),
+		strings.NewReader(body),
+	)
+}
+
+func sendWriteRequest(t testing.TB, router *route.Router, ts []prompb.TimeSeries) {
+	req, err := getHTTPWriteRequest(&prompb.WriteRequest{Timeseries: ts})
```
antekresic, line 105: Should we use our pool here as well?

cevian: I don't think so since this is the "client" generating the request and it wouldn't use the same pool or be on the same machine

cevian:
Yes, verified that it finds the bugs in both 165 and 166.

PR #174, opened 2020-08-03
Fix PromQL Differences for delta

cevian:
- Pass down the error handling better and log it.
- Also make use-docker flag only apply to the database.
- Improve promql tests to catch more cases. Some of the
  cases don't pass. To be fixed in a followup commit.

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -184,7 +184,9 @@ impl GapfillDeltaTransition {
         if self.window.back().map_or(false, |(prev, _)| *prev > time) {
             elog!(Error, "inputs must be in ascending time order")
         }
-        self.window.push_back((time, val));
+        if time >= self.current_window_min {
+            self.window.push_back((time, val));
+        }
     }
 
     fn in_current_window(&self, time: TimestampTz) -> bool {
--- a/pkg/pgmodel/end_to_end_tests/main_test.go
+++ b/pkg/pgmodel/end_to_end_tests/main_test.go
@@ -40,15 +40,17 @@ func TestMain(m *testing.M) {
 		flag.Parse()
 		ctx := context.Background()
 		_ = log.Init("debug")
-		if !testing.Short() && *useDocker {
+		if !testing.Short() {
 			var err error
 
-			pgContainerTestDataDir = generatePGTestDirFiles()
+			if *useDocker {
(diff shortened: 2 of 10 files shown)


extension/src/lib.rs
```diff
@@ -184,7 +184,9 @@ impl GapfillDeltaTransition {
         if self.window.back().map_or(false, |(prev, _)| *prev > time) {
             elog!(Error, "inputs must be in ascending time order")
         }
-        self.window.push_back((time, val));
+        if time >= self.current_window_min {
+            self.window.push_back((time, val));
+        }
```
JLockerman, line 189: shouldn't this be a `elog(Error, ...)` in the aggregate functions?

pkg/pgmodel/end_to_end_tests/main_test.go
```diff
@@ -40,15 +40,17 @@ func TestMain(m *testing.M) {
 		flag.Parse()
 		ctx := context.Background()
 		_ = log.Init("debug")
-		if !testing.Short() && *useDocker {
+		if !testing.Short() {
```
JLockerman, line 43: Should we have a warning if we're `!Short && !useDocker`?

JLockerman: In general, I'm not sure I agree with this change. From what I recall, when we originally added `useDocker` it was to be a strict subset of `!Short`, so you would add it if you wanted long tests but not the docker tests. Do we no longer think that's a needed case, or do we think that separating out into `end_to_end_tests` suffices?

JLockerman: (I know at least me and @davidkohn88 don't always have docker running because of it's CPU implications, so being able to run only those tests that don't use docker seems like it might be useful still)

cevian: I use it to run tests against a local db instance. That's very useful for debugging. In that case I wan't to run this test with use-docker=0. This was actually the use-case I created -use-docker for originally

extension/src/lib.rs
cevian: Not if you have a step of 30m and a range of 5m. Then every time you move windows you'll still have. 25 minutes of data to throw out

PR #175, opened 2020-08-03
Fix for min/max start/end timestamp values for remote read

### LINKED ISSUE #125: series endpoint via remote read with no start time causes PG error
Gets a timezone out of range error

https://timescaledb.slack.com/archives/C011FC0PPC5/p1592251003082800

antekresic:
When start/end timestamp values are not sent to a query, Prometheus sets them
min/max timestamp values. These values are outside of Prometheus date/time
ranges and cause errors when used in SQL queries. This fix limits those values
to year 1970 for min and year 3000 for max.

Closes #125

### DIFF
--- a/pkg/pgmodel/query_builder.go
+++ b/pkg/pgmodel/query_builder.go
@@ -6,6 +6,7 @@ package pgmodel
 
 import (
 	"fmt"
+	"math"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


pkg/pgmodel/query_builder.go
```diff
@@ -53,6 +53,11 @@ const (
 	AND time >= '%[4]s'
 	AND time <= '%[5]s'
 	GROUP BY s.id`
+
+	// Jan 1st, 1970
+	millisecondMin = 0
```
cevian, line 58: I don't love limiting us to 1970 and 3000. There will def be use cases that want to insert at least older data I'd suggest setting 
```
minTime = time.Unix(math.MinInt64/1000+62135596801, 0)
maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999)
```

pkg/pgmodel/query_builder.go
```diff
@@ -386,6 +391,12 @@ func toMilis(t time.Time) int64 {
 }
 
 func toRFC3339Nano(milliseconds int64) string {
+	if milliseconds < millisecondMin {
```
cevian, line 394: My suggestion here is 
```
if milliseconds == minTime {
   return '-Infinity'
}
if milliseconds == maxTime {
   return 'Infinity'
}
```

PR #177, opened 2020-08-03
Use multi-valued inserts rather than batches

davidkohn88:
Multi-valued inserts are significantly more performant than batched
single valued inserts, it also simplifies the code for finding how
many duplicates there were etc.

### DIFF
--- a/pkg/pgmodel/pgx.go
+++ b/pkg/pgmodel/pgx.go
@@ -167,18 +167,13 @@ func (t *SampleInfoIterator) Next() bool {
 }
 
 // Values returns the values for the current row
-func (t *SampleInfoIterator) Values() ([]interface{}, error) {
+func (t *SampleInfoIterator) Values() (time.Time, float64, SeriesID) {
 	info := t.sampleInfos[t.sampleInfoIndex]
 	sample := info.samples[t.sampleIndex]
-	row := []interface{}{
-		model.Time(sample.Timestamp).Time(),
-		sample.Value,
-		info.seriesID,
-	}
 	if t.minSeen > sample.Timestamp {
 		t.minSeen = sample.Timestamp
 	}
-	return row, nil
+	return model.Time(sample.Timestamp).Time(), sample.Value, info.seriesID
 }
 
 // Err returns any error that has been encountered by the CopyFromSource. If
@@ -232,7 +227,7 @@ func newPgxInserter(conn pgxConn, cache MetricCache, cfg *Cfg) (*pgxInserter, er
 	numCopiers := maxProcs*ConnectionsPerProc - maxProcs
 	toCopiers := make(chan copyRequest, numCopiers)
 	for i := 0; i < numCopiers; i++ {
-		go runCopyFrom(conn, toCopiers)
+		go runInserter(conn, toCopiers)
 	}
 
 	inserter := &pgxInserter{
@@ -630,7 +625,7 @@ func (h *insertHandler) flushPending() {
 	h.pending = pendingBuffers.Get().(*pendingBuffer)
 }
 
-func runCopyFrom(conn pgxConn, in chan copyRequest) {
+func runInserter(conn pgxConn, in chan copyRequest) {
 	for {
 		req, ok := <-in
 		if !ok {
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/pgx.go
```diff
@@ -647,112 +647,68 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+	times := make([]time.Time, 0, numRows)
+	vals := make([]float64, 0, numRows)
+	series := make([]int, 0, numRows)
+	for req.data.batch.Next() {
+		row, _ := req.data.batch.Values()
+		times = append(times, row[0].(time.Time))
+		vals = append(vals, row[1].(float64))
+		series = append(series, int(row[2].(SeriesID)))
+	}
+	queryString := fmt.Sprintf("INSERT INTO %s(time, value, series_id) SELECT * FROM unnest($1::TIMESTAMPTZ[], $2::DOUBLE PRECISION[], $3::INT[]) a ON CONFLICT DO NOTHING", pgx.Identifier{dataSchema, req.table}.Sanitize())
```
davidkohn88, line 703: this now should be bigint[]

pkg/pgmodel/pgx.go
```diff
@@ -647,112 +647,68 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+	for i := range req.data.batch.sampleInfos {
+		numRows += len(req.data.batch.sampleInfos[i].samples)
 	}
-	var resp pgx.BatchResults
-	resp, err = conn.SendBatch(context.Background(), insertBatch)
+	times := make([]time.Time, 0, numRows)
+	vals := make([]float64, 0, numRows)
+	series := make([]int64, 0, numRows)
+	for req.data.batch.Next() {
+		row, _ := req.data.batch.Values()
```
cevian, line 698: Can we make data.batch.Values() actually return the typed values instead of []interface() so we don't need to do the typecast below?

pkg/pgmodel/pgx.go
```diff
@@ -647,112 +647,68 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
...
+		times = append(times, row[0].(time.Time))
+		vals = append(vals, row[1].(float64))
+		series = append(series, int64(row[2].(SeriesID)))
+	}
```
cevian, line 702: can we check that len(times) == numRows? and throw an error if not aligned?

pkg/pgmodel/pgx.go
```diff
@@ -636,9 +636,9 @@ func runCopyFrom(conn pgxConn, in chan copyRequest) {
 		if !ok {
```
cevian, line 636: I'd change the name of this function since it's no longer doing a Copy

pkg/pgmodel/pgx.go
JLockerman: Ok. In a future version we're going to remove the iterator entirely; now that we're not doing CopyFrom we don't need it

pkg/pgmodel/pgx.go
JLockerman: it'll be a panic but sure. (It's a panic because `numRows` is gathered by iterating through the same slices; if there's an error it's a non-recoverable bug)

pkg/pgmodel/pgx.go
davidkohn88: should we just do that now? Or is that too much pre-release so we're waiting?

pkg/pgmodel/pgx_test.go
```diff
@@ -54,9 +57,33 @@ func (m *mockPGXConn) UseDatabase(dbName string) {
...
+			end += 1
+		}
+		tableName := sql[len("INSERT INTO "):end]
+		m.CopyFromTableName = append(m.CopyFromTableName, tableName)
+
+		times := arguments[0].([]time.Time)
+		vals := arguments[1].([]float64)
+		series := arguments[2].([]int64)
+
+		m.Times = append(m.Times, times...)
```
cevian, line 77: should we actually use these for checks?

pkg/pgmodel/pgx_test.go
```diff
@@ -608,19 +631,15 @@ func TestPGXInserterInsertData(t *testing.T) {
...
 
-			for i, metric := range mock.CopyFromTableName {
-				name := metric[1]
-				for metric, tableName := range mockMetrics.metricCache {
-					if tableName == name {
-						name = metric
-					}
-				}
-				result := c.rows[name]
+			// for i, metric := range mock.CopyFromTableName {
```
cevian, line 634: do we want to do any check with mock.Times etc here?

JLockerman: nah, the data is never populated anyway

PR #184, opened 2020-08-04
Update tests to detect labels issue

JLockerman:
Our labels were too well behaved; the order in which we queried them
was always the order they were stored. This adds another label to
ensure that there's some noise in how we query.

### DIFF
--- a/pkg/pgmodel/sql_reader.go
+++ b/pkg/pgmodel/sql_reader.go
@@ -208,10 +208,12 @@ func (q *pgxQuerier) getLabelsForIds(ids []int64) (lls labels.Labels, err error)
 	numHits := q.labels.GetValues(keys, values)
 
 	if numHits < len(ids) {
-		err = q.fetchMissingLabels(keys[numHits:], ids[numHits:], values[numHits:])
+		var numFetches int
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/sql_reader.go
```diff
@@ -222,13 +224,13 @@ func (q *pgxQuerier) getLabelsForIds(ids []int64) (lls labels.Labels, err error)
 	return
 }
 
-func (q *pgxQuerier) fetchMissingLabels(misses []interface{}, missedIds []int64, newLabels []interface{}) error {
+func (q *pgxQuerier) fetchMissingLabels(misses []interface{}, missedIds []int64, newLabels []interface{}) (numNewLabels int, err error) {
```
cevian, line 227: nit: should we return newLabes as []interface{} as a return to make it clear this is the thing returned? I always get confused which things are returned out of here. Also a comment may help.

JLockerman: I tried that the first time I made this edit, it didn't end up working because this is the suffix of a slice whose other values are also needed. I'll look into a better fix in a latter PR.

PR #185, opened 2020-08-04
Change migration system and add tests

cevian:


cmd/timescale-prometheus/version_info.go
```diff
@@ -1,6 +1,20 @@
 package main
 
 var (
-	Version    = "0.1.0-beta.1"
+	// Rules for Versioning:
+	// A release version cannot contains `dev` in it's pre-release tag.
```
davidkohn88, line 5: nit: contain not contains

cmd/timescale-prometheus/version_info.go
```diff
@@ -1,6 +1,20 @@
...
+	// The next development cycle MUST meet two requirements:
+	// 1) The development cycle version must be higher in semver than the release.
+	//    Thus you either need to increase the patch or pre-release tag. You
+	//    SHOULD always increase the version by the smallest amount possible
+	//    (e.g. prefer 0.1.2 to 0.2.0)
+	// 2) It must include `dev` as last part of prelease.
```
davidkohn88, line 11: prelease->pre-release?

cmd/timescale-prometheus/version_info.go
```diff
@@ -1,6 +1,20 @@
...
+	//
+	// Example if releasing 0.1.1, the next development cycle is 0.1.2-dev
+	// Example if releasing 0.1.3-beta the next development cycle is 0.1.3-beta.1.dev
+	//
+	// When introducing a new SQL migration script always increment the numeral after
+	// the `dev` tag. Add the SQL file corresponding to the /new/ version.
+	Version    = "0.1.0-beta.2-dev.0"
```
davidkohn88, line 18: I'm confused by this, isn't that dev.0 supposed to be for sql files not for the version? I thought that pre-release versions had dev as the last thing in the tag?

davidkohn88: (I might be wrong about this, but then the comment should probably be clarified?)

JLockerman: @davidkohn88 
This is to handle upgrades from `master` to `master`; after developing version `0.1.1` the next master release will be `0.1.2-dev.0`, the one after will be `0.1.2-dev.1`, the one after `0.1.2-dev.2` etc.
This allows us to determine which `master` version a given user was on when they come to us with bugs.

pkg/pgmodel/migrate.go
```diff
@@ -34,24 +35,21 @@ const (
...
-	toc = map[string][]string{
+	tableOfContets = map[string][]string{
 		"idempotent": {
 			"base.sql",
 			"matcher-functions.sql",
 		},
 	}
-	migrateMutex = &sync.Mutex{}
+	migrateMutex            = &sync.Mutex{}
+	migrationFileNameRegexp = regexp.MustCompile(`([[:digit:]]*)-[[:word:]]*.sql`)
```
JLockerman, line 52: (note to self e.g. `6-foo.sql`, actually, should that be a comment?)
@cevian should this be
```
([[:digit:]]+)-[[:word:]]*.sql
```
do we use this for files that don't have a version?

pkg/pgmodel/migrate.go
```diff
@@ -105,16 +103,74 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
...
-	dbVersion, err := getDBVersion(db)
+
+	dbVersion, err := getDBVersion(t.db)
 	if err != nil {
 		return fmt.Errorf("failed to get the version from database: %w", err)
 	}
 
-	// If already at correct version, nothing to migrate.
+	// If already at correct version, nothing to migrate on proper release.
+	// On dev versions, idempotent files need to be reapplied.
```
JLockerman, line 149: how come idempotent files need to be reapplied even if we're on the correct version?
(If it's just for testing, we should probably have a different flag to force the reapplication)

pkg/pgmodel/migrate.go
```diff
@@ -313,47 +366,90 @@ func replaceSchemaNames(r io.ReadCloser) (string, error) {
 	return s, err
 }
 
+//A migration file is inside a directory that is a semver version number. The filename itself has the format
+//<migration file number)-<description>.sql. That file correspond to the semver of <dirname>.<migration file number>
+//where the migration file number is always part of prerelease tag.
+//All app versions >= (inclusive) migration files's semver will include the migration file
```
JLockerman, line 372: maybe add something like
```
// That is if we're on version `0.1.1-dev.3` then we'll include all sql files up to `0.1.1-dev/3-foo.sql`
```

pkg/pgmodel/migrate.go
```diff
@@ -313,47 +366,90 @@ func replaceSchemaNames(r io.ReadCloser) (string, error) {
...
+	n, err := fmt.Sscanf(matches[1], "%d", &migrationFileNumber)
+	if n != 1 || err != nil {
+		return nil, fmt.Errorf("unable to parse the migration file name %v: %w", fileName, err)
+	}
+
+	migrationFileVersion, err := semver.Make(dirName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse version from directory %v: %w", dirName, err)
+	}
+	migrationNumberPrVersion, err := semver.NewPRVersion(fmt.Sprintf("%d", migrationFileNumber))
```
JLockerman, line 388: I'm inclined to read `PrVersion` as "dull request version" instead of "pre-release version", maybe `DevVersion` instead?

pkg/pgmodel/end_to_end_tests/migrate_test.go
```diff
@@ -40,3 +44,151 @@ func TestMigrateTwice(t *testing.T) {
...
+	for rows.Next() {
+		var value string
+		err = rows.Scan(&value)
+		if err != nil {
+			t.Fatal(err)
+		}
+		found = append(found, value)
+	}
+	if !reflect.DeepEqual(expected, found) {
+		t.Errorf("wrong values in DB\nexpected:\n\t%v\ngot:\n\t%v", expected, found)
```
JLockerman, line 64: I've found in the pgx tests that for anything with control characters (such as newlines) it pays to go through the array line-by-line and print each as a string; that way on an error you can put both versions into a diff too and it'll tell you exactly where the difference is. I've found that wen everything is on a single line, like here, the output is much less readable.

pkg/pgmodel/end_to_end_tests/migrate_test.go
```diff
@@ -40,3 +44,151 @@ func TestMigrateTwice(t *testing.T) {
...
+			"migration 0.10.1=1",
+			"idempotent 1",
+			"idempotent 2",
+			"migration 0.10.1=2",
+			"idempotent 1",
+			"idempotent 2",
+			"migration 0.10.2-beta=1",
+			"idempotent 1",
+			"idempotent 2",
+		}
```
JLockerman, line 107: I like this test, but keeping all the values in one static slice makes in incredibly difficult to determine what the individual migrations are supposed to do. I recommend  starting with the log values for the initial migration and `append`ing the later values as we need them, e.g.
```go
expected := []string{
	"setup",
	"idempotent 1",
	"idempotent 2",
}

mig := pgmodel.NewMigrator(db, test_migrations.MigrationFiles, testTOC)

err := mig.Migrate(semver.MustParse("0.1.1"))
if err != nil {
	t.Fatal(err)
}

verifyLogs(t, db, expected)

...

expected = append(expected,
	"migration 0.2.0",
	"idempotent 1",
	"idempotent 2",
)

err = mig.Migrate(semver.MustParse("0.2.0"))
if err != nil {
	t.Fatal(err)
}
verifyLogs(t, db, expected)
...
```
That'll make it very simple to determine what the effect of each migration should be.

pkg/pgmodel/migrate.go
cevian: This is just on dev/master versions. We don't want to require bumping versions for idempotent changes.

davidkohn88: Please add this clarification to the comment!

cmd/timescale-prometheus/version_info.go
```diff
@@ -1,6 +1,20 @@
 package main
 
 var (
-	Version    = "0.1.0-beta.1"
+	// Rules for Versioning:
+	// A release version cannot contain `dev` in it's pre-release tag.
```
davidkohn88, line 5: it's -> its

PR #186, opened 2020-08-05
Fix extension installed check for already migrated DB schema

### LINKED ISSUE #183: ExtensionIsInstalled is not set correctly on restart
ExtensionIsInstalled is not set when restarting because `dbVersion.Compare(appVersion) == 0 `

antekresic:
Closes #183

### DIFF
--- a/pkg/pgmodel/end_to_end_tests/migrate_test.go
+++ b/pkg/pgmodel/end_to_end_tests/migrate_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/timescale/timescale-prometheus/pkg/internal/testhelpers"
+	"github.com/timescale/timescale-prometheus/pkg/pgmodel"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


antekresic:
@cevian would this test catch that? Checking the flag after second migration:
https://github.com/timescale/timescale-prometheus/pull/186/files#diff-84dae51ce95be25c580ac1f2b9adc643R45

Not really sure what test we could do except that.

pkg/pgmodel/end_to_end_tests/migrate_test.go
```diff
@@ -37,6 +38,12 @@ func TestMigrateTwice(t *testing.T) {
 	}
 	testhelpers.WithDB(t, *testDatabase, testhelpers.NoSuperuser, func(db *pgxpool.Pool, t testing.TB, connectURL string) {
 		performMigrate(t, connectURL)
+		if *useExtension && !pgmodel.ExtensionIsInstalled {
+			t.Errorf("extension is not installed, expected it to be installed")
+		}
```
cevian, line 43:
```suggestion
		}
		//reset the flag to make sure it's set correctly again.
		pgmodel.ExtensionIsInstalled = false
```

PR #188, opened 2020-08-07
Add concurrent endpoint tests

archen2019:
* Moved each endpoint to separate file
* Created testRequestConcurrent function for testing endpoints concurrently


pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -673,3 +57,63 @@ func testRequest(req *http.Request, handler http.Handler, client *http.Client, c
 		}
 	}
 }
+
+func testRequestConcurrent(reqs []*http.Request, logs []string, handler http.Handler, client *http.Client, comparator resultComparator) func(*testing.T) {
```
cevian, line 61: I would consider changing reqs `[]*http.Request, logs []string` to a `cases []*testCase` where the testCase struct contains one http.Request and a log string.

PR #195, opened 2020-08-17
Fixup error handling of sql_reader

JLockerman:
Make sure errors are propagated or handled in all cases. Notably `appendTsRows()` now bubbles up any error it may encounter.

### DIFF
--- a/pkg/pgmodel/sql_reader.go
+++ b/pkg/pgmodel/sql_reader.go
@@ -117,7 +117,7 @@ func (q *pgxQuerier) Query(query *prompb.Query) ([]*prompb.TimeSeries, error) {
 
 	results, err := buildTimeSeries(rows, q)
 
-	return results, nil
+	return results, err
 }
 
 func (q *pgxQuerier) LabelNames() ([]string, error) {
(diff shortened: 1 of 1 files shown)


pkg/pgmodel/sql_reader.go
```diff
@@ -357,21 +364,24 @@ func (q *pgxQuerier) querySingleMetric(metric string, filter metricTimeRangeFilt
...
 	// TODO this allocation assumes we usually have 1 row, if not, refactor
-	tsRows := appendTsRows(make([]timescaleRow, 0, 1), rows)
-	return tsRows, topNode, nil
+	tsRows, err := appendTsRows(make([]timescaleRow, 0, 1), rows)
+	return tsRows, topNode, err
 }
 
-func appendTsRows(out []timescaleRow, in pgx.Rows) []timescaleRow {
+func appendTsRows(out []timescaleRow, in pgx.Rows) ([]timescaleRow, error) {
+	if in.Err() != nil {
```
cevian, line 372: I think this should be at the end of the function after the in.Next calls, not the beginning

JLockerman: I think we need to check in the beginning also, in case it comes in with an error

PR #196, opened 2020-08-17
Probe database for connection limit, and make it configurable

### LINKED ISSUE #187: Use too much connections
* timescaledb
  * docker install - latest
  * 16G max-connection 500
* tow timescale/timescale-prometheus - docker installed latest
  * used all connections, still reporting `FATAL: sorry, too many clients already (SQLSTATE 53300)`
* promethes - 10 node-exporter

timescale-prometheus use prepared statement, so can not use transaction polling to reduce the connections.

JLockerman:
This commit contains two changes:
  1. At startup, we now probe the database to determine the number of
     connections it can handle in order to ensure we don't go over
     that limit
  2. We add a command line flag to override this setting for instances
     where our heuristic is incorrect.

This should make the connector a better-behaved client of the database,
making it much less likely to starve out other clients, or itself.

fixes #187

### DIFF
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 
+	pgx "github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 
 	"github.com/timescale/timescale-prometheus/pkg/clockcache"
@@ -21,18 +22,20 @@ import (
 
 // Config for the database
 type Config struct {
-	host             string
-	port             int
-	user             string
-	password         string
-	database         string
-	sslMode          string
-	dbConnectRetries int
-	AsyncAcks        bool
-	ReportInterval   int
-	LabelsCacheSize  uint64
-	MetricsCacheSize uint64
-	SeriesCacheSize  uint64
+	host                    string
+	port                    int
+	user                    string
+	password                string
+	database                string
+	sslMode                 string
+	dbConnectRetries        int
+	AsyncAcks               bool
+	ReportInterval          int
+	LabelsCacheSize         uint64
+	MetricsCacheSize        uint64
+	SeriesCacheSize         uint64
+	WriteConnectionsPerProc int
+	MaxConnections          int
 }
 
 // ParseFlags parses the configuration flags specific to PostgreSQL and TimescaleDB
@@ -48,6 +51,8 @@ func ParseFlags(cfg *Config) *Config {
 	flag.IntVar(&cfg.ReportInterval, "tput-report", 0, "interval in seconds at which throughput should be reported")
 	flag.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "maximum number of labels to cache")
 	flag.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", pgmodel.DefaultMetricCacheSize, "maximum number of metric names to cache")
+	flag.IntVar(&cfg.WriteConnectionsPerProc, "db-writer-connection-concurrency", 4, "maximum number of database connections per go process writing to the database")
+	flag.IntVar(&cfg.MaxConnections, "db-connections-max", -1, "maximum connections that can be open at once, defaults to 80% of the max the DB can handle")
 	return cfg
 }
 
@@ -65,17 +70,14 @@ type Client struct {
 // NewClient creates a new PostgreSQL client
 func NewClient(cfg *Config, readHist prometheus.ObserverVec) (*Client, error) {
 	connectionStr := cfg.GetConnectionStr()
-
-	maxProcs := runtime.GOMAXPROCS(-1)
-	if maxProcs <= 0 {
-		maxProcs = runtime.NumCPU()
+	minConnections, maxConnections, numCopiers, err := cfg.GetNumConnections()
+	if err != nil {
+		log.Error("err configuring number of connections", util.MaskPassword(err.Error()))
+		return nil, err
 	}
-	if maxProcs <= 0 {
-		maxProcs = 1
-	}
-	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxProcs*pgmodel.ConnectionsPerProc, maxProcs))
+	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
 
-	log.Info("msg", util.MaskPassword(connectionStr))
+	log.Info("msg", util.MaskPassword(connectionStr), "numCopiers", numCopiers)
 
 	if err != nil {
 		log.Error("err creating connection pool for new client", util.MaskPassword(err.Error()))
@@ -88,6 +90,7 @@ func NewClient(cfg *Config, readHist prometheus.ObserverVec) (*Client, error) {
 		AsyncAcks:       cfg.AsyncAcks,
 		ReportInterval:  cfg.ReportInterval,
 		SeriesCacheSize: cfg.SeriesCacheSize,
+		NumCopiers:      numCopiers,
 	}
 	ingestor, err := pgmodel.NewPgxIngestorWithMetricCache(connectionPool, cache, &c)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/pgclient/client.go
```diff
@@ -114,6 +114,50 @@ func (cfg *Config) GetConnectionStr() string {
 		cfg.host, cfg.port, cfg.user, cfg.database, cfg.password, cfg.sslMode)
 }
 
+func (cfg *Config) GetNumConnections() (min int, max int, err error) {
```
cevian, line 117: why is this modifying the cfg.ConnectionsPerProc variable? This shouldn't be called more than once correct? So this isn't about memoization?

There is some weird behavior if this is called more than once: you'll never hit a bunch of branches like line 147 on subsequent runs.

I'd rather not have this modify cfg state if possible

pkg/pgclient/client.go
```diff
@@ -48,6 +50,7 @@ func ParseFlags(cfg *Config) *Config {
 	flag.IntVar(&cfg.ReportInterval, "tput-report", 0, "interval in seconds at which throughput should be reported")
 	flag.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "maximum number of labels to cache")
 	flag.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", pgmodel.DefaultMetricCacheSize, "maximum number of metric names to cache")
+	flag.IntVar(&cfg.ConnectionsPerProc, "connections-per-proc", -1, "maximum number of database connections per go process")
```
davidkohn88, line 53: I would keep this parameter but add a second that is a total max connections to sort of act like a failsafe, like if I scale up my instance I don't want to swamp my db completely, so basically, I want to be able to say give me 4 connections per proc up to 100 for the database, so then I can, for instance, use an autoscaler.

pkg/pgclient/client.go
```diff
@@ -114,6 +115,50 @@ func (cfg *Config) GetConnectionStr() string {
...
+	err = row.Scan(dbMaxConnections)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	if dbMaxConnections <= 1 {
+		log.Warn("msg", "database can only handle 1 connection")
+		return dbMaxConnections, dbMaxConnections, 1, nil
+	}
+	min = maxProcs
+	if dbMaxConnections <= min {
```
davidkohn88, line 146: I dont' know that half the connections is right, I'd say 80% is probably closer as a sort of failsafe here.

pkg/pgmodel/sql_ingest.go
```diff
@@ -28,9 +28,10 @@ const (
 )
 
 type Cfg struct {
-	AsyncAcks       bool
-	ReportInterval  int
-	SeriesCacheSize uint64
+	AsyncAcks          bool
+	ReportInterval     int
+	SeriesCacheSize    uint64
+	ConnectionsPerProc int
```
cevian, line 34: I think it would be better to make this `IngestorConnections` or `copierConnections` and have all those calculations be done in pgClient. This would mean we wouldn't have to calculate maxProcs in the pgmodel code etc

pkg/pgclient/client.go
```diff
@@ -114,6 +117,47 @@ func (cfg *Config) GetConnectionStr() string {
 		cfg.host, cfg.port, cfg.user, cfg.database, cfg.password, cfg.sslMode)
 }
 
+func (cfg *Config) GetNumConnections() (min int, max int, perProc int, err error) {
```
cevian, line 120: I think this should return min, max, and copierConnections.

pkg/pgclient/client.go
```diff
@@ -48,6 +51,8 @@ func ParseFlags(cfg *Config) *Config {
 	flag.IntVar(&cfg.ReportInterval, "tput-report", 0, "interval in seconds at which throughput should be reported")
 	flag.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "maximum number of labels to cache")
 	flag.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", pgmodel.DefaultMetricCacheSize, "maximum number of metric names to cache")
+	flag.IntVar(&cfg.ConnectionsPerProc, "db-connections-per-proc", 5, "maximum number of database connections per go process")
+	flag.IntVar(&cfg.MaxConnections, "max-db-connections", -1, "maximum connections that can be open at once, defaults to 80% of the max the DB can handle")
```
cevian, line 55:
```suggestion
	flag.IntVar(&cfg.MaxConnections, "db-connections-max", -1, "maximum connections that can be open at once, defaults to 80% of the max the DB can handle")
```

pkg/pgmodel/sql_ingest.go
```diff
@@ -67,9 +67,9 @@ func newPgxInserter(conn pgxConn, cache MetricCache, cfg *Cfg) (*pgxInserter, er
 		maxProcs = 1
 	}
```
cevian, line 68: All the maxProcs code above here can now be deleted

pkg/pgclient/client.go
```diff
@@ -117,44 +117,48 @@ func (cfg *Config) GetConnectionStr() string {
...
 		}
-		if dbMaxConnections <= 1 {
+		if max <= 1 {
 			log.Warn("msg", "database can only handle 1 connection")
-			return dbMaxConnections, dbMaxConnections, 1, nil
+			return 1, 1, 1, nil
 		}
-		dbMaxConnections = int(0.8 * float32(dbMaxConnections))
+		// we try to only use 80% the database connections
+		max = int(0.8 * float32(max))
```
cevian, line 143: shouldn't this be the min(int(0.8 * float32(max)), perProc*maxProcs)?

cevian: or maybe this should be below the if?

pkg/pgclient/client.go
```diff
@@ -117,44 +117,48 @@ func (cfg *Config) GetConnectionStr() string {
...
-	// we try to only use 80% the database connections, but we need at least 1 connection per go process
-	if dbMaxConnections/maxProcs < perProc {
-		perProc = dbMaxConnections / maxProcs
-		if perProc < 1 {
-			perProc = 1
-		}
+
+	// we leave one connection per-core for non-copier usages
+	numCopiers = perProc*maxProcs - maxProcs
+	if numCopiers+maxProcs >= max {
```
cevian, line 157:
```suggestion
	if numCopiers+maxProcs > max {
```

pkg/pgclient/client.go
```diff
@@ -117,44 +117,48 @@ func (cfg *Config) GetConnectionStr() string {
...
+		log.Warn("msg", fmt.Sprintf("cannot handle desired number of copiers. wanted %v, has %v", numCopiers, max/2))
```
cevian, line 158:
```suggestion
		log.Warn("msg", fmt.Sprintf("had to reduct the number of copiers due to connection limits: wanted %v, reduced to %v", numCopiers, max/2))
```

pkg/pgclient/client.go
```diff
@@ -65,15 +70,12 @@ type Client struct {
...
-	}
-	if maxProcs <= 0 {
-		maxProcs = 1
+	minConnections, maxConnections, numCopiers, err := cfg.GetNumConnections()
+	if err != nil {
+		log.Error("err configuring number of connections", util.MaskPassword(err.Error()))
+		return nil, err
 	}
-	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxProcs*pgmodel.ConnectionsPerProc, maxProcs))
+	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
```
cevian, line 78: I would add a log-line with the minConnections, maxConnections and numCopiers here. Would ease debugging on client machines

pkg/pgclient/client.go
JLockerman: I don't think so; if the user wants more connections for e.g. readers, it should be allowed.

pkg/pgclient/client.go
JLockerman: line 80 should have the min and max at least
```
log.Info("msg", util.MaskPassword(connectionStr))
```

pkg/pgclient/client.go
cevian: So then the `db-connections-per-proc` is only used for calculating the numCopiers? Is that right? If so we should make that explicit in the description of the option

PR #207, opened 2020-08-20
Add upgrade tests

JLockerman:
Test that upgrading from the previous version results in the same
database.

### DIFF
--- a/pkg/pgmodel/upgrade_tests/upgrade_test.go
+++ b/pkg/pgmodel/upgrade_tests/upgrade_test.go
@@ -1,0 +1,499 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+package upgrade_tests
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"reflect"
+	"testing"
+	"time"
+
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+func toPreviousVersion(version *semver.Version) {
+	if len(version.Pre) > 0 {
+		lastPreDigit := &version.Pre[len(version.Pre)-1]
+		if lastPreDigit.VersionNum > 0 {
+			lastPreDigit.VersionNum -= 1
+			return
+		}
+
+		if len(version.Pre) > 2 {
```
cevian, line 193: why are we skipping 2 here. Why not just go backward in Pre slice order and decrease any version that is numeric?

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -0,0 +1,497 @@
...
+	"prom_data",
+	"prom_data_series",
+	"prom_info",
+	"prom_metric",
+	"prom_series",
+	"public",
+	"timescaledb_information",
+}
+
+var ourSchemas []string = []string{
```
cevian, line 356: add `_prom_catalog`

cevian:
Note: a followup PR will fix testing the pre-install scripts

Issue #202, opened 2020-08-20
[beta.2] use more connection and more cpu


wenerme:
```sql
select datname,count(*) FROM pg_stat_activity group by datname;
```

| datname | count |
| :--- | :--- |
| instanceA | 200 |
| timescale | 2 |
| instanceB | 801 |
| postgres | 1 |
| NULL | 6 |

instanceB's cpu is normal

instanceA's cpu is very high, and report logs

```
{"caller":"write.go:70","err":"The insert routine has previously failed with failed to connect to `host=10.0.1.1 user=username database=instanceA`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))","level":"warn","msg":"Error sending samples to remote storage","num_samples":100,"ts":"2020-08-20T08:21:03.129Z"}
```

and a lot duplicate log

```
{"caller":"sql_ingest.go:524","duplicate_count":6,"level":"warn","msg":"duplicate data in sample","row_count":6,"table":"ifInDiscards","ts":"2020-08-20T08:22:07.953Z"}
```

wenerme:
after increase db connections, db restarted, the error still exists,  timescale-prometheus can not recover from the error

still reporting `server error (FATAL: sorry, too many clients already (SQLSTATE 53300))`

when

| datname | count |
| :--- | :--- |
| instanceA | 167 |
| timescale | 2 |
| instanceB | 196 |
| postgres | 1 |
| NULL | 6 |

```sql
show max_connections;
# 1800
```

have to restart the instance, then everything goes normal, but the connection still high

| datname | count |
| :--- | :--- |
| instanceA | 392 |
| timescale | 2 |
| instanceB | 498 |
| postgres | 1 |
| NULL | 6 |

wenerme:
Dose timescale-prometheus still use prepare statement ? I don't find any code related to prepare, if timescale-prometheus do not use prepare, so I can setup a pool for tsdb.

timescale-prometheus use a lot connections, and this is risky, not scaleable, increase the max_connection is a temporary fix, when add more tenant/instance,  about 500 connection per tenant is unacceptable.

wenerme:
after long running ,connection reduced to about 100,

| datname | count |
| :--- | :--- |
| instanceA | 101 |
| instanceB | 161 |

instanceA samples about 1200/s
instanceB samples about 900/s

![image](https://user-images.githubusercontent.com/1777211/90779604-9969bb00-e330-11ea-8ed9-1385f76d004c.png)

[...]

PR #209, opened 2020-08-24
Cap max number of connections used by default

cevian:
Cap to 50 connections

### DIFF
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -146,8 +146,11 @@ func (cfg *Config) GetNumConnections() (min int, max int, numCopiers int, err er
 			log.Warn("msg", "database can only handle 1 connection")
 			return 1, 1, 1, nil
 		}
-		// we try to only use 80% the database connections
+		// we try to only use 80% the database connections, capped at 50
 		max = int(0.8 * float32(max))
+		if max > 50 {
+			max = 50
+		}
 	}
 
 	// we want to leave some connections for non-copier usages, so in the event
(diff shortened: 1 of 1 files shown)


cevian:
Addresses #202

pkg/pgclient/client.go
```diff
@@ -161,7 +166,7 @@ func (cfg *Config) GetNumConnections() (min int, max int, numCopiers int, err er
 
 	numCopiers = perProc * maxProcs
 	// we leave one connection per-core for non-copier usages
-	if numCopiers+maxProcs > max {
+	if numCopiers > max-maxProcs {
 		log.Warn("msg", fmt.Sprintf("had to reduce the number of copiers due to connection limits: wanted %v, reduced to %v", numCopiers, max/2))
 		numCopiers = max / 2
```
davidkohn88, line 171: shouldn't this just be max-maxProcs? rather than max/2?

pkg/pgclient/client.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/timescale/timescale-prometheus/pkg/util"
 )
 
+const maxDefaultConnections = 50
```
JLockerman, line 24: I don't see any reason to have this up here instead of in the function

pkg/pgclient/client.go
```diff
@@ -161,7 +166,7 @@ func (cfg *Config) GetNumConnections() (min int, max int, numCopiers int, err er
 
 	numCopiers = perProc * maxProcs
 	// we leave one connection per-core for non-copier usages
-	if numCopiers+maxProcs > max {
+	if numCopiers > max-maxProcs {
```
JLockerman, line 169: if we're doing this, do you want to set the default `db-writer-connection-concurrency` back to `5`?

pkg/pgclient/client.go
cevian: no because maxProcs may be > max. Changed comment.

PR #212, opened 2020-08-26
 Add real dataset end-to-end samples and tests

antekresic:
Dataset will be run behind a extended test flag and scheduled to be tested
daily via a separate Github Action.


.github/workflows/go-scheduled.yml
```diff
@@ -0,0 +1,43 @@
+name: Daily scheduled extended Go tests
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
JLockerman, line 5: 😍

JLockerman: Note that this is midnight UTC, is that the time we want?
(I could see it being an issue because it's "only" 5pm PST)

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -11,23 +13,82 @@ import (
...
+func (s *sample) UnmarshalJSON(data []byte) error {
+	var temp dummySample
+	if err := json.Unmarshal(data, &temp); err != nil {
+		if err = json.Unmarshal(data, &s.Timestamp); err != nil {
+			return json.Unmarshal(data, &s.Raw)
+		}
+		return nil
+	}
+	*s = sample(temp)
+	return nil
```
JLockerman, line 39: Is this really the best way to write this code?
What do you think of something like
```suggestion
	var temp dummySample
	err := json.Unmarshal(data, &temp)
	if == nil {
		*s = sample(temp)
		return nil
	}
	err = json.Unmarshal(data, &s.Timestamp)
	if err == nil {
		return nil
	}
	return json.Unmarshal(data, &s.Raw)
```
?

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -11,23 +13,82 @@ import (
...
+
+	if fmt.Sprintf("%v", s.Metric) != fmt.Sprintf("%v", other.Metric) {
+		return false
+	}
+
+	// fast path
+
+	if fmt.Sprintf("%v", s.Values) == fmt.Sprintf("%v", other.Values) {
+		return true
+	}
```
JLockerman, line 58: Is this really faster? That'd be disturbing...

pkg/pgmodel/end_to_end_tests/promql_query_endpoint_test.go
```diff
@@ -18,17 +17,40 @@ import (
...
 	Warnings []string  `json:"warnings,omitempty"`
 }
 
+func (q queryResponse) Equal(other queryResponse) bool {
+	if q.Status != other.Status {
+		return false
+	}
+	if fmt.Sprintf("%v", q.Warnings) != fmt.Sprintf("%v", other.Warnings) {
+		return false
+	}
```
JLockerman, line 36: Why not just `reflect.DeepEquals` if we don't want a manual loop?

pkg/pgmodel/end_to_end_tests/promql_query_endpoint_test.go
```diff
@@ -272,59 +295,2693 @@ func TestPromQLQueryEndpoint(t *testing.T) {
...
+		var (
+			requestCases []requestCase
+			req          *http.Request
+			err          error
+		)
+		for _, c := range testCases {
+			if c.realDataset {
+				if !*extendedTest {
+					continue
+				}
```
JLockerman, line 2955: since it looks like all the test cases are `realDataset`, can we just skip this entire function if `!*extendedTest`

cevian:
Another note: since the data in samples.go is just used for tests. It does not /need/ to be part of the binary. Thus go generate isn't necessary. We can have it just live in a testdata dir. we already have that in `pkg/pgmodel/testdata`

JLockerman:
>One thing I wonder is how much we can shrink samples.go by compressing it, maybe using go generate. Right now at 37 MB it's nearly 3x the size of the rest of the repo combined.

On further thought, it may be even better to just dump the protobufs and compress that, then read it back in programmatically instead of compiling it in...

pkg/pgmodel/end_to_end_tests/promql_query_endpoint_test.go
antekresic: Most of them are but not all, there are existing test cases which work on the previous dataset.

.github/workflows/go-scheduled.yml
antekresic: Changing it to 6am UTC which is roughly start of shift for European timezone.

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
antekresic: Yup, updated.

pkg/pgmodel/end_to_end_tests/samples_test.go
```diff
@@ -0,0 +1,175 @@
+package end_to_end_tests
```
cevian, line 1: nit: datasets_test.go a better name?

pkg/pgmodel/end_to_end_tests/samples_test.go
```diff
@@ -0,0 +1,175 @@
...
+		})
+		i++
+		time = startTime + (timeDelta * int64(i))
+	}
+
+	return samples
+}
+
+// generateRealTimeseries is used to read the real-world dataset from an
+// external file.
```
cevian, line 149: Let's add an acknowledgment in the comments to Julius' tool and say we got the data from his URLs

pkg/pgmodel/end_to_end_tests/promql_query_endpoint_test.go
```diff
@@ -272,59 +297,2693 @@ func TestPromQLQueryEndpoint(t *testing.T) {
...
-				t.Fatalf("unable to create PromQL query request: %s", err)
-			}
-			requestCases = append(requestCases, requestCase{req, fmt.Sprintf("%s (instant query, ts=start+30sec)", c.name)})
-		}
-		testMethod := testRequestConcurrent(requestCases, instantQuery, client, queryResultComparator)
-		tester.Run("test instant query endpoint", testMethod)
-
-		requestCases = nil
-		for _, c := range testCases {
+		{
```
cevian, line 300: Let's acknowledge in comments where these queries came from and a link to the repo

pkg/pgmodel/end_to_end_tests/promql_query_endpoint_test.go
```diff
@@ -272,59 +297,2693 @@ func TestPromQLQueryEndpoint(t *testing.T) {
...
+
+		start := time.Unix(startTime/1000, 0)
+		end := time.Unix(endTime/1000, 0)
+		var (
+			requestCases []requestCase
+			req          *http.Request
+			err          error
+		)
+		for _, c := range testCases {
+			if c.realDataset {
```
cevian, line 2954: Instead of having this logic where `TestPromQLQueryEndpoint` runs different queries depending on *extendedTest or not. I'd prefer a `TestPromQLQueryEndpointGeneratedDataset` and `TestPromQLQueryRealDataset' where the latter has t.Skip() called when !*extendendedTest. Both these functions would define their own testCases and then call a common function that would take in testCases and run the appropriate tests. Also start and end times can then be clearly defined for each TestPromQL function

antekresic:
Removed it, will merge once the tests pass.

PR #215, opened 2020-08-26
Improve documentation

cevian:
The goal of this commit is to streamline our documentation.
Namely we wanted to:

- Make instructions for various installation methods more streamlined.
  We now show all installation methods in one place in the README
  with links to more detailed instructions.
- Docker and Binary instructions become more complete and self-contained
  at the cost of some duplication.
- Make the main README shorter and more to-the-point
- Make the links to the SQL stuff more prominent and easy to find.
  Some of this is done but more can be done later.


antekresic (approved):
Looks good. Maybe we can add some more details on how to compile and install the extra extension for people that are not used to it. 

On the other hand, they would have to do it for timescale as well so maybe its not such a big deal.

PR #220, opened 2020-08-27
 Change name of drop_chunks procedure to execute_maintenance

cevian:


pkg/pgmodel/migrations/sql/versions/dev/0.1.0-beta.3.dev/1-drop_procedure_named_drop_chunks.sql
```diff
@@ -0,0 +1 @@
+DROP PROCEDURE SCHEMA_PROM.drop_chunks();
```
JLockerman, line 1: newline?

davidkohn88, line 1: Probably should be a drop if exists just in case.

PR #228, opened 2020-08-31
Fix docker-compose

JLockerman:
There were two issues:
1. There's a race-condition between connector-start and database-start:
   if the connector starts too quickly (and it often does) it will try
   to connect to the database before the database is ready and the
   connection will fail. We fix this by adding a probe step with a retry
   loop to make sure there is a database listening.
2. Using SSL with postgres on docker requires the user to set up their
   own SSL certificate before starting docker. Since communication
   between the connector and the database in the docker-compose workflow
   is restricted to a single machine we simply tell the connector that
   it need not use SSL.


cmd/timescale-prometheus/main.go
```diff
@@ -162,6 +165,14 @@ func main() {
 	log.Info("msg", "Version:"+version.Version+"; Commit Hash: "+version.CommitHash)
 	log.Info("config", util.MaskPassword(fmt.Sprintf("%+v", cfg)))
 
+	err = waitForDb(cfg)
+	if err != nil {
```
Harkishen-Singh, line 169: Since `waitForDb` only returns an error, I think it will be better to have them on same line. 
Much like,
```
if err = waitForDb(cfg); err != nil {
  ...
}
```

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
 	return cfg, nil
 }
 
+var errTimedOut = fmt.Errorf("timed out waiting to connect to DB")
```
Harkishen-Singh, line 343: Since `errTimeOut` is just used once, it will be better to declare these as constants at the top. However, an error in the top of a main file might look odd, so maybe, functions related to wait can be shifted to `utils` and the error being declared there.

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
...
+
+func waitForDb(cfg *config) error {
+	maxTimeout := 60 * time.Second
+	addr := fmt.Sprintf("%s:%d", cfg.pgmodelCfg.Host, cfg.pgmodelCfg.Port)
```
Harkishen-Singh, line 347: `var` declaration will be preferred in top of a function. It will save from allocation of any memory in runtime if the var is not used.

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
...
+		if err != nil {
+			return err
+		} else if done {
+			return nil
+		}
+		if time.Since(start) > maxTimeout {
+			return errTimedOut
+		}
+		log.Info("msg", "retrying DB connection")
+		time.Sleep(1 * time.Second)
```
Harkishen-Singh, line 387: Not sure, but won't 1 second be too small for a wait?

Harkishen-Singh (changes requested):

cmd/timescale-prometheus/main.go
cevian: I think putting it at top of file is fine

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
...
+	maxTimeout := 60 * time.Second
+	addr := fmt.Sprintf("%s:%d", cfg.pgmodelCfg.Host, cfg.pgmodelCfg.Port)
+
+	tryConnect := func() (done bool, err error) {
+		conn, err := net.DialTimeout("tcp", addr, 1*time.Second)
+		if err == nil {
+			conn.Close()
+			return true, nil
+		} else if err == syscall.ECONNREFUSED {
+			log.Info("msg", "connection to DB failed", "err", err.Error())
```
cevian, line 355:
```suggestion
			log.Info("msg", "DB not yet ready, retrying...", "reason", err.Error())
```

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
...
+			return false, nil
+		} else if e, ok := err.(net.Error); ok && e.Temporary() || e.Timeout() {
+			log.Info("msg", "connection to DB failed", "err", err.Error())
+			return false, nil
+		} else if strings.Contains(err.Error(), "connection refused") { //unfortunately the only protable to detect this
+			log.Info("msg", "connection to DB failed", "err", err.Error())
```
cevian, line 361:
```suggestion
			log.Info("msg", "DB not yet ready, retrying...", "reason", err.Error())
```

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
 	return cfg, nil
 }
 
+var errTimedOut = fmt.Errorf("timed out waiting to connect to DB")
+
+func waitForDb(cfg *config) error {
+	maxTimeout := 60 * time.Second
+	addr := fmt.Sprintf("%s:%d", cfg.pgmodelCfg.Host, cfg.pgmodelCfg.Port)
+
+	tryConnect := func() (done bool, err error) {
```
cevian, line 349: why is this a closure instead of a top-level function? I'd prefer this to be a top-level func

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
 	return cfg, nil
 }
 
+var errTimedOut = fmt.Errorf("timed out waiting to connect to DB")
+
+func waitForDb(cfg *config) error {
```
cevian, line 345: Fundamentally I don't think this should return an error. If our probe failed we should still try at least once to connect to the db using the psql protocol. That way any bugs in the probe still allow the rest of the system to work.

cmd/timescale-prometheus/main.go
```diff
@@ -329,6 +340,54 @@ func parseFlags() (*config, error) {
...
+	maxTimeout := 60 * time.Second
```
cevian, line 346: This should be a cli flag and we need to make sure that a value of 0 turns off the probe.

pkg/pgclient/client.go
```diff
@@ -23,8 +23,8 @@ import (
 
 // Config for the database
 type Config struct {
-	host                    string
-	port                    int
+	Host                    string
```
cevian, line 26: nit: I would expose  Config.TcpAddr() (string) instead of making these public. but will leave up to you.

cevian (changes requested):
Mostly minor changes.

cmd/timescale-prometheus/main.go
Harkishen-Singh: Yes, and 60 secs be the default value.

cmd/timescale-prometheus/main.go
JLockerman: So I actually disagree with that one; I think the hit to readability from the singleline version outweighs the pollution reduction of not having to define `err` outside.

JLockerman: (incidentally, I'd consider the [go style guide's](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) attitude towards error handling a tacit admission that the single-line form is usually bad style)

PR #229, opened 2020-09-01
Support logic for timescale_prometheus_extra updates and telemetry fixes

cevian:


pkg/internal/testhelpers/containers.go
```diff
@@ -143,12 +144,22 @@ func StartPGContainerWithImage(ctx context.Context, image string, testDataDir st
...
 	}
 
-	if testDataDir != "" {
-		req.BindMounts = map[string]string{
-			testDataDir: "/testdata",
+	if withExtension {
+		req.Cmd = []string{
+			"-c", "shared_preload_libraries=timescaledb",
+			"-c", "local_preload_libraries=pgextwlist",
+			"-c", "extwlist.extensions=timescale_prometheus_extra,timescaledb",
```
JLockerman, line 151: This bugs me; there's a non-local dependency between the image being used and the `withExtension` flag where the flag needs to set iff the extension contains `pgextwlist` and `timescaledb`. If it wasn't for the fact that we need to use multiple versions for the upgrade tests, I'd be inclined to hardcode the images:tag and dispatch purely based on `withExtension`. Even in the current version, I think we should consider hardcoding the image we're looking at, dispatched by `withExtension`, and only take in the tag portion as an argument. (This would also make it easier to update the image we use, if we ever choose to do so. Right now it's written in a couple of different places)

pkg/internal/testhelpers/containers.go
```diff
@@ -295,13 +310,11 @@ func StartConnectorWithImage(ctx context.Context, image string, printLogs bool,
 }
 
 func StopContainer(ctx context.Context, container testcontainers.Container, printLogs bool) {
-	defer func() {
```
JLockerman, line 298: 🤦 sorry about that one

pkg/pgmodel/migrate.go
```diff
@@ -457,14 +460,132 @@ func (t *Migrator) upgradeVersion(tx pgx.Tx, from, to semver.Version) error {
...
+
+	currentVersion, isInstalled, err := getInstalledExtensionVersion(conn)
+	if err != nil {
+		return fmt.Errorf("Could not get the installed extension version: %w", err)
+	}
+
+	if !isInstalled {
+		_, extErr := conn.Exec(context.Background(),
+			fmt.Sprintf("CREATE EXTENSION IF NOT EXISTS timescale_prometheus_extra WITH SCHEMA %s VERSION '%s'",
+				extSchema, getSqlVersion(newVersion)))
```
JLockerman, line 488: One of these days we should decide on a policy for what gets inlined and what gets put in a `const` at the top. (My vote is everything that's used once should be inlined, it's easier to read, but once it's used twice it should be a `const` so it's easy to edit all usages)

pkg/pgmodel/migrate.go
```diff
@@ -457,14 +460,132 @@ func (t *Migrator) upgradeVersion(tx pgx.Tx, from, to semver.Version) error {
...
+				getSqlVersion(newVersion)))
+		if err != nil {
+			return err
+		}
 		ExtensionIsInstalled = true
+		return nil
+	}
+}
+
+func getAvailableExtensionVersions(conn *pgxpool.Conn) (semver.Versions, error) {
```
JLockerman, line 519: it bugs me a little that both this function, which reads from the DB, and `getNewExtensionVersion` below which does not, are both called `getFoo`, it feels like we should have a distinct prefix for stuff that reads. Maybe `fetch` as in `fetchAvailableExtensionVersions` and `fetchInstalledExtensionVersion`?

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -107,16 +108,25 @@ func getUpgradedDbInfo(t *testing.T) (upgradedDbInfo dbInfo) {
...
-			doIngest(t, ingestor, preUpgradeData1, preUpgradeData2, postUpgradeData1, postUpgradeData2)
+			doIngest(t, ingestor, preUpgradeData1, preUpgradeData2)
+		},
+		/* postUpgrade */
+		func(container testcontainers.Container, _ string, db *pgxpool.Pool, tmpDir string) {
+			ingestor, err := pgmodel.NewPgxIngestor(db)
+			if err != nil {
+				t.Fatalf("error connecting to DB: %v", err)
+			}
+			defer ingestor.Close()
```
JLockerman, line 127: why is there an upgrade here? This version is specifically _not_ supposed to have any upgrades, and just go immediately to the final version

JLockerman: (if you want to upgrade the extension independant of the connector, that should be a separate test)

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -229,78 +245,123 @@ func withDBAndConnector(t testing.TB, DBName string, connectorImage string,
...
+		t.Fatal("Error setting up container", err)
 	}
-	f(dbContainer, tmpDir, connectorHost, connectorPort)
+
+	defer testhelpers.StopContainer(ctx, dbContainer, *printLogs)
+
+	t.Logf("upgrading versions %v => %v", prevVersion, version.Version)
+	// TODO this should be NoSuperuser expect for the extwlist error
+	connectURL := testhelpers.PgConnectURL(*testDatabase, testhelpers.Superuser)
+	migrateToVersion(t, connectURL, version.Version, "azxtestcommit")
```
JLockerman, line 304: (with the need to reset the pristine DB as well, performing the migrate here may be forced by the reset logic, so this comment may be irrelevant. Still, if we do end up having the discussion I think it's better that we do so explicitly in a well-recorded location instead of just in-person or on slack, so I'm leaving it for now)

so here's the thing: these lines are the most important lines of this file; they define how we're doing the upgrade, and thus the validity of the tests. That being the case, I think they should be high in the call-hierarchy as possible, as any issue the tests encounter will almost certainly relate to them, and this is going to be the place we debug from. As such I think putting them inside this function is a false economy giving the illusion encapsulation for no real benefit.

pkg/pgmodel/upgrade_tests/upgrade_test.go
JLockerman: (if it's just a restart, comments are wrong)

JLockerman (approved):
no blocking issues

pkg/pgmodel/migrate.go
cevian: I agree with that policy

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -71,8 +73,8 @@ func getUpgradedDbInfo(t *testing.T) (upgradedDbInfo dbInfo) {
 	// TODO we could probably improve performance of this test by 2x if we
 	//      gathered the db info in parallel. Unfortunately our db runner doesn't
 	//      support this yet
-	prevImage := "timescale/timescale-prometheus:" + prevVersion.String()
-	withDBAndConnector(t, *testDatabase, prevImage,
+	withDBStartingAtOldVersionAndUpgrading(t, *testDatabase, prevVersion,
```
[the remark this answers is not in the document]
cevian: You have a better idea?

pkg/pgmodel/upgrade_tests/upgrade_test.go
cevian: Fixed comment

antekresic (approved):

PR #230, opened 2020-09-02
Fix golangci-lint issues and integrate the same on CI

Harkishen-Singh:

This PR addresses some issues related to `golangci-lint` along with adding lint checks in CI since they were not being captured by our bot. Probably, since most bots tend to comment on the new changes, that might be a reason why these errors were not caught yet.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,6 +33,15 @@ jobs:
             dep ensure
         fi
 
+    - name: Format
+      run: |
+        diff -u <(echo -n) <(gofmt -d .)
+
+    - name: Lint
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0
+        /home/runner/go/bin/golangci-lint run --skip-dirs=pkg/promql --skip-dirs=pkg/promb
+
     - name: Build
       run: go build -v ./...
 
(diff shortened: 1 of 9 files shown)


JLockerman (commented):
Looks good to me for the changes to our code.
@cevian we need a policy decision on what to do with the code we copied from prom; I kinda want it to be as similar as possible to ease diffing when we want to upgrade.

Harkishen-Singh:
> I wish there was a way to just golint our changes but I don't think there is?
I think that a regex should do that work.

Harkishen-Singh:
> all the changes under pkg/promql need to be rolled back 

Done!

pkg/promql/functions_test.go
```diff
@@ -67,6 +67,6 @@ func TestFunctionList(t *testing.T) {
 
 	for i := range parser.Functions {
 		_, ok := FunctionCalls[i]
-		testutil.Assert(t, ok, (fmt.Sprintf("function %s exists in parser package, but not in promql package", i)))
+		testutil.Assert(t, ok, fmt.Sprintf("function %s exists in parser package, but not in promql package", i))
```
cevian, line 70: This error exists in upstream and should also be rolled back

cevian (approved):
One small comment to address otherwise good to merge

pkg/promql/functions_test.go
Harkishen-Singh: oh, I thought its just us, so I let it be. Anyways, rolling back that too.

Harkishen-Singh: Done!

JLockerman (approved):

cevian (approved):
Please clean up the commit history before merging

Harkishen-Singh:
@cevian I hope this looks good.

cevian (approved):

PR #232, opened 2020-09-02
Fix codeclimate

JLockerman:
We have no desire to evaluate code that's simply copied from prometheus and test code can violate good style for legitimate reasons.


cevian (approved):

Harkishen-Singh:
> We have no desire to evaluate code that's simply copied from Prometheus 

Given that we are now having a bit different version of promql (like topNode, etc), this is expected to increase as we add support for more nodes in `topNode` way. In such scenarios, the difference might be significant and evaluations can be good.
Does that make sense?

@JLockerman @cevian

JLockerman:
We'll see. Before this change the false-positive rate was overwhelming enough that the tool wasn't worth having at all, so we'll have to evaluate how useful it is first. If it is useful, we'll see if we can narrow down the exclude on the `promql` directory to be worthwhile. (My bet is that it will take factoring the code so that the code we add is in separate files than the code from upstream which, to my eye, sounds like a good idea anyway for diffing purposes...)

PR #234, opened 2020-09-03
Add HTTP header checking to end-to-end integration tests

antekresic:
This commit adds additional tests which check that HTTP responses from
Prometheus and the connector contain matching HTTP headers.

### DIFF
--- a/pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
+++ b/pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
@@ -5,17 +5,18 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net/http"
-	"net/http/httptest"
 	"reflect"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/prometheus/common/model"
 )
 
 type requestCase struct {
-	req *http.Request
-	log string
+	tsReq   *http.Request
+	promReq *http.Request
+	log     string
 }
 
 type sample struct {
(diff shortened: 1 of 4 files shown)


JLockerman (changes requested):
I don't think the `.bak` and `.stash` files should be included

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -122,30 +126,35 @@ func testRequest(req *http.Request, handler http.Handler, client *http.Client, c
...
 		wg := sync.WaitGroup{}
 
 		for i := 0; i < len(perm); i++ {
-			req := requestCases[perm[i]].req
+			tsReq := requestCases[perm[i]].tsReq
+			promReq := requestCases[perm[i]].promReq
 			log := requestCases[perm[i]].log
 			wg.Add(1)
-			go func() {
+			go func(log string) {
```
cevian, line 141: why was this change necessary? Log is created inside the loop so it should be local to the closure

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -177,3 +186,39 @@ func testRequestConcurrent(requestCases []requestCase, handler http.Handler, cli
 		wg.Wait()
 	}
 }
+
+func compareHTTPHeaders(t *testing.T, first, second http.Header) {
```
cevian, line 190:
```suggestion
func compareHTTPHeaders(t *testing.T, expected, actual http.Header) {
```

pkg/pgmodel/end_to_end_tests/promql_label_endpoint_test.go
```diff
@@ -75,28 +77,39 @@ func TestPromQLLabelEndpoint(t *testing.T) {
...
+		}
 		labelNamesHandler := api.Labels(apiConfig, queryable)
 		labelValuesHandler := api.LabelValues(apiConfig, queryable)
 
 		router := route.New()
 		router.Get("/api/v1/label/:name/values", labelValuesHandler.ServeHTTP)
 		router.Get("/api/v1/labels", labelNamesHandler.ServeHTTP)
 
-		apiURL := fmt.Sprintf("http://%s:%d/api/v1", testhelpers.PromHost, testhelpers.PromPort.Int())
+		ts := httptest.NewServer(router)
```
cevian, line 90: Is this the actual change that fixed the issue? If so I would explain in the commit message a bit more....

cevian (approved):
Overall I really like these changes. I have a hard time understanding what was wrong before and what's the actual fix here. I'd put more of a description in the commit message.

pkg/pgmodel/end_to_end_tests/promql_label_endpoint_test.go
antekresic: Yep, will update it.

pkg/pgmodel/end_to_end_tests/promql_endpoint_integration_test.go
antekresic: I'll remove that, its leftover, not used anymore.

JLockerman (approved):

PR #237, opened 2020-09-03
Update lock handling for migrate

JLockerman:
Prior to this commit migration was gated on leader election. This had
three issues:

 1. Multiple writers that did not engage in leader-election would race
    to migrate the database.
 2. Read-only connectors would not check that the database was at a
    valid version
 3. There was nothing preventing the database from being updated out
    from under the connectors.

To fix these issues, this commit introduces a new lease acquired by
every instance of the connector, regardless of whether they read or
write, and gates migration on this lease. Since Every migration is gated
on this one lock, this has two effects:

 1. We guarantee mutual exclusion for migrators.
 2. No migration can happen if there's another connector attached to the
    database.

This means that if a user wishes to update the connector version, they
must first shut down all connectors communicating with the database,
then bring up the connector performing the migration, and finally, only
after the migration has started bring up any other connectors they want
running.

### DIFF
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -18,6 +18,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	pgx "github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/jamiealquiza/envy"
@@ -41,15 +42,19 @@ type config struct {
 	electionInterval  time.Duration
 	migrate           bool
 	stopAfterMigrate  bool
+	useVersionLease   bool
 	corsOrigin        *regexp.Regexp
 }
 
 const (
 	promLivenessCheck = time.Second
+	schemaLockId      = 0x4D829C732AAFCEDE // chosen randomly
 )
 
 var (
-	elector *util.Elector
+	elector            *util.Elector
+	appVersion         = pgmodel.VersionInfo{Version: version.Version, CommitHash: version.CommitHash}
+	migrationLockError = fmt.Errorf("Could not acquire migration lock. Ensure there are no other connectors running and try again.")
 )
 
 func main() {
@@ -69,52 +74,17 @@ func main() {
 	log.Info("config", util.MaskPassword(fmt.Sprintf("%+v", cfg)))
 
 	promMetrics := api.InitMetrics()
-	elector, err = initElector(cfg, promMetrics)
 
+	client, err := createClient(cfg, promMetrics)
 	if err != nil {
-		errStr := fmt.Sprintf("Aborting startup because of elector init error: %s", util.MaskPassword(err.Error()))
-		log.Error("msg", errStr)
+		log.Error("msg", "aborting startup due to error", "err", util.MaskPassword(err.Error()))
 		os.Exit(1)
 	}
 
-	if elector == nil {
-		log.Warn(
-			"msg",
-			"No adapter leader election. Group lock id is not set. "+
-				"Possible duplicate write load if running adapter in high-availability mode",
-		)
+	if client == nil {
+		os.Exit(0)
 	}
 
-	appVersion := pgmodel.VersionInfo{Version: version.Version, CommitHash: version.CommitHash}
-
-	// migrate has to happen after elector started
-	if cfg.migrate {
-		err = migrate(&cfg.pgmodelCfg, appVersion, promMetrics)
-
-		if err != nil {
-			log.Error("msg", fmt.Sprintf("Aborting startup because of migration error: %s", util.MaskPassword(err.Error())))
-			os.Exit(1)
-		}
-		if cfg.stopAfterMigrate {
-			log.Info("msg", "Migration successful, exiting")
-			os.Exit(0)
-		}
-	} else {
-		log.Info("msg", "Skipping migration")
-	}
-
-	if err := checkDependencies(&cfg.pgmodelCfg, appVersion); err != nil {
-		log.Error("msg", fmt.Sprintf("Aborting startup because of dependency error: %s", util.MaskPassword(err.Error())))
-		os.Exit(1)
-	}
-
-	// client has to be initiated after migrate since migrate
-	// can change database GUC settings
-	client, err := pgclient.NewClient(&cfg.pgmodelCfg)
-	if err != nil {
-		log.Error(util.MaskPassword(err.Error()))
-		os.Exit(1)
-	}
 	defer client.Close()
 
 	apiConf := &api.Config{AllowedOrigin: cfg.corsOrigin}
@@ -158,6 +128,7 @@ func parseFlags() (*config, error) {
 	flag.BoolVar(&cfg.restElection, "leader-election-rest", false, "Enable REST interface for the leader election")
 	flag.DurationVar(&cfg.electionInterval, "scheduled-election-interval", 5*time.Second, "Interval at which scheduled election runs. This is used to select a leader and confirm that we still holding the advisory lock.")
 	flag.StringVar(&migrateOption, "migrate", "true", "Update the Prometheus SQL to the latest version. Valid options are: [true, false, only]")
+	flag.BoolVar(&cfg.useVersionLease, "use-schema-version-lease", true, "Prevent race conditions during migration")
 	envy.Parse("TS_PROM")
 	flag.Parse()
 
@@ -182,6 +153,100 @@ func parseFlags() (*config, error) {
 	return cfg, nil
 }
 
+func createClient(cfg *config, promMetrics *api.Metrics) (*pgclient.Client, error) {
+	var schemaVersionLease *util.PgAdvisoryLock
+	if cfg.useVersionLease {
+		// migration lock logic
+		// we don't want to upgrade the schema version while we still have connectors
+		// attached who think the schema is at the old version. To prevent this, as
+		// best we can, each normal connection attempts to grab a (shared) advisory
+		// lock on schemaLockId, and we attempt to grab an exclusive lock on it
+		// before running migrate. This implies that migration must be run when no
+		// other connector is running.
+		schemaVersionLease, err := util.NewPgAdvisoryLock(schemaLockId, cfg.pgmodelCfg.GetConnectionStr())
+		if err != nil {
+			log.Error("msg", "error creating schema version lease", "err", err)
+			os.Exit(1)
+		}
+		// after the client has started it's in charge of maintaining the leases
+		defer schemaVersionLease.Close()
+	}
+
+	migration_success := true
+	if cfg.migrate {
+		err := migrate(&cfg.pgmodelCfg, appVersion, schemaVersionLease)
+		migration_success = err != nil
+		if err != nil && err != migrationLockError {
+			return nil, fmt.Errorf("migration error: %w", err)
+		}
+
+		if cfg.stopAfterMigrate {
+			if err != nil {
+				return nil, err
+			}
+			log.Info("msg", "Migration successful, exiting")
+			return nil, nil
+		}
+	} else {
+		log.Info("msg", "Skipping migration")
+	}
+
+	if schemaVersionLease != nil {
+		// Grab a lease to protect version checking and client creation. This
+		// lease will be released when the schemaVersionLease is closed.
+		locked, err := schemaVersionLease.GetSharedAdvisoryLock()
+		if err != nil {
+			return nil, fmt.Errorf("could not acquire schema version lease due to: %w", err)
+		}
+		if !locked {
+			return nil, fmt.Errorf("could not acquire schema version lease. is a migration in progress?")
+		}
+	}
+
+	// Check the database is on the correct version.
+	// This must be done even if we attempted a migrate; if we failed to acquire
+	// the lock we'll start up as-if we were never supposed to migrate in the
+	// first place. This is also needed as checkDependencies populates our
+	// extension metadata
+	err := checkDependencies(&cfg.pgmodelCfg, appVersion)
+	if err != nil {
+		err = fmt.Errorf("dependency error: %w", err)
+		if !migration_success {
+			log.Error("msg", "Unable to run migrations; failed to acquire the lock. If the database is on an incorrect version, ensure there are no other connectors running and try again.", "err", err)
+		}
+		return nil, err
+	}
+
+	// Election must be done after migration and version-checking: if we're on
+	// the wrong version we should not participate in leader-election.
+	elector, err = initElector(cfg, promMetrics)
+
+	if err != nil {
+		return nil, fmt.Errorf("elector init error: %w", err)
+	}
+
+	if elector == nil {
+		log.Warn(
+			"msg",
+			"No adapter leader election. Group lock id is not set. "+
+				"Possible duplicate write load if running multiple connectors",
+		)
+	}
+
+	leasingFunction := getSchemaLease
+	if !cfg.useVersionLease {
+		leasingFunction = nil
+	}
+	// client has to be initiated after migrate since migrate
+	// can change database GUC settings
+	client, err := pgclient.NewClient(&cfg.pgmodelCfg, leasingFunction)
+	if err != nil {
+		return nil, fmt.Errorf("client creation error: %w", err)
+	}
+
+	return client, nil
+}
+
 func initElector(cfg *config, metrics *api.Metrics) (*util.Elector, error) {
 	if cfg.restElection && cfg.haGroupLockID != 0 {
 		return nil, fmt.Errorf("Use either REST or PgAdvisoryLock for the leader election")
@@ -195,7 +260,8 @@ func initElector(cfg *config, metrics *api.Metrics) (*util.Elector, error) {
 	if cfg.prometheusTimeout == -1 {
 		return nil, fmt.Errorf("Prometheus timeout configuration must be set when using PG advisory lock")
 	}
-	lock, err := util.NewPgLeaderLock(cfg.haGroupLockID, cfg.pgmodelCfg.GetConnectionStr())
+
+	lock, err := util.NewPgLeaderLock(cfg.haGroupLockID, cfg.pgmodelCfg.GetConnectionStr(), getSchemaLease)
 	if err != nil {
 		return nil, fmt.Errorf("Error creating advisory lock\nhaGroupLockId: %d\nerr: %s\n", cfg.haGroupLockID, err)
 	}
@@ -213,19 +279,29 @@ func initElector(cfg *config, metrics *api.Metrics) (*util.Elector, error) {
 	return &scheduledElector.Elector, nil
 }
 
-func migrate(cfg *pgclient.Config, appVersion pgmodel.VersionInfo, metrics *api.Metrics) error {
-	shouldWrite, err := isWriter()
-	if err != nil {
-		metrics.LeaderGauge.Set(0)
-		return fmt.Errorf("isWriter check failed: %w", err)
-	}
-	if !shouldWrite {
-		metrics.LeaderGauge.Set(0)
-		log.Debug("msg", fmt.Sprintf("Election id %v: Instance is not a leader. Won't update", elector.ID()))
-		return nil
+func migrate(cfg *pgclient.Config, appVersion pgmodel.VersionInfo, leaseLock util.AdvisoryLock) error {
+	// At startup migrators attempt to grab the schema-version lock. If this
+	// fails that means some other connector is running. All is not lost: some
+	// other connector may have migrated the DB to the correct version. We warn,
+	// then start the connector as normal. If we are on the wrong version, the
+	// normal version-check code will prevent us from running.
+
+	if leaseLock != nil {
+		locked, err := leaseLock.GetAdvisoryLock()
+		if err != nil {
+			return fmt.Errorf("error while acquiring migration lock %w", err)
+		}
+		if !locked {
+			return migrationLockError
+		}
+		defer func() {
+			_, err := leaseLock.Unlock()
+			if err != nil {
+				log.Error("msg", "error while releasing migration lock", "err", err)
+			}
+		}()
 	}
 
-	metrics.LeaderGauge.Set(1)
 	db, err := pgxpool.Connect(context.Background(), cfg.GetConnectionStr())
 	if err != nil {
 		return fmt.Errorf("Error while trying to open DB connection: %w", err)
@@ -251,18 +327,23 @@ func checkDependencies(cfg *pgclient.Config, appVersion pgmodel.VersionInfo) err
 	return pgmodel.CheckDependencies(db, appVersion)
 }
 
-func isWriter() (bool, error) {
-	if elector != nil {
-		shouldWrite, err := elector.IsLeader()
-		return shouldWrite, err
-	}
-	return true, nil
-}
-
 func compileAnchoredRegexString(s string) (*regexp.Regexp, error) {
 	r, err := regexp.Compile("^(?:" + s + ")$")
 	if err != nil {
 		return nil, err
 	}
 	return r, nil
+}
+
+// Except for migration, every connection that communicates with the DB must be
+// guarded by an instante of the schema-version lease to ensure that no other
+// connector can migrate the DB out from under it. We do not bother to release
+// said lease; in such and event the connector will be shutdown anyway, and
+// connection-death will close the connection.
+func getSchemaLease(ctx context.Context, conn *pgx.Conn) error {
+	err := util.GetSharedLease(ctx, conn, schemaLockId)
+	if err != nil {
+		return err
+	}
+	return pgmodel.CheckSchemaVersion(ctx, conn, appVersion)
 }
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -35,7 +35,7 @@ type config struct {
 	telemetryPath     string
 	pgmodelCfg        pgclient.Config
 	logLevel          string
-	haGroupLockID     int
+	haGroupLockID     int64
 	restElection      bool
 	prometheusTimeout time.Duration
 	electionInterval  time.Duration
@@ -152,7 +152,7 @@ func parseFlags() (*config, error) {
 	var migrateOption string
 	flag.StringVar(&corsOriginFlag, "web-cors-origin", ".*", `Regex for CORS origin. It is fully anchored. Example: 'https?://(domain1|domain2)\.com'`)
 	flag.StringVar(&cfg.logLevel, "log-level", "debug", "The log level to use [ \"error\", \"warn\", \"info\", \"debug\" ].")
-	flag.IntVar(&cfg.haGroupLockID, "leader-election-pg-advisory-lock-id", 0, "Unique advisory lock id per adapter high-availability group. Set it if you want to use leader election implementation based on PostgreSQL advisory lock.")
+	flag.Int64Var(&cfg.haGroupLockID, "leader-election-pg-advisory-lock-id", 0, "Unique advisory lock id per adapter high-availability group. Set it if you want to use leader election implementation based on PostgreSQL advisory lock.")
 	flag.DurationVar(&cfg.prometheusTimeout, "leader-election-pg-advisory-lock-prometheus-timeout", -1, "Adapter will resign if there are no requests from Prometheus within a given timeout (0 means no timeout). "+
 		"Note: make sure that only one Prometheus instance talks to the adapter. Timeout value should be co-related with Prometheus scrape interval but add enough `slack` to prevent random flips.")
 	flag.BoolVar(&cfg.restElection, "leader-election-rest", false, "Enable REST interface for the leader election")
@@ -195,7 +195,7 @@ func initElector(cfg *config, metrics *api.Metrics) (*util.Elector, error) {
 	if cfg.prometheusTimeout == -1 {
 		return nil, fmt.Errorf("Prometheus timeout configuration must be set when using PG advisory lock")
 	}
-	lock, err := util.NewPgAdvisoryLock(cfg.haGroupLockID, cfg.pgmodelCfg.GetConnectionStr())
+	lock, err := util.NewPgLeaderLock(cfg.haGroupLockID, cfg.pgmodelCfg.GetConnectionStr())
 	if err != nil {
 		return nil, fmt.Errorf("Error creating advisory lock\nhaGroupLockId: %d\nerr: %s\n", cfg.haGroupLockID, err)
 	}
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -1,0 +1,36 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/timescale/timescale-prometheus/pkg/log"
+	"github.com/timescale/timescale-prometheus/pkg/runner"
+	"github.com/timescale/timescale-prometheus/pkg/version"
+)
+
+func main() {
+	logLevel := flag.String("log-level", "debug", "The log level to use [ \"error\", \"warn\", \"info\", \"debug\" ].")
+	cfg := &runner.Config{}
+	cfg, err := runner.ParseFlags(cfg)
+	if err != nil {
+		fmt.Println("Version: ", version.Version, "Commit Hash: ", version.CommitHash)
+		fmt.Println("Fatal error: cannot parse flags ", err)
+		os.Exit(1)
+	}
+	err = log.Init(*logLevel)
+	if err != nil {
+		fmt.Println("Version: ", version.Version, "Commit Hash: ", version.CommitHash)
+		fmt.Println("Fatal error: cannot start logger", err)
+		os.Exit(1)
+	}
+	err = runner.Run(cfg)
+	if err != nil {
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
--- a/cmd/timescale-prometheus/main_test.go
+++ b/cmd/timescale-prometheus/main_test.go
@@ -7,81 +7,10 @@ import (
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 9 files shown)


JLockerman:
@davidkohn88 @cevian One thing I did not yet include in the PR is retries for failed lock acquisition, or incorrect versions. I'm curious for you thoughts on this: should we handle this at the connector level, or leave that for whatever is running the connector (keep in mind that since migrations can potentially take a while, there isn't really a perfect time to wait, so there's always going to the possibility that the next level will have to handle the connector failing to start). If you think we should handle it, care to suggest a timeout? If we have no better ideas, I'll probably go with `1 second` sleeps until at most a minute or so has passed.

cevian (commented):
Partial review

pkg/pgmodel/migrate.go
```diff
@@ -125,6 +125,18 @@ func CheckDependencies(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
 	return checkExtensionVersion(db)
 }
 
+func CheckDBVersion(ctx context.Context, conn *pgx.Conn, versionInfo VersionInfo) error {
```
cevian, line 128: I think this needs to be re-written to use pgmodel.CheckDependencies instead?

cmd/timescale-prometheus/main.go
```diff
@@ -406,19 +414,22 @@ func initElector(cfg *config) (*util.Elector, error) {
 	return &scheduledElector.Elector, nil
 }
 
-func migrate(cfg *pgclient.Config, appVersion pgmodel.VersionInfo) error {
-	shouldWrite, err := isWriter()
+func migrate(cfg *pgclient.Config, appVersion pgmodel.VersionInfo, leaseLock util.AdvisoryLock) error {
+	locked, err := leaseLock.GetAdvisoryLock()
```
cevian, line 418: Don't we need both a leader lock and a schema lock here? Why is it safe to get rid of the leader lock check? That seems strange. I'd expect: first get a leader lock, then get a shared advisory lock, then iff a migration is necessary get a exclusive lock, then migrate.

cevian (changes requested):
A bunch of comments. Think this requires a bit more work.

pkg/util/lock.go
```diff
@@ -152,36 +221,35 @@ func (l *PgAdvisoryLock) connCleanUp() {
...
-	defer l.mutex.RUnlock()
-	l.connCleanUp()
-}
-
-// Locked returns if the instance was able to obtain the leader lock.
-func (l *PgAdvisoryLock) Locked() bool {
-	l.mutex.RLock()
-	defer l.mutex.RUnlock()
-	return l.obtained
+func (l *PgAdvisoryLock) Unlock() error {
```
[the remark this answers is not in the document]
JLockerman: naming conventions seems to be `Lock/Unlock`, `LockShared/UnlockShared` [¯\_(ツ)_/¯](https://www.postgresql.org/docs/12/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS)

pkg/util/lock.go
```diff
@@ -111,7 +85,70 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+}
+
+// Release releases the already obtained locks.
+func (l *PgLeaderLock) Release() error {
+	l.mutex.Lock()
+	if !l.obtained {
+		return fmt.Errorf("can't release while not holding the lock")
+	}
+	defer l.mutex.Unlock()
+	rows, err := l.conn.Query(context.Background(), "SELECT pg_advisory_unlock($1)", l.groupLockID)
```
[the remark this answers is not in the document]
JLockerman: lock-ordering: I want to grab to check `l.obtained` and release the lock atomically, which requires I hold the mutex for the entire thing, and I also want `Unlock()` to be thread-safe, so it needs to grab the mutex as well. I'll add a comment

cmd/timescale-prometheus/main.go
```diff
@@ -197,10 +198,17 @@ func main() {
 		)
 	}
 
+	leaseLock, err := util.NewPgAdvisoryLock(leaseLockId, cfg.pgmodelCfg.GetConnectionStr())
```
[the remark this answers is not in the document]
JLockerman: how about `schemaVersionLease`? I think that most-accurately describes how it's being used

cmd/timescale-prometheus/main_test.go
```diff
@@ -252,59 +228,6 @@ func TestInitElector(t *testing.T) {
 	}
 }
 
-func TestMigrate(t *testing.T) {
```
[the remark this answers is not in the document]
JLockerman: with the refactor the only thing it's testing is that we grab the lock at the start of the migrate and release it at the end. The race condition detector does a better job of testing that.

cmd/timescale-prometheus/main.go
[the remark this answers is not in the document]
JLockerman: Hmm, I never considered that there was a valid reason to do that. Will update

JLockerman: under the scheme where we grab an exclusive lock as the first thing we do when we try to migrate, it's gratuitously redundant: if we succeed in grabbing the migration lock we'll be the one to migrate regardless

pkg/util/lock.go
```diff
@@ -111,7 +85,70 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	l.mutex.Lock()
+	if !l.obtained {
+		return fmt.Errorf("can't release while not holding the lock")
+	}
+	defer l.mutex.Unlock()
+	rows, err := l.conn.Query(context.Background(), "SELECT pg_advisory_unlock_all()")
+	if err != nil {
+		return err
+	}
+	rows.Close()
```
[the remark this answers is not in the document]
JLockerman: if we're not using the `rows` I don't think there's any reason to leave the connection open longer...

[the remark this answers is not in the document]
JLockerman: if we're not using the rows, I'm not sure there's a reason to leave the connection open longer...

pkg/util/lock.go
```diff
@@ -111,7 +85,70 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	l.mutex.Lock()
+	if !l.obtained {
+		return fmt.Errorf("can't release while not holding the lock")
+	}
+	defer l.mutex.Unlock()
+	rows, err := l.conn.Query(context.Background(), "SELECT pg_advisory_unlock($1)", l.groupLockID)
+	if err != nil {
+		return err
+	}
+	rows.Close()
```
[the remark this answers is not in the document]
JLockerman: I'm not sure it's an error to call this function when you don't hold the lock

cmd/timescale-prometheus/main.go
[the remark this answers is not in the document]
JLockerman: update: the way the logic works is the a migrating connector will unconditionally attempt to migrate. In the event that it fails to acquire the lock, it'll perform the same version checking as any other connector, and abort if the DB is at the wrong version

pkg/pgmodel/migrate.go
JLockerman: we deliberately don't want to check the extension as that can race, see the updated code

pkg/util/lock.go
```diff
@@ -111,7 +85,70 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	}
+	err = checkConnection(lockConn)
+	if err != nil {
+		log.Error("msg", "Connection pool returned invalid connection", "err", err)
+		return getConn(connStr, cur+1, maxRetries)
+	}
+	return lockConn, nil
+}
+
+func (l *PgAdvisoryLock) GetAdvisoryLock() (bool, error) {
```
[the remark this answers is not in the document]
JLockerman: I want to keep them as `GetAdvisoryLock` and `GetSharedAdvisoryLock` to be consistent with postgres, and the go mutex `Lock`/`RLock`

pkg/util/lock.go
```diff
@@ -111,30 +114,167 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	return l.getSharedAdvisoryLock()
+}
+
+func (l *PgAdvisoryLock) getSharedAdvisoryLock() (bool, error) {
+	return l.runLockFuncion("SELECT pg_try_advisory_lock_shared($1)")
+}
+
+func (l *PgAdvisoryLock) runLockFuncion(query string) (bool, error) {
```
cevian, line 229:
```suggestion
func (l *PgAdvisoryLock) runLockFunction(query string) (bool, error) {
```

pkg/util/lock.go
```diff
@@ -111,30 +114,167 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	return l.unlockShared()
+}
+
+func (l *PgAdvisoryLock) unlockShared() (bool, error) {
+	return l.runLockFuncion("SELECT SELECT pg_advisory_unlock_shared($1)")
+}
+
+func runLockFuncion(ctx context.Context, conn *pgx.Conn, query string, lockId int64) (bool, error) {
```
cevian, line 266:
```suggestion
func runLockFunction(ctx context.Context, conn *pgx.Conn, query string, lockId int64) (bool, error) {
```

pkg/util/lock.go
```diff
@@ -111,30 +114,167 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
 	if !rows.Next() {
 		err = rows.Err()
 		if err != nil {
-			return false, fmt.Errorf("error while trying to read response rows from `pg_try_advisory_lock` function: %v", err)
+			return false, fmt.Errorf("error while trying to read response rows from locking function: %w", err)
 		}
-		return false, fmt.Errorf("missing response row from `pg_try_advisory_lock` function")
+		return false, fmt.Errorf("missing response row from locking function")
 	}
 	var result bool
```
cevian, line 279: I think we should check rows.Err() after the scan to make sure no other errors came up

cmd/timescale-prometheus/main.go
```diff
@@ -31,25 +32,28 @@ import (
...
+	electionInterval   time.Duration
+	migrate            bool
+	stopAfterMigrate   bool
+	lockWhileMigrating bool
+	corsOrigin         *regexp.Regexp
 }
 
 const (
 	promLivenessCheck = time.Second
+	schemaLockId      = 0x4D829C732AAFCEDE // chosen randomly
```
cevian, line 51: nit: should we move this into util/lock.go?

cevian (changes requested):

cmd/timescale-prometheus/main.go
```diff
@@ -181,6 +152,80 @@ func parseFlags() (*config, error) {
...
+
+	if cfg.migrate {
+		err = migrate(cfg, appVersion, schemaVersionLease)
+		if err != nil {
+			return nil, fmt.Errorf("migration error: %w", err)
+		}
+
+		if cfg.stopAfterMigrate {
+			log.Info("msg", "Migration successful, exiting")
+			return nil, nil
```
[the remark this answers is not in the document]
JLockerman: I like keeping them all in `main` we handle that case in the
```go
if client == nil {
    os.Exit(0)
}
```
around line 83

cmd/timescale-prometheus/main.go
```diff
@@ -212,20 +258,32 @@ func initElector(cfg *config, metrics *api.Metrics) (*util.Elector, error) {
...
+		}
+		defer func() {
+			_, err := leaseLock.Unlock()
+			if err != nil {
+				log.Error("msg", "error while releasing migration lock", "err", err)
+				os.Exit(1)
+			}
+		}()
+		if !locked {
+			log.Warn("msg", "Unable to run migrations do to lock failure. If the database is on an incorrect version, ensure there are no other connectors running and try again.")
```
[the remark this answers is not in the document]
JLockerman: have a look at commit `nits 1` to see what that entails

cmd/timescale-prometheus/main_test.go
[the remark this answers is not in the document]
JLockerman: yah, the leadership is now completely independant of migration; the migrator may _potentially_ become the leader, but they're fully independant locks

pkg/pgmodel/migrate.go
```diff
@@ -111,18 +111,33 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
...
+	err = CheckDBVersion(context.Background(), conn.Conn(), versionInfo)
+	if err != nil {
+		return err
+	}
+
+	return checkExtensionVersion(db)
+}
+
+// check the DB schema version without checking the extension
+func CheckDBVersion(ctx context.Context, conn *pgx.Conn, versionInfo VersionInfo) error {
```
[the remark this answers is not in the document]
JLockerman: `CheckSchemaVersion` sounds fine to me

pkg/pgmodel/migrate.go
```diff
@@ -225,7 +240,17 @@ func ensureVersionTable(db *pgxpool.Pool) error {
 
 func getDBVersion(db *pgxpool.Pool) (semver.Version, error) {
```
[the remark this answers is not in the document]
JLockerman: it's used in `Migrate()` in `migrate.go:159`

pkg/util/lock.go
JLockerman: it's sad that even with the diff in front of me it took a good minute for me to figure out what changed 😞

pkg/util/lock.go
JLockerman: I don't think my `t` is broken...

pkg/util/lock.go
JLockerman: I'm going to double check the pgx code, but I'm _pretty_ sure that the err is only populated on `Next()` and `Scan` is "pure" in the sense that it reads no new data from the DB and just tries to decode what it's already got

cmd/timescale-prometheus/main.go
JLockerman: I don't think so. `lock.go` describes the abstract locking mechanism, but defining which lock we use it up to the application. (if anything I'm inclined to move `PgLeaderLock` to `election.go`, or else rename it to describe what it is instead of how it's used, but I can't think of a goon name, and this PR is large enough already)

pkg/util/lock.go
JLockerman: The resolution was to switch to `QueryRow()` and avoid all issues

pkg/pgmodel/migrate.go
```diff
@@ -225,7 +241,17 @@ func ensureVersionTable(db *pgxpool.Pool) error {
...
```
cevian, line 242:
```suggestion
func getSchemaVersion(db *pgxpool.Pool) (semver.Version, error) {
```

pkg/pgmodel/migrate.go
```diff
@@ -225,7 +241,17 @@ func ensureVersionTable(db *pgxpool.Pool) error {
...
-	res, err := db.Query(context.Background(), getVersion)
+	conn, err := db.Acquire(context.Background())
+	if err != nil {
+		return version, err
+	}
+	defer conn.Release()
+	return getDBVersionOnConnection(context.Background(), conn.Conn())
+}
+
+func getDBVersionOnConnection(ctx context.Context, db *pgx.Conn) (semver.Version, error) {
```
cevian, line 252:
```suggestion
func getSchemaVersionOnConnection(ctx context.Context, db *pgx.Conn) (semver.Version, error) {
```

pkg/util/lock.go
```diff
@@ -111,81 +114,175 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+
+	cfg, err := pgx.ParseConfig(connStr)
 	if err != nil {
-		return false, err
+		return nil, fmt.Errorf("error parsing config connection: %w", err)
 	}
-	defer rows.Close()
-	if !rows.Next() {
-		err = rows.Err()
+
```
cevian, line 192: question: why don't we use cfg.AfterConnect here instead?

pkg/util/lock.go
```diff
@@ -111,81 +114,175 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	lockConn, err := pgx.ConnectConfig(ctx, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("error getting DB connection: %w", err)
+	}
+
```
cevian, line 197: The old code also did a SELECT 1 here and then retried if it didn't work. Is that OK to leave out now?

pkg/util/lock.go
```diff
@@ -111,81 +114,175 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	if l.afterConnect != nil {
+		err = l.afterConnect(ctx, lockConn)
 		if err != nil {
-			return false, fmt.Errorf("error while trying to read response rows from `pg_try_advisory_lock` function: %v", err)
+			log.Error("msg", "Invalid connection", "err", err)
```
cevian, line 201:
```suggestion
			log.Error("msg", "Lock connection initiation failed", "err", err)
```

cevian (commented):
Mostly nits now

cmd/timescale-prometheus/main.go
```diff
@@ -181,6 +153,94 @@ func parseFlags() (*config, error) {
...
 
+func createClient(cfg *config, promMetrics *api.Metrics) (*pgclient.Client, error) {
+	// migration lock logic
+	// we don't want to upgrade the schema version while we still have connectors
+	// attached who think the schema is at the old version. To prevent this, as
+	// best we can, each normal connection attempts to grab a (shared) advisory
+	// lock on schemaLockId, and we attempt to grab an exclusive lock on it
+	// before running migrate. This implies that migration must be run when no
+	// other connector is running.
+	schemaVersionLease, err := util.NewPgAdvisoryLock(schemaLockId, cfg.pgmodelCfg.GetConnectionStr())
```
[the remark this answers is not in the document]
JLockerman: oh, that's interesting, let me try that out...

pkg/pgmodel/migrate.go
JLockerman: I think this change should happen in the PR that adds timescale-version checking

pkg/util/lock.go
JLockerman: it's got a weird function type intended for much more low-level use than we want (it exposes the underlying `PgConn`)

pkg/util/lock.go
JLockerman: it is if we're doing the lease, that's an equivalent check. I need to readd it if we're not

pkg/util/lock.go
```diff
@@ -111,81 +117,181 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
+	Unlock() (bool, error)
+	UnlockShared() (bool, error)
+	Close()
+}
+
+// PgAdvisoryLock is a AdvisoryLock
+var _ AdvisoryLock = (*PgAdvisoryLock)(nil)
+
+// NewPgAdvisoryLock creates a new instance with specified lock ID, connection pool and lock timeout.
+func NewPgAdvisoryLock(groupLockID int64, connStr string) (*PgAdvisoryLock, error) {
```
cevian, line 175: don't we need to make sure afterconnect isn't nil in here too?

JLockerman: I think the usage pattern is different enough that it's not as critical (it usually won't be unconnected like the leader election can be) but why not

Harkishen-Singh:
Will review this in morning. After that, I think we can go ahead with merge (if intended).

JLockerman:
For anyone coming to this thing late: it's the first two commits that contain the relevant changes, the latter is refactoring to enable more testing (though we do want more opinions on the new `main()` structure)

JLockerman:
(be warned, this PR has already been in development for a while now, so there's a fairly high bar to get further changes into the code touched by the first two commits)

cmd/timescale-prometheus/main.go
```diff
@@ -3,265 +3,34 @@
...
 		fmt.Println("Fatal error: cannot start logger", err)
 		os.Exit(1)
 	}
-	log.Info("msg", "Version:"+version.Version+"; Commit Hash: "+version.CommitHash)
-	log.Info("config", util.MaskPassword(fmt.Sprintf("%+v", cfg)))
-
-	promMetrics := api.InitMetrics()
-	elector, err = initElector(cfg, promMetrics)
-
+	err = timescale_prometheus.Run(cfg)
```
Harkishen-Singh, line 31: Are we planning to run multiple instances of timescale-prometheus? If not, then this seems wierd since timescale-prometheus's main file does not have the main code rather, some file in pkg has.

cmd/timescale-prometheus/main.go
```diff
@@ -3,265 +3,34 @@
...
-)
-
 func main() {
-	cfg, err := parseFlags()
+	logLevel := flag.String("log-level", "debug", "The log level to use [ \"error\", \"warn\", \"info\", \"debug\" ].")
+	cfg := &timescale_prometheus.Config{}
+	cfg, err := timescale_prometheus.ParseFlags(cfg)
 	if err != nil {
 		fmt.Println("Version: ", version.Version, "Commit Hash: ", version.CommitHash)
 		fmt.Println("Fatal error: cannot parse flags ", err)
```
Harkishen-Singh, line 22: Shouldn't we be using `Fprintf()` here (and below) with `os.Stderr` than going this way?

cmd/timescale-prometheus/main.go
Harkishen-Singh: If yes, then lets make this line and the following line in a single line.

pkg/pgclient/client.go
```diff
@@ -66,15 +66,25 @@ type Client struct {
 	metricCache   *pgmodel.MetricNameCache
 }
 
+//
```
Harkishen-Singh, line 69: Better to add an explanation than leaving it empty to slient go-vet.

pkg/pgmodel/migrate.go
```diff
@@ -111,18 +111,34 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
...
+// the extension, are set up correctly. This will set the ExtensionIsInstalled
+// flag and thus should only be called once, at initialization.
 func CheckDependencies(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
+	conn, err := db.Acquire(context.Background())
+	if err != nil {
+		return fmt.Errorf("failed to check schema version: %w", err)
+	}
+	defer conn.Release()
+	err = CheckSchemaVersion(context.Background(), conn.Conn(), versionInfo)
+	if err != nil {
```
Harkishen-Singh, line 124: Can be on the same line. Simple error returns on new lines add unnecessarily to the code length.

pkg/pgmodel/migrate.go
```diff
@@ -223,9 +239,19 @@ func ensureVersionTable(db *pgxpool.Pool) error {
 	return nil
 }
 
-func getDBVersion(db *pgxpool.Pool) (semver.Version, error) {
+func getSchemaVersion(db *pgxpool.Pool) (semver.Version, error) {
+	var version semver.Version
```
Harkishen-Singh, line 243: Sorry, but I didn't get where are we initializing this.

pkg/pgmodel/migrate.go
```diff
@@ -223,9 +239,19 @@ func ensureVersionTable(db *pgxpool.Pool) error {
...
+	conn, err := db.Acquire(context.Background())
+	if err != nil {
+		return version, err
+	}
+	defer conn.Release()
+	return getSchemaVersionOnConnection(context.Background(), conn.Conn())
```
Harkishen-Singh, line 249: IMO context does not need to start from here since this is not a dependent parent. This would have been the case if a cancel was being called in some part of this function, which is not the case.

pkg/timescale_prometheus/main.go
```diff
@@ -0,0 +1,339 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
Harkishen-Singh, line 1: It seems odd that the main (runner) of the repo is in pkg than the actual cmd. Is there any reason or plan more than this? Like, multiple calls to `Run()`?

Harkishen-Singh: If there are plans, then the design of Run can be better. Will look into that after confirming the answer for the above question.

Harkishen-Singh (changes requested):
First pass. Few more passes to go before completion.

pkg/pgmodel/migrate.go
```diff
@@ -111,18 +111,34 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
 	return nil
 }
 
-//CheckDependcies makes sure all project dependencies are set up correctly
+// CheckDependcies makes sure all project dependencies, including the DB schema
```
antekresic, line 114:
```suggestion
// CheckDependencies makes sure all project dependencies, including the DB schema
```

pkg/pgmodel/migrate.go
```diff
@@ -111,18 +111,34 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
...
+	defer conn.Release()
+	err = CheckSchemaVersion(context.Background(), conn.Conn(), versionInfo)
+	if err != nil {
+		return err
+	}
+
+	return checkExtensionVersion(db)
+}
+
+// check the DB schema version without checking the extension
```
antekresic, line 131:
```suggestion
// CheckSchemaVersion checks the DB schema version without checking the extension.
```

pkg/pgmodel/migrate.go
antekresic: Looks like this is just a placeholder value to send when an error occurs that should be ignored.

antekresic (dismissed):
The code looks OK, I'm just unsure about the naming of the package as  `timescale-prometheus` for the main runner. 

First of all, its not descriptive enough and just looks odd when importing it in the project with a stutter `.../timescale-prometheus/pkg/timescale-prometheus`

Secondly, if we leave it like that, it will have to change in the near future because of the name change.

I suggest changing the package to something like `runner`.

cmd/timescale-prometheus/main.go
JLockerman: No. I still think inlining non-trivial functions into the error handling conditional is bad style

pkg/pgclient/client.go
JLockerman: Looks like a bad merge removed the comment...

pkg/pgmodel/migrate.go
JLockerman: __This__ one actually should be on the same line

pkg/pgmodel/migrate.go
JLockerman: Yeah, it's pretty silly. In my defense, it was like that when I found it 😉

pkg/timescale_prometheus/main.go
JLockerman: Testing nonsense: we can't use `internal` packages in `cmd` and we can't use binary packages in `end_to_end` tests

cmd/timescale-prometheus/main.go
JLockerman: This one is going to postponed to a later PR where we switch our default log format to match prometheus's, issue #261.

pkg/pgmodel/migrate.go
JLockerman: yeah, we should really be more principled about how we pipe through contexts, and will probably go through and clean them after 0.1.0

Harkishen-Singh (dismissed):
Another pass. Will give one more before completion soon.

pkg/timescale_prometheus/main.go
Harkishen-Singh: This file required renaming.

Harkishen-Singh (commented):
lgtm for me except a few doubts.

pkg/pgclient/client.go
```diff
@@ -39,13 +39,13 @@ type Config struct {
...
 func ParseFlags(cfg *Config) *Config {
-	flag.StringVar(&cfg.host, "db-host", "localhost", "The TimescaleDB host")
-	flag.IntVar(&cfg.port, "db-port", 5432, "The TimescaleDB port")
-	flag.StringVar(&cfg.user, "db-user", "postgres", "The TimescaleDB user")
-	flag.StringVar(&cfg.password, "db-password", "", "The TimescaleDB password")
-	flag.StringVar(&cfg.database, "db-name", "timescale", "The TimescaleDB database")
-	flag.StringVar(&cfg.sslMode, "db-ssl-mode", "require", "The TimescaleDB connection ssl mode")
-	flag.IntVar(&cfg.dbConnectRetries, "db-connect-retries", 0, "How many times to retry connecting to the database")
+	flag.StringVar(&cfg.Host, "db-host", "localhost", "The TimescaleDB host")
+	flag.IntVar(&cfg.Port, "db-port", 5432, "The TimescaleDB port")
```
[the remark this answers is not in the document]
JLockerman: Hmm, in this case I was thinking that `TimescaleDB` is functioning as an adjective modifying `host` and `port` so it's `The <TimescaleDB> port` as uposed to `The <DB> port` or `The <other> port`. That said, I am pretty terrible at naming things, so @cevian @davidkohn88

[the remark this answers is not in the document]
JLockerman: oh, wait, that isn't part of this PR. This, and related comments are deferred to a followup
(@Harkishen-Singh feel free to open a `Clean cmd arguments help`  PR)

pkg/pgclient/client.go
```diff
@@ -39,13 +39,13 @@ type Config struct {
...
+	flag.StringVar(&cfg.User, "db-user", "postgres", "The TimescaleDB user")
+	flag.StringVar(&cfg.Password, "db-password", "", "The TimescaleDB password")
+	flag.StringVar(&cfg.Database, "db-name", "timescale", "The TimescaleDB database")
+	flag.StringVar(&cfg.SslMode, "db-ssl-mode", "require", "The TimescaleDB connection ssl mode")
+	flag.IntVar(&cfg.DbConnectRetries, "db-connect-retries", 0, "How many times to retry connecting to the database")
```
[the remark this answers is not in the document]
JLockerman: deferred

pkg/pgclient/client.go
```diff
@@ -39,13 +39,13 @@ type Config struct {
...
 	flag.BoolVar(&cfg.AsyncAcks, "async-acks", false, "Ack before data is written to DB")
 	flag.IntVar(&cfg.ReportInterval, "tput-report", 0, "interval in seconds at which throughput should be reported")
 	flag.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "maximum number of labels to cache")
```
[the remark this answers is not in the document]
JLockerman: deferred

pkg/pgclient/client.go
```diff
@@ -66,15 +66,26 @@ type Client struct {
...
+type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
+
 // NewClient creates a new PostgreSQL client
-func NewClient(cfg *Config) (*Client, error) {
+func NewClient(cfg *Config, schemaLocker LockFunc) (*Client, error) {
 	connectionStr := cfg.GetConnectionStr()
 	minConnections, maxConnections, numCopiers, err := cfg.GetNumConnections()
 	if err != nil {
-		log.Error("err configuring number of connections", util.MaskPassword(err.Error()))
+		log.Error("msg", "configuring number of connections", "err", util.MaskPassword(err.Error()))
```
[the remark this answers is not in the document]
JLockerman: dunno, and it can change over time, so better safe (anything that transitively uses the CFG can potentially report the password needs to be assumed to report sensitive info, which gets to be nearly everything in client)

pkg/pgclient/client.go
```diff
@@ -66,15 +66,26 @@ type Client struct {
...
 		return nil, err
 	}
-	connectionPool, err := pgxpool.Connect(context.Background(), connectionStr+fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
+
+	pgConfig, err := pgxpool.ParseConfig(connectionStr + fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
```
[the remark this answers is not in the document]
JLockerman: eh. I think this one reads fine, and it's 6 of one, half of dozen of the other

pkg/util/lock.go
```diff
@@ -111,81 +117,182 @@ func (l *PgAdvisoryLock) TryLock() (bool, error) {
...
-	if !l.obtained {
-		return fmt.Errorf("can't release while not holding the lock")
-	}
 	defer l.mutex.Unlock()
-	rows, err := l.conn.Query(context.Background(), "SELECT pg_advisory_unlock_all()")
+	return l.unlockShared()
+}
+
+func (l *PgAdvisoryLock) unlockShared() (bool, error) {
+	return l.runLockFunction("SELECT SELECT pg_advisory_unlock_shared($1)")
```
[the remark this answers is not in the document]
JLockerman: nope, that's a bug (and you can see how useful this function is that we never encounter it 😅 )

pkg/pgmodel/migrate.go
```diff
@@ -111,18 +111,34 @@ func Migrate(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
...
 }
 
-//CheckDependcies makes sure all project dependencies are set up correctly
+// CheckDependencies makes sure all project dependencies, including the DB schema
+// the extension, are set up correctly. This will set the ExtensionIsInstalled
+// flag and thus should only be called once, at initialization.
 func CheckDependencies(db *pgxpool.Pool, versionInfo VersionInfo) (err error) {
+	conn, err := db.Acquire(context.Background())
+	if err != nil {
+		return fmt.Errorf("failed to check schema version: %w", err)
```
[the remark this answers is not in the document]
JLockerman: it can't be; it has an interior error it needs to wrap (I don't think it's worth creating it's own error type, realistically we're never going to branch on it)

cevian (approved):

Issue #235, opened 2020-09-03
Verify timescaledb version at start and warn on any inconsistency


Harkishen-Singh:
We aim to connect and carry out operations on timescaledb only when the version is above `1.7.3`. Otherwise, this could end up as connection/compatibility issues.

PR #239, opened 2020-09-04
Adds support for checking Timescaledb version

Harkishen-Singh:

Fixes: #235 

This PR adds support to check Timescaledb extension version at the start in order to warn the user in terms of incompatibility.

### DIFF
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -86,6 +86,7 @@ func main() {
 	}
 
 	appVersion := pgmodel.VersionInfo{Version: version.Version, CommitHash: version.CommitHash}
+
 	// migrate has to happen after elector started
 	if cfg.migrate {
 		err = migrate(&cfg.pgmodelCfg, appVersion, promMetrics)
--- a/pkg/pgmodel/extension.go
+++ b/pkg/pgmodel/extension.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/blang/semver/v4"
 	"github.com/jackc/pgx/v4"
@@ -20,19 +21,54 @@ var (
 	ExtensionIsInstalled = false
 )
 
-// checkExtensionVersion checks for the correct version and enables the extension if
+// checkExtensionsVersion checks for the correct version and enables the extension if
 // it is at the right version
-func checkExtensionVersion(conn *pgxpool.Pool) error {
-	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn)
-	if err != nil {
-		return fmt.Errorf("Could not get the installed extension version: %w", err)
+func checkExtensionsVersion(conn *pgxpool.Pool) error {
+	if err := checkTimescaleDBVersion(conn); err != nil {
+		return err
 	}
+	if err := checkTimescalePrometheusExtraVersion(conn); err != nil {
+		return err
+	}
+	return nil
+}
 
+func checkTimescaleDBVersion(conn *pgxpool.Pool) error {
+	timescaleVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, "timescaledb")
+	if err != nil {
+		return fmt.Errorf("could not get the installed extension version: %w", err)
+	}
+	if !isInstalled {
+		return fmt.Errorf("timescaledb is not installed: %w", err)
+	}
+	switch version.VerifyTimescaleVersion(timescaleVersion) {
+	case version.Warn:
+		safeRanges := strings.Split(version.TimescaleVersionRangeString.Safe, " ")
+		log.Warn(
+			"msg",
+			fmt.Sprintf(
+				"Might lead to incompatibility issues due to TimescaleDB version. Expected version within %s to %s.",
+				safeRanges[0], safeRanges[1],
+			),
+			"Installed Timescaledb version:", timescaleVersion.String(),
+		)
+	case version.Err:
+		safeRanges := strings.Split(version.TimescaleVersionRangeString.Safe, " ")
+		return fmt.Errorf("incompatible Timescaledb version: %s. Expected version within %s to %s", timescaleVersion.String(), safeRanges[0], safeRanges[1])
+	case version.Safe:
+	}
+	return nil
+}
+
+func checkTimescalePrometheusExtraVersion(conn *pgxpool.Pool) error {
+	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, "timescale_prometheus_extra")
+	if err != nil {
+		return fmt.Errorf("could not get the installed extension version: %w", err)
+	}
 	if !isInstalled {
 		ExtensionIsInstalled = false
 		return nil
 	}
-
 	if version.ExtVersionRange(currentVersion) {
 		ExtensionIsInstalled = true
 	} else {
@@ -58,7 +94,7 @@ func migrateExtension(conn *pgxpool.Pool) error {
 		return nil
 	}
 
-	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn)
+	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, "timescale_prometheus_extra")
 	if err != nil {
 		return fmt.Errorf("Could not get the installed extension version: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


cmd/timescale-prometheus/main.go
```diff
@@ -147,6 +152,7 @@ var (
 	writeThroughput     = util.NewThroughputCalc(tickInterval)
 	elector             *util.Elector
 	lastRequestUnixNano = time.Now().UnixNano()
+	safeVersion         = semver.Version{Major: 1, Minor: 7, Patch: 3}
```
cevian, line 155: This should go into pkg/version/versions.go and be a semver range just like `ExtVersionRange `

cmd/timescale-prometheus/main.go
```diff
@@ -444,6 +458,26 @@ func checkDependencies(cfg *pgclient.Config, appVersion pgmodel.VersionInfo) err
 	return pgmodel.CheckDependencies(db, appVersion)
 }
 
+// isDBVersionSafe verifies if the compatible version of the database.
+func isDBVersionSafe(cfg *pgclient.Config) (bool, error) {
```
cevian, line 462: This should go into pkg/pgmodel/extension.go and should be like checkExtensionVersion (call new func `checkTimescaleVersion`). We should have 3 buckets Everything `>=1.7.3 && <2.0.0` is fine. `>=1.7.0 && <1.7.3` is a warning `<1.7.0 || >=2.0.0` is an error.

cevian: We should also refactor `func fetchInstalledExtensionVersion` in extensions.go to take the extension name so we could reuse that code.

cevian (changes requested):
Some notes on where to move some of this stuff.

cmd/timescale-prometheus/main.go
```diff
@@ -198,6 +204,14 @@ func main() {
 	}
 
 	appVersion := pgmodel.VersionInfo{Version: version.Version, CommitHash: version.CommitHash}
+	isSafe, err := isDBVersionSafe(&cfg.pgmodelCfg)
+	if err != nil {
+		log.Error("err", errors.Wrap(err, "db-version safe check"))
```
JLockerman, line 209: @cevian are we using `errors` or are we migrating to `fmt.Errorf("%w")`?

Harkishen-Singh: IMHO errors should be better since we want to approach a modular way. This makes it much easy to propagate down the error stack to the cause.

cmd/timescale-prometheus/main.go
Harkishen-Singh: Done!

cmd/timescale-prometheus/main.go
Harkishen-Singh: Done and used the refactored `func fetchInstalledExtensionVersion` way.

cmd/timescale-prometheus/main.go
cevian: we made the decision to use `fmt.Errorf("%w")` in this project because it is becoming the standard

pkg/pgmodel/extension.go
```diff
@@ -117,22 +150,25 @@ func fetchAvailableExtensionVersions(conn *pgxpool.Pool) (semver.Versions, error
...
 
-func fetchInstalledExtensionVersion(conn *pgxpool.Pool) (semver.Version, bool, error) {
+func fetchInstalledExtensionVersion(conn *pgxpool.Pool, extensionName string) (semver.Version, bool, error) {
 	var versionString string
-	err := conn.QueryRow(context.Background(),
-		"SELECT extversion  FROM pg_extension WHERE extname='timescale_prometheus_extra'").Scan(&versionString)
-	if err != nil {
+	if err := conn.QueryRow(
+		context.Background(),
+		fmt.Sprintf("SELECT extversion FROM pg_extension WHERE extname='%s';", extensionName),
```
cevian, line 157:
```suggestion
		"SELECT extversion FROM pg_extension WHERE extname=$1", extensionName,
```
Using parameters is better practice. Among other things its safer again SQL-injection attacks. Sometimes we can't use it, but we should try to use parameters whenever we can.

pkg/pgmodel/extension.go
```diff
@@ -23,11 +25,42 @@ var (
 // checkExtensionVersion checks for the correct version and enables the extension if
 // it is at the right version
 func checkExtensionVersion(conn *pgxpool.Pool) error {
-	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn)
+	timescaleVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, "timescaledb")
 	if err != nil {
-		return fmt.Errorf("Could not get the installed extension version: %w", err)
+		fmt.Fprintln(os.Stderr, "could not get the installed extension version: %w", err)
+		os.Exit(1)
```
cevian, line 31: Why are we exiting here? Can't we just return the error? (multiple places in function)

pkg/pgmodel/extension.go
```diff
@@ -23,11 +25,42 @@ var (
 // checkExtensionVersion checks for the correct version and enables the extension if
 // it is at the right version
 func checkExtensionVersion(conn *pgxpool.Pool) error {
-	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn)
+	timescaleVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, "timescaledb")
```
cevian, line 28: Nit: would prefer to break up this function into 2: one for each extension

cevian (changes requested):
Mostly looks good, but some changes required.

pkg/pgmodel/extension.go
```diff
@@ -23,11 +25,42 @@ var (
...
+	if !isInstalled {
+		fmt.Fprintln(os.Stderr, "Timescaledb is not installed", err)
+		os.Exit(1)
+	}
+	switch version.VerifyTimescaleVersion(timescaleVersion) {
+	case version.Warn:
+		safeRanges := strings.Split(version.TimescaleVersionRangeString.Safe, " ")
+		log.Warn(
+			"warn",
+			"Installed Timescaledb version:", timescaleVersion.String(),
```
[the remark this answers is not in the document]
Harkishen-Singh: Oh, got it! Thank you for the explanation.

pkg/pgmodel/extension.go
Harkishen-Singh: Ahh, got it.

pkg/pgmodel/extension.go
Harkishen-Singh: I thought that if the Timescaledb version is in error range, then we should not allow doing operations. Fixing that to just returning errors.

pkg/pgmodel/extension.go
Harkishen-Singh: That would be good.

Harkishen-Singh:
@cevian done!

cevian (approved):
Great work

Harkishen-Singh:
@cevian have cleaned the history. A final approval needed before I rebase.

pkg/util/util.go
```diff
@@ -10,12 +10,15 @@ import (
...
+	PromNamespace              = "ts_prom"
 	maskPasswordReplaceString1 = "password=$1'****'"
-	maskPasswordRegex2         = regexp.MustCompile(`password:(\s*?)([^\s]+?)( |$)`)
 	maskPasswordReplaceString2 = "password:$1****$3"
-	PromNamespace              = "ts_prom"
+)
+
+var (
+	maskPasswordRegex1 = regexp.MustCompile(`password=(\s*?)'([^']+?)'`)
+	maskPasswordRegex2 = regexp.MustCompile(`password:(\s*?)([^\s]+?)( |$)`)
```
Harkishen-Singh, line 21: I don't think I did this. Maybe the rebasing messed this. Should I undo?

antekresic: Do you have any automated tools that fixed this up? Its better like this anyways.

antekresic (approved):

cevian (approved):
Can you please clean up the commit message? We don't keep a history of fixups like that but rather try to have descriptive msgs.

Harkishen-Singh:
@cevian done! I hope the current commit message is satisfactory.

cevian:
@Harkishen-Singh  sorry to be annoying about this but the commit message should be something like 
```
Adds support for checking Timescaledb version

On startup we check for the correct TimescaleDB version. 
````

Harkishen-Singh:
@cevian done!

PR #246, opened 2020-09-08
Update PromQL from upstream 00ee73e: 28th Aug,20

Harkishen-Singh:

This PR updates PromQL and Promb from upstream as of commit 00ee73e: 28th Aug,20

Note: I am facing issues with `go mod vendor` since it was saying `github.com/prometheus/prometheus v2.5 +incompatible`. So, there remains 12 syntax errors in `engine.go` which is part of updated prometheus code, but due to incompatibility, it could not be copied to vendor. Any help related to this is appreciated.

Update: The above issue of incompatibility is now fixed!


pkg/prompb/remote.pb.go
```diff
@@ -68,7 +68,7 @@ type WriteRequest struct {
 	XXX_sizecache        int32        `json:"-"`
 }
 
-func (m *WriteRequest) Reset()         { *m = WriteRequest{Timeseries: m.Timeseries[:0]} }
```
cevian, line 71: These are our changes. They shouldn't be backed out. Wonder if we should move these function to a different go file.

pkg/prompb/remote.pb.go
```diff
@@ -937,12 +937,7 @@ func (m *WriteRequest) Unmarshal(dAtA []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			if len(m.Timeseries) < cap(m.Timeseries) {
-				m.Timeseries = m.Timeseries[:len(m.Timeseries)+1]
-				m.Timeseries[len(m.Timeseries)-1].Reset()
-			} else {
-				m.Timeseries = append(m.Timeseries, TimeSeries{})
-			}
+			m.Timeseries = append(m.Timeseries, TimeSeries{})
```
cevian, line 940: Our changes should not be changed back

pkg/prompb/types.pb.go
```diff
@@ -146,7 +146,7 @@ type TimeSeries struct {
 	XXX_sizecache        int32    `json:"-"`
 }
 
-func (m *TimeSeries) Reset()         { *m = TimeSeries{Labels: m.Labels[:0], Samples: m.Samples[:0]} }
+func (m *TimeSeries) Reset()         { *m = TimeSeries{} }
```
cevian, line 149: our intentional change. should not be undone.

pkg/prompb/types.pb.go
```diff
@@ -255,7 +255,7 @@ type Labels struct {
 	XXX_sizecache        int32    `json:"-"`
 }
 
-func (m *Labels) Reset()         { *m = Labels{Labels: m.Labels[:0]} }
+func (m *Labels) Reset()         { *m = Labels{} }
```
cevian, line 258: our intentional change. should not be undone.

pkg/prompb/types.pb.go
```diff
@@ -1384,12 +1384,7 @@ func (m *TimeSeries) Unmarshal(dAtA []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			if len(m.Labels) < cap(m.Labels) {
-				m.Labels = m.Labels[:len(m.Labels)+1]
-				m.Labels[len(m.Labels)-1].Reset()
-			} else {
-				m.Labels = append(m.Labels, Label{})
-			}
+			m.Labels = append(m.Labels, Label{})
```
cevian, line 1387: our intentional change. should not be undone.

pkg/prompb/types.pb.go
```diff
@@ -1423,12 +1418,7 @@ func (m *TimeSeries) Unmarshal(dAtA []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			if len(m.Samples) < cap(m.Samples) {
-				m.Samples = m.Samples[:len(m.Samples)+1]
-				m.Samples[len(m.Samples)-1].Reset()
-			} else {
-				m.Samples = append(m.Samples, Sample{})
-			}
+			m.Samples = append(m.Samples, Sample{})
```
cevian, line 1421: our intentional change. should not be undone.

pkg/prompb/types.pb.go
```diff
@@ -1634,12 +1624,7 @@ func (m *Labels) Unmarshal(dAtA []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			if len(m.Labels) < cap(m.Labels) {
-				m.Labels = m.Labels[:len(m.Labels)+1]
-				m.Labels[len(m.Labels)-1].Reset()
-			} else {
-				m.Labels = append(m.Labels, Label{})
-			}
+			m.Labels = append(m.Labels, Label{})
```
cevian, line 1627: our intentional change. should not be undone.

cevian (changes requested):
Some changes override out intentional changes and should be undone. But mostly this looks ok

Harkishen-Singh:
Ahh, I just copied from upstream for promb assuming that it was generated so we wouldn't have edited it. I just kept in mind to make sure our intentional changes in promql should not be reverted.

Anyways, will revert the places where you have commented in the review.

pkg/prompb/remote.pb.go
Harkishen-Singh: Yeah, moving to a different file will be safer. I will keep our custom changes in `custom.pb.go` in the same dir so that it remains safe to know when changes happen.

pkg/promql/test.go
```diff
@@ -78,8 +78,8 @@ type QuerierWrapper struct {
 }
 
 func (t *QuerierWrapper) Select(b bool, sh *storage.SelectHints, _ []parser.Node, m ...*labels.Matcher) (storage.SeriesSet, parser.Node, storage.Warnings, error) {
-	ss, w, err := t.Querier.Select(b, sh, m...)
-	return ss, nil, w, err
+	ss := t.Querier.Select(b, sh, m...)
+	return ss, nil, nil, nil
```
Harkishen-Singh, line 82: @cevian pls see if this is okay.

Harkishen-Singh:
The `storage.SeriesSet` design has been changed in upstream. Now, it returns only SeriesSet (unexpandedseriesset). I have not made those changes in our code, rather worked around it to skip that change as it will be breaking on our side. There are some points (our way of `q.Select` in `series.go` I guess) to be discussed before doing those changes.

Harkishen-Singh:
Fixed all tests! Up for depth review especially for the querier.Select. Added promql testsdata from upstream that was missing.

Any idea why `go generate` is failing?

pkg/promql/test.go
```diff
@@ -568,10 +568,11 @@ func (t *Test) clear() {
...
 	opts := EngineOpts{
-		Logger:     nil,
-		Reg:        nil,
-		MaxSamples: 10000,
-		Timeout:    100 * time.Second,
+		Logger:                   nil,
+		Reg:                      nil,
+		MaxSamples:               10000,
+		Timeout:                  100 * time.Second,
+		NoStepSubqueryIntervalFn: func(int64) int64 { return durationMilliseconds(1 * time.Minute) },
```
cevian, line 575: We should maybe add an feature-request issue to make this configurable

pkg/pgmodel/sql_reader.go
```diff
@@ -86,15 +86,14 @@ func (q *pgxQuerier) LabelsCacheCapacity() int {
 }
 
 // entry point from our own version of the prometheus engine
-func (q *pgxQuerier) Select(mint int64, maxt int64, sortSeries bool, hints *storage.SelectHints, path []parser.Node, ms ...*labels.Matcher) (storage.SeriesSet, parser.Node, storage.Warnings, error) {
+func (q *pgxQuerier) Select(mint int64, maxt int64, sortSeries bool, hints *storage.SelectHints, path []parser.Node, ms ...*labels.Matcher) (storage.SeriesSet, parser.Node) {
 	rows, topNode, err := q.getResultRows(mint, maxt, hints, path, ms)
-
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil
```
cevian, line 92: It seems like this error is getting lost. Should we return an errorSeriesSet?

pkg/api/series.go
```diff
@@ -71,12 +71,12 @@ func series(queryable *query.Queryable) http.HandlerFunc {
...
 		var warnings storage.Warnings
 		for _, mset := range matcherSets {
-			s, _, wrn, err := q.Select(false, nil, nil, mset...)
-			warnings = append(warnings, wrn...)
-			if err != nil {
-				respondError(w, http.StatusUnprocessableEntity, err, "execution")
-				return
-			}
+			s, _ := q.Select(false, nil, nil, mset...)
+			//warnings = append(warnings, wrn...)
```
cevian, line 75: can't we grab these warning from s.Warnings() now?

pkg/prompb/custom.pb.go
```diff
@@ -0,0 +1,5 @@
+package prompb
```
cevian, line 1: Like these a lot.

pkg/prompb/remote.pb.go
```diff
@@ -68,7 +68,6 @@ type WriteRequest struct {
 	XXX_sizecache        int32        `json:"-"`
 }
```
cevian, line 70: I'd maybe leave a comment here that the Reset function was moved to custom.pb.go. Same for all places it moved

pkg/prompb/remote.proto
```diff
@@ -0,0 +1,82 @@
+// Copyright 2016 Prometheus Team
```
cevian, line 1: I'm not sure we should move the proto files in here. It is incorrect to try to recompile them into Go because of our optimizations.

pkg/prompb/rpc.proto
```diff
@@ -0,0 +1,76 @@
+// Copyright 2017 Prometheus Team
```
cevian, line 1: same as above for .proto files

pkg/prompb/types.proto
```diff
@@ -0,0 +1,85 @@
+// Copyright 2017 Prometheus Team
```
cevian, line 1: this file too for .proto

pkg/promql/engine.go
```diff
@@ -743,22 +728,17 @@ func (ng *Engine) populateSeries(ctx context.Context, querier Querier, s *parser
...
-				return fmt.Errorf("Assigning topNode twice")
-			}
-			set, topNode, wrn, err = querier.Select(false, hints, path, n.LabelMatchers...)
-			warnings = append(warnings, wrn...)
-			if err != nil {
-				level.Error(ng.logger).Log("msg", "error selecting series set", "err", err)
-				return err
+				return fmt.Errorf("assigning topNode twice")
 			}
+			set, topNode = querier.Select(false, hints, path, n.LabelMatchers...)
```
cevian, line 733: hmm (for another PR) should we consider making topNode also part of the SeriesSet?

pkg/pgmodel/sql_reader.go
Harkishen-Singh: yeah, that will be better. I wonder how I missed that.

pkg/promql/test.go
Harkishen-Singh: Ahh, can you explain this bit more? Is this for upstream?

pkg/api/series.go
Harkishen-Singh: Ahh, I missed those debug comments. Anyways, the warnings will now be handled in the SeriesSet. I kept Warnings as nil in pgxSeriesSet since in upstream, the `blockBaseSeriesSet` has the same behaviour.

But still, being on the safe side, we can make a way to pass warnings through `(*pgxSeriesSet).Warnings()`.

pkg/prompb/remote.proto
Harkishen-Singh: In that case, should I let them be as they were?

pkg/promql/engine.go
Harkishen-Singh: We anyhow need to keep topNode in evaluator for checks in `ev.eval()`. (I think so, not sure).

Harkishen-Singh: Sure, will open an issue for this.

Harkishen-Singh:
@cevian made changes as per the planings on this PR.

pkg/api/series.go
cevian: Yes let's do that. (pass it up)

pkg/api/series.go
```diff
@@ -71,12 +71,12 @@ func series(queryable *query.Queryable) http.HandlerFunc {
...
+			//if err != nil {
```
cevian, line 76: We should probably check s.Err() here? and not just comment out?

pkg/pgmodel/sql_reader.go
cevian: This still needs to be done

cevian (changes requested):
A few small things. will ask for a review from Josh + Ante

pkg/api/marshal.go
```diff
@@ -43,7 +43,7 @@ func marshalCommonFooter(out *errorWrapper, warnings []string) {
 }
 
 func marshalMatrixData(out *errorWrapper, data promql.Matrix) {
-	out.WriteStrings(`{"resultType":"`, parser.ValueTypeMatrix, `","result":[`)
+	out.WriteStrings(`{"resultType":"`, string(parser.ValueTypeMatrix), `","result":[`)
```
JLockerman, line 46: Please don't change code independant of the PR at hand: it makes it more difficult to review PRs, and to track down where changes happen.

JLockerman (changes requested):

go.mod
```diff
@@ -3,30 +3,78 @@ module github.com/timescale/timescale-prometheus
...
+	github.com/gophercloud/gophercloud v0.12.0 // indirect
+	github.com/gregjones/httpcache v0.0.0-20180305231024-9cad4c3443a7 // indirect
+	github.com/grpc-ecosystem/grpc-gateway v1.14.8
+	github.com/hashicorp/consul/api v1.6.0 // indirect
+	github.com/hashicorp/go-hclog v0.14.1 // indirect
+	github.com/hashicorp/go-immutable-radix v1.2.0 // indirect
+	github.com/hashicorp/golang-lru v0.5.4 // indirect
+	github.com/hashicorp/serf v0.9.4 // indirect
+	github.com/imdario/mergo v0.3.5 // indirect
+	github.com/influxdata/influxdb v1.8.2 // indirect
```
JLockerman, line 38: I don't love this `go.mod change` is there a way we can shrink it?

JLockerman: There doesn't seem to be any significant new dependencies...

pkg/pgmodel/series_set.go
```diff
@@ -99,6 +99,9 @@ func (p *pgxSeriesSet) Err() error {
 	return nil
 }
 
+// Warnings implements storage.SeriesSet.
```
JLockerman, line 102: this comment is likely wrong

pkg/api/series.go
JLockerman: (even if we do only ever return `nil` in our current `SeriesSet` we should still retain the logic to forward the warnings here for ease of future changes)

JLockerman (commented):
Besides what's mentioned in earlier comments, the changes look good!
The only other big thing is that these need real commit messages: at a minimum the first commit needs to describe _why_ we're doing this upgrade, and what the commit range we're including is; the second should describe the kinds of changes necessitated by the new code we're merging in ([this article](https://chris.beams.io/posts/git-commit/) may be helpful, seems to be what the DB team is using now).

pkg/api/marshal.go
Harkishen-Singh: This PR has two commits: 1st for changes related to just PromQL and the 2nd for changes that are dependent on the PromQL update. This line is part of the 2nd commit.

go.mod
Harkishen-Singh: I am not sure that manual intervention will do this. Any suggestions?

pkg/api/series.go
Harkishen-Singh: Agreed.

pkg/pgmodel/series_set.go
Harkishen-Singh: Wrong in sense?

pkg/api/series.go
Harkishen-Singh: This is addressed. Hence, resolving.

pkg/api/series.go
```diff
@@ -71,9 +71,9 @@ func series(queryable *query.Queryable) http.HandlerFunc {
 		var sets []storage.SeriesSet
 		var warnings storage.Warnings
 		for _, mset := range matcherSets {
-			s, _, wrn, err := q.Select(false, nil, nil, mset...)
-			warnings = append(warnings, wrn...)
-			if err != nil {
+			s, _ := q.Select(false, nil, nil, mset...)
+			warnings = append(warnings, s.Warnings()...)
+			if s.Err() != nil {
 				respondError(w, http.StatusUnprocessableEntity, err, "execution")
```
antekresic, line 77: Should this be changed to `s.Err()` instead of `err`?

antekresic (dismissed):
Seems like the tests are failing, I believe its connected to my comment. After that is fixed, should be good to go, along with the commit message changes that @JLockerman mentioned previously.

pkg/api/series.go
Harkishen-Singh: Missed that! Addressed now.

cevian (approved):

JLockerman (approved):

PR #249, opened 2020-09-10
Move router and metrics to api package

### LINKED ISSUE #206: Make end-to-end test use router
Move router into pkg/api from main. Makes sure our router is set up correctly.

antekresic:
Moving the router and metrics to separate package outside of main package
enables us to reuse it for our end-to-end tests and use the main router for
testing.

Closes #206

### DIFF
--- a/cmd/timescale-prometheus/main.go
+++ b/cmd/timescale-prometheus/main.go
@@ -18,18 +18,14 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/prometheus/common/route"
-
 	"github.com/jackc/pgx/v4/pgxpool"
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/jamiealquiza/envy"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/timescale/timescale-prometheus/pkg/api"
 	"github.com/timescale/timescale-prometheus/pkg/log"
 	"github.com/timescale/timescale-prometheus/pkg/pgclient"
 	"github.com/timescale/timescale-prometheus/pkg/pgmodel"
-	"github.com/timescale/timescale-prometheus/pkg/query"
 	"github.com/timescale/timescale-prometheus/pkg/util"
 	"github.com/timescale/timescale-prometheus/pkg/version"
 )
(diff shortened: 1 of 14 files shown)


pkg/api/metrics.go
```diff
@@ -18,8 +122,95 @@ type Metrics struct {
...
+		QueryBatchDuration:  queryBatchDuration,
+		FailedQueries:       failedQueries,
+		ReceivedQueries:     receivedQueries,
+		HTTPRequestDuration: httpRequestDuration,
+	}
+
+	return metrics
+}
+
+func InitClientMetrics(client *pgclient.Client) *Metrics {
```
cevian, line 169: Optional additional refactor: I really think the code would look a lot better if this was in pkg/pgclient/metrics.go and looked like the existing pkg/pgmodel/metrics.go

There is no reason this needs to be part of of the metrics struct we pass around we can just make this local to pkg/pgclient

pkg/api/metrics.go
```diff
@@ -1,10 +1,114 @@
...
 )
 
+const (
+	tickInterval = time.Second
+)
+
+var (
+	metrics *Metrics
+
+	leaderGauge = prometheus.NewGauge(
```
cevian, line 18: nit and optional refactor : do we need these vars? can't we just init the struct fields in `InitMetrics` and then register the fields? i.e.

```
InitMetrics() {
...
metrics = &Metrics{
          receivedSamples: prometheus.NewCounter(
		prometheus.CounterOpts{
			Name:      "received_samples_total",
			Help:      "Total number of received samples.",
		},
	)
...
prometheus.MustRegister(metrics.receivedSamples)
...
}
```

cevian (approved):
This looks great and cleans up a lot of code. Added 2 optional refactors you could do on top of this to make the code even cleaner. But I'll leave it up to you if you think worthwhile.

pkg/api/metrics.go
```diff
@@ -18,8 +122,95 @@ type Metrics struct {
...
+	CachedLabels        prometheus.CounterFunc
+	MetricNamesCacheCap prometheus.GaugeFunc
+	LabelsCacheCap      prometheus.GaugeFunc
+}
+
+func InitMetrics() *Metrics {
+	if metrics != nil {
+		return metrics
+	}
+	prometheus.MustRegister(leaderGauge)
```
Harkishen-Singh, line 138: `MustRegister()` takes on args of metrics. Instead of calling on separate lines repeatedly, we should pass all these in just one MustRegister. This will lessen the complexity of stack trace, where multiple function calls are a overheads.

pkg/api/metrics.go
```diff
@@ -18,8 +122,95 @@ type Metrics struct {
...
+
+	labelsCacheCap := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Namespace: util.PromNamespace,
+		Name:      "label_cache_capacity",
+		Help:      "Total number of label-id to label mappings cache.",
+	}, func() float64 {
+		return float64(client.LabelsCacheCapacity())
+	})
+
+	prometheus.MustRegister(cachedMetricNames)
```
Harkishen-Singh, line 205: Sames goes over here as well.

Harkishen-Singh (changes requested):

pkg/api/metrics.go
Harkishen-Singh: Like this:

https://github.com/prometheus/prometheus/blob/a282d2509971174301408c5a5f96946c478dfb0f/tsdb/db.go#L266

antekresic: Thanks for the tip, fixed!

pkg/api/metrics.go
antekresic: Good catch!

pkg/api/metrics.go
antekresic: Makes sense, will do it.

antekresic:
@cevian implemented the refactoring you recommended. Could you take another look before merging?

cmd/timescale-prometheus/main.go
```diff
@@ -221,97 +109,16 @@ func main() {
...
-		InvalidWriteReqs:    invalidWriteReqs,
-	}
-	writeHandler := timeHandler(httpRequestDuration, "write", api.Write(client, elector, &promMetrics))
-	router.Post("/write", writeHandler)
-
-	readHandler := timeHandler(httpRequestDuration, "read", api.Read(client, &promMetrics))
-	router.Get("/read", readHandler)
-	router.Post("/read", readHandler)
-
+	pgclient.InitClientMetrics(client)
```
cevian, line 119: Nit: can we put this call inside of pgclient.NewClient?

cevian (approved):
One nit but otherwise LGTM

pkg/pgmodel/end_to_end_tests/promql_write_endpoint_test.go
```diff
@@ -186,30 +181,14 @@ func verifyTimeseries(t testing.TB, db *pgxpool.Pool, tsSlice []prompb.TimeSerie
...
 
-func getWriteRouter(t testing.TB, db *pgxpool.Pool) *route.Router {
-	r, err := pgmodel.NewPgxIngestor(db)
+func sendConcurrentWrites(t testing.TB, db *pgxpool.Pool, queues int, metricGroups int, totalRequests int, duplicates bool) {
+	router, err := buildRouter(db)
+
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("Unable to send concurrent writes, error building router: %s", err)
+		return
```
Harkishen-Singh, line 189: Not sure, but isn't `t.Fatalf()` automatically exit?

Harkishen-Singh:
Looks good. Didn't give an approval since not completely familiar with some parts. But, Mat and Josh have approved, so it should be good to go ahead.

PR #251, opened 2020-09-10
[WIP] Batch inserts to multiple tables

JLockerman:
This commit optimistically batches multiple INSERTs together into a single transaction if the inserter finds there are multiple requests available. This should reduce the number of transactions performed under high load, and may improve performance.

### DIFF
--- a/pkg/pgmodel/sql_ingest.go
+++ b/pkg/pgmodel/sql_ingest.go
@@ -495,8 +495,8 @@ func inserterGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, 
 hot_gather:
 	for len(batch) < cap(batch) {
 		select {
-		case req := <-in:
-			batch = append(batch, req)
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/sql_ingest.go
```diff
@@ -522,34 +573,47 @@ func tryRecovery(conn pgxConn, req copyRequest, err error) error {
...
+		vals := make([]float64, 0, numRows)
+		series := make([]int64, 0, numRows)
+		for req.data.batch.Next() {
+			time, val, serie := req.data.batch.Values()
+			times = append(times, time)
+			vals = append(vals, val)
+			series = append(series, int64(serie))
+		}
+		if len(times) != numRows {
+			panic("invalid insert request")
```
Harkishen-Singh, line 594: I hope, this isn't too much of a blocking code, what it might lead to sample gaps when prometheus queries most recently written samples.

PR #252, opened 2020-09-10
[WIP] Batchyer inserts2

JLockerman:
Like PR #251 except with a timeout to ensure some minimal number of requests are gathered

### DIFF
--- a/pkg/pgmodel/metrics.go
+++ b/pkg/pgmodel/metrics.go
@@ -33,6 +33,14 @@ var (
 			Name:      "decompress_min_unix_time",
 			Help:      "Earliest decdompression time",
 		}, []string{"table"})
+	numInsertsPerBatch = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: util.PromNamespace,
+			Name:      "inserts_per_batch",
+			Help:      "number of INSERTs in a single transaction",
+			Buckets:   []float64{1, 2, 3, 4, 5, 10, 20, 30, 40, 50, 75, 100},
+		},
+	)
 )
 
 func init() {
@@ -40,4 +48,5 @@ func init() {
 	prometheus.MustRegister(duplicateWrites)
 	prometheus.MustRegister(decompressCalls)
 	prometheus.MustRegister(decompressEarliest)
+	prometheus.MustRegister(numInsertsPerBatch)
 }
--- a/pkg/pgmodel/sql_ingest.go
+++ b/pkg/pgmodel/sql_ingest.go
@@ -495,8 +495,8 @@ func inserterGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, 
 hot_gather:
 	for len(batch) < cap(batch) {
 		select {
-		case req := <-in:
-			batch = append(batch, req)
+		case r2 := <-in:
+			batch = append(batch, r2)
 		default:
 			break hot_gather
 		}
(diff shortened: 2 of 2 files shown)


pkg/pgmodel/sql_ingest.go
```diff
@@ -465,31 +465,85 @@ func (h *insertHandler) flushPending() {
...
+}
+
+func doInsertOrFallback(conn pgxConn, reqs ...copyRequest) {
+	err := doInsert(conn, reqs...)
+	if err != nil {
+		insertBatchErrorFallback(conn, reqs...)
+		return
+	}
+
+	for i := range reqs {
```
Harkishen-Singh, line 517: Why not directly reference the req in reqs, instead of using index. Same for `insertBatchErrorFallback()`.

pkg/pgmodel/sql_ingest.go
```diff
@@ -522,34 +576,48 @@ func tryRecovery(conn pgxConn, req copyRequest, err error) error {
...
+		vals := make([]float64, 0, numRows)
+		series := make([]int64, 0, numRows)
+		for req.data.batch.Next() {
+			time, val, serie := req.data.batch.Values()
+			times = append(times, time)
+			vals = append(vals, val)
+			series = append(series, int64(serie))
+		}
+		if len(times) != numRows {
+			panic("invalid insert request")
```
Harkishen-Singh, line 598: Should we return error than panicking? The user might not be expecting a panic here.

pkg/pgmodel/sql_ingest.go
```diff
@@ -522,34 +576,48 @@ func tryRecovery(conn pgxConn, req copyRequest, err error) error {
...
 	}
+	defer results.Close()
 
-	if int64(numRows) != ct.RowsAffected() {
-		log.Warn("msg", "duplicate data in sample", "table", req.table, "duplicate_count", int64(numRows)-ct.RowsAffected(), "row_count", numRows)
-		duplicateSamples.Add(float64(int64(numRows) - ct.RowsAffected()))
-		duplicateWrites.Inc()
+	for _, _ = range numRowsPerInsert {
+		_, err := results.Exec()
+		if err != nil {
```
Harkishen-Singh, line 613: same line

pkg/pgmodel/metrics.go
```diff
@@ -33,11 +33,20 @@ var (
 			Name:      "decompress_min_unix_time",
 			Help:      "Earliest decdompression time",
 		}, []string{"table"})
+	numInsertsPerBatch = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: util.PromNamespace,
+			Name:      "inserts_per_batch",
+			Help:      "number of INSERTs in a single transaction",
+			Buckets:   []float64{1, 2, 3, 4, 5, 10, 20, 30, 40, 50, 75, 100},
```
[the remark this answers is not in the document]
JLockerman: these aren't the amount of datapoints per-INSERT it's the number of INSERTs per-batch, which has a hard limit of 100 in this experiment

pkg/pgmodel/sql_ingest.go
```diff
@@ -465,31 +465,85 @@ func (h *insertHandler) flushPending() {
...
+		case _ = <-timeout:
+			break hot_gather
+		}
+	}
+	return batch, true
+}
+
+func doInsertOrFallback(conn pgxConn, reqs ...copyRequest) {
+	err := doInsert(conn, reqs...)
+	if err != nil {
```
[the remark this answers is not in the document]
JLockerman: I still think that's bad style

pkg/pgmodel/sql_ingest.go
JLockerman: if this occurs it means the number of timestamps we're trying to insert do not equal the number of datapoints we're trying to insert. The only way that can occur is if there's an internal error in the connector, in which case shutting it down before it can cause any more damage is the best thing to do.

JLockerman:
@Harkishen-Singh it doesn't pay to review these `[WIP]` PR to heavily; this one happens to be an experiment we're running on Forge, to tell us something about the insert pattern there. When we learn what we want from that, then we may open a real PR with the actual changes.

pkg/pgmodel/sql_ingest.go
Harkishen-Singh: Agreed. Just meant for a smoother error handling.

PR #257, opened 2020-09-15
Fix leader election

JLockerman:
The test for eligibility for leader election is based on how long it's been since the client has received a WriteReq from prometheus. Therefore we need to record this time even if the client is not the leader.


cevian (approved):
Ok testing will be in #237

JLockerman:
(or after #237, it's already getting long)

PR #259, opened 2020-09-16
Add randomness to chunks when creating

### LINKED ISSUE #250: Need to add randomness to chunk interval upon creation
Randomness prevents compression jobs from getting synchronized. `_prom_catalog.set_chunk_interval_on_metric_table` already adds randomness but for new hypertables randomness is not correctly added when they are first created.

cevian:
Previously randomness was only added when chunk sizes were changed.
Fixes #250.


davidkohn88 (approved):
LGTM. This should help some of the sawtooth stuff we were seeing and might also help smooth vacuum work etc.

Harkishen-Singh (approved):
lgtm. However, will like the comment to be `// Avoid randomness in chunk interval size by setting explicitly.`

JLockerman (approved):
It's not a particularly good random function, but I doubt it'll matter for this usecase

PR #262, opened 2020-09-17
Add end-to-end tests using Docker

antekresic:
This PR adds end-to-end tests by spinning up the TimescaleDB and Prometheus container using Docker. It also spins up timescale-prometheus connector thats is build using the source code. Prometheus is configured to use the connector for remote read and write requests.

Once the environment is up, connector will be backfilled using a test dataset. After that there will be a series of requests sent to both the connector and Prometheus to compare the results and make sure that all the systems have been integrated properly.


cevian (commented):
Can we also put in the changes to make this run on CI?

scripts/end_to_end_tests.sh
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
+
+set -euf -o pipefail
+
+SCRIPT_DIR=$(pwd)
```
JLockerman, line 5: This will get you the directory the script is _running_ in, and is redundant with `pwd` the point of the
```bash
SCRIPT_DIR=$(dirname ${0})
```
is to give us the location of the script _file_, so we have a canonical location in the `timescale-prometheus` file-tree from which we can find other files (issue being that the script could have been called from an arbitrary location, so we have no real way to get from `pwd` to our other stuff)

JLockerman (commented):
Looks good, mostly just a couple of cleanliness nits, with two high-level ones:

1. I don't think we should upload both the `gz` and `sz` version of `real-dataset`; we should either use the `real-dataset.sz` in the other end-to-end tests, or write a small `gz`to`sz` converter we can pipe `real-dataset.gz` through for this test.
2. We need to add this test to the github actions ^__^

scripts/end_to_end_tests.sh
```diff
@@ -0,0 +1,122 @@
...
+
+compare_connector_and_prom() {
+    QUERY=${1}
+    CONNECTOR_OUTPUT=$(curl "http://${CONNECTOR_URL}/api/v1/${QUERY}")
+    PROM_OUTPUT=$(curl "http://${PROM_URL}/api/v1/${QUERY}")
+    echo "ran: ${QUERY}"
+    echo " connector response: ${CONNECTOR_OUTPUT}"
+    echo "prometheus response: ${PROM_OUTPUT}"
+    if [ "${CONNECTOR_OUTPUT}" != "${PROM_OUTPUT}" ]; then
+        echo "mismatched output"
```
Harkishen-Singh, line 88: Should we maintain a count? Like, `x out of y` passed.

scripts/end_to_end_tests.sh
```diff
@@ -0,0 +1,122 @@
...
+wait_for_connector() {
+    echo "waiting for connector"
+
+    sleep 10
+
+    ${SCRIPT_DIR}/wait-for.sh ${CONNECTOR_URL} -t 60 -- echo "connector may be ready..."
+
+    sleep 10
+
+    ${SCRIPT_DIR}/wait-for.sh ${CONNECTOR_URL} -t 60 -- echo "connector ready"
```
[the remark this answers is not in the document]
Harkishen-Singh: also, shouldn't this be a recursive loop with a timeout/count limit?

scripts/end_to_end_tests.sh
```diff
@@ -0,0 +1,122 @@
...
+
+# Check that backfilled dataset is present in both sources.
+compare_connector_and_prom "query_range?query=demo_disk_usage_bytes%7Binstance%3D%22demo.promlabs.com%3A10002%22%7D&start=$DATASET_START_TIME&end=$DATASET_END_TIME&step=30s"
+compare_connector_and_prom "query?query=demo_cpu_usage_seconds_total%7Binstance%3D%22demo.promlabs.com%3A10000%22%2Cmode%3D%22user%22%7D&time=$DATASET_START_TIME"
+# Check that connector metrics are scraped.
+compare_connector_and_prom "query?query=ts_prom_received_samples_total&time=$START_TIME"
+# Check that connector is up.
+compare_connector_and_prom "query?query=up&time=$START_TIME"
+# Check series endpoint matches on connector series.
+compare_connector_and_prom "series?match%5B%5D=ts_prom_sent_samples_total"
```
Harkishen-Singh, line 106: I think we are going to run these in a loop with values from a list?

JLockerman: nah, the bash ends up less readable if you try that

scripts/end_to_end_tests.sh
JLockerman: on further investigation, something like
```bash
SCRIPT_DIR=$(cd $(dirname ${0}) && pwd)
```
may work better

scripts/end_to_end_tests.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+set -euf -o pipefail
+
+PASSED=0
+FAILED=0
+
+SCRIPT_DIR=$(dirname $(readlink -f ${0}))
```
JLockerman, line 8: unfortunately, the `readlink -f` trick doesn't work on macos by default 😓

JLockerman (approved):
generally looks good!
There is one potential issue on macos: the `readlink -f` trick doesn't work to find the absolute location of a path. (So far I haven't found any portable way besides `cd`, though os-detection should also work)

cevian (approved):
LGTM although I'm no bash expert so others can probably give a better review

PR #264, opened 2020-09-21
Ensure fate-sharing for the migration lock

JLockerman:
This commit switches migration to use the same connection that holds the migration-lock instead of a separate connection pool. Using a separate pool has a failure mode where the connection holding the lock dies, but the pool remains alive and keeps migrating, potentially allowing a race among multiple migrators. If the same connection is used, the connection dying will also force migration to stop.

### DIFF
--- a/pkg/pgmodel/end_to_end_tests/main_test.go
+++ b/pkg/pgmodel/end_to_end_tests/main_test.go
@@ -110,7 +110,12 @@ func performMigrate(t testing.TB, connectURL string) {
 		t.Fatal(err)
 	}
 	defer migratePool.Close()
-	err = Migrate(migratePool, pgmodel.VersionInfo{Version: version.Version, CommitHash: "azxtestcommit"})
+	conn, err := migratePool.Acquire(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer conn.Release()
+	err = Migrate(conn.Conn(), pgmodel.VersionInfo{Version: version.Version, CommitHash: "azxtestcommit"})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	pgx "github.com/jackc/pgx/v4"
-	"github.com/jackc/pgx/v4/pgxpool"
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/jamiealquiza/envy"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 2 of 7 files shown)


pkg/pgmodel/end_to_end_tests/main_test.go
```diff
@@ -110,7 +110,12 @@ func performMigrate(t testing.TB, connectURL string) {
 		t.Fatal(err)
 	}
 	defer migratePool.Close()
-	err = Migrate(migratePool, pgmodel.VersionInfo{Version: version.Version, CommitHash: "azxtestcommit"})
+	conn, err := migratePool.Acquire(context.Background())
```
davidkohn88, line 113: Do we need a separate pool? Isn't this always only going to be one connection?

pkg/runner/runner.go
```diff
@@ -160,7 +159,11 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
 
 	migration_success := true
 	if cfg.Migrate {
-		err := migrate(&cfg.PgmodelCfg, appVersion, schemaVersionLease)
+		conn, err := schemaVersionLease.Conn()
```
cevian, line 162: schemaVersionLease allowed to be nil here, so this doesn't work

pkg/runner/runner.go
```diff
@@ -308,11 +305,11 @@ func migrate(cfg *pgclient.Config, appVersion pgmodel.VersionInfo, leaseLock *ut
 }
 
 func checkDependencies(cfg *pgclient.Config, appVersion pgmodel.VersionInfo) error {
```
cevian, line 307: Should this also use the schemaLeaseLock connection for fate-sharing?

cevian (changes requested):
Mostly looks good but one issue with incorrect nil handling

pkg/pgmodel/end_to_end_tests/main_test.go
davidkohn88: Seems like maybe it is a single connection, in which case, the naming is the problem.

JLockerman: we don't, but it's easier than refactoring the test more deeply

pkg/runner/runner.go
JLockerman: blah, forgot we allowed that 😞

pkg/runner/runner.go
JLockerman: it doesn't hurt, but it matters less: all the Client connections will perform their own checking, so this check is really just so we can print a nice error message in the event we have the wrong version; it's not a correctness hazard.

JLockerman:
note: I want to wait until PR #262 is merged before this is, that way we'll have testing for the new behavior in CI 😋

pkg/runner/runner.go
JLockerman: it was easy so I changed it

cevian (approved):

PR #266, opened 2020-09-24
Add log format flag and default to logfmt formatting for logs

### LINKED ISSUE #261: Clean Log Formatting
Prometheus defaults to outputting logs in logfmt, with an option to switch it to JSON. Since the only reason we're using JSON output is to match Prometheus, we should switch our defaults to match. We also need to decide what we do about our error message [for when log initialization fails](https://github.com/timescale/timescale-prometheus/blob/4adc895803edad460619572a5b78f70fa45b7419/cmd/timescale-prometheus/main.go#L58-L59): on the one hand, all our logs go to `stderr`, on the other hand, if we're committing to structured logging, we need to make sure we don't output error messages which don't match the format.

antekresic:
This commit adds a new log format flag which can be used for setting the format
of the logs to either logfmt (default) or json. It also cleans up logger flags
and initalization logic a bit.

Closes #261

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -6,6 +6,7 @@
 package log
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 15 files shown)


pkg/log/log.go
```diff
@@ -25,16 +26,35 @@ var (
...
-	logLevelOption, err := parseLogLevel(logLevel)
+// Config represents a logger configuration used upon initialization.
+type Config struct {
+	Level  string
+	Format string
+}
+
+// ParseFlags parses the configuration flags for logging.
+func ParseFlags(cfg *Config) *Config {
+	flag.StringVar(&cfg.Level, "log-level", "debug", "The log level to use [ \"error\", \"warn\", \"info\", \"debug\" ].")
```
Harkishen-Singh, line 37: How about: `Available log levels: "error", "warn", "info", "debug"`?

pkg/log/log.go
```diff
@@ -25,16 +26,35 @@ var (
...
+	flag.StringVar(&cfg.Format, "log-format", "logfmt", "The log format to use [ \"logfmt\", \"json\" ].")
```
Harkishen-Singh, line 38: Similarly, `Available log formats: "logfmt", "json"`

pkg/log/log.go
```diff
@@ -25,16 +26,35 @@ var (
...
+	return cfg
+}
+
+// Init starts logging given the configuration. By default, it uses logfmt format
+// and minimum logging level.
+func Init(cfg Config) error {
+	var l log.Logger
+	if cfg.Format == "json" {
+		l = log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
+	} else {
```
Harkishen-Singh, line 48: We should also check for `logfmt` to avoid those cases where the user puts a log format that matches none.

Harkishen-Singh (changes requested):

pkg/log/log.go
```diff
@@ -25,16 +26,35 @@ var (
 	)
 )
 
-// Init starts logging given the minimum log level
-func Init(logLevel string) error {
-	l := log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
-	logLevelOption, err := parseLogLevel(logLevel)
+// Config represents a logger configuration used upon initialization.
+type Config struct {
```
cevian, line 30: Can we use the Config from prometheus/common/promlog instead?

pkg/log/log.go
antekresic: IMO its clear enough how it was.

pkg/log/log.go
```diff
@@ -25,16 +26,39 @@ var (
...
+func ParseFlags(cfg *Config) *Config {
+	flag.StringVar(&cfg.Level, "log-level", "debug", "The log level to use [ \"error\", \"warn\", \"info\", \"debug\" ].")
+	flag.StringVar(&cfg.Format, "log-format", "logfmt", "The log format to use [ \"logfmt\", \"json\" ].")
+	return cfg
+}
+
+// Init starts logging given the configuration. By default, it uses logfmt format
+// and minimum logging level.
+func Init(cfg Config) error {
+	var l log.Logger
```
cevian, line 45: can we just use promlog.New() here?

pkg/log/log.go
antekresic: Matching the previous format for log flag description which is clear enough IMO.

cevian (approved):

Harkishen-Singh (approved):
lgtm!

PR #268, opened 2020-09-25
Move extension into promscale_extension repo

cevian:
The extension was separated out of the Promscale repo because
it is optional code and has a different release cadence.


antekresic (approved):

Harkishen-Singh:
I feel like promscale is the parent repo of all projects that are related to the same. We should keep in a single repo than splitting across multiple repos. Keeps things at one place and easier to manage. Many organisations do this way, like Kubernetes, though Kubernetes has a completely different release cycle.

Harkishen-Singh:
Agreed with the reason. We can go ahead.

Harkishen-Singh (approved):

JLockerman (approved):

PR #269, opened 2020-09-25
Change telemtry prefix

cevian:


pkg/pgmodel/end_to_end_tests/migrate_test.go
```diff
@@ -150,7 +150,7 @@ func TestMigrateTwice(t *testing.T) {
 		}
 
 		var versionString string
-		err := db.QueryRow(context.Background(), "SELECT value FROM _timescaledb_catalog.metadata WHERE key='timescale_prometheus_version'").Scan(&versionString)
+		err := db.QueryRow(context.Background(), "SELECT value FROM _timescaledb_catalog.metadata WHERE key='promscale_version'").Scan(&versionString)
```
Harkishen-Singh, line 153: Not sure but, should this be both promscale and Timescale-prometheus with priority of promscale version over Timescale-prometheus? Reason being the existing users do not have promscale.

cevian: This is just a test so existing users don't matter

Harkishen-Singh (approved):

PR #279, opened 2020-10-01
Make Timescaledb extension optional

cevian:


pkg/pgmodel/extension.go
```diff
@@ -152,11 +150,11 @@ func fetchAvailableExtensionVersions(conn *pgx.Conn) (semver.Versions, error) {
...
 }
 
-func fetchInstalledExtensionVersion(conn *pgx.Conn, extensionName string) (semver.Version, bool, error) {
+func fetchInstalledExtensionVersion(tx pgx.Tx, extensionName string) (semver.Version, bool, error) {
 	var versionString string
-	if err := conn.QueryRow(
+	if err := tx.QueryRow(
 		context.Background(),
-		"SELECT extversion FROM pg_extension WHERE extname=$1;",
+		"SELECT extversion FROM pg_extension WHERE extName=$1;",
```
JLockerman, line 157: I think this should remain `extname`; that's how it's documented

pkg/pgmodel/extension.go
```diff
@@ -38,7 +38,7 @@ func checkTimescaleDBVersion(conn *pgx.Conn) error {
 		return fmt.Errorf("could not get the installed extension version: %w", err)
 	}
 	if !isInstalled {
-		return fmt.Errorf("timescaledb is not installed: %w", err)
+		return nil
```
Harkishen-Singh, line 41: Should we not warn here? Like, `running promscale on pure postgres. features of timescaledb hence unavailable`

pkg/pgmodel/extension.go
```diff
@@ -76,62 +76,70 @@ func checkPromscaleExtensionVersion(conn *pgx.Conn) error {
...
-	ExtensionIsInstalled = false
+func migrateExtension(conn *pgx.Conn, extName string, extSchemaName string, validRange semver.Range, rangeString string) error {
+	availableVersions, err := fetchAvailableExtensionVersions(conn, extName)
 	if err != nil {
 		return err
 	}
 	if len(availableVersions) == 0 {
-		log.Warn("msg", "the promscale extension is not available, proceeding without extension")
-		return nil
+		return fmt.Errorf("the extension is not available")
```
Harkishen-Singh, line 85: `the extension is not available` => `promscale extension unavailable`

pkg/pgmodel/extension.go
```diff
@@ -176,27 +182,27 @@ func fetchInstalledExtensionVersion(conn *pgx.Conn, extensionName string) (semve
...
+	if extName == "promscale" && v.String() == "0.1.0" {
 		return "0.1"
 	}
 	return v.String()
 }
 
-// getNewExtensionVersion returns the highest version allowed by ExtVersionRange
-func getNewExtensionVersion(availableVersions semver.Versions) (semver.Version, bool) {
+// getNewExtensionVersion returns the highest version allowed by validRange
+func getNewExtensionVersion(availableVersions semver.Versions, validRange semver.Range) (semver.Version, bool) {
```
Harkishen-Singh, line 201: Should we rename this to `getNewVersionWithinRange()`?

pkg/pgmodel/extension.go
```diff
@@ -76,62 +76,70 @@ func checkPromscaleExtensionVersion(conn *pgx.Conn) error {
...
 		}
-		return checkExtensionsVersion(conn)
+		return nil
 	}
 
 	comparator := currentVersion.Compare(newVersion)
 	if comparator > 0 {
 		//currentVersion greater than what we can handle, don't use the extension
-		log.Warn("msg", "the promscale extension at a greater version than supported by the connector: %v > %v, proceeding without extension", currentVersion, newVersion)
+		return fmt.Errorf("the extension at a greater version than supported by the connector: %v > %v", currentVersion, newVersion)
```
Harkishen-Singh, line 111: Why did we not specify promscale extension here?

Harkishen-Singh (changes requested):

pkg/pgmodel/extension.go
cevian: no because this now runs for both promscale and timescale

pkg/pgmodel/extension.go
cevian: I don't think so since there is no other function that doesn't take a range

pkg/pgmodel/extension.go
cevian: because this now runs for both extensions. The returned errors are wrapped in parent errors that specify which extension

Harkishen-Singh: Oh, then its fine.

pkg/pgmodel/extension.go
Harkishen-Singh: Oh yes.

Harkishen-Singh (approved):
lgtm!

antekresic (approved):
:+1:

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1154,35 +1189,72 @@ GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) TO prom_writer
...
+AS $func$
+BEGIN
+        IF NOT SCHEMA_CATALOG.is_timescaledb_installed() THEN
+                    RETURN QUERY
+                    SELECT
+                        m.id,
+                        m.metric_name,
+                        m.table_name,
+                        SCHEMA_CATALOG.get_metric_retention_period(m.metric_name) as retention_period,
+                        NULL as chunk_interval,
```
davidkohn88, line 1206: I think I'd cast my nulls here. I know it will do it implicitly on the way out, but to prevent misordering bugs and the like it seems like a good idea.

davidkohn88 (approved):
LGTM One minor nit, but I think it looks fine.

pkg/runner/runner.go
```diff
@@ -140,6 +142,17 @@ func Run(cfg *Config) error {
 }
 
 func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, error) {
+	//The TimescaleDB migration has to happen before other connections
+	//are open. Also, it has to happen as the first command on a connection
+	//thus, we cannot rely on a migration lock here. Instead we assume
+	//that upgrading TimescaleDB will not break existing connectors.
```
JLockerman, line 148: I'm confused by this. IIUC one of two things must be true

1. The TimescaleDB migration must be the only connection open on the DB. In this case that guarantee will ensure that there are no other connectors.
2. The migration only has to be the first command on its connection, in which case we should be able to hold the lock on a different connection

If neither of these are the case, I'd be a little concerned; while I don't think the DB will have many breaking changes after 2.0 at minimum that upgrade is something we need to be careful about.

cevian: 2 is true

JLockerman (approved):

PR #280, opened 2020-10-02
Adds support for statistics in prometheus metrics in a global view

Harkishen-Singh:

This PR adds the following stats to the `prom_info` schema:
1. System-stats: includes number of series (approx for less intensiveness), metrics, label_keys, label_pairs
2. Cardinality of all metrics
3. Cardinality of label keys
4. Cardinality of label pairs (can be expensive when high series cardinality)
5. Approx samples count per metric

Approx values are considered to avoid high resource usage.


pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1193,6 +1233,41 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM _prom_catalog.label
+    ) AS num_label_pairs;
+
+CREATE OR REPLACE VIEW prom_info.metric_cardinality AS
+    SELECT metric_name, row_count(format('prom_series.%I', table_name)::regclass)  AS cardinality
+    FROM _prom_catalog.metric ORDER BY cardinality DESC;
+
+CREATE OR REPLACE VIEW prom_info.label_key_cardinality AS
```
cevian, line 1260: I  think the right query here is actually 
`SELECT k.key AS label_key, coalesce(cnt,0)
  FROM _prom_catalog.label_key k
  LEFT JOIN (SELECT l.key key, count(*) cnt  FROM _prom_catalog.label l GROUP BY key) as l ON (k.key = l.key)`

The reason the above query is better is because it's a single scan of the label table

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1193,6 +1233,41 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+    ) AS num_series_approx,
+    (
+        SELECT count(*) FROM _prom_catalog.metric
+    ) AS num_metric,
+    (
+        SELECT count(*) FROM _prom_catalog.label_key
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM _prom_catalog.label
+    ) AS num_label_pairs;
```
cevian, line 1254:
```suggestion
    ) AS num_labels;
```

cevian (changes requested):
This is a good first commit but we need some improvement

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -418,6 +418,21 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
 AS $function$ SELECT unnest(label_array) $function$;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.label_unnest(anyarray) to prom_reader;
 
+-- row_count is used to count the cardinality of any table. It is mainly used
+-- for knowing the cardinality of a metric.
+CREATE OR REPLACE FUNCTION public.row_count(table_name REGCLASS)
```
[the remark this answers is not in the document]
Harkishen-Singh: row_count aims to be a function that just simplifies the use case of count. As per the current update in doc, the only approximation is number of samples calculation for which we are using `hypertable_approximate_row_count`. Hence, `row_count` is only used in metric_cardinality calculation i.e., `prom_series<metric_name>` where the rows are very less in numbers, thus not being costly.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -471,6 +486,31 @@ $$
 LANGUAGE SQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_or_create_label_key_pos(text, text) to prom_writer;
 
+-- label_cardinality returns the cardinality of a label_pair in the series table.
+-- In simple terms, it means the number of times a label_pair/label_matcher is used
+-- across all the series.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_cardinality(label_id BIGINT)
```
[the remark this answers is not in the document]
Harkishen-Singh: Agreed. The aim of is to give the functionality which the users can take advantage of though, we can just provide this as function with warning in our docs in terms of performance in high cardinality. Being costly, we are not using in our views (the existing view in the PR is discontinued), so this more on the user side now.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1193,6 +1233,41 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+
+CREATE OR REPLACE VIEW prom_info.metric_cardinality AS
+    SELECT metric_name, row_count(format('prom_series.%I', table_name)::regclass)  AS cardinality
+    FROM _prom_catalog.metric ORDER BY cardinality DESC;
+
+CREATE OR REPLACE VIEW prom_info.label_key_cardinality AS
+    SELECT a.key AS label_key, cardinality(ARRAY(SELECT value FROM _prom_catalog.label b WHERE b.key=a.key))
+    FROM _prom_catalog.label_key a order by cardinality desc;
+
+CREATE OR REPLACE VIEW prom_info.label_cardinality AS
```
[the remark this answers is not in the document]
Harkishen-Singh: Agree to this in a previous comment.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -471,6 +486,31 @@ $$
...
+    RETURNS BIGINT
+    LANGUAGE plpgsql
+AS
+$$
+DECLARE
+    occurrence BIGINT;
+    la RECORD;
+    contains BOOLEAN;
+BEGIN
+    occurrence = 0;
```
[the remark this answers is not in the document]
Harkishen-Singh: This works well when I do a simple query, but gives a wrong answer when I use this in a function. Any reason for this?

```
create or replace function _prom_catalog.label_cardinality(label_id int)
    returns int
    language plpgsql
as
$$
begin
    return (SELECT count(*) FROM _prom_catalog.series s WHERE s.labels > array[label_id]);
end;
$$;
```

doing `select _prom_catalog.label_cardinality(1)` gives 1027 but in query gives 4.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1193,6 +1233,41 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+        SELECT count(*) FROM _prom_catalog.metric
+    ) AS num_metric,
+    (
+        SELECT count(*) FROM _prom_catalog.label_key
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM _prom_catalog.label
+    ) AS num_label_pairs;
+
+CREATE OR REPLACE VIEW prom_info.metric_cardinality AS
```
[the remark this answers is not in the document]
Harkishen-Singh: That would be good. We can keep num series as accurate as that should not be approx.

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
Harkishen-Singh: Sure. Normally stats refers to more than one cal, however since we had only one here, so kept cardinality. Ayways, let's keep _stats.

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
cevian: given that it returns a singe query this should not be a plpgsql func but a sql func (more efficient). Also the operator is `@>` not `>`

The code should be something like 

```
create or replace function _prom_catalog.label_cardinality(label_id int)
    returns int
    language sql
as
$$
SELECT count(*) FROM _prom_catalog.series s WHERE s.labels @> array[label_id];
$$ STABLE PARALLEL SAFE;
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -483,6 +498,31 @@ $$
 LANGUAGE SQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_or_create_label_key_pos(text, text) to prom_writer;
 
+-- label_cardinality returns the cardinality of a label_pair in the series table.
+-- In simple terms, it means the number of times a label_pair/label_matcher is used
+-- across all the series.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_cardinality(label_id BIGINT)
```
cevian, line 504: If this is a public function, it should be in SCHEMA_PROM not SCHEMA_CATALOG

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
cevian: Yes but this is used in a view that lists all metrics, so the cost will be very high. I don't see the point of exposing the exact version to users in a view -- users can very easliy just run `SELECT count(*) FROM prom_data_series.<metric_name>` themselves. Getting an approximate value is much harder for users, and so we should provide that.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1279,6 +1319,39 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+    (
+        SELECT count(*) FROM _prom_catalog.label_key
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM _prom_catalog.label
+    ) AS num_labels;
+
+CREATE OR REPLACE VIEW prom_info.metric_stats AS
+    SELECT metric_name,
+    row_count(format('prom_series.%I', table_name)::regclass) AS num_series,
```
cevian, line 1344: yeah this is much more useful as an approximation.

cevian (changes requested):

Harkishen-Singh:
Done @cevian

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1430,6 +1449,34 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+    (
+        SELECT count(*) FROM SCHEMA_CATALOG.label_key
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM SCHEMA_CATALOG.label
+    ) AS num_labels;
+
+CREATE OR REPLACE VIEW SCHEMA_INFO.metric_stats AS
+    SELECT metric_name,
+    SCHEMA_PROM.approximate_row_count(format('prom_series.%I', table_name)::regclass) AS num_series,
```
cevian, line 1469:
```suggestion
    SCHEMA_PROM.approximate_row_count(format('prom_series.%I', table_name)::regclass) AS num_series_approx,
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1430,6 +1449,34 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+    (SELECT row_estimate FROM public.hypertable_approximate_row_count(format('prom_data.%I',table_name)::regclass)) AS num_samples_approx
+    FROM SCHEMA_CATALOG.metric ORDER BY metric_name;
+
+CREATE OR REPLACE VIEW SCHEMA_INFO.label_key_stats AS
+    SELECT k.key AS label_key,
+    COALESCE(cnt, 0) AS cardinality
```
cevian, line 1475: actually should this column simply be added to the existing `SCHEMA_INFO.label` view? Seems like this would belong there.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -430,6 +430,14 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
 AS $function$ SELECT unnest(label_array) $function$;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.label_unnest(anyarray) to prom_reader;
 
+-- approximate_row_count is used to count the approximate number of rows in a table (non-hypertable).
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.approximate_row_count(table_name REGCLASS)
```
cevian, line 434: Looks great

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -483,6 +491,17 @@ $$
 LANGUAGE SQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_or_create_label_key_pos(text, text) to prom_writer;
 
+-- label_cardinality returns the cardinality of a label_pair id in the series table.
+-- In simple terms, it means the number of times a label_pair/label_matcher is used
+-- across all the series.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_cardinality(label_id INT)
```
cevian, line 497:
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_PROM.label_cardinality(label_id INT)
```
since this is a public function

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1430,6 +1449,34 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
...
+        SELECT count(*) FROM SCHEMA_CATALOG.label_key
+    ) AS num_label_keys,
+    (
+        SELECT count(*) FROM SCHEMA_CATALOG.label
+    ) AS num_labels;
+
+CREATE OR REPLACE VIEW SCHEMA_INFO.metric_stats AS
+    SELECT metric_name,
+    SCHEMA_PROM.approximate_row_count(format('prom_series.%I', table_name)::regclass) AS num_series,
+    (SELECT row_estimate FROM public.hypertable_approximate_row_count(format('prom_data.%I',table_name)::regclass)) AS num_samples_approx
```
cevian, line 1470: For getting around the vanilla PG issue I suggest creating a wrapper function called SCHEMA_CATALOG.data_approximate_row_count() that calls either public.hypertable_approximate_row_count() or  SCHEMA_PROM.approximate_row_count() depending on SCHEMA_CATALOG.is_timescaledb_installed().

Search for `SCHEMA_CATALOG.is_timescaledb_installed()` for some examples.

cevian (changes requested):
The views look better. You are really close. You should think about adding golden file tests. Look at pgmodel/testdata/sql/views.sql and golden_file_tests.go

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: sure, that would be better

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: I over came this issue by making metric_stats to be created by a function where we can get more controls over pgsql commands. But this suggestion is definitely better. Thank you.

Harkishen-Singh: how about naming `data_approximate_row_count` => `safe_approximate_row_count`?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1429,9 +1483,30 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.label AS
     lk.value_column_name,
     lk.id_column_name,
     ARRAY(SELECT value FROM SCHEMA_CATALOG.label l WHERE l.key = lk.key ORDER BY value)
-      AS values
-  FROM SCHEMA_CATALOG.label_key lk;
+      AS values,
+    cardinality(ARRAY(SELECT value FROM SCHEMA_CATALOG.label l WHERE l.key = lk.key))
+  FROM SCHEMA_CATALOG.label_key lk ORDER BY cardinality DESC;
```
cevian, line 1488: hmm the values and cardinality columns use the same array, but now uses two scans. For efficiency we should do something like,
```
...
   va.vals as values,
   cardinality(va.values) as num_values
 FROM SCHEMA_CATALOG.label_key lk 
 INNER JOIN LATERAL(SELECT key, array_agg(value ORDER BY value) as vals FROM SCHEMA_CATALOG.label GROUP BY key) 
        AS va ON (va.key = lk.key) 
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -430,6 +430,29 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
...
+    LANGUAGE SQL
+AS
+$$
+    SELECT reltuples::BIGINT FROM pg_class WHERE oid=table_name;
+$$ STABLE PARALLEL SAFE;
+
+-- safe_approximate_row_count returns the approximate row count of a hypertable if timescaledb is installed
+-- else returns the approximate row count in the normal table. This prevents errors in approximate count calculation
+-- if timescaledb is not installed, which is the case in plain postgres support.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.safe_approximate_row_count(table_name REGCLASS) RETURNS BIGINT
```
cevian, line 445: This only works on hypertables so the name should reflect that we only use it on data tables.
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.data_table_approximate_row_count(table_name REGCLASS) RETURNS BIGINT
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1353,10 +1387,30 @@ $func$
 LANGUAGE PLPGSQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) TO prom_writer;
 
+-- CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.metric_stats_view() RETURNS BOOLEAN
```
cevian, line 1390: This is no longer used so should be deleted

pkg/pgmodel/testdata/expected/info_view-postgres.out
```diff
@@ -37,12 +37,12 @@ SELECT id , metric_name , table_name, retention_period, chunk_interval > interva
 (2 rows)
```
cevian, line 37: We should add queries to prom_info.metric_stats and prom_info.system_stats and the label_cardinality function to these tests

cevian (changes requested):

davidkohn88 (commented):
Looks good other than a couple small bits, agree with Mats suggestion on teh cardinality bit.

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: > This only works on hypertables

Only the `public.hypertable_approximate_row_count` works on hypertables, not `schema_prom.approximate_row_count`, and hence gave it a generic name.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -430,6 +430,29 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
...
+-- safe_approximate_row_count returns the approximate row count of a hypertable if timescaledb is installed
+-- else returns the approximate row count in the normal table. This prevents errors in approximate count calculation
+-- if timescaledb is not installed, which is the case in plain postgres support.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.safe_approximate_row_count(table_name REGCLASS) RETURNS BIGINT
+    LANGUAGE PLPGSQL
+AS
+$$
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() THEN
+        RETURN (SELECT row_estimate FROM public.hypertable_approximate_row_count(table_name));
```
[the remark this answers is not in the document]
Harkishen-Singh: So should I let this be or do some check of existence (if possible) for hypertable_approximate_row_count()?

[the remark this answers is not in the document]
davidkohn88: Let me check how we do this elsewhere, I think we just avoid schema qualifying it and let the search path take care of it. and if you've installed timescale in a different schema you just have to put it in your search path.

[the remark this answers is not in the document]
davidkohn88: Elsewhere we use un-schema-qualified names, I think that's what we should do here, any thoughts @cevian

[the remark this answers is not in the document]
Harkishen-Singh: can you explain? Didn't get this @davidkohn88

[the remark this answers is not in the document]
Harkishen-Singh: done, resolving

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -430,6 +430,29 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
...
+-- if timescaledb is not installed, which is the case in plain postgres support.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.safe_approximate_row_count(table_name REGCLASS) RETURNS BIGINT
+    LANGUAGE PLPGSQL
+AS
+$$
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() THEN
+        RETURN (SELECT row_estimate FROM hypertable_approximate_row_count(table_name));
+    END IF;
+    RETURN (SELECT SCHEMA_PROM.approximate_row_count(table_name));
```
davidkohn88, line 453: I don't understand why we have a separate function here instead of just writing the sql in this function?

Harkishen-Singh: Its like to have a function that is smart enough to know that when demanded for approx row cunt, then based on the situation (timescaledb or vanilla postgres), decide and give the rows. this reduces the callers length and makes thing concise.

As per discussion, we should check if its a hypertable than checking if timescaledb installed.

pkg/pgmodel/testdata/expected/info_view-postgres.out
```diff
@@ -37,12 +37,12 @@ SELECT id , metric_name , table_name, retention_period, chunk_interval > interva
...
 
 SELECT * FROM prom_info.label ORDER BY key;
-    key    | value_column_name | id_column_name |        values         
------------+-------------------+----------------+-----------------------
- __name__  | __name__          | __name___id    | {cpu_total,cpu_usage}
- namespace | namespace         | namespace_id   | {dev,production}
- new_tag   | new_tag           | new_tag_id     | {foo}
- new_tag_2 | new_tag_2         | new_tag_2_id   | {bar}
- node      | node              | node_id        | {brain,pinky}
+    key    | value_column_name | id_column_name |        values         | cardinality
```
cevian, line 40: the column name should be num_values. not sure why it's still cardinality? Did you not update the golden file?

Harkishen-Singh: the postgres view got left out, and I guess this wasn't caught since these tests are skipped if running on vanilla postgres

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -431,6 +431,24 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.label_unnest(label_array anyarray)
...
 
+-- safe_approximate_row_count returns the approximate row count of a hypertable if timescaledb is installed
+-- else returns the approximate row count in the normal table. This prevents errors in approximate count calculation
+-- if timescaledb is not installed, which is the case in plain postgres support.
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.safe_approximate_row_count(table_name_input REGCLASS) RETURNS BIGINT
+    LANGUAGE PLPGSQL
+AS
+$$
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() THEN
```
cevian, line 442:
```suggestion
    IF SCHEMA_CATALOG.is_timescaledb_installed()  AND (SELECT count(*) > 0 from _timescaledb_catalog.hypertable WHERE format('%I.%I', schema_name, table_name)::regclass=table_name_input) THEN
```

cevian: Note the two changes:
1. We put both the conditions in one IF, simplifying into one IF/ELSE instead of IF/IF/ElSE/ELSE
2. instead of casting the regclass to text, we cast the text to regclass and we use format. Format makes sure we avoid any possible sql injection.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1525,15 +1552,38 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.metric AS
      *
     FROM SCHEMA_CATALOG.metric_view();
 
+DROP VIEW IF EXISTS SCHEMA_INFO.label;
```
cevian, line 1555: The drop view shouldn't go into the idempotent file. Rather it has to go into a migration file. Take a look at https://github.com/timescale/promscale/pull/291/files#diff-3a439eaeefc867aa848294a901f6931fa90f20ed559f28c3c8db20c3d1ada565 for an example of a migration file. Note that this requires bumping the version (https://github.com/timescale/promscale/pull/291/files#diff-73b8ec8a05213857da6e9598e34c5d7a18a6e5349250908ae6710578236977f2R36)

cevian (changes requested):

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: > cast the text to regclass and we use format

This is great :+1:

cevian (approved):

PR #283, opened 2020-10-05
Add simple function to retrieve a labels_array

cevian:
Retrieves a labels_array from a series_id.


pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -640,6 +640,20 @@ COMMENT ON FUNCTION SCHEMA_PROM.jsonb(labels SCHEMA_PROM.label_array)
...
+RETURNS SCHEMA_PROM.label_array AS $$
+    SELECT
+        labels
+    FROM
+        SCHEMA_CATALOG.series
+    WHERE id = series_id
+$$
+LANGUAGE SQL STABLE PARALLEL SAFE;
+COMMENT ON FUNCTION SCHEMA_PROM.labels(series_id BIGINT)
+IS 'converts a  series_id into a labels array';
```
Harkishen-Singh, line 654: Wondering if this should be `fetches labels array for the given series id` since there is no conversion in this case. So logical mismatch.

Harkishen-Singh (changes requested):

pkg/pgmodel/migrations/sql/idempotent/base.sql
cevian: Good suggestion

JLockerman (approved):
looks reasonable, but could you add a test of the output?

pkg/pgmodel/end_to_end_tests/functions_test.go
```diff
@@ -244,7 +244,17 @@ func TestSQLJsonLabelArray(t *testing.T) {
 
 					if labelSet.Fingerprint() != fingerprintRes {
 						t.Fatalf("Json not equal: id %v\n got\n%v\nexpected\n%v", seriesID, string(jsonRes), string(jsonOrig))
+					}
 
+					err = db.QueryRow(context.Background(), "SELECT jsonb(labels($1))",
+						seriesID).Scan(&jsonRes)
```
Harkishen-Singh, line 250: Not important but I do love seeing this on a single line.

Harkishen-Singh (approved):
Except that small comment for single line, everything lgtm.

PR #286, opened 2020-10-05
Update architecture diagram

cevian:


akulkarni:
I believe the image needs to be cropped

akulkarni (approved):

PR #291, opened 2020-10-08
Add SQL API for managing metric compression setting

### LINKED ISSUE #150: Do not enable compression in connector (make ApacheOnly compatible)
Hi, 

I am experiencing the "err starting ingestor" bug/feature and it's not clear to me how to disable compression.

Downloading the pre-built timescale-prometheus shows no options to disable compression on runtime.  Reading your very thorough documentation on compression, it seems like:

ALTER TABLE <table_name> SET (timescaledb.compress = false);

Should work..  Is this correct?

antekresic:
This commit provides the necessary SQL API for enabling/disabling metric
compression system-wide and on a per-metric basis. It also adds a mechanism
for detecting if compression is not available (as in the case of OSS version
of TimescaleDB) and disables it system-wide.

Closes #150


pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -886,6 +894,118 @@ LANGUAGE SQL VOLATILE;
...
+    INTO STRICT can_compress;
+
+    --if already set to desired value, nothing to do
+    IF SCHEMA_CATALOG.get_metric_compression_setting(metric_name) = new_compression_setting THEN
+        RETURN TRUE;
+    END IF;
+
+    --if compression is missing, cannot enable it
+    IF NOT can_compress AND new_compression_setting THEN
+        RAISE NOTICE 'Cannot enable metrics compression, feature not found';
```
cevian, line 969: This should be an error not a NOTICE

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -886,6 +894,118 @@ LANGUAGE SQL VOLATILE;
...
+            ALTER TABLE SCHEMA_DATA.%I SET (
+                timescaledb.compress,
+                timescaledb.compress_segmentby = 'series_id',
+                timescaledb.compress_orderby = 'time, value'
+            ); $$, metric_table_name);
+        --chunks where the end time is before now()-1 hour will be compressed
+        PERFORM add_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+    ELSE
+        PERFORM remove_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name));
+        PERFORM decompress_chunk(i) from show_chunks(format('SCHEMA_DATA.%I', metric_table_name)) i;
```
cevian, line 990: we also need to ` ALTER TABLE SCHEMA_DATA.%I SET( timescaledb.compress=false);`

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -886,6 +894,118 @@ LANGUAGE SQL VOLATILE;
...
+        WHERE h.schema_name = 'SCHEMA_DATA' 
+        AND h.table_name = metric_table_name)
+    INTO result;
+    RETURN result;
+END
+$$
+LANGUAGE PLPGSQL STABLE PARALLEL SAFE;
+GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_metric_compression_setting(TEXT) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.set_default_compression_setting(compression_setting BOOLEAN)
```
cevian, line 929: This doesn't change the setting on existing tables. I do think we need to add a `default_compression BOOLEAN NOT NULL DEFAULT true` to `SCHEMA_CATALOG.metric` similar to `default_chunk_interval` to tell which metrics follow the default and which do not.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -886,6 +894,118 @@ LANGUAGE SQL VOLATILE;
...
+    END IF;
+
+    --use get_or_create_metric_table_name because we want to be able to set /before/ any data is ingested
+    --needs to run before update so row exists before update.
+    SELECT table_name 
+    INTO STRICT metric_table_name
+    FROM SCHEMA_CATALOG.get_or_create_metric_table_name(set_metric_compression_setting.metric_name);
+
+    IF new_compression_setting THEN
+        EXECUTE format($$
```
cevian, line 980: I think we need to create a function for this so the code to enable compression is not duplicated both here and on line 88

cevian (changes requested):
Some changes needed

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -624,6 +624,109 @@ func TestInsertCompressed(t *testing.T) {
...
+		_, err = db.Exec(context.Background(), "SELECT prom_api.set_default_compression_setting(false)")
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = db.QueryRow(context.Background(), "SELECT _prom_catalog.get_default_compression_setting()").Scan(&compressionEnabled)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if compressionEnabled {
+			t.Error("compression should be disabled by default, was not")
```
cevian, line 654:
```suggestion
			t.Error("compression should have been disabled")
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -886,6 +886,158 @@ LANGUAGE SQL VOLATILE;
...
+    IF NOT can_compress THEN
+        RETURN FALSE;
+    END IF;
+
+    SELECT table_name 
+    INTO STRICT metric_table_name
+    FROM SCHEMA_CATALOG.get_or_create_metric_table_name(metric_name);
+
+    SELECT EXISTS ( 
+        SELECT FROM _timescaledb_config.bgw_policy_compress_chunks p
```
cevian, line 909: This should be based on h.compressed_hypertable_id IS NOT NULL. Not on the bgw policy. This should now work given that we set the  timescaledb.compress = false in the off code.

cevian (approved):
Great work. One of the SQL queries needs to change, but if the change works, feel free to merge

Issue #270, opened 2020-09-25
Broken links to a 'master' directory that no longer exists


Paul-Sorensen-Checkfront:
Various markdown files in this repo point to files/folders that no longer exist.

cevian:
I think all links have now been fixed

PR #295, opened 2020-10-08
Replace broken relative links in documentation with working full URLs.   See:  Issue #270

andersje:
as i was trying to set up promscale and the extensions, this is one instance i found of issue #270 -- I'm pretty sure i got the links right, can someone validate?


cevian (approved):
Thanks @andersje, this looks great

antekresic:
Would you mind squashing the commits so we can get this merged?

docs/binary.md
```diff
@@ -5,10 +5,10 @@
 You can download pre-built binaries for the Promscale
 Connector [on our release page](/releases).
```
antekresic, line 6: This link needs to be changed to https://github.com/timescale/promscale/releases as well.

antekresic (changes requested):
Would you mind updating the releases page link as well? And also squashing the commits into one so we can get this merged?

Thanks!

docs/binary.md
andersje: done!  :)

cevian:
Thanks @andersje

PR #297, opened 2020-10-08
Fix views on Vanilla PG and add tests

cevian:


pkg/pgmodel/testdata/expected/views.out
```diff
@@ -1,22 +1,20 @@
 \set ON_ERROR_STOP 1
 SELECT _prom_catalog.get_or_create_metric_table_name('cpu_usage');
-psql:/testdata/sql/views.sql:4: NOTICE:  adding not-null constraint to column "time"
-DETAIL:  Time dimensions cannot have NULL values
-psql:/testdata/sql/views.sql:4: NOTICE:  adding index _compressed_hypertable_2_series_id__ts_meta_sequence_num_idx ON _timescaledb_internal._compressed_hypertable_2 USING BTREE(series_id, _ts_meta_sequence_num)
  get_or_create_metric_table_name 
 ---------------------------------
- (1,cpu_usage)
+ (1,cpu_usage,t)
```
JLockerman, line 5: how come this changed? I don't see anything in the SQL that should have changed it

JLockerman:
In general looks good to me, will rereview in the morning when awake

pkg/pgmodel/testdata/expected/views.out
cevian: This test was silently broken for a while (see first commit) so the golden file was out of date.

antekresic (approved):
Looks good :+1:

pkg/pgmodel/testdata/expected/views.out
JLockerman: ah, ouch.

JLockerman (approved):

PR #298, opened 2020-10-09
Fix bug with deletion of metric name labels.

cevian:
Previously, when all the data for a metric was dropped for data
retention, the label for the metric name was also dropped.
This caused problems since metric name ids are checked in CHECK
constraints and shouldn't change. It seems a bad idea to
let labels ids for metric names change in general. So,
we simply don't delete them now.


pkg/pgmodel/end_to_end_tests/drop_test.go
```diff
@@ -381,3 +381,134 @@ func TestSQLDropMetricChunk(t *testing.T) {
...
+			t.Errorf("unexpected series count: %v", count)
+		}
+
+		err = db.QueryRow(context.Background(), `SELECT count(*) FROM _prom_catalog.label`).Scan(&count)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if count != 1 {
+			t.Errorf("unexpected label count: %v", count)
```
Harkishen-Singh, line 466: I know we do not have this throughout the repository code, but places like these should use `testutil` and then `testutil.Equals(t, expected, actual, message, args...)`. I will want things like these to go in for cleanup times so that debugging is easier.

Harkishen-Singh (approved):
lgtm except the cleanup thing which can be a todo as its across the entire code.

pkg/pgmodel/end_to_end_tests/drop_test.go
```diff
@@ -381,3 +381,134 @@ func TestSQLDropMetricChunk(t *testing.T) {
...
 }
+
+// Tests case that all metric data was dropped and then the metric came back alive
+func TestSQLDropAllMetricData(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+	if !*useTimescaleDB {
+		t.Skip("This test only runs on installs with TimescaleDB")
+	}
```
JLockerman, line 392: It looks like we run the DELETE independent of whether timescaledb is installed, should we have a vanilla prom test?

JLockerman (approved):

antekresic (approved):

PR #300, opened 2020-10-12
Fix logging output to match rest of the project

antekresic:


cevian (approved):
Thanks for going through everything

PR #303, opened 2020-10-14
Fix error reporting to prevent panic

cevian:
Previously, runInserterRoutine reported errors to two places:
1) the errChan originally starting the inserter
2) The errChan associated with the request it's handling

The reporting in (1) caused a send-on-closed-channel error
since the request that originally started the insert routine
is actually done after the first error and closes its errChan.

Another words, errors should always be reported on the errChan
assoicated with the request being handled, since we know that
request is still not done. As a plus, that's semantically more
correct too.

Thus, we get rid of (1) and keep (2). We also get rid of passing
errChan not associated with a request down -- this simplifies code
and prevents this type of error in the future.

### DIFF
--- a/pkg/pgmodel/sql_ingest.go
+++ b/pkg/pgmodel/sql_ingest.go
@@ -233,7 +233,7 @@ func (p *pgxInserter) InsertData(rows map[string][]samplesInfo) (uint64, error) 
 }
 
 func (p *pgxInserter) insertMetricData(metric string, data []samplesInfo, finished *sync.WaitGroup, errChan chan error) {
-	inserter := p.getMetricInserter(metric, errChan)
+	inserter := p.getMetricInserter(metric)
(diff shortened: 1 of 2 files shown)


antekresic (commented):
Looks good but needs fixing tests.

antekresic (approved):

JLockerman (approved):

JLockerman:
perhaps we should log the error 🤔

cevian:
@JLockerman the error will be logged by reporting method 2, which remains.

JLockerman:
@cevian I know, but since this is a more fundamental failure than just a single request failing (usually it means the table can't be created IIRC) I wonder if it should be called out especially

cevian:
@JLockerman the issue is this is retried for every request. I don't want to pollute the log too much

JLockerman:
@cevian fair enough

PR #304, opened 2020-10-14
Epoch-Based Cache Validation

JLockerman:
Epoch-Based cache validation

Previously our SeriesID cache would never delete entries. This could
cause issues as SeriesIDs are deleted from the database if we see that
no data table uses them; in the event that the series is resurrected
after that, new rows in that series would contain an ID that did not
refer to any series set stored. 

In order to fix this, we introduce an epoch mechanism to validate that
the SeriesIDs and INSERT is adding to the DB could not have been
deleted. At a high level, the mechanism works as follows:
  - The database stores an epoch counter that monotonically increases
    whenever SeriesIDs are deleted.
  - The SeriesID cache stores the smallest epoch at which it read an
    ID from the DB. This epoch is sent to the inserter along with the
    data to be inserted.
  - As the first statement of the INSERT transaction, the inserter
    validates that the epoch at the DB is low enough that none of the
    SeriesIDs being inserted could have been deleted. If the epoch
    isn't valid the INSERT transaction is aborted.

Choosing which epochs are valid for insertions has subtle performance
implications. While we could dictate that the epoch we read must match
the epoch we see during an INSERT exactly, this would cause thundering
herd aborts whenever we delete SeriesIDs. Instead, we allow the epoch
at insert to be 1 greater than the one we saw at read time, and delay
the deletion of SeriesIDs until after they've been unused for at least
a full epoch. Graphically, with R being the earliest point in the
epoch numberline at which we read, and I being the point we write, we
allow read/insert combinations like:
```
  R              R
0|1|2|3   or   0|1|2|3
  I                I
```

but not ones like
```
  R              R
0|1|2|3   or   0|1|2|3
      I        I
```

Similarly, if M is the first time we mark a SeriesID as being deletable
because there are no data columns that reference it, and D is the
point at which we actually delete it (note that it is these deletes
that increment the epoch, so they properly happen /between/ epochs. We
allow mark/delete combinations like:
```
M
0|1|2|3
     D
```

but not ones like
```
M                M              M
0|1|2|3   or   0|1|2|3   or   0|1|2|3
 D                  D          D
```

Over all the valid read/insert and mark/delete combinations nest
something like one of:
```
M R            M R
0|1|2|3   or   0|1|2|3
  I  D             ID
```

So a recheck to see if SeriesIDs are still live before actual deletion
ensures that no dead IDs are inserted.

This PR is ~not yet~ complete. Still to come are:

- [x] Documentation for the new epoch scheme
- [x] Test that we can resurrect a fully deleted series
- [x] Test that our cache-trashing code works correctly
- [x] should the series view now include a `delete_epoch IS NULL` qual?
- [x] A real commit message


pkg/pgmodel/sql_ingest.go
```diff
@@ -711,6 +753,7 @@ func (h *insertHandler) setSeriesIds(sampleInfos []samplesInfo) (string, error)
 		}
 
 		batch.Queue("BEGIN;")
+		batch.Queue(getEpochSQL)
```
cevian, line 756: Do we need this once per series? or just once per call? It seems the latter since the epoch can't go down?

pkg/pgmodel/sql_ingest.go
```diff
@@ -721,42 +764,104 @@ func (h *insertHandler) setSeriesIds(sampleInfos []samplesInfo) (string, error)
...
+}
+
+func (h *insertHandler) refreshSeriesCache() {
+	newEpoch, err := h.getServerEpoch()
+	if err != nil {
+		// we don't have any great place to report this error, and if the
+		// connection recovers we can still make progress, so we'll just log it
+		// and continue execution
+		msg := fmt.Sprintf("error refreshing the series cache for %s", h.metricTableName)
+		log.Error("msg", msg, "err", err)
```
cevian, line 822:
```suggestion
		log.Error("msg", msg, "metric_table", h.metricTableName, "err", err)
```

pkg/pgmodel/sql_ingest.go
```diff
@@ -721,42 +764,104 @@ func (h *insertHandler) setSeriesIds(sampleInfos []samplesInfo) (string, error)
...
+
+	if newEpoch != h.seriesCacheEpoch {
+		h.seriesCache = map[string]SeriesID{}
+		h.seriesCacheEpoch = newEpoch
+	}
+}
+
+func (h *insertHandler) getServerEpoch() (Epoch, error) {
+	// using a batch to get QueryRow
+	batch := h.conn.NewBatch()
```
cevian, line 836: This does not need to be a batch.

cevian (changes requested):

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1054,94 +1070,136 @@ BEGIN
...
+            INTO last_epoch_time, deletion_epoch, next_epoch;
+        IF last_epoch_time + '1 hour' < now() THEN
+                RETURN;
+        END IF;
+
+        EXECUTE format($query$
+            -- recheck that the sries IDs we might delete are actually dead
+            WITH dead_series AS (
+                SELECT id FROM SCHEMA_DATA_SERIES.%1$I
+                    WHERE delete_epoch <= deletion_epoch
```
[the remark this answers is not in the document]
JLockerman: I really hope that isn't necessary, `<=` implies `NOT NULL` anyway

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -744,12 +745,16 @@ BEGIN
    --the data table could be locked during label key creation
    --and must be locked before the series parent according to lock ordering
    EXECUTE format($query$
-        LOCK TABLE ONLY SCHEMA_DATA.%1$I IN ACCESS SHARE MODE
+        LOCK TABLE ONLY SCHEMA_DATA.%1$I IN ROW EXCLUSIVE MODE
```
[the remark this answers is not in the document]
JLockerman: we're going to be locking it like this to do the UPDATE, and getting it early seemed less likely to deadlock

pkg/pgmodel/sql_ingest.go
JLockerman: we probably don't, but I don't really want to make the Id-reading code more complicated right now

pkg/pgmodel/sql_ingest.go
JLockerman: yeah, I'm just sing it to get `QueryRow` (halves the size of this function)

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1054,94 +1070,136 @@ BEGIN
...
+                SELECT id FROM SCHEMA_DATA_SERIES.%1$I
+                    WHERE delete_epoch <= deletion_epoch
+                        AND NOT EXISTS (
+                            SELECT 1 FROM SCHEMA_DATA.%1$I
+                            WHERE id = label_id LIMIT 1
+                        )
+            ), deleted_series AS (
+                DELETE FROM SCHEMA_DATA_SERIES.%1$I
+                WHERE delete_epoch <= deletion_epoch
+                    AND id IN dead_series -- concurrency means we need this qual in both
```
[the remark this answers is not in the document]
JLockerman: My bet is that our tests never exercise this code, and since it's in a `format()` it's only checked when run

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1037,15 +1042,26 @@ LANGUAGE PLPGSQL VOLATILE;
 COMMENT ON FUNCTION SCHEMA_PROM.reset_metric_compression_setting(TEXT)
 IS 'resets the compression setting for a specific metric to using the default';
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.epoch_abort()
+RETURNS VOID AS $func$
+BEGIN
+    RAISE EXCEPTION 'epoch to old to continue INSERT';
```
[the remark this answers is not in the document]
JLockerman: Looking through Appendix A, I don't think there's a better state than `raise_exception`

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
JLockerman: in my very sketchy experiments it looks like postgres prefers the `NOT NULL` index regardless

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1054,94 +1070,136 @@ BEGIN
...
+        ELSE
+                EXECUTE format($$ DELETE FROM SCHEMA_DATA.%I WHERE time < %L $$, metric_table, older_than);
+        END IF;
+
+        -- wait for current insertions to be done, and ensure that all future
+        -- insertions will see the epoch change
+        LOCK TABLE SCHEMA_CATALOG.ids_epoch IN ACCESS EXCLUSIVE MODE;
+
+        UPDATE SCHEMA_CATALOG.ids_epoch
+            SET (current_epoch, last_update_time) = (current_epoch + 1, now());
```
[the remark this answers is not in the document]
JLockerman: sigh, yeah, that's what `next_epoch` was for, I guess I never did the second half

pkg/pgmodel/migrations/sql/preinstall/003-tables.sql
```diff
@@ -23,9 +23,14 @@ INSERT INTO public.prom_installation_info(key, value) VALUES
...
     id bigint NOT NULL,
     metric_id int NOT NULL,
-    labels SCHEMA_PROM.label_array NOT NULL --labels are globally unique because of how partitions are defined
+    labels SCHEMA_PROM.label_array NOT NULL, --labels are globally unique because of how partitions are defined
+    delete_epoch bigint -- epoch after which this row can be deleted
 ) PARTITION BY LIST(metric_id);
 CREATE INDEX series_labels_id ON SCHEMA_CATALOG.series USING GIN (labels);
+CREATE INDEX series_deleted
+    ON SCHEMA_CATALOG.series
+    USING GIN (labels) -- no default operator for bigint so no epoch
```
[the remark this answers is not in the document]
JLockerman: I was thinking this index would be needed for either the deadness scan, or the label read, but looking at it now I don't think it is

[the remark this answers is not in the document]
JLockerman: our WHERE clauses
```
WHERE delete_epoch IS NOT NULL
            AND labels = (SELECT * FROM cte)

WHERE delete_epoch IS NULL
                AND series_id IN (SELECT * FROM confirmed_drop_series)

WHERE delete_epoch <= deletion_epoch
    AND NOT EXISTS (
        SELECT 1 FROM SCHEMA_DATA.%1$I
        WHERE id = label_id LIMIT 1
    )

WHERE delete_epoch <= deletion_epoch
                    AND id IN (SELECT id FROM dead_series) -- (I'm not sure both this and the above is needed
```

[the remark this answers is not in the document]
JLockerman: The first and second ones must be driven based on `series_labels_id` because they're looking for `NULL`s. The third and fourth should be driven by `delete_epoch`, and having `id` could change them to and IndexOnly scans.

I think this means the index should be btree `(delete_epoch, id)`

pkg/pgmodel/migrations/sql/preinstall/003-tables.sql
```diff
@@ -23,9 +23,14 @@ INSERT INTO public.prom_installation_info(key, value) VALUES
 CREATE TABLE SCHEMA_CATALOG.series (
     id bigint NOT NULL,
     metric_id int NOT NULL,
-    labels SCHEMA_PROM.label_array NOT NULL --labels are globally unique because of how partitions are defined
+    labels SCHEMA_PROM.label_array NOT NULL, --labels are globally unique because of how partitions are defined
+    delete_epoch bigint -- epoch after which this row can be deleted
 ) PARTITION BY LIST(metric_id);
 CREATE INDEX series_labels_id ON SCHEMA_CATALOG.series USING GIN (labels);
+CREATE INDEX series_deleted
```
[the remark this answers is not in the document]
JLockerman: Note that we don't currently do so for `series_labels_id`, so if we do need to, that may be a bug there...

pkg/pgmodel/migrations/sql/preinstall/003-tables.sql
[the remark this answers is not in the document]
JLockerman: in my experiment, it looks like the index gets created
```SQL
mat=> create table test_table_root(i int, j int) partition by list (i);
CREATE TABLE

mat=> create index on test_table_root (i) where j is not null;
CREATE INDEX

mat=> create table test_table_child(i int, j int);
CREATE TABLE

mat=> ALTER TABLE test_table_root ATTACH PARTITION test_table_child FOR VALUES IN (1);
ALTER TABLE

mat=> \d+ test_table_root
                        Partitioned table "public.test_table_root"
 Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description 
--------+---------+-----------+----------+---------+---------+--------------+-------------
 i      | integer |           |          |         | plain   |              | 
 j      | integer |           |          |         | plain   |              | 
Partition key: LIST (i)
Indexes:
    "test_table_root_i_idx" btree (i) WHERE j IS NOT NULL
Partitions: test_table_child FOR VALUES IN (1)

mat=> \d+ test_table_child
                             Table "public.test_table_child"
 Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description 
--------+---------+-----------+----------+---------+---------+--------------+-------------
 i      | integer |           |          |         | plain   |              | 
 j      | integer |           |          |         | plain   |              | 
Partition of: test_table_root FOR VALUES IN (1)
Partition constraint: ((i IS NOT NULL) AND (i = 1))
Indexes:
    "test_table_child_i_idx" btree (i) WHERE j IS NOT NULL
Access method: heap
```

pkg/pgmodel/migrations/sql/preinstall/003-tables.sql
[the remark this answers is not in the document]
JLockerman: in my experiment, it looks like the index gets created

```SQL
mat=> create table test_table_root(i int, j int) partition by list (i);
CREATE TABLE

mat=> create index on test_table_root (i) where j is not null;
CREATE INDEX

mat=> create table test_table_child(i int, j int);
CREATE TABLE

mat=> ALTER TABLE test_table_root ATTACH PARTITION test_table_child FOR VALUES IN (1);
ALTER TABLE

mat=> \d+ test_table_root
                        Partitioned table "public.test_table_root"
 Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description 
--------+---------+-----------+----------+---------+---------+--------------+-------------
 i      | integer |           |          |         | plain   |              | 
 j      | integer |           |          |         | plain   |              | 
Partition key: LIST (i)
Indexes:
    "test_table_root_i_idx" btree (i) WHERE j IS NOT NULL
Partitions: test_table_child FOR VALUES IN (1)

mat=> \d+ test_table_child
                             Table "public.test_table_child"
 Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description 
--------+---------+-----------+----------+---------+---------+--------------+-------------
 i      | integer |           |          |         | plain   |              | 
 j      | integer |           |          |         | plain   |              | 
Partition of: test_table_root FOR VALUES IN (1)
Partition constraint: ((i IS NOT NULL) AND (i = 1))
Indexes:
    "test_table_child_i_idx" btree (i) WHERE j IS NOT NULL
Access method: heap
```

pkg/pgmodel/sql_ingest.go
```diff
@@ -467,8 +484,8 @@ func (h *insertHandler) handleReq(req insertDataRequest) bool {
 	// we fill in any SeriesIds we have in cache now so we can free any Labels
 	// that are no longer needed, and because the SeriesIds might get flushed.
 	// (neither of these are that critical at the moment)
-	h.fillKnowSeriesIds(req.data)
-	needsFlush := h.pending.addReq(req)
+	_, epoch := h.fillKnowSeriesIds(req.data)
```
davidkohn88, line 487: fillKnow -> fillKnown?

pkg/pgmodel/migrations/sql/preinstall/003-tables.sql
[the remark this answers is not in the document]
davidkohn88: Yeah, the docs say that for declaratively partitioned tables those indexes will be created.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1054,94 +1071,138 @@ BEGIN
...
-                 ORDER BY time ASC
-                 LIMIT 1
+        EXECUTE format($query$
+            -- recheck that the sries IDs we might delete are actually dead
+            WITH dead_series AS (
+                SELECT id FROM SCHEMA_DATA_SERIES.%1$I
+                    WHERE delete_epoch <= deletion_epoch
+                        AND NOT EXISTS (
+                            SELECT 1 FROM SCHEMA_DATA.%1$I
+                            WHERE id = label_id LIMIT 1
```
davidkohn88: Should this be label id? not series id?

davidkohn88: And why isn't this using the same optimizations as above?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1054,94 +1071,138 @@ BEGIN
...
+                    WHERE series_exists.labels && ARRAY[labels.label_id]
+                    LIMIT 1
+                )
             )
-        )
-        DELETE FROM SCHEMA_CATALOG.label
-        WHERE id IN (SELECT * FROM confirmed_drop_labels) AND key != '__name__';
-    $query$, metric_table) USING label_array;
+            DELETE FROM SCHEMA_CATALOG.label
+            WHERE id IN (SELECT * FROM confirmed_drop_labels) AND key != '__name__';
```
davidkohn88, line 1184: Are we not using the same epoching stuff here? Seems like we'd run into the same issues?

davidkohn88: Aren't labels not particular to a given series? I thought they were shared? But maybe I'm wrong...

davidkohn88 (commented):
There are a few questions here, I might just have a few things wrong about our approach though.

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: it should be `series_id` 😅 
didn't think of it

JLockerman: thoughI'm not sure it's relevant, since we don't have a natural time qual...

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: We never store label_ids in promscalem, so they're safe to delete as soon as there's no series that references them

davidkohn88: Really? no caching for that? Okay. and also, what I was really asking here was whether looking for usage of a label in a given series is going to work or if we need to look everywhere, which could be a bug in the old code...

davidkohn88: Ah. I see why now on not storing them in promscale memory, but I don't get the other bit...

JLockerman: label has `UNIQUE(key, value)` which certainly __suggests__ they can be shared across series...
@cevian?

pkg/pgmodel/sql_ingest.go
cevian: I think this is pretty bad performance wise. I'd rather get this fixed now

cevian (changes requested):
Left some comments. Obviously this needs a lot more testing

pkg/pgmodel/sql_ingest.go
```diff
@@ -443,14 +454,20 @@ func (h *insertHandler) hasPendingReqs() bool {
...
+				return false
+			}
 
-	h.handleReq(req)
+			h.handleReq(req)
 
-	return true
+			return true
+		case <-h.seriesCacheRefresh.C:
+			h.refreshSeriesCache()
```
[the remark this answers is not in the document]
JLockerman: I don't think so; the cache is necissarily empty at that time, and the code __should__ be able to handle the nonce epoch, but then, I __am__ seeing a epoch-check bug...

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -744,12 +745,16 @@ BEGIN
...
-        LOCK TABLE ONLY SCHEMA_DATA.%1$I IN ACCESS SHARE MODE
+        LOCK TABLE ONLY SCHEMA_DATA.%1$I IN ROW EXCLUSIVE MODE
     $query$, table_name);
 
    EXECUTE format($query$
-    WITH CTE AS (
+    WITH cte AS (
         SELECT SCHEMA_CATALOG.get_or_create_label_array($1, $2, $3)
+    ), updates AS (
+        UPDATE series SET delete_epoch = NULL
```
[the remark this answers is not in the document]
davidkohn88: Yeah. that looks better to me.

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
JLockerman: my earlier comment was incorrect; I'd confused `SCHEMA_DATA` and `SCHEMA_DATA_SERIES`

pkg/pgmodel/end_to_end_tests/drop_test.go
```diff
@@ -452,16 +531,25 @@ func TestSQLDropAllMetricData(t *testing.T) {
...
 		if count != 1 {
+			t.Errorf("unexpected series count: %v", count)
+		}
+
+		err = db.QueryRow(context.Background(), `SELECT count(*) FROM _prom_catalog.label`).Scan(&count)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if count != 2 {
```
cevian, line 552: we should be able to avoid restarting the ingestor at line 570 near (//Restart ingestor to avoid stale cache issues.) now that the cache issues have been solved :)

pkg/pgmodel/end_to_end_tests/drop_test.go
```diff
@@ -332,53 +333,131 @@ func TestSQLDropMetricChunk(t *testing.T) {
...
+			if count != 2 {
+				t.Errorf("unexpected row count: %v @ %v", count, loc)
+			}
+
+			err = db.QueryRow(context.Background(), `SELECT count(*) FROM _prom_catalog.series`).Scan(&count)
+			if err != nil {
+				t.Error(loc, err)
+			}
+
+			// one of the series should be removed
```
cevian, line 424: how is this working after the first call to drop_metric_chunks. I would of expected at least 2 calls necessary before this is true

pkg/pgmodel/end_to_end_tests/drop_test.go
```diff
@@ -5,6 +5,7 @@ package end_to_end_tests
```
cevian, line 5: We should also add a test where the series is really truly deleted and the ingestor's cache handling correctly handles that and re-creates a new series

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1098,21 +1078,68 @@ BEGIN
...
+        FROM SCHEMA_CATALOG.ids_epoch LIMIT 1
+        INTO last_epoch_time, deletion_epoch, next_epoch;
+
+    -- we don't want to delete too soon
+    IF ran_at < last_epoch_time + '1 hour' THEN
+        RETURN;
+    END IF;
+
+    EXECUTE format($query$
+        -- recheck that the sries IDs we might delete are actually dead
```
cevian, line 1120:
```suggestion
        -- recheck that the series IDs we might delete are actually dead
```

pkg/pgmodel/sql_ingest.go
```diff
@@ -719,44 +768,107 @@ func (h *insertHandler) setSeriesIds(sampleInfos []samplesInfo) (string, error)
...
+func (h *insertHandler) refreshSeriesCache() {
+	newEpoch, err := h.getServerEpoch()
+	if err != nil {
+		// we don't have any great place to report this error, and if the
+		// connection recovers we can still make progress, so we'll just log it
+		// and continue execution
+		msg := fmt.Sprintf("error refreshing the series cache for %s", h.metricTableName)
+		log.Error("msg", msg, "metric_table", h.metricTableName, "err", err)
+		// Trash the cache just in case an epoch change occured, seems safer
+		h.seriesCache = map[string]SeriesID{}
```
cevian, line 842: should we also add h.seriesCacheEpoch = -1 here?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1123,26 +1150,100 @@ BEGIN
...
         WHERE id IN (SELECT * FROM confirmed_drop_labels) AND key != '__name__';
     $query$, metric_table) USING label_array;
 
-   IF SCHEMA_CATALOG.is_timescaledb_installed() THEN
-        PERFORM drop_chunks(table_name=>metric_table, schema_name=> 'SCHEMA_DATA', older_than=>older_than, cascade_to_materializations=>FALSE);
-   ELSE
-        EXECUTE format($$ DELETE FROM SCHEMA_DATA.%I WHERE time < %L $$, metric_table, older_than);
-   END IF;
-   RETURN true;
+    -- wait for current insertions to be done, and ensure that all future
```
cevian, line 1163: is it worth avoiding this lock and the epoch increment if there are no series rows where delete_epoch IS NOT NULL?

cevian (commented):
Left a few comments. I think this needs more end-to-end tests and some docs. I'd like to see a simplified proof in a README somewhere in here (at least an intuition. 

Also a question: should the series view now include a delete_epoch IS NULL qual

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: @cevian `drop_metric_chunks()` is called in a loop 6 times [here](https://github.com/timescale/promscale/pull/304/files#diff-59a0f996197a51fe1276a3e186e4ab912e5dc3673ee9bf2c0699549834cb1605R397-R404) before this function is ever called

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: 🤔 
@cevian I might duplicate this test then; I think it's good to have a test where one ingestor deletes the Id and another one resurrects it...

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: I really should get a new keyboard 😅

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: @cevian I think the logic would have to be
```sql
count(deleted_series) = 0 AND count(delete_epoch IS NOT NULL) = 0
```
but I think we should hold off for now: this is already getting complicated to test, and I don't think the once-an-hour barrier will be that large a performance issue

pkg/pgmodel/sql_ingest.go
JLockerman: ehh, why not

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: 👍 
I also want to add a test that the cache-trashing code works correctly...

JLockerman:
> I think this needs more end-to-end tests and some docs. I'd like to see a simplified proof in a README somewhere in here 
yup, added to the TODO list at the top

> should the series view now include a delete_epoch IS NULL qual
Oh, that's interesting. I'm inclined to say "yes" though I want @davidkohn88 to weigh in.

JLockerman:
> should the series view now include a delete_epoch IS NULL qual

note that there is a migration issue in that old views won't contain the qual unless we figure out a way to migrate them

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: > We should also add a test where the series is really truly deleted and the ingestor's cache handling correctly handles that and re-creates a new series

added to the end of `TestSQLDropMetricChunk()`

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: > we should be able to avoid restarting the ingestor at line 570 

It turns out this is more difficult than I thought; the Ingestor is stuck on an old epoch until it refreshes, so it cannot perform new Inserts once the epoch changes. I think our options are to add a way for the Ingestor to tell the per-metric goroutines to trash their caches (I don't love increasing the complexity of those messages, but it might be a good thing to have anyway) or just test it in the unit tests

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: > I also want to add a test that the cache-trashing code works correctly...

added in `sql_tests.go`

cevian, line 5: Do we have an end-to-end test with an epoch abort? That should be doable end-to-end? Can we make sure that a retry works?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1098,21 +1078,68 @@ BEGIN
...
+        FROM SCHEMA_CATALOG.ids_epoch
+        WHERE delete_epoch IS NULL
+            AND id IN (SELECT * FROM confirmed_drop_series)
+    $query$, metric_table, older_than, check_time);
+END
+$func$
+LANGUAGE PLPGSQL VOLATILE;
+
+
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.clear_old_ids(
```
cevian, line 1099:
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_expired_series(
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1037,50 +1043,24 @@ LANGUAGE PLPGSQL VOLATILE;
...
+        INTO db_epoch;
+    RAISE EXCEPTION 'epoch % to old to continue INSERT, current: %',
+        user_epoch, db_epoch
+        USING ERRCODE='PS001';
+END;
+$func$ LANGUAGE PLPGSQL VOLATILE;
+COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
+IS 'ABORT an INSERT transaction due to the ID epoch being out of date';
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.mark_unused_ids(
```
cevian, line 1060:
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.mark_unused_series(
```

pkg/pgmodel/sql_ingest.go
```diff
@@ -715,44 +764,111 @@ func (h *insertHandler) setSeriesIds(sampleInfos []samplesInfo) (string, error)
...
+			return "", epoch, err
 		}
 	}
 
-	return tableName, nil
+	return tableName, epoch, nil
 }
 
-func (p *pendingBuffer) addReq(req insertDataRequest) bool {
+func (h *insertHandler) refreshSeriesCache() {
```
cevian, line 832: I think this must be called if the insert fails with a 'PS001' error. That error should cause a refresh and the propagate up. That makes sure that a retry doesn't hit this same error again.

cevian (changes requested):
Getting close. A few comments. Most important I think we should test the abort case and also we must refresh on abort.

pkg/pgmodel/end_to_end_tests/drop_test.go
JLockerman: e2e showing an abort is [here](https://github.com/timescale/promscale/pull/304/files#diff-59a0f996197a51fe1276a3e186e4ab912e5dc3673ee9bf2c0699549834cb1605R518-R521) getting a successful retry has to be done in the unit tests right now because we don't want to wait the cache refresh interval.

pkg/pgmodel/sql_ingest.go
JLockerman: I don't see a way to reasonably do without adding a backedge to our gouroutine communication graph, and for something that likely to deadlock I'd rather put it in its own PR, with time to think through all of test cases we should have. I'm also hesitant to delay this PR any longer than it has been; right this bug can cause unrecoverable data corruption, and  fixing that, even if we introduce a brief DoS in an unlikely desynchronization event, seems like enough of an improvement that I don't think we should delay releasing it unless we absolutely have to.

JLockerman: (@cevian)

cevian (approved):

PR #305, opened 2020-10-15
Make example docker-compose clearer

fayak:
This simply makes the example docker-compose.yml clearer to read, normalize it a bit with common docker-compose usage, and avoid a pitfall with `db-ssl-mode` being the only option not set with env var

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.0'
 
 services:
   db:
-    image: timescale/timescaledb:latest-pg12
+    image: timescaledev/promscale-extension:latest-pg12
     ports:
       - 5432:5432/tcp
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,33 +1,36 @@
+version: '3.0'
+
 services:
   db:
+    image: timescale/timescaledb:latest-pg12
+    ports:
+      - 5432:5432/tcp
     environment:
       POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,33 +1,36 @@
+version: '3.0'
+
 services:
   db:
+    image: timescale/timescaledb:latest-pg12
```
blagojts, line 5: Can you make the db use `timescaledev/promscale-extension:latest-pg12`? It's the image with the promscale extension included for better performance?

blagojts (approved):
Thank you for this, it makes the compose file much better. I'll approve the PR but If you have the time can you change the db image used?

docker-compose.yml
fayak: Oh I had no idea this image exists haha, I thought I had to have 2 containers :sweat_smile: 

That could help people !

blagojts: You still need the two containers. 
timescale/promscale is the prometheus connector 
timescaledev/promscale-extension is postgres+timescaledb+"promscale-extension" 

This extension is a postgres extension that contains code that helps promscale do ingest and querying multiple times faster.

blagojts: you should revert to using the two containers, just change the image for the db

fayak: Hum, I feel like this could benefit from having a bit more of documentation. I'll update it and push it

blagojts (approved):

PR #309, opened 2020-10-19
Add code documentation for query/read path of the connector

antekresic:
This commit also contains some light refactoring to make the code easier to
read and understand.

### DIFF
--- a/pkg/pgmodel/querier.go
+++ b/pkg/pgmodel/querier.go
@@ -19,11 +19,17 @@ type Reader interface {
 // Querier queries the data using the provided query data and returns the
 // matching timeseries.
 type Querier interface {
+	// Query returns resulting timeseries for a query.
 	Query(*prompb.Query) ([]*prompb.TimeSeries, error)
+	// Select returns a series set that matches the supplied query parameters.
 	Select(mint int64, maxt int64, sortSeries bool, hints *storage.SelectHints, path []parser.Node, ms ...*labels.Matcher) (storage.SeriesSet, parser.Node)
+	// LabelNames returns all the distinct label names in the system.
 	LabelNames() ([]string, error)
+	// LabelValues returns all the distinct values for a given label name.
 	LabelValues(labelName string) ([]string, error)
+	// NumCachedLabels returns the number of labels cached in the system.
 	NumCachedLabels() int
+	// LabelsCacheCapacity returns the capacity of the labels cache.
 	LabelsCacheCapacity() int
 }
 
--- a/pkg/pgmodel/sql_reader.go
+++ b/pkg/pgmodel/sql_reader.go
@@ -5,6 +5,7 @@ package pgmodel
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 
@@ -26,6 +27,7 @@ const (
 	getMetricsTableSQL = "SELECT table_name FROM " + catalogSchema + ".get_metric_table_name_if_exists($1)"
 	getLabelNamesSQL   = "SELECT distinct key from " + catalogSchema + ".label"
 	getLabelValuesSQL  = "SELECT value from " + catalogSchema + ".label WHERE key = $1"
+	getLabelsSQL       = "SELECT (labels_info($1::int[])).*"
 )
 
 // NewPgxReaderWithMetricCache returns a new DBReader that reads from PostgreSQL using PGX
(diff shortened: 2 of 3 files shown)


blagojts (approved):

pkg/pgmodel/querier.go
```diff
@@ -19,11 +19,17 @@ type Reader interface {
 // Querier queries the data using the provided query data and returns the
 // matching timeseries.
 type Querier interface {
+	// Query returns resulting timeseries for a query.
 	Query(*prompb.Query) ([]*prompb.TimeSeries, error)
+	// Select returns a series set that matches the supplied query parameters.
```
Harkishen-Singh, line 24: Should we be more specific with `query parameters`, like `... matches the supplied series hints`

pkg/pgmodel/sql_reader.go
```diff
@@ -77,15 +79,18 @@ func (q *pgxQuerier) HealthCheck() error {
...
 	return q.labels.Len()
 }
 
+// LabelsCacheCapacity implements the Querier interface.
 func (q *pgxQuerier) LabelsCacheCapacity() int {
 	return q.labels.Cap()
 }
 
-// entry point from our own version of the prometheus engine
+// Select implements the Querier interface. It is the entry point from our
```
Harkishen-Singh, line 92: Maybe, `from` => `for`?

pkg/pgmodel/sql_reader.go
```diff
@@ -119,137 +125,30 @@ func (q *pgxQuerier) Query(query *prompb.Query) ([]*prompb.TimeSeries, error) {
...
-	defer rows.Close()
-
-	labelNames := make([]string, 0)
-
-	for rows.Next() {
-		var labelName string
-		if err := rows.Scan(&labelName); err != nil {
-			return nil, err
+// fromLabelMatchers parses protobuf label matchers to Prometheus label matchers.
+func fromLabelMatchers(matchers []*prompb.LabelMatcher) ([]*labels.Matcher, error) {
```
Harkishen-Singh, line 129: Can we rename this to `ProtoLabelMatcherstoLabelMatchers()`? and vice-versa for the other one? Makes it easier to understand, just like how its named in the upstream.

pkg/pgmodel/sql_reader.go
```diff
@@ -259,8 +158,10 @@ type timescaleRow struct {
 	err      error
 }
 
+// getResultRows fetches the result row datasets from the database using the
+// supplied query parameters.
 func (q *pgxQuerier) getResultRows(startTimestamp int64, endTimestamp int64, hints *storage.SelectHints, path []parser.Node, matchers []*labels.Matcher) ([]timescaleRow, parser.Node, error) {
-
+	// Build a subquery per metric matcher.
 	metric, cases, values, err := buildSubQueries(matchers)
```
Harkishen-Singh, line 165: Should we rename `buildSubQueries` to `buildMetricQuery` as subquery confuses me with the ones in our promql package?

pkg/pgmodel/sql_reader.go
```diff
@@ -439,3 +361,142 @@ func (errorSeriesSet) Next() bool                   { return false }
 func (errorSeriesSet) At() storage.Series           { return nil }
 func (e errorSeriesSet) Err() error                 { return e.err }
 func (e errorSeriesSet) Warnings() storage.Warnings { return nil }
+
+// LabelNames implements the Querier interface. It returns all distinct
+// label names available in the database.
```
Harkishen-Singh, line 366: I do feel mentioning `labels table` is more useful for the contributors that `in the database`. Then, it sounds in lines of prometheus tsdb design, but actually being a table here.

Harkishen-Singh (changes requested):
Initial pass.

pkg/pgmodel/querier.go
antekresic: I think the code shows that itself. I personally don't think that change helps with the readability,

pkg/pgmodel/sql_reader.go
antekresic: I don't think that specific detail is relevant at this level of code documentation since its a implementation detail.

pkg/pgmodel/sql_reader.go
antekresic: The problem is that its not a metric query, its a sub-query clause so calling it a metric query makes no sense.

pkg/pgmodel/sql_reader.go
antekresic: I personally think that is too long and too verbose given how the function signature already shows that info. We can rename it if more people feel otherwise. I don't think we have the other version of the function, its just this one.

Harkishen-Singh (approved):

PR #311, opened 2020-10-20
REFACTOR sql_tests.go to use the new mock for sql_ingest tests

JLockerman:
`sql_tests.go` has two mock database connections: over time the old `mockPGXConn` has grown ad hoc logic for the various cases under test, but no real unifying principle, which made it difficult to determine what the exact behavior of the test should be and how to change tests, when the code under test changed. The newer `sqlRecorder` is much simpler, containing the query string, input, and output for each query we make to the database, and validating we make the queries we expected as we go; this makes it very simple to determine what broke when things change.

We always intended to refactor the remaining tests to use `sqlRecorder`, but it wasn't a high enough priority to finish. Well, now PR #304 is blocked on editing `sql_tests.go` to match the new logic, and it's simpler to do this refactor first.

This PR is organized into three commits:

1. Editing `TestPGXInserterInsertSeries` to use the new mock.
2. Editing `TestPGXInserterInsertData` to use the new mock.
3. Rearranging where in the file the mocks live, now that `sqlRecorder` is the main mock.

The PR is best read one commit at a time; it can be overwhelming to look through all the code at once.

Though this is "just" a refactoring PR, it does block PR #304, which contains some rather important bugfixes, so it should be prioritized accordingly.

### DIFF
--- a/pkg/pgmodel/sql_test.go
+++ b/pkg/pgmodel/sql_test.go
@@ -557,110 +557,168 @@ func TestPGXInserterInsertSeries(t *testing.T) {
 	}
 }
 
-func createRows(x int) map[string][]samplesInfo {
-	return createRowsByMetric(x, 1)
-}
-
-func createRowsByMetric(x int, metricCount int) map[string][]samplesInfo {
-	ret := make(map[string][]samplesInfo)
-	i := 0
-
-	metrics := make([]string, 0, metricCount)
-
-	for metricCount > i {
-		metrics = append(metrics, fmt.Sprintf("metric_%d", i))
-		i++
-	}
-
-	i = 0
-
-	for i < x {
-		metricIndex := i % metricCount
-
-		ret[metrics[metricIndex]] = append(ret[metrics[metricIndex]], samplesInfo{samples: []prompb.Sample{{}}})
-		i++
-	}
-	return ret
-}
-
 func TestPGXInserterInsertData(t *testing.T) {
 	testCases := []struct {
-		name           string
-		rows           map[string][]samplesInfo
-		queryNoRows    bool
-		queryErr       map[int]error
-		copyFromResult int64
-		copyFromErr    error
-		metricsGetErr  error
-		metricsSetErr  error
+		name          string
+		rows          map[string][]samplesInfo
+		sqlQueries    []sqlQuery
+		metricsGetErr error
 	}{
 		{
 			name: "Zero data",
+			sqlQueries: []sqlQuery{
+				{sql: "CALL _prom_catalog.finalize_metric_creation()"},
+			},
 		},
 		{
 			name: "One data",
-			rows: createRows(1),
+			rows: map[string][]samplesInfo{
+				"metric_0": {{samples: make([]prompb.Sample, 1)}},
+			},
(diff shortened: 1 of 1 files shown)


pkg/pgmodel/sql_test.go
```diff
@@ -446,136 +519,191 @@ func TestPGXInserterInsertSeries(t *testing.T) {
...
+				{
+					sql:     "SELECT table_name, possibly_new FROM _prom_catalog.get_or_create_metric_table_name($1)",
+					args:    []interface{}{"metric_0"},
+					results: rowResults{},
+					err:     fmt.Errorf("create table error"),
+				},
+			},
 		},
 		{
-			name:           "Not all data inserted",
```
JLockerman, line 534: I believe this test case is subsumed under the `Copy from error` one now that we've switched to `INSERT ON UPDATE DO NOTHING`: the only source of errors is an `INSERT` error; if we modify the wrong number of rows that just implies duplicates.

pkg/pgmodel/sql_test.go
```diff
@@ -446,136 +519,191 @@ func TestPGXInserterInsertSeries(t *testing.T) {
...
-			rows:           createRows(5),
-			copyFromResult: 4,
```
JLockerman, line 536: I believe this test case is subsumed under the `Copy from` error one now that we've switched to `INSERT ON UPDATE DO NOTHING`: the only source of errors is an `INSERT` error; if we modify the wrong number of rows that just implies duplicates.

pkg/pgmodel/sql_test.go
```diff
@@ -446,136 +519,191 @@ func TestPGXInserterInsertSeries(t *testing.T) {
...
+			name: "Metrics get error",
+			rows: map[string][]samplesInfo{
+				"metric_0": {{samples: make([]prompb.Sample, 1)}},
+			},
 			metricsGetErr: fmt.Errorf("some metrics error"),
-		},
-		{
-			name:          "Metrics set error",
-			rows:          createRows(1),
-			metricsSetErr: fmt.Errorf("some metrics error"),
```
JLockerman, line 551: Failing to set the metric table name is now [explicitly not an error](https://github.com/timescale/promscale/blob/1c852ea723e19b1a4bd63f3c8b00b3fe905c6a09/pkg/pgmodel/sql_ingest.go#L421-L422l) (frankly, I think the metric cache should no longer be returning errors anyway, I just never go around to removing that, partly due to these tests.)

antekresic (approved):
Looks good! 

I do have one thing that sticks out to me. Personally, it looks confusing when we are calling 
`finalize_metric_creation` ([example here](https://github.com/timescale/promscale/pull/311/files#diff-97e96c6f51a58f31570ebf9a16f3dc05a0693c63d696dc90871406d0a44d60feR564))
before we call the metric creation function or even if we are not inserting any data. Is there any reason why its setup like this and not included in the create metric function itself?

JLockerman:
@antekresic `finalize_metric_creation()` contains specifically those things that are too slow to do during the critical path of metric insertion. As to why it appears in the tests; we call `finalize_metric_creation()` whenever we start a new ingestor to finish any metric creation that may be left over from an old connector.

PR #318, opened 2020-10-22
Refresh on abort

JLockerman:
Add a backchannel from the `inserter` to the `insertHandler` to notify the `insertHandler` to notify the `insertHandler` that it should refresh its series cache in the event that we see an epoch abort. Though the code the code is simple it's slightly fraught in that it adds a back edge to our communication graph which _will_ deadlock if we do it wrong.

A possible extension for this is to send the data back along with the refresh signal so we can retry without aborting the entire insert. If we do this, need to be careful that the insert doesn't cycle infinitely.

~Currently includes the commit from PR #316 as tests fail without it.~


pkg/pgmodel/sql_ingest.go
```diff
@@ -579,6 +584,20 @@ func tryRecovery(conn pgxConn, req copyRequest, err error) error {
 		return err
 	}
 
+	// if there was an epoch abort, notify the requester that they should update
+	// their cache
+	if pgErr.Code == "PS001" {
+		// notify the requester that they should update their cache.
+		// this _must never block_: since it is a back edge blocking will cause
+		// deadlocks
+		// TODO should we also return the data so it INSERT can be retried?
```
cevian, line 593: I'd remove this TODO. I don't see it on even our longer-term roadmap since Prometheus itself retries. And this should happen ~never

pkg/pgmodel/sql_ingest.go
```diff
@@ -462,6 +465,8 @@ func (h *insertHandler) blockingHandleReq() bool {
 			return true
 		case <-h.seriesCacheRefresh.C:
 			h.refreshSeriesCache()
+		case <-h.epochAbort:
```
cevian, line 468: I might put this first in the SELECT since we want to do this before processing any more input if it's non-empty

cevian (approved):
Looks. Good. two comments to address before merge

Issue #310, opened 2020-10-20
Password not masked in runner.go:104


chrisgraf:
I downloaded the latest binary and ran:
```
./promscale -db-user=prometheus -db-name=prometheus -db-password=mypassword
```
resulting in: 
```
level=info ts=2020-10-20T06:34:41.096Z caller=runner.go:103 msg="Version:0.1.0; Commit Hash: "
level=info ts=2020-10-20T06:34:41.096Z caller=runner.go:104 config="&{ListenAddr::9201 TelemetryPath:/metrics PgmodelCfg:{Host:localhost Port:5432 User:prometheus Password:mypassword Database:prometheus SslMode:require DbConnectRetries:0 AsyncAcks:false ReportInterval:0 LabelsCacheSize:10000 MetricsCacheSize:10000 SeriesCacheSize:0 WriteConnectionsPerProc:4 MaxConnections:-1} LogCfg:{Level:debug Format:logfmt} HaGroupLockID:0 RestElection:false PrometheusTimeout:-1ns ElectionInterval:5s Migrate:true StopAfterMigrate:false UseVersionLease:true CorsOrigin:^(?:.*)$ InstallTimescaleDB:true}"
level=warn ts=2020-10-20T06:34:41.254Z caller=migrate.go:128 msg="could not install promscale: the extension is not available. continuing without extension"
level=warn ts=2020-10-20T06:34:41.256Z caller=runner.go:241 msg="No adapter leader election. Group lock id is not set. Possible duplicate write load if running multiple connectors"
level=warn ts=2020-10-20T06:34:41.264Z caller=client.go:186 msg="had to reduce the number of copiers due to connection limits: wanted 64, reduced to 25"
level=info ts=2020-10-20T06:34:41.276Z caller=client.go:90 msg="host=localhost port=5432 user=prometheus dbname=prometheus password='****' sslmode=require connect_timeout=10" numCopiers=25 pool_max_conns=50 pool_min_conns=16
level=debug ts=2020-10-20T06:34:41.280Z caller=query_engine.go:13 msg="Lookback delta is zero, setting to default value" value=5m0s
level=info ts=2020-10-20T06:34:41.280Z caller=runner.go:123 msg="Starting up..."
level=info ts=2020-10-20T06:34:41.280Z caller=runner.go:124 msg=Listening addr=:9201
```
So in client.go:90 it is correctly masked but not so in runner.go:104 (I replaced my true password by 'mypassword')

[...]

PR #319, opened 2020-10-23
Mask password while printing config

Harkishen-Singh:

Fixes: #310


antekresic (changes requested):
Can we combine 2nd and 3rd regex since the only difference is capital letter? 

Also, this needs a test case to make sure its covered.

antekresic (approved):

PR #323, opened 2020-10-27
Add read-only mode to the connector

antekresic:
Read-only mode is added to run against a read replica. In this mode, migration
and leader election is disabled. Write route will be missing and return 404
error code.

### DIFF
--- a/cmd/promscale/main.go
+++ b/cmd/promscale/main.go
@@ -14,16 +14,16 @@ import (
 
 func main() {
 	cfg := &runner.Config{}
-	cfg, err := runner.ParseFlags(cfg)
+	cfg, err := runner.ParseFlags(cfg, os.Args[1:])
 	if err != nil {
 		fmt.Println("Version: ", version.Version, "Commit Hash: ", version.CommitHash)
-		fmt.Println("Fatal error: cannot parse flags ", err)
+		fmt.Println("Fatal error: cannot parse flags: ", err)
 		os.Exit(1)
 	}
 	err = log.Init(cfg.LogCfg)
 	if err != nil {
 		fmt.Println("Version: ", version.Version, "Commit Hash: ", version.CommitHash)
-		fmt.Println("Fatal error: cannot start logger", err)
+		fmt.Println("Fatal error: cannot start logger: ", err)
 		os.Exit(1)
 	}
 	err = runner.Run(cfg)
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -14,7 +14,14 @@ import (
 
 func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client, elector *util.Elector) http.Handler {
 	router := route.New()
+
 	writeHandler := timeHandler(metrics.HTTPRequestDuration, "write", Write(client, elector, metrics))
+
+	// If we are running in read-only mode, log and send NotFound status.
+	if apiConf.ReadOnly {
+		writeHandler = withWarnLog("trying to send metrics to write API while connector is in read-only mode", http.NotFoundHandler())
+	}
+
 	router.Post("/write", writeHandler)
 
 	readHandler := timeHandler(metrics.HTTPRequestDuration, "read", Read(client, metrics))
@@ -45,6 +52,13 @@ func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client, 
 	router.Get("/healthz", Health(client))
 
 	return router
+}
+
+func withWarnLog(msg string, handler http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		log.Warn("msg", msg)
+		handler.ServeHTTP(w, r)
+	}
 }
 
 // timeHandler uses Prometheus histogram to track request time
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -42,6 +42,7 @@ type Config struct {
 	UseVersionLease    bool
 	CorsOrigin         *regexp.Regexp
 	InstallTimescaleDB bool
+	ReadOnly           bool
 }
 
 const (
@@ -56,7 +57,7 @@ var (
 	startupError       = fmt.Errorf("startup error")
 )
 
-func ParseFlags(cfg *Config) (*Config, error) {
+func ParseFlags(cfg *Config, args []string) (*Config, error) {
 	pgclient.ParseFlags(&cfg.PgmodelCfg)
 	log.ParseFlags(&cfg.LogCfg)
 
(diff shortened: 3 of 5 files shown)


pkg/pgmodel/sql_ingest.go
```diff
@@ -30,6 +31,10 @@ const (
 	getEpochSQL              = "SELECT current_epoch FROM " + catalogSchema + ".ids_epoch LIMIT 1"
 )
 
+var (
+	noOpIngestError = errors.New("trying to ingest new data in read-only mode")
```
Harkishen-Singh, line 35: I think `fmt.Errorf()` might be preferred as that's used mostly across the codebase.

pkg/runner/runner.go
```diff
@@ -94,13 +97,28 @@ func ParseFlags(cfg *Config) (*Config, error) {
...
+
+	if cfg.ReadOnly {
+		if cfg.HaGroupLockID != 0 {
+			return nil, fmt.Errorf("Invalid option for HA group lock ID, cannot enable HA mode and read-only mode")
+		}
+		cfg.Migrate = false
+		cfg.StopAfterMigrate = false
+		cfg.UseVersionLease = false
+		cfg.InstallTimescaleDB = false
+		cfg.PgmodelCfg.ReadOnly = true
```
Harkishen-Singh, line 109: If the user provides input for these, we should I guess error then and say that these are not supported, else users will be confused of the flag as they provided but we disabled it. A generic error like `Flag not supported in read-only`

pkg/runner/runner.go
```diff
@@ -94,13 +97,28 @@ func ParseFlags(cfg *Config) (*Config, error) {
...
+	}
 	return cfg, nil
 }
 
 func Run(cfg *Config) error {
 	log.Info("msg", "Version:"+version.Version+"; Commit Hash: "+version.CommitHash)
 	log.Info("config", util.MaskPassword(fmt.Sprintf("%+v", cfg)))
 
+	if cfg.ReadOnly {
+		log.Info("msg", "Migration disabled for read-only mode")
```
Harkishen-Singh, line 119: `Migration` => `Migrations` ?

pkg/runner/runner.go
```diff
@@ -73,8 +74,10 @@ func ParseFlags(cfg *Config) (*Config, error) {
 	flag.StringVar(&migrateOption, "migrate", "true", "Update the Prometheus SQL to the latest version. Valid options are: [true, false, only]")
 	flag.BoolVar(&cfg.UseVersionLease, "use-schema-version-lease", true, "Prevent race conditions during migration")
 	flag.BoolVar(&cfg.InstallTimescaleDB, "install-timescaledb", true, "Install or update the TimescaleDB extension")
+	flag.BoolVar(&cfg.ReadOnly, "read-only", false, "Read-only mode. Don't write to database. Useful when pointing adapter to read replica")
 	envy.Parse("TS_PROM")
-	flag.Parse()
+	// Ignore errors; CommandLine is set for ExitOnError.
+	_ = flag.CommandLine.Parse(args)
```
Harkishen-Singh, line 80: Not sure, but I think error is returned here if there is any parsing issue or a `--help` is passed. In both the cases (especially for later), we should stop or just return here.

pkg/runner/runner.go
```diff
@@ -235,7 +253,7 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
 		return nil, fmt.Errorf("elector init error: %w", err)
 	}
 
-	if elector == nil {
+	if elector == nil && !cfg.ReadOnly {
```
Harkishen-Singh, line 256: not related to this PR, but do we have plans for **multiple readers**?

Harkishen-Singh (changes requested):

cmd/promscale/main.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func main() {
 	cfg := &runner.Config{}
-	cfg, err := runner.ParseFlags(cfg)
+	cfg, err := runner.ParseFlags(cfg, os.Args[1:])
```
Harkishen-Singh, line 17: I couldn't get the need for this, so if you can tell that.

JLockerman (commented):
Over all 👍  I think this is a reasonable way to do this, though it is starting to feel like the ingestor and reader should be moved out of client to the top level runner.

We'll need some end-to-end tests against a read-only DB, but maybe that should wait until we have timescaledb 2.0 migrations; we'll definitely need tests for that 🤔

pkg/runner/runner.go
antekresic: `ExitOnError` means the program will call `os.Exit` on an error when parsing flags. This is the default behavior, this line change does the same thing as `flag.Parse`, only takes in arguments manually instead of using `os.Args`.

https://golang.org/pkg/flag/#ExitOnError

pkg/runner/runner.go
antekresic: I don't think we thought about using multiple readers at the same time, the use-case did not occur yet.

cmd/promscale/main.go
antekresic: The need is to have arguments passed into the function so we can use them for testing purposes. Without this, the function is very hard to test and its getting more complicated so it needs testing IMO.

pkg/runner/runner.go
antekresic: Added checks.

Harkishen-Singh (approved):
lgtm!

pkg/pgmodel/sql_ingest.go
```diff
@@ -312,6 +316,22 @@ func (p *pgxInserter) getMetricInserter(metric string) chan insertDataRequest {
 	return inserter.(chan insertDataRequest)
 }
 
+type noOpInserter struct{}
+
+func (*noOpInserter) InsertNewData(rows map[string][]samplesInfo) (uint64, error) {
+	return 0, noOpIngestError
```
cevian, line 322: Hmm. If I understand correctly this error will return an HTTP 500 response? The problem is Prometheus will classify this as a Recoverable error(https://github.com/prometheus/prometheus/blob/master/storage/remote/client.go#L200)  and will retry (https://github.com/prometheus/prometheus/blob/master/storage/remote/queue_manager.go#L921).

But I don't think we want retrying here? 

I think a status 400 0r 404 is more appropriate here.

Honestly this may be better done at the mux level when setting up routing rather than thru a no-op inserter. But I'll leave that up to you.

cevian (changes requested):
There is a blocking issue with the HTTP status code returned.

pkg/pgmodel/sql_ingest.go
Harkishen-Singh: AFAIK, there is plan to make 4xx errors as recoverable errors and just the 5xx as unrecoverable. at least, that's what the discussion says on the link I mentioned below (I knew that as I am working on that issue)

related issue: https://github.com/prometheus/prometheus/issues/7078

antekresic: @Harkishen-Singh thats for rate limiting errors it seems (429 code). Will change this to do 404 Not Found on write in read-mode. That one shouldn't be recoverable IMO.

cevian: yeah making 404 recoverable would just be wrong

pkg/api/router.go
```diff
@@ -14,8 +14,12 @@ import (
 
 func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client, elector *util.Elector) http.Handler {
 	router := route.New()
-	writeHandler := timeHandler(metrics.HTTPRequestDuration, "write", Write(client, elector, metrics))
-	router.Post("/write", writeHandler)
+
+	// If we are running in read-only mode, don't handle write route.
+	if !apiConf.ReadOnly {
+		writeHandler := timeHandler(metrics.HTTPRequestDuration, "write", Write(client, elector, metrics))
```
cevian, line 20: Optional: can we replace with a handler that logs the 404 and then returns 404. Would making the logging more useful in that case. But I like this A LOT more.

cevian (approved):
Approved with one optional suggestion. This makes a ton more sense

cevian (approved):

PR #324, opened 2020-10-27
Timescale 2.0

JLockerman:
This commit updates Promscale's SQL to support TimescaleDB 2.0 but does not yet add support for distributed hypertables, those will be added in a future commit.


pkg/pgmodel/migrations/sql/preinstall/001-users.sql
```diff
@@ -6,6 +6,25 @@ DO $$
...
+            $inner$
+                BEGIN
+                    CREATE ROLE prom_reader;
+                EXCEPTION WHEN duplicate_object THEN
+                    RETURN;
+                END
+            $inner$;
+        $dist$);
+    EXCEPTION WHEN SQLSTATE '0A000' THEN
+        -- we're not the acess node, just return
```
antekresic, line 23:
```suggestion
        -- we're not the access node, just return
```

antekresic (approved):
:+1: minor typo nit, other than that looks LGTM!

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1003,11 +1013,22 @@ BEGIN
                 timescaledb.compress_segmentby = 'series_id',
                 timescaledb.compress_orderby = 'time, value'
             ); $$, metric_table_name);
+
         --chunks where the end time is before now()-1 hour will be compressed
-        PERFORM add_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+        IF new_compress_policy THEN
```
cevian, line 1018: Nit: I think this is more clear. But I'll leave up to you.
```suggestion
        IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1486,35 +1516,67 @@ BEGIN
...
-                    (h.schema_name = 'SCHEMA_DATA' AND h.table_name = m.table_name);
+        IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
+            RETURN QUERY
+            SELECT
+                m.id,
+                m.metric_name,
+                m.table_name,
+                SCHEMA_CATALOG.get_metric_retention_period(m.metric_name) as retention_period,
+                (
+                    SELECT _timescaledb_internal.to_interval(interval_length)
```
cevian, line 1527: can we use the new info view for this https://docs.timescale.com/v2.0/api#timescaledb_information-dimensions

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1003,11 +1013,22 @@ BEGIN
...
         --chunks where the end time is before now()-1 hour will be compressed
-        PERFORM add_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+        IF new_compress_policy THEN
+            PERFORM add_compression_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+        ELSE
+            PERFORM add_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+        END IF;
     ELSE
-        PERFORM remove_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name));
+        IF new_compress_policy THEN
```
cevian, line 1024:
```suggestion
       IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
```

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -35,16 +35,18 @@ import (
...
+	useTimescale2 = flag.Bool("use-timescale2", false, "use TimescaleDB 2.0")
+	printLogs     = flag.Bool("print-logs", false, "print TimescaleDB logs")
 )
 
 var cleanImage = "timescale/timescaledb:latest-pg12"
 var prevDBImage = "timescale/timescaledb:latest-pg12"
 
 func TestMain(m *testing.M) {
 	var code int
+	// TODO Timescale 2.0 and regular postgres
```
cevian, line 49: what's this todo for?

cevian (approved):
Looks good. I do prefer the version switch in the SQL. I think it's much more clear what's going on.

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: I thought it was too much last week when I tried it, but I like it better now 🤔

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: 🤩

pkg/pgmodel/upgrade_tests/upgrade_test.go
JLockerman: @cevian currently we only test upgrading a DB with timescale-1.X to another DB with timescale-1.X we don't test that upgrades work on vanilla postgres or timescale-2. We need to decide if and how we want to do those

PR #325, opened 2020-10-27
Batch inserts into txns to reduce the txn rate

cevian:


pkg/pgmodel/sql_ingest.go
```diff
@@ -552,7 +555,18 @@ func runInserter(conn pgxConn, in chan copyRequest) {
...
+			//for loop since may need to merge multiple table
+			for i < len(insertBatch)-2 && insertBatch[i].table == insertBatch[i+1].table {
+				insertBatch[i].data.absorb(insertBatch[i+1].data)
+				insertBatch[i+1].data.release()
+				//delete element i+1
+				copy(insertBatch[i+1:], insertBatch[i+2:])
+				insertBatch[len(insertBatch)-1] = copyRequest{}
+				insertBatch = insertBatch[:len(insertBatch)-1]
+			}
+		}
```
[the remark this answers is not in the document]
JLockerman: based on the [equivalent rust stdlib code](https://doc.rust-lang.org/stable/src/alloc/vec.rs.html#3034-3053) I think a more efficient way to do this is something like
```go
// invariant: there's a gap-free prefix containing data we want to insert
//             in table order, Following this is a gap which once contained
//             batches we compacted away, following this is the
//             uncompacted data.
dst := 0
for src := 1; src < len(insertBatch); src++ {
    if insertBatch[dst].table == insertBatch[src].table {
        insertBatch[dst].data.absorb(insertBatch[src].data)
        insertBatch[src].data.release()
    } else {
        dst += 1
        if dst != src {
            insertBatch[dst] = insertBatch[src]
        }
    }
}
for i := dst; i < len(insertBatch); i++ {
    insertBatch[i] = copyRequest{}
}
insertBatch = insertBatch[:dst]
```

JLockerman (commented):
I need to do a fuller review, but some prelim thoughts.
How's the performance improvement from this?

pkg/pgmodel/sql_ingest.go
```diff
@@ -562,19 +576,31 @@ func runInserter(conn pgxConn, in chan copyRequest) {
...
+var getBatchMutex = &sync.Mutex{}
+
 func inserterGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, bool) {
+	//This mutex is not for safety, but rather for better batching.
+	//It guarantees that only one copier is reading from the channel at one time
+	//This ensures bigger batches as well as less spread of a
+	//single http request to multiple copiers, decreasing latency via less sync
+	//overhead especially on low-pressure systems.
+	getBatchMutex.Lock()
+	defer getBatchMutex.Unlock()
```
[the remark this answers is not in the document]
JLockerman: Just thinking out loud:
The traditional way to solve this are either a `recvBatch()` function (not provided by go) or to send multiple things per message (not really applicable here). This is going to increase contention a little but **a.** probably not much more than the channel already does, and **b.** this code is IO (probably latecy) bound and not CPU bound, so it's unlikely to matter.

pkg/pgmodel/sql_ingest.go
```diff
@@ -539,38 +542,115 @@ func runInserter(conn pgxConn, in chan copyRequest) {
...
+
+func inserterGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, bool) {
+	//This mutex is not for safety, but rather for better batching.
+	//It guarantees that only one copier is reading from the channel at one time
+	//This ensures bigger batches as well as less spread of a
+	//single http request to multiple copiers, decreasing latency via less sync
+	//overhead especially on low-pressure systems.
+	getBatchMutex.Lock()
+	defer getBatchMutex.Unlock()
+	req, ok := <-in
```
JLockerman, line 589: still thinking out loud:  I don't love that we're sleeping while holding the lock though we'd need it for the "hot" gather anyway, and I don't see a better way to get the same performance.

To summarize my understanding:
The lock changes the coordination scheme to something like a token model (sorry I don't know the actual name). One thread has the token, and is allowed to read from the channel. Once that thread fills up (either receives enough messages, or spends enough time) it releases the token, allowing another thread to start receiving, and forwards its messages to the DB. One way of thinking about this is that by default we send batches at 50Hz, but if the batches get too large we start sending faster to pin the batches at `maxCopyRequestsPerTxn`.

JLockerman: The "bad" outcome for this is that we'll round-robin across threads in the low-contention setting: after a thread sends a batch it'll release the lock, and, since go mutexen are approximately fair, the woken thread will be the last thread to get a new batch.

One way to fix this would be to add yet another layer of indirection: have this batching be done by a distributor thread that distributes to one or more inserter routines, starting with exactly one routine, and having the distributor spin up more as needed. I not sure this is worth the complexity until we actually see a performance issue.

pkg/pgmodel/sql_ingest.go
```diff
@@ -539,38 +542,115 @@ func runInserter(conn pgxConn, in chan copyRequest) {
...
 // certain errors are recoverable, handle those we can
 //   1. if the table is compressed, decompress and retry the insertion
-func insertErrorFallback(conn pgxConn, req copyRequest, err error) error {
-	err = tryRecovery(conn, req, err)
-	if err != nil {
-		log.Warn("msg", fmt.Sprintf("time out while processing error for %s", req.table), "err", err.Error())
-		return err
+func insertErrorFallback(conn pgxConn, err error, reqs ...copyRequest) error {
+	for _, req := range reqs {
+		err = tryRecovery(conn, err, req)
```
JLockerman, line 640: Just verifying: do we test that calling `decompressChunks()` succeeds when the chunks aren't compressed?

JLockerman: this comment may still be relevant

JLockerman (approved):

PR #326, opened 2020-10-28
Adds support for complete deletion of series based on matchers

Harkishen-Singh:
This PR adds support for deletion of series based on matches. This happens mostly through a public API as

`_prom_catalog.delete_series_from_metric(<series_id>, <metric_name_where_for_series_id>);`


Harkishen-Singh:
Buggy GitHub in terms of reviewers addition!

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1455,6 +1455,32 @@ $func$
...
+    num_chunks_touched bigint := 0;
+BEGIN
+    FOR delete_stmt in
+        SELECT FORMAT('DELETE FROM %1$I.%2$I WHERE series_id = %3$s', schema_name, table_name, series_id)
+        FROM (
+            SELECT (COALESCE(chc, ch)).* FROM pg_class c
+                INNER JOIN pg_namespace n ON c.relnamespace = n.oid
+                INNER JOIN _timescaledb_catalog.chunk ch ON (ch.schema_name, ch.table_name) = (n.nspname, c.relname)
+                LEFT JOIN _timescaledb_catalog.chunk chc ON ch.compressed_chunk_id = chc.id
+            WHERE c.oid IN (SELECT show_chunks(metric_name)::oid)
```
davidkohn88, line 1474: @cevian Do we need to do anything different here with metric_name to make it safe? I don't really think so I think if we get one that isn't right we will just throw an error.

Only thing I don't know is whether we need to get a metric name that is the right length from another function or something or if that happens in Promscale somewhere before this happens?

davidkohn88 (commented):
SQL looks good to me, a couple questions I just added around whether we need to mod the name to our shortened version in some cases or if that's handled in the Promscale stuff and I missed it.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1455,6 +1455,32 @@ $func$
...
+        ) a
+    LOOP
+        EXECUTE delete_stmt;
+        num_chunks_touched = num_chunks_touched + 1;
```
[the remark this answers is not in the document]
Harkishen-Singh: that is right! So, if I am right, we just need to check if a row with that series id is present in the chunk which is just about to be touched by the delete query?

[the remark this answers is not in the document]
davidkohn88: Definitely avoid an extra scan, use: https://www.postgresql.org/docs/12/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS which should tell you the number of rows affected.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1483,6 +1483,33 @@ $func$
...
+    END LOOP;
+    DELETE FROM SCHEMA_CATALOG.series WHERE id=series_id;
+    RETURN num_chunks_touched;
```
cevian, line 1509: right now this is a count of all the chunks in the table, whether or not they have data that was deleted. Is this right? Why do we need this at all?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1483,6 +1483,33 @@ $func$
 LANGUAGE PLPGSQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) TO prom_writer;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_series_from_metric(series_id bigint, metric_name text)
```
cevian, line 1486: TODO for next round of changes -- make this function vanilla PG compatible

pkg/pgmodel/end_to_end_tests/delete_test.go
```diff
@@ -0,0 +1,292 @@
...
+				testutil.Equals(t, 1025, len(deletedSeriesIDs), "delete all series does not match in", m.name)
+				testutil.Equals(t, 62, len(touchedMetrics), "delete all metrics does not match in", m.name)
+			}
+		})
+	}
+}
+
+func TestDeleteMixins(t *testing.T) {
+	if !*useTimescaleDB {
+		t.Skip("skipping delete tests: delete api is based on chunks that are absent without timescaledb.")
```
cevian, line 171: We should make this vanilla PG compatible. I'll discuss how with you.

pkg/pgmodel/end_to_end_tests/delete_test.go
```diff
@@ -0,0 +1,292 @@
+package end_to_end_tests
```
cevian, line 1: need to add a test that specifies a time range for the deletion and fails

cevian (changes requested):
Good start but need some changed. Also this should be tested after 2.0 compatiibility PR is in to make sure nothing breaks with 2.0

pkg/pgmodel/migrations/sql/idempotent/base.sql
davidkohn88: There's a function for that right?

pkg/api/delete.go
```diff
@@ -0,0 +1,87 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		var (
+			metricsTouched     []string
+			totalChunksTouched int
+			seriesDeleted      []pgmodel.SeriesID
+		)
+		if err := r.ParseForm(); err != nil {
+			respondError(w, http.StatusBadRequest, err, "bad_data")
+			return
+		}
```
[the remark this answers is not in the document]
Harkishen-Singh: Maybe more of a warning, right?

pkg/api/delete.go
```diff
@@ -0,0 +1,87 @@
...
+				totalChunksTouched,
+			),
+		)
+	}
+}
+
+func distinctValues(slice interface{}) []string {
+	var (
+		temp = make(map[string]struct{})
+		keys []string
```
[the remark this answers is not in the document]
Harkishen-Singh: Ah, I always feel like it makes the code clean. Anyways, will do.

pkg/api/delete.go
```diff
@@ -0,0 +1,87 @@
...
+			matchers, err := parser.ParseMetricSelector(s)
+			if err != nil {
+				respondError(w, http.StatusBadRequest, err, "bad_data")
+				return
+			}
+			pgDeleter := client.Deleter().PgDeleter()
+			touchedMetrics, deletedSeriesIDs, chunksTouched, err := pgDeleter.DeleteSeries(matchers)
+			if err != nil {
+				respondErrorWithMessage(w, http.StatusInternalServerError, err, "deleting_series",
+					fmt.Sprintf("partial delete: deleted %s series IDs from %s metrics, touching %d chunks in total",
```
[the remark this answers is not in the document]
Harkishen-Singh: I wrote it from my mind, so definitely no. Do you want that to be same?

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
Harkishen-Singh: We cannot say that the seriesIDs coming in belong to the same metric name. That means, the sql function now no longer supports accepting metric name, rather fetch it dynamically.

pkg/pgmodel/delete.go
```diff
@@ -0,0 +1,95 @@
...
+		err                error
+		metricsTouched     = make(map[string]struct{})
+	)
+	metricNames, seriesIDMatrix, err := pgDel.getMetricNameSeriesIDFromMatchers(matchers)
+	if err != nil {
+		return nil, nil, -1, fmt.Errorf("delete-series: %w", err)
+	}
+	for metricIndex, metricName := range metricNames {
+		seriesIDs := seriesIDMatrix[metricIndex]
+		for _, seriesID := range seriesIDs {
```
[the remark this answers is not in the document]
Harkishen-Singh: As long as the second parameter is metric name, the caller need to ensure that the seriesIDs belong to that metric only, else it would be clumsy. Maybe, getting rid of it solves the matter here?

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: Yes. The reason it's there is that users can know the efficiency of the query or like why does that endpoint took so long to respond (if it does).

pkg/api/delete.go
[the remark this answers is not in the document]
Harkishen-Singh: As far as I see, prometheus does not respond in message. Is it same for you?

pkg/pgmodel/delete.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+const (
+	queryDeleteSeries = "select _prom_catalog.delete_series_from_metric($1, $2)"
+)
+
+type DBDeleter struct {
```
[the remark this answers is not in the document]
Harkishen-Singh: > just have a constructor for PgDelete

Should we have a constructor that does nothing but return an object? or should I just go with directly using the struct?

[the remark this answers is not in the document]
Harkishen-Singh: > just have a constructor for PgDelete

Should we prefer directly using the defined structs in the package over a constructor that does nothing but an object return?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1483,6 +1483,33 @@ $func$
...
+RETURNS BIGINT
+LANGUAGE PLPGSQL
+AS
+$$
+DECLARE
+    delete_stmt text;
+    num_chunks_touched bigint := 0;
+BEGIN
+    FOR delete_stmt in
+        SELECT FORMAT('DELETE FROM %1$I.%2$I WHERE series_id = %3$s', schema_name, table_name, series_id)
```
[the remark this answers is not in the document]
Harkishen-Singh: this causes an error.

```
ERROR:  syntax error at or near "$1"
LINE 1: ..._internal.compress_hyper_257_300_chunk WHERE series_id IN $1
                                                                     ^
QUERY:  DELETE FROM _timescaledb_internal.compress_hyper_257_300_chunk WHERE series_id IN $1
CONTEXT:  PL/pgSQL function delete_series_from_metric(bigint[],text) line 18 at EXECUTE
```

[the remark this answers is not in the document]
Harkishen-Singh: figured it out using `any`

cevian (changes requested):

pkg/pgmodel/delete.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/timescale/promscale/pkg/log"
+)
+
+const queryDeleteSeries = "SELECT _prom_catalog.delete_series_from_metric($1, $2)"
+
+var (
+	ErrTimeBasedDeletion = fmt.Errorf("time based series deletion is unsupported")
+	MinTimeProm          = time.Unix(math.MinInt64/1000+62135596801, 0).UTC()
```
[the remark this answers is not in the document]
Harkishen-Singh: They are the general minTime and maxTime from or code, but the existing ones are not uniquely used. Hence, had to create them again based on the requirement. I will take these to `util` and then import into places that require it.

pkg/api/delete.go
```diff
@@ -0,0 +1,97 @@
...
+func distinctValues(slice interface{}) []string {
+	temp := make(map[string]struct{})
+	switch elem := slice.(type) {
+	case []string:
+		for _, element := range elem {
+			if _, ok := temp[element]; !ok {
+				temp[element] = struct{}{}
+			}
+		}
+	case []pgmodel.SeriesID:
```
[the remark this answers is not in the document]
Harkishen-Singh: > test for a Stringer interface here

Sorry, but I didn't get this.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1483,6 +1483,35 @@ $func$
 LANGUAGE PLPGSQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) TO prom_writer;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_series_from_metric(series_ids bigint[], name text)
```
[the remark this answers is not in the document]
Harkishen-Singh: Done. Please see if it's satisfactory.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1516,6 +1483,48 @@ $func$
...
+                SELECT (COALESCE(chc, ch)).* FROM pg_class c
+                    INNER JOIN pg_namespace n ON c.relnamespace = n.oid
+                    INNER JOIN _timescaledb_catalog.chunk ch ON (ch.schema_name, ch.table_name) = (n.nspname, c.relname)
+                    LEFT JOIN _timescaledb_catalog.chunk chc ON ch.compressed_chunk_id = chc.id
+                WHERE c.oid IN (SELECT show_chunks(format('%I.%I','SCHEMA_DATA', metric_table))::oid)
+                ) a
+        LOOP
+            EXECUTE delete_stmt USING series_ids;
+            GET DIAGNOSTICS rows_affected = ROW_COUNT;
+            num_chunks_touched = num_chunks_touched + rows_affected;
```
davidkohn88, line 1512: This is no longer num_chunks_touched, but rather a row count...should we call it that to avoid being confusing with naming?

Harkishen-Singh: sure, but won't the row_count denote the number of rows that were updated (or deleted)?

davidkohn88: (though it's only counting compressed rows once, so it's a bit misleading if we call it that too)

Harkishen-Singh: yes, it should be changed to `num_rows_affected`. I will update once this PR is reviewed so that it comes along with other changes if required.

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1516,6 +1483,48 @@ $func$
...
+    ELSE
+        FOR delete_stmt IN
+            SELECT FORMAT('DELETE FROM SCHEMA_DATA.%1$I WHERE series_id = ANY($1)', metric_table)
+        LOOP
+            EXECUTE delete_stmt USING series_ids;
+            GET DIAGNOSTICS rows_affected = ROW_COUNT;
+            num_chunks_touched = num_chunks_touched + rows_affected;
+        END LOOP;
+    END IF;
+    DELETE FROM SCHEMA_CATALOG.series WHERE id = ANY($1);
```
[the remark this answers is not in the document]
cevian: Maybe we should create a function for this and then we could call it both here and from SCHEMA_CATALOG.mark_unused_series but I'm not sure it's worth it.

```
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_series(
   series_ids bigint[]
) RETURNS VOID AS 
$func$
 UPDATE SCHEMA_DATA_SERIES.%1$I SET delete_epoch = current_epoch+1
  FROM SCHEMA_CATALOG.ids_epoch
        WHERE delete_epoch IS NULL
            AND id = ANY(series_ids);
$func$
LANGUAGE SQL VOLATILE;

File: pkg/pgmodel/migrations/sql/idempotent/base.sql (hunk 16 above)
Comment (cevian) [reply pkg/pgmodel/migrations/sql/idempotent/base.sql:1512]: yes I like `num_rows_affected` or `num_rows_deleted`

File: pkg/pgmodel/migrations/sql/idempotent/base.sql [hunk 18]
```diff
@@ -1516,6 +1483,48 @@ $func$
...
+            num_chunks_touched = num_chunks_touched + rows_affected;
+        END LOOP;
+    ELSE
+        FOR delete_stmt IN
+            SELECT FORMAT('DELETE FROM SCHEMA_DATA.%1$I WHERE series_id = ANY($1)', metric_table)
+        LOOP
+            EXECUTE delete_stmt USING series_ids;
+            GET DIAGNOSTICS rows_affected = ROW_COUNT;
+            num_chunks_touched = num_chunks_touched + rows_affected;
+        END LOOP;
```
cevian, line 1521:
```suggestion

            EXECUTE FORMAT('DELETE FROM SCHEMA_DATA.%1$I WHERE series_id = ANY($1)', metric_table) USING series_ids;
            GET DIAGNOSTICS rows_affected = ROW_COUNT;
            num_chunks_touched = num_chunks_touched + rows_affected;

```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1645,30 +1630,17 @@ CREATE OR REPLACE VIEW SCHEMA_INFO.metric_stats AS
 CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delay_compression_job(ht_table name, new_start timestamptz) RETURNS VOID
 AS $$
 DECLARE
-    bgw_job_id int;
+  bgw_job_id int;
```
cevian, line 1633: These changes look wrong. Bad rebase?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1545,59 +1554,35 @@ BEGIN
...
-                chs.number_compressed_chunks as compressed_chunks
-            FROM SCHEMA_CATALOG.metric m
-            LEFT JOIN timescaledb_information.hypertable hi ON
-                        (hi.table_schema = 'SCHEMA_DATA' AND hi.table_name = m.table_name)
-            LEFT JOIN timescaledb_information.compressed_hypertable_stats chs ON
-                        (chs.hypertable_name = format('%I.%I', 'SCHEMA_DATA', m.table_name)::regclass)
-            LEFT JOIN _timescaledb_catalog.hypertable h ON
-                        (h.schema_name = 'SCHEMA_DATA' AND h.table_name = m.table_name);
-        END IF;
+        RETURN QUERY
```
cevian, line 1557: These changes look wrong. Bad rebase?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1280,19 +1256,10 @@ BEGIN
...
-                );
-            ELSE
-                PERFORM drop_chunks(
-                    table_name=>metric_table,
-                    schema_name=> 'SCHEMA_DATA',
-                    older_than=>older_than,
-                    cascade_to_materializations=>FALSE
-                );
-            END IF;
+            PERFORM drop_chunks(table_name=>metric_table,
```
cevian, line 1259: This is wrong. Bad rebase?

cevian (changes requested):
Some suggestions to fix the series delete. There is a bad rebase here I think.

pkg/pgmodel/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
Harkishen-Singh: It does simplify the work, so yes we should go for a function for this IMO.

pkg/pgmodel/migrations/sql/idempotent/base.sql
davidkohn88: Do we care about the compressed rows thing? We could do a returning with our stats column that has the number of rows and use that to get the actual number of rows deleted?

pkg/pgmodel/end_to_end_tests/delete_test.go
```diff
@@ -0,0 +1,481 @@
...
```
cevian, line 1: we should add a test that sends a request to the delete endpoint. Currently the endpoint itself isn't tested

cevian (changes requested):

pkg/pgmodel/migrations/sql/idempotent/base.sql
davidkohn88: @cevian thoughts?

pkg/pgmodel/end_to_end_tests/delete_test.go
Harkishen-Singh: Done.

pkg/pgmodel/delete.go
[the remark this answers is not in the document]
cevian: Please add a comment either in this or the refactor pr with the source of where it is in the prom code

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -462,6 +462,21 @@ BEGIN
 END;
 $$;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_series(
```
cevian, line 465: Let's change the name to reflect that this deletes on the row for the series not all the info from the data tables as well.
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.delete_series_catalog_row(
```

cevian (approved):
LGTM except for 2 minor comments. Great work

pkg/pgmodel/migrations/sql/idempotent/base.sql
Harkishen-Singh: Definitely!

pkg/pgmodel/delete.go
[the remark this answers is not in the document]
Harkishen-Singh: Definitely! I have added this as TODO in the description of the refactor PR.

cevian (approved):

PR #331, opened 2020-11-04
Fix update old series view issue with no metrics ingested

### LINKED ISSUE #330: Unit-tests for fresh install/database
As in community channel, the 0.1.1-dev version has a `foreach null issue` on a fresh db. So, we need to have a dedicated test suite that works the installation of promscale schemas and normal functionality on a fresh database. This can be related to upgrades as well.

antekresic:
Fixes #330

This is a WIP mostly because the testing setup is too slow to run twice (once with data, once with no data ingested) and needs some work to get it running faster.


JLockerman (approved):
The SQL looks good.

CC @cevian on the fact that we're changing an upgrade script. Since it's a non-functional change I think it should be allowed

pkg/pgmodel/upgrade_tests/upgrade_test.go
```diff
@@ -421,7 +444,7 @@ func toPreviousVersion(version *semver.Version) {
 	version.Pre = nil
 
 	if version.Patch > 0 {
-		version.Patch -= 1
+		version.Patch = 0
```
cevian, line 447: This is not correct in general. But I think we are gonna change this logic anyway and this is needed for the PR so I think its ok

cevian (approved):
Looks good. I think we can optimize the testing later.

cevian:
Note PR #336 is relevant here.

PR #332, opened 2020-11-04
Split tests into seperate jobs to run in parallel

JLockerman:
Speeding up CI

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -37,11 +37,6 @@ jobs:
       run: |
         diff -u <(echo -n) <(gofmt -d .)
 
-    - name: Lint
-      run: |
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0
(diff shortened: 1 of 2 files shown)


JLockerman:
Looking at [our actions](https://github.com/timescale/promscale/actions) it seems like this change reduces the test time from around 10m to around 5m

JLockerman:
if we switch to testing each module individually instead of using `go test ./...` we can parallelize even more

antekresic:
Could we use some caching to reduce the go module downloading? 

https://github.com/actions/cache/blob/main/examples.md#go---modules

JLockerman:
@antekresic almost certainly, though since a build on.y takes ~30s I'm not sure how much it'll help

antekresic (approved):
I think this looks good to merge, once you get rid of the unused function.

antekresic:
@cevian what do you think about this one? I like it, anything that can speed CI is a good change in my book.

JLockerman:
One issue is that I'm not sure how much it's actually speeding things up at

JLockerman:
(whoops, wrong button)

JLockerman:
Currently two outstanding issues:
1. it's using the wrong docker image for most of the tests
2. I'm not sure if it's actually speeding up CI

antekresic:
Superseded by #395

PR #334, opened 2020-11-05
Partial Multinode Support

JLockerman:
This commit adds initial support for multinode: it enables storing metrics as distributed hypertables when Promscale is run against a multinode cluster, and adds testing that this works as intended.

Multinode support is not considered complete because compression policies do not yet work on multinode clusters; only after this is added, either upstream or with a custom Promscale compression runner will multinode metrics be ready to use. This feature is tracked in issue #347.

We currently do not support migrating an existing single-node Promscale instance to multiple nodes, and adding additional nodes once Promscale is running requires some manually intervention to set up the schemas and users on the data nodes.


pkg/internal/testhelpers/containers.go
```diff
@@ -130,118 +199,263 @@ func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
...
 		image = "postgres:12"
 	}
-	return StartPGContainerWithImage(ctx, image, testDataDir, "", printLogs, usePromscaleExtension, useTimescaleDB)
+
+	return StartPGContainerWithImage(ctx, image, testDataDir, "", printLogs, extensionState)
 }
 
-func StartPGContainerWithImage(ctx context.Context, image string, testDataDir string, dataDir string, printLogs bool, usePromscaleExtension bool, useTimescaleDB bool) (testcontainers.Container, error) {
-	containerPort := nat.Port("5432/tcp")
+func StartPGContainerWithImage(ctx context.Context,
```
cevian, line 229: I wonder if this would be cleaner if we separate this into a single-node and multi-node function that calls helpers. Not sure.

pkg/internal/testhelpers/containers.go
```diff
@@ -130,118 +199,263 @@ func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
...
 	}
 
-	pgPort, err = container.MappedPort(ctx, containerPort)
-	if err != nil {
-		return nil, err
+	if multinode {
+		containerPort := nat.Port("5433/tcp")
+		_, err := startContainer(containerPort)
+		if err != nil {
+			c.Close()
```
cevian, line 400: why not do this in a defer?

pkg/internal/testhelpers/containers.go
```diff
@@ -79,35 +150,90 @@ func GetReadOnlyConnection(t testing.TB, DBName string) *pgxpool.Pool {
...
 
-	err = db.Close(context.Background())
+	ourDb, err := pgx.Connect(context.Background(), PgConnectURL(DBName, Superuser))
 	if err != nil {
 		return nil, err
 	}
 
-	//some docker images may have timescaledb installed in a template, which we don't always want
-	dbDrop, err := pgx.Connect(context.Background(), PgConnectURL(DBName, Superuser))
+	err = func() error {
```
cevian, line 194: since I don't see any defers here why is this run inside a closure?

pkg/internal/testhelpers/containers.go
JLockerman: because if we _don't_ error we still need the container around

pkg/internal/testhelpers/containers.go
JLockerman: line 232: if we didn't have the closure we'd have to remember to add the Close() for every return (I think I'm going to split out this closure into a real function)

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -100,7 +101,17 @@ BEGIN
         END IF;
 
         IF SCHEMA_CATALOG.is_timescaledb_installed() AND SCHEMA_CATALOG.get_default_compression_setting() THEN
-            PERFORM SCHEMA_PROM.set_compression_on_metric_table(r.table_name, TRUE);
+            IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
```
cevian, line 104: We should create an is_distributed function because this code repeats here and at line 157

pkg/internal/testhelpers/containers.go
```diff
@@ -130,24 +236,145 @@ func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
...
+	case VanillaPostgres:
 		image = "postgres:12"
 	}
-	return StartPGContainerWithImage(ctx, image, testDataDir, "", printLogs, usePromscaleExtension, useTimescaleDB)
+
+	return StartPGContainerWithImage(ctx, image, testDataDir, "", printLogs, extensionState)
 }
 
-func StartPGContainerWithImage(ctx context.Context, image string, testDataDir string, dataDir string, printLogs bool, usePromscaleExtension bool, useTimescaleDB bool) (testcontainers.Container, error) {
+func StartPGContainerWithImage(ctx context.Context,
```
cevian, line 266: This now starts multiple containers. The name should reflect that
```suggestion
func StartPGClusterWithImage(ctx context.Context,
```

pkg/internal/testhelpers/containers.go
```diff
@@ -130,24 +236,145 @@ func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
...
+	if multinode {
+		networkName, closer, err := createMultinodeNetwork()
+		if err != nil {
+			return nil, c, err
+		}
+		c.Append(closer)
+		networks = []string{networkName}
+	}
+
+	startContainer := func(containerPort nat.Port) (testcontainers.Container, error) {
```
cevian, line 287: This has a code smell. You are only creating this closure to avoid passing in all these parameters in multiple places. We should probably create an object (ClusterConfig?) that stores these configs and it can have StartCluster and StartInstance methods.

pkg/internal/testhelpers/containers.go
```diff
@@ -158,20 +385,25 @@ func StartPGContainerWithImage(ctx context.Context, image string, testDataDir st
...
-		SkipReaper: false, /* switch to true not to kill docker container */
+		Networks:       networks,
+		NetworkAliases: map[string][]string{"promscale-network": {"db" + containerPort.Port()}},
+		SkipReaper:     false, /* switch to true not to kill docker container */
 	}
 	req.Cmd = []string{
 		"-c", "max_connections=100",
+		"-c", "port=" + containerPort.Port(),
+		"-c", "max_prepared_transactions=150",
+		"-i",
```
cevian, line 396: what does the -i do?

pkg/internal/testhelpers/containers.go
```diff
@@ -209,39 +441,120 @@ func StartPGContainerWithImage(ctx context.Context, image string, testDataDir st
...
 	if err != nil {
-		return nil, err
+		return nil, closer, err
 	}
 
-	return container, nil
+	return container, closer, nil
+}
+
+func enableDataNodes(db *pgx.Conn, database string) error {
```
cevian, line 502:
```suggestion
func attachDataNodesOnAN(db *pgx.Conn, database string) error {
```

pkg/internal/testhelpers/containers.go
```diff
@@ -130,24 +236,145 @@ func (s stdoutLogConsumer) Accept(l testcontainers.Log) {
...
+
 	containerPort := nat.Port("5432/tcp")
+	container, err := startContainer(containerPort)
+	if err != nil {
+		_ = c.Close()
+		return nil, nil, err
+	}
+
+	if multinode {
+		db, err := pgx.Connect(context.Background(), PgConnectURL(defaultDB, Superuser))
```
cevian, line 327: why not do the connect and shutdown within enableDataNodes if connection never used anywhere else?

cevian (changes requested):

pkg/internal/testhelpers/containers.go
JLockerman: hmm, it's not always a cluster though. How about `StartDatabaseWithImage(...)`

pkg/internal/testhelpers/containers.go
JLockerman: I agree that this function needs to be refactored; honestly I think most of `testhelpers` needs to be (in particular the way we're storing `host` and `port` as global state has bitten us a few times). However, I think it's only as much a priority as the other refactors we have on the list, like moving `end_to_end_tests` or breaking up `sql_ingest.go`; for the same reason we decided to prioritize 2.0 support over finishing those, I think we should do the same here, and revisit this after multinode is working. (I'm not entirely sure that a config struct solves the problem here: config structs make things less fraught for callers since they give arguments names. However, here we should only ever have one caller, since this is a private helper function, and most of the arguments are internal details. I think the issue we're actually running into is that the container organization has implications to our test logic, but we don't have a good way to separate out container-starting from container-organization because testcontainers requires both of them at container start)

(I also wonder if there are ways we can refactor this to share containers across more tests, it might help with some of our CI issues).

And because I'm so easily nerd-sniped:
I'm not entirely sure that a lamda capturing the contextual arguments is bad style in the general case: it's pretty much equivalent to the partial function evaluation
```haskell
startConainer = startPGInstance ctx image testDataDir dataDir extensionState printLogs networks
```
though with a slightly uglier style since go doesn't have partial evaluation. This contains strictly more information than a config struct would; it says not only that these function calls have the same arguments, it says they _must_ have identical arguments, and only the remaining inputs can change. With a config struct it's ambiguous which of the arguments are set for convenience and which are set out of necessity.

pkg/internal/testhelpers/containers.go
JLockerman: > Allows remote clients to connect via TCP/IP (Internet domain) connections. Without this option, only local connections are accepted. This option is equivalent to setting listen_addresses to * in postgresql.conf or via -h.

I figured it pays to be explicit about this, in case we decide to tighten up the `postgresql.conf`

pkg/internal/testhelpers/containers.go
JLockerman: if we're going that direction I think `attachDataNode()`; I think it should be clear we're attaching to the db we're passing in

pkg/internal/testhelpers/containers.go
JLockerman: I want to use the same function in [`DbSetup`](https://github.com/timescale/promscale/pull/334/files#diff-a9a488c67e814258e1b5878fa9c9160b6b011e45d10772a918341c983d185204R198) and there the connection is used for other things (not married to that, I can move the connection into the other branch as well)

pkg/internal/testhelpers/containers.go
cevian: Let's add this to the refactorpalooza list then

cevian (approved):
LGTM. Need a cleanup of commits. And clear messaging in PR description and commits that compression isn't included but will be in a followup PR.

Issue #329, opened 2020-11-02
Better auto-upgrade policies for dependencies


cevian:
A few things to do (all up for discussion) 

- [x] Prevent auto-upgrades to extension versions higher than default version of extension
- [x] Prevent auto-upgrades across major versions (guarded by flag)
- [x] Prevent auto-upgrades to pre-release versions (guarded by flag; possibly with exception of default version?)
- [x] Have a catch-all auto-upgrade flag for all extensions
- [x] Make sure all CLI flags settable via helm (done?)
- [ ] Document production setting options 
    - use immutable docker tags (preferred I think)
    - turn off auto-upgrade

franck102:
Hi, I would also suggest:
1. Having any upgrade turned off by default
2. Having installation and upgrades of the timescaledb extension turned off by default

I do understand the desire to make things easy for users, however with the current defaults the risk is high to see a properly tested solution deployed to a production environment, only to fail at a random later time as a new release of either extension is made available. IMHO promscale shouldn't ever try to modify the timescaledb extension: after all we don't expect the timescale installer to update Postgres, or the Postgres installer to upgrade Kubernetes :)

It is also not quite clear whether using a completely specified timescaledb image tag (timescaledb-single.image.tag) in the timescale-single chart would prevent the promscale go command from trying to upgrade the timescale extension?

franck102:
Is there any proposed workaround for installing promscale today using helm?
The latest chart available from the repository, 0.1.1, doesn't support setting container args for the promscale container, so we cannot pass -install-timescaledb=false & -migrate=false, promscale upgrades tsdb to 2.0.0-rc2 and the deployment doesn't work.

I created a local copy of the latest promscale chart, specifying the arguments, and the promscale container now fails to start with this error:
`level=error ts=2020-11-09T07:28:42.988Z caller=runner.go:110 msg="aborting startup due to error" err="dependency error: db schema version is incorrect: expected 0.1.1, got 0.1.0-beta.5"`

Can the auto-install / auto-upgrade code in promscale be disabled until this is properly fixed? Promscale is currently breaking our TimescaleDB deployment, which is used by other services.

franck102:
Trying to install with the  -install-timescaledb=false & -migrate=false args, against a fresh 1.7.4 timescaledb install, promscale fails to start with this message:

level=info ts=2020-11-09T13:20:55.340Z caller=runner.go:103 msg="Version:0.1.1; Commit Hash: "
[...]

PR #335, opened 2020-11-05
Fix TimescaleDB version accepted

cevian:
Previously, the versions accepted the 2.0.0-rc which created problems
when being auto-upgraded. Fix this for now. But a fuller fix for
issue #329 will be in a separate PR.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -42,7 +42,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.0.0",
+		Safe: ">=1.7.3 <1.7.99",
 		Warn: ">=1.7.0 <1.7.3",
 	}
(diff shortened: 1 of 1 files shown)


pkg/version/version.go
```diff
@@ -42,7 +42,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.0.0",
+		Safe: ">=1.7.3 <1.7.99",
```
JLockerman, line 45: it feels a bit weird conditioning against a version that doesn't actually exist. Maybe we can use `<2.0.0-alpha` instead? That way we can be compatible with any theoretical future point releases.

JLockerman (commented):
basically looks good, though I do wonder if using `2.0.0-alpha` as the first version we're not compatible with might be better

Harkishen-Singh (approved):
Looks good to me, the comment which josh gave is extra safe (and meaningful). I am happy with either way because I don't think (afaik) there's gonna be any version after 1.7.

cevian:
Superseded by 328?

PR #336, opened 2020-11-06
Specify the version to test upgrades from explicitly

cevian:
Instead of trying to determine the previous version for
the update tests, set it explicitly. Also set it as
far back as reasonable. Moving it back always adds
more code to what is tested, since more upgrade scripts
are run.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -33,8 +33,9 @@ var (
 	// since an app version must uniquely determine the state of the schema.
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
-	Version    = "0.1.3-dev.1"
-	CommitHash = ""
+	Version                    = "0.1.3-dev.1"
+	CommitHash                 = ""
+	EarliestUpgradeTestVersion = "0.1.0"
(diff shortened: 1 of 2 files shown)


pkg/version/version.go
```diff
@@ -33,8 +33,9 @@ var (
 	// since an app version must uniquely determine the state of the schema.
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
-	Version    = "0.1.2-dev.0"
-	CommitHash = ""
+	Version                     = "0.1.2-dev.0"
+	CommitHash                  = ""
+	VersionToUpgradeFromInTests = "0.1.0"
```
JLockerman, line 38: oy that name. What do you think of `EarliestCompatibleVersion` or `EarliestTestedVesrsion` or, dare I say it, `EarliestSupportedVersion`, or something like that?

JLockerman (commented):
Looks food to me!

pkg/version/version.go
antekresic: `EarliestUpgradeableVersion`?

antekresic (approved):

PR #338, opened 2020-11-09
Auto upgrade fixes

cevian:
Related to #329

### DIFF
--- a/pkg/pgmodel/extension.go
+++ b/pkg/pgmodel/extension.go
@@ -121,14 +121,14 @@ func migrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		return fmt.Errorf("the extension is not available")
 	}
 
-	newVersion, ok := getNewExtensionVersion(availableVersions, validRange)
-	if !ok {
-		return fmt.Errorf("the extension is not available at the right version, need version: %v", rangeString)
-	}
-
 	currentVersion, isInstalled, err := fetchInstalledExtensionVersion(conn, extName)
 	if err != nil {
 		return fmt.Errorf("Could not get the installed extension version: %w", err)
+	}
+
+	newVersion, ok := getNewExtensionVersion(extName, availableVersions, validRange, isInstalled, currentVersion)
+	if !ok {
+		return fmt.Errorf("the extension is not available at the right version, need version: %v", rangeString)
 	}
 
 	if !isInstalled {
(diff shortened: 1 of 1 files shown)


pkg/pgmodel/extension.go
```diff
@@ -234,10 +256,30 @@ func getSqlVersion(v semver.Version, extName string) string {
...
 // getNewExtensionVersion returns the highest version allowed by validRange
-func getNewExtensionVersion(availableVersions semver.Versions, validRange semver.Range) (semver.Version, bool) {
+func getNewExtensionVersion(extName string,
+	availableVersions semver.Versions,
+	defaultVersion semver.Version,
+	validRange semver.Range,
+	validCurrentVersion bool,
+	currentVersion semver.Version) (semver.Version, bool) {
 	//sort higher extensions first
 	sort.Sort(sort.Reverse(availableVersions))
```
Harkishen-Singh, line 266: I don't see availableVersion is already sorted. So, shouldn't this be `sort.Sort(sort.Reverse(sort.SliceStabe(callback)))` ? Or a normal sort and then negative for loop.

pkg/pgmodel/extension.go
```diff
@@ -234,11 +256,20 @@ func getSqlVersion(v semver.Version, extName string) string {
...
 	printedWarning := false
 	for i := range availableVersions {
+		/* skip any versions above the default version for auto-upgrade */
+		if availableVersions[i].GT(defaultVersion) {
```
Harkishen-Singh, line 270: Could you please mention the significance of the `defaultVersion` here? The name default does not self explain the matter to me.

Harkishen-Singh:
Also, should we have some unit-test here? Like, making a mock version that is greater in major and then trying to see if it is accepted or not by the function. WDYT?

pkg/pgmodel/extension.go
cevian: Default version is specified by the extension itself when installed into the system. It is usually the latest version but not always. It is installed by CREATE EXTENSION if no version is explicitly specified.

pkg/pgmodel/extension.go
cevian: Semver.Versions has Sort.Interface implemented. So this will sort in descending version order.

Harkishen-Singh (approved):
Lgtm!

JLockerman (approved):
LGTM

JLockerman:
We might want to add some more explanatory comments at some point (such as what `defaultVersion` is) but those can happily wait to a followup PR

PR #344, opened 2020-11-11
Add import endpoint

antekresic:
This new endpoint is supposed to be an easier way to ingest data using Promscale
by sending JSON streaming metric data.

### DIFF
--- a/pkg/api/write.go
+++ b/pkg/api/write.go
@@ -1,7 +1,9 @@
 package api
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -14,6 +16,7 @@ import (
 	io_prometheus_client "github.com/prometheus/client_model/go"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel"
+	"github.com/timescale/promscale/pkg/prompb"
 	"github.com/timescale/promscale/pkg/util"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/api/write.go
```diff
@@ -148,6 +158,96 @@ func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
...
+		numSamples, err := writer.Ingest(req.GetTimeseries(), req)
+		if err != nil {
+			log.Warn("msg", "Error importing samples to remote storage", "err", err, "num_samples", numSamples)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	})
+}
+
+func toWriteRequest(i importPayload, req *prompb.WriteRequest) *prompb.WriteRequest {
```
cevian, line 220:
```suggestion
func appendImportPayload(req *prompb.WriteRequest, i importPayload) *prompb.WriteRequest {
```

pkg/api/write.go
```diff
@@ -148,6 +158,96 @@ func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
...
+	log.Error("msg", msg, "err", err)
 	http.Error(w, err, http.StatusBadRequest)
 }
+
+func buildImportError(w http.ResponseWriter, err string) {
+	buildError(w, err, "Import header validation error")
+}
+
+type importPayload struct {
+	Metric     map[string]string `json:"metric"`
```
JLockerman, line 174: is this to insert the same values to multiple metrics, or does it store the labels key-value mappings, or something else?

pkg/api/write.go
```diff
@@ -148,6 +158,96 @@ func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
...
+				log.Error("msg", "Import error", "err", err.Error())
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+			if len(i.Timestamps) != len(i.Values) {
+				log.Error("msg", "Import error", "err", "Mismatch number of timestamps and values in payload")
+				http.Error(w, "Mismatch number of timestamps and values in payload", http.StatusInternalServerError)
+				return
+			}
```
JLockerman, line 206: This feels like a bit of an awkward constraint to impose; is this better than having the input be an array of `(time, value)` pairs?

pkg/api/write.go
JLockerman: Looking at the code, it seems to be the label key-value mappings. Can we name this `labels` or `series`?

JLockerman (commented):
Over all it looks good, though I have some questions about the JSON format. If we do have some constraints forcing it to be the way it is, we should document them.

PR #345, opened 2020-11-12
Add JSON streaming payload support to write endpoint

antekresic:


pkg/api/write.go
```diff
@@ -100,6 +89,54 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
...
+		if err != nil {
+			return nil, err, "Snappy decode error"
+		}
+
+		req = pgmodel.NewWriteRequest()
+		if err := proto.Unmarshal(reqBuf, req); err != nil {
+			pgmodel.FinishWriteRequest(req)
+			return nil, err, "Protobuf unmarshal error"
+		}
+	case "application/json":
```
cevian, line 112: we should def support snappy (optionally) in this path as well

cevian (commented):
Looks great. One comment about snappy. Let's resolve the json format issue with @JLockerman

pkg/api/write.go
JLockerman: @cevian I'd recommend doing that in a separate PR if it takes any non-trivial time

pkg/api/write.go
```diff
@@ -45,23 +48,10 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
...
-		compressed, err := ioutil.ReadAll(r.Body)
-		if err != nil {
-			log.Error("msg", "Read error", "err", err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
-		}
-
-		reqBuf, err := snappy.Decode(nil, compressed)
+		req, err, logMsg := loadWriteRequest(r)
+		defer r.Body.Close()
```
cevian, line 52: defers usually go after the err check. Is this a special case?

antekresic:
@cevian @JLockerman updated the JSON format and added snappy support for it. Could you take another look?

There is one subtle thing with snappy here, since we are using JSON streaming format, you need to use snappy streaming format as well or it won't work. Write request uses snappy's block format, if you use that you would just get an error about `corrupt input`.

pkg/api/write.go
```diff
@@ -151,3 +192,65 @@ func buildWriteError(w http.ResponseWriter, err string) {
...
+
+	i.Value, ok = v[1].(float64)
+	if !ok {
+		return fmt.Errorf("cannot convert value to float64 type for sample: %v", v)
+	}
+
+	return nil
+}
+
+func toWriteRequest(i importPayload, req *prompb.WriteRequest) *prompb.WriteRequest {
```
cevian, line 233:
```suggestion
func appendWriteRequest(i importPayload, req *prompb.WriteRequest) *prompb.WriteRequest {
```

pkg/api/write.go
```diff
@@ -151,3 +192,65 @@ func buildWriteError(w http.ResponseWriter, err string) {
...
+
+	if len(v) != 2 {
+		return fmt.Errorf("invalid sample, should contain timestamp and value: %v", v)
+	}
+
+	floatTimestamp, ok := v[0].(float64)
+	if !ok {
+		return fmt.Errorf("cannot convert timestamp to int64 type for sample: %s", data)
+	}
+	i.Timestamp = int64(floatTimestamp)
```
cevian, line 223: can we make sure the floatTimestamp has no decimal part. We don't want to silently drop precision here as that implies the input is wrong

pkg/api/write_test.go
```diff
@@ -23,10 +24,174 @@ import (
...
+		{
+			name:        "invalid timestamp",
+			input:       `["1",2]`,
+			shouldError: true,
+		},
+		{
+			name:        "invalid value",
+			input:       `[1,"2"]`,
+			shouldError: true,
+		},
```
cevian, line 70: add a test for [10.1, 10.2] should fail due to decimal in time

cevian (approved):
A few small changes

pkg/api/write.go
antekresic: Actually, this is my bad, we don't need to close it for requests. Will remove.

JLockerman:
> Since we are using JSON streaming format, you need to use snappy streaming format as well or it won't work.

@antekresic could you explain this one a bit more?

antekresic:
```There are actually two Snappy formats: block and stream. They are related, but different: trying to decompress block-compressed data as a Snappy stream will fail, and vice versa. The block format is the Decode and Encode functions and the stream format is the Reader and Writer types.```

https://pkg.go.dev/github.com/golang/snappy

**Comment by JLockerman** [comment]
@antekresic yeah, I meant on the JSON side. As far as I'm aware JSON is a pure-text format with only one encoding, so it shouldn't matter what the compression layer above it does, we should be able to decompress it, then feed the resulting data into a decoder.

That said, I've been burned before by the fact that everyone else only uses one of the snappy formats while go uses both. Would you be willing to test against something like [snzip's defaults](https://formulae.brew.sh/formula/snzipl) see if it works? If the only way to use this would be go, it'd be less great...

**Comment by JLockerman** [comment]
> Would you be willing to test against something like snzip's defaults see if it works? 

(@antekresic by that I mean just run it once, not add it to the test suite or anything)

**Review by JLockerman — APPROVED** [review]
Besides the above it looks good.
Unless we have concrete plans to add textual timestamps in the near future, we should open a tracking issue so users have a place to request them.

**Comment by antekresic** [comment]
My main concern is the "sharp edge" this creates with the snappy format. I would be willing to make it sub-optimal performance-wise for this JSON case to avoid it causing support issues.

What do you guys think?

**Comment by JLockerman** [comment]
I think we should use the snappy format that the C++ library uses; that seems to be the most commonly used snappy.

**Comment by JLockerman** [comment]
> ... we should open a tracking issue so users have a place to request them.

Textual timestamp tracking issue open at #352.

File: pkg/api/write.go [hunk 6]
```diff
@@ -100,6 +90,69 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
 	})
 }
 
+func loadWriteRequest(r *http.Request) (*prompb.WriteRequest, error, string) {
+	var req *prompb.WriteRequest
+
+	snappyEncoding := strings.Contains(r.Header.Get("Content-Encoding"), "snappy")
```
JLockerman, line 96: can be moved into the json case

pkg/api/write.go
```diff
@@ -123,31 +190,117 @@ func validateWriteHeaders(w http.ResponseWriter, r *http.Request) bool {
...
+}
+
+func detectSnappyStreamFormat(input []byte) bool {
+	if len(input) < 10 {
 		return false
 	}
 
-	if !strings.HasPrefix(remoteWriteVersion, "0.1.") {
-		buildWriteError(w, fmt.Sprintf("unexpected Remote-Write-Version %s, expected 0.1.X", remoteWriteVersion))
+	if string(input[0:10]) != "\xff\x06\x00\x00\x73\x4e\x61\x50\x70\x59" {
```
JLockerman, line 301: we should add a link to the framing format description https://github.com/google/snappy/blob/01a566f825e6083318bda85c862c859e198bd98a/framing_format.txt#L68-L81, and describe why we think filtering based on this will work.
(I'm not entirely sure it _does_ work; `0xff, 0x06` is a completely valid varint representing 895 and it may be possible to construct a valid snappy-encoded buffer starting with this string. That said, I don't think it's likely, and we can safely ignore it)

JLockerman (approved):

pkg/api/write.go
```diff
@@ -100,6 +90,69 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
...
+			return nil, err, "Protobuf unmarshal error"
+		}
+	case "application/json":
+		var i importPayload
+		var body io.Reader = r.Body
+		if snappyEncoding {
+			buf := make([]byte, 10)
+			_, err := r.Body.Read(buf)
+			if err != nil {
+				return nil, err, "payload too short or corrupt"
```
cevian, line 117:
```suggestion
				return nil, err, "snappy payload too short or corrupt"
```

pkg/api/write.go
```diff
@@ -100,6 +90,69 @@ func Write(writer pgmodel.DBInserter, elector *util.Elector, metrics *Metrics) h
...
+		req = pgmodel.NewWriteRequest()
+		if err := proto.Unmarshal(reqBuf, req); err != nil {
+			pgmodel.FinishWriteRequest(req)
+			return nil, err, "Protobuf unmarshal error"
+		}
+	case "application/json":
+		var i importPayload
+		var body io.Reader = r.Body
+		if snappyEncoding {
+			buf := make([]byte, 10)
```
cevian, line 114: can we break this all out into a separate function? e.g. decodeSnappy?

cevian (approved):
Minor code cleanup comments + agree with some of Josh's suggestions.

pkg/api/write.go
antekresic: `0xff` is the byte that signals snappy stream format, any chunk that starts with that must mean stream encoding. I doubt you can have a valid snappy buffer starting with that byte without it being a stream. I've checked the whole "magic chunk" which needs to be at the start of stream. If what you are saying is possible, it would be an issue with snappy itself IMO.

Will add a comment there to clear up what we are checking for.

PR #349, opened 2020-11-15
Adds support for prom-migrator

Harkishen-Singh:
This PR adds support for prom-migrator. Prom-migrator is a migration tool for prometheus data, that can migrate from one remote storage to another using the remote-read and remote-write endpoints.


pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		blockRef.SetDescription(fmt.Sprintf("fetching time-range: %d mins...", timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+		result, err := fetch.start(context.Background(), []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, "", ".*")})
+		if err != nil {
+			return fmt.Errorf("remote-read run: %w", err)
+		}
+		if len(result.Timeseries) == 0 {
+			rr.plan.DecrementBlockCount()
+			i += timeRangeMinutesDelta + 1
+			if timeRangeMinutesDelta < MaxTimeRangeDeltaLimit.Milliseconds() {
+				timeRangeMinutesDelta *= 2
```
blagojts, line 88: add a comment why

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+			continue
+		}
+		qResultSize := result.Size() // Assuming each character corresponds to one byte.
+		rr.plan.UpdateLastPushedMaxt(i)
```
blagojts, line 93: push hasn't happened

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		blockRef.SetDescription(fmt.Sprintf("received %d MB with delta %d mins...", qResultSize/megaByte, timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+		blockRef.Timeseries = result.Timeseries
+		rr.sigBlockRead <- struct{}{}
+		<-rr.sigBlockWrite
+		if err := rr.plan.Clear(); err != nil {
+			return fmt.Errorf("remote-read run: %w", err)
+		}
+		switch {
+		case qResultSize < ResponseDataSizeHalfLimit && timeRangeMinutesDelta >= MaxTimeRangeDeltaLimit.Milliseconds():
```
blagojts, line 102: explain case

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		Matchers:         ms,
+	}
+	result, err := f.clientCopy.Read(context, readRequest)
+	if err != nil {
+		return nil, fmt.Errorf("executing client-read: %w", err)
+	}
+	return result, nil
+}
+
+func toLabelMatchers(matchers []*labels.Matcher) ([]*prompb.LabelMatcher, error) {
```
blagojts, line 151: why is this needed and the []*prompb.LabelMatchers are not directly created?

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		mint:       mint,
+		maxt:       maxt,
+		clientCopy: rr.client,
+	}
+	rr.fetchersRunning.Add(1)
+	return f
+}
+
+// start starts fetching the samples from remote read storage as client based on the matchers.
+func (f *fetch) start(context context.Context, matchers []*labels.Matcher) (*prompb.QueryResult, error) {
```
blagojts, line 134: looks like the function starts and finishes fetching the samples, maybe it can be called fetch

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		default:
+			i += timeRangeMinutesDelta + 1
+			timeRangeMinutesDelta /= 2
+		}
+	}
+	log.Info("msg", "reader is down")
+	return nil
+}
+
+type fetch struct {
```
blagojts, line 116: the fetcher in the current implementation seems like it is not needed, it is only used once to call the start method and discarded, can be a function

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
...
+				timeRangeMinutesDelta *= 2
+			}
+			continue
+		}
+		qResultSize := result.Size() // Assuming each character corresponds to one byte.
+		rr.plan.UpdateLastPushedMaxt(i)
+		blockRef.SetDescription(fmt.Sprintf("received %d MB with delta %d mins...", qResultSize/megaByte, timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+		blockRef.Timeseries = result.Timeseries
+		rr.sigBlockRead <- struct{}{}
+		<-rr.sigBlockWrite
```
blagojts, line 97: go channels auto block if they reached capacity. sigBlockWrite can be removed, and if the writer hasn't processed the previous signal then `rr.sigBlockRead <- struct{}{} ` will block

pkg/migration-tool/writer/writer.go
```diff
@@ -0,0 +1,153 @@
...
+		blockRef.SetDescription("pushing...", 1)
+		ts := timeseriesRefToTimeseries(blockRef.Timeseries)
+		if err := rw.sendSamplesWithBackoff(context.Background(), ts, &buf); err != nil {
+			return fmt.Errorf("remote-write run: %w", err)
+		}
+		if err := blockRef.Done(); err != nil {
+			return fmt.Errorf("remote-write: %w", err)
+		}
+		rw.sigBlockWrite <- struct{}{}
+		if !readerUp.Load() {
```
blagojts, line 78: can the sigBlockRead channel be used to signal whether the reader is up or not? it will close upon existing the Reader.Run function

pkg/migration-tool/writer/writer.go
```diff
@@ -0,0 +1,153 @@
...
+	log.Info("msg", "writer is up")
+	defer func() {
+		wg.Done()
+	}()
+	for {
+		if _, ok := <-rw.sigBlockRead; !ok {
+			break
+		}
+		var buf []byte
+		blockRef := rw.plan.CurrentBlock()
```
blagojts, line 68: Since the plan doesn't reuse the created Block structures, just creates a new one on Clear, why not just pass a *Block in the read channel, and the writer will consume it. That will simplify Plan to not have to keep track of the current block, and it will be easier to expand the code to use multiple writers in the future

blagojts (changes requested):
Overal structure is good, I like the organization of the code. I think some simplifications can be made, left as comments throughout the code

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,83 @@
...
+type config struct {
+	mint     int64
+	maxt     int64
+	readURL  string
+	writeURL string
+}
+
+func main() {
+	conf := new(config)
+	flag.Int64Var(&conf.mint, "mint", -1, "Minimum timestamp for carrying out data migration. Setting this value less than zero will indicate all data upto the maxt. Setting mint and maxt less than zero will migrate all data available in the read storage.")
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, I couldn't get this. Do you mean the int64 or even supporting decimal timestamps or normal UTC time?

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+func (p *Plan) UpdateLastPushedMaxt(t int64) {
+	p.lastPushedT.Store(t)
+}
+
+func (p *Plan) DecrementBlockCount() {
+	p.blockCounts.Sub(1)
+}
+
+// CreateBlock creates a new block and returns reference to the block for faster write and read operations.
+func (p *Plan) CreateBlock(mint, maxt int64) (reference *Block) {
```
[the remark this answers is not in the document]
Harkishen-Singh: Yeah, that would be good for being extra safe. Thank you.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	reference = &Block{
+		ID:              p.blockCounts.Add(1),
+		percent:         percent,
+		pbarInitDetails: fmt.Sprintf("block-%d time-range: %d mins | mint: %d | maxt: %d", p.blockCounts.Load(), (maxt-mint)/time.Minute.Milliseconds(), mint, maxt),
+		pbar: progressbar.NewOptions(
+			6,
+			//progressbar.OptionSetRenderBlankState(true),
+			progressbar.OptionOnCompletion(func() {
+				fmt.Fprint(os.Stderr, "\n")
```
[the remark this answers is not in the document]
Harkishen-Singh: I copied this from the reference docs :sweat_smile:

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+			}),
+		),
+		mint: mint,
+		maxt: maxt,
+	}
+	reference.SetDescription(fmt.Sprintf("fetching time-range: %d mins...", (maxt-mint)/time.Minute.Milliseconds()), 1)
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, didn't get this. The block is initialized here.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+	p.currentBlock = reference
```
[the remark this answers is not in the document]
Harkishen-Singh: I don't think its required as after the recent commits, the currentBlock will no longer be there. We will directly pass the address to the writer than giving it exclusive access.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+	pbarInitDetails string
+	pbar            *progressbar.ProgressBar
+	Timeseries      []*prompb.TimeSeries
+}
+
+func (b *Block) SetData(timeseries []*prompb.TimeSeries) {
+	b.Timeseries = timeseries
+}
+
+func (b *Block) SetDescription(description string, proceed int) {
```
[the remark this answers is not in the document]
Harkishen-Singh: > Does a block's description change while it lives

Yes. The description is the message to the left of the block. It provides real-time updates as to what that block is doing. These include: ['fetching...', 'received %MB', 'preparing to push', 'pushing...', 'pushed']

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,174 @@
+package reader
+
+import (
+	"context"
+	"fmt"
+	"go.uber.org/atomic"
```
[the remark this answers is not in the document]
Harkishen-Singh: Aimed for strict types. Using `sync/atomic` will lead to type conversions (from interface) which may make code look untidy.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Changed in recent commits.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: `labels.Matchers` is not accepted in `promb.Query` unless in `promb.LabelMatcher` form. That's why we need to convert.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: For now, yes that is best. But I think the reason why this was done is that fetch was planned to be able to divide 1 big request into multiple small requests and hence this was created in the way it is. Now, I don't see that coming up, so better to just keep this as a function. Thank you.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+type Plan struct {
+	mux          sync.RWMutex
+	Mint         int64
+	Maxt         int64
+	currentBlock *Block
+	blockCounts  atomic.Int64
+	lastPushedT  atomic.Int64
+}
+
+// CreatePlan creates a planner in the provided planner-path. It checks for planner config in the
```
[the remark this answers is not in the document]
Harkishen-Singh: This is no longer the case. Recent commits have made the tool stateless, thus no need for planner-path.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,126 @@
...
+	if !p.currentBlock.done {
+		if err := p.currentBlock.Done(); err != nil {
+			return fmt.Errorf("clear: %w", err)
+		}
+	}
+	p.currentBlock = nil
+	return nil
+}
+
+// Block provides the metadata of a block.
```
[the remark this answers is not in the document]
Harkishen-Singh: Added the description.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Done in recent commits.

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Yeah, that was the plan during refactoring. Done now.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Yes I agree, but the lib I am using for progress-bar has some issues with concurrency, so I have to be extra careful that at a single time, only one block write to progress bar. If I do that way, it will to 2 blocks writing, one by the writer and the other one waiting in memory for the next write iteration.

cevian (commented):
It seems like you are still making changes to get rid of currentBlock and add tests. I will wait for those to review

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/storage/remote"
+)
+
+var interner = sync.Map{}
+
+const Megabyte = 1024 * 1024
+
+// CreateReadClient creates a new read client that can be used to fetch promb samples.
+func CreateReadClient(name, urlString string, readTimeout model.Duration) (remote.ReadClient, error) {
```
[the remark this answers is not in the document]
blagojts: and I think it's ok to either move the validation to the NewReadClient function, or create a separate function that would validate the url and return a ClientConfig (but still in the remote package)

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	"sync"
+
+	"github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/storage/remote"
+)
+
+var interner = sync.Map{}
```
blagojts, line 15: what's this? and I would avoid global state

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+		Samples: make([]prompb.Sample, 0),
+	}}
+	interner.Store(labels.MetricName, ps)
+	return ps, nil
+}
+
+// MergeWithTimeseries merges the new series with the time-series.
+func MergeWithTimeseries(ts *[]*prompb.TimeSeries, newTS *prompb.TimeSeries) {
+	*ts = append(*ts, newTS)
+}
```
blagojts, line 91: do we really need this function?

pkg/migration-tool/writer/writer.go
```diff
@@ -24,49 +21,49 @@ const (
...
+		if !ok {
 			break
 		}
 		var buf []byte
-		blockRef := rw.plan.CurrentBlock()
-		blockRef.SetDescription("pushing...", 1)
+		blockRef.SetDescription(fmt.Sprintf("pushing %.2f...", float64(blockRef.Bytes())/float64(utils.Megabyte)), 1)
+		ps, err := utils.GetorGenerateProgressTimeseries(rw.progressMetricName, rw.migrationJobName)
+		if err != nil {
+			return fmt.Errorf("get or create progress time-series: %w", err)
```
blagojts, line 63: can be done on init and stored in plan? so here it will only be updated and appended?

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+	"time"
+
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/schollz/progressbar/v3"
+	"github.com/timescale/promscale/pkg/migration-tool/utils"
+)
+
+// createBlock creates a new block and returns reference to the block for faster write and read operations.
+func (p *Plan) createBlock(mint, maxt int64) (reference *Block, err error) {
```
cevian, line 14: if this is a method of Plan it should be in planner.go. In general I'd like to see a better separation between blocks and plans

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,105 @@
...
+		),
+		mint: mint,
+		maxt: maxt,
+	}
+	reference.SetDescription(fmt.Sprintf("fetching time-range: %d mins...", timeRangeInMinutes), 1)
+	p.currentBlock = reference
+	return
+}
+
+func (p *Plan) validateT(mint, maxt int64) error {
```
cevian, line 43: this should be in planner.go

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,96 @@
...
+		close(rr.sigBlockWrite)
+	}()
+	log.Info("msg", "reader is up")
+	for rr.plan.ShouldProceed() {
+		blockRef, err, timeRangeMinutesDelta := rr.plan.NextBlock()
+		if err != nil {
+			return fmt.Errorf("remote-run run: %w", err)
+		}
+		blockRef.SetDescription(fmt.Sprintf("fetching time-range: %d mins...", timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+		result, err := rr.fetch(context.Background(), blockRef.Mint(), blockRef.Maxt(), []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")})
```
cevian, line 60: I'd expect Fetch(c *Client) to be a method of Block

pkg/migration-tool/writer/writer.go
```diff
@@ -0,0 +1,147 @@
...
+	// snappy uses len() to see if it needs to allocate a new slice. Make the
+	// buffer as long as possible.
+	if buf != nil {
+		buf = buf[0:cap(buf)]
+	}
+	compressed := snappy.Encode(buf, data)
+	return compressed, nil
+}
+
+func timeseriesRefToTimeseries(tsr []*prompb.TimeSeries) (ts []prompb.TimeSeries) {
```
cevian, line 141: can't say I love this. Why don't we store []prompb.TimeSeries instead of []*prompb.TimeSeries to begin with

cevian (changes requested):
Very good work. Some comments for overall improvements. The thing I am most worried about is correct handling of boundary conditions on time. e.g. the time+1, equal vs <= etc. and code organization.

pkg/migration-tool/utils/utils.go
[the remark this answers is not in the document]
Harkishen-Singh: The structs/interface being created are from upstream package, so cannot be moved there. The `NewReadClient()` is from upstream remote package.

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+		Timeout: readTimeout,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("new read-client: %w", err)
+	}
+	return rc, nil
+}
+
+// CreateWriteClient creates a new write client that can be used to push promb samples.
+func CreateWriteClient(name, urlString string, readTimeout model.Duration) (remote.WriteClient, error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: From upstream package. These constructor functions are just to make the job easier. The actual structs are in prometheus remote package.

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+func validateConf(conf *config) error {
+	switch {
+	case strings.TrimSpace(conf.readURL) == "" && strings.TrimSpace(conf.writeURL) == "":
+		return fmt.Errorf("remote read storage url and remote write storage url must be specified. Without these, data migration cannot begin")
+	case strings.TrimSpace(conf.readURL) == "":
+		return fmt.Errorf("remote read storage url needs to be specified. Without read storage url, data migration cannot begin")
+	case strings.TrimSpace(conf.writeURL) == "":
+		return fmt.Errorf("remote write storage url needs to be specified. Without write storage url, data migration cannot begin")
+	case conf.mint > conf.maxt:
```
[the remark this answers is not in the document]
Harkishen-Singh: I think we should allow instant query like migrations too. But, they may not be that useful.

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	ps := &ProgressSeries{&prompb.TimeSeries{
+		Labels:  labelsToLabelsProto(lset),
+		Samples: make([]prompb.Sample, 0),
+	}}
+	interner.Store(labels.MetricName, ps)
+	return ps, nil
+}
+
+// MergeWithTimeseries merges the new series with the time-series.
+func MergeWithTimeseries(ts *[]*prompb.TimeSeries, newTS *prompb.TimeSeries) {
```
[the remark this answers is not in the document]
Harkishen-Singh: No we don't need this. In recent push, these have been removed.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,178 @@
...
+	var found bool
+	if remoteWriteStorageReadURL != "" && !ignoreProgress {
+		lastPushedMaxt, found, err := plan.fetchLastPushedMaxt()
+		if err != nil {
+			return nil, false, fmt.Errorf("create plan: %w", err)
+		}
+		if found && lastPushedMaxt > plan.Mint && lastPushedMaxt <= plan.Maxt {
+			log.Warn("msg", fmt.Sprintf("progress-metric found on the write storage. Last push was on %d. "+
+				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, lastPushedMaxt+1, plan.Maxt, (plan.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
+			plan.Mint = lastPushedMaxt + 1
```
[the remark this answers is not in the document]
Harkishen-Singh: We are assuming that the read-storage are including both the time-range boundaries. I don't think so. WDYT, should we proceed with just the next (+1) timestamp or continue with the previous?

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+	plan.lastMaxT = plan.Mint - 1 // Since block creation starts with (lastMaxT + 1), the first block will miss one millisecond if we do not subtract from here.
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, but I don't understand how it can get out of sync. Anyways, will set to int_max and do a check at NextBlock for it.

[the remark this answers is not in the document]
Harkishen-Singh: Also, I think it should be int_min, else the ShouldProceed() will return false.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,178 @@
...
+		readClient:         rc,
+		progressMetricName: progressMetricName,
+	}
+	plan.blockCounts.Store(0)
+	plan.lastPushedT.Store(0)
+	if ignoreProgress {
+		log.Info("msg", "ignoring progress-metric. Continuing migration with the provided time-range.")
+	}
+	var found bool
+	if remoteWriteStorageReadURL != "" && !ignoreProgress {
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes, we should. That will help avoid confusion for the users.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,178 @@
...
+				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, lastPushedMaxt+1, plan.Maxt, (plan.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
+			plan.Mint = lastPushedMaxt + 1
+		}
+	}
+	plan.lastMaxT = plan.Mint - 1 // Since block creation starts with (lastMaxT + 1), the first block will miss one millisecond if we do not subtract from here.
+	if plan.Mint > plan.Maxt {
+		return nil, false, fmt.Errorf("invalid: mint greater than maxt")
+	}
+	if plan.Mint == plan.Maxt && !found {
+		// progress_metric does not exists. This means that the migration is expected to be an instant query.
```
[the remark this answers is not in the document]
Harkishen-Singh: Migration which the user just wants to have a single timestamp to migrate. Just to fill all cases, though this looks wierd.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,178 @@
...
+		return plan, true, nil
+	}
+
+	return plan, true, nil
+}
+
+// fetchLastPushedMaxt fetches the maxt of the last block pushed to remote-write storage. At present, this is developed
+// for a single migration job (i.e., not supporting multiple migration metrics and successive migrations).
+func (p *Plan) fetchLastPushedMaxt() (lastPushedMaxt int64, found bool, err error) {
+	query, err := utils.CreatePrombQuery(p.Mint, p.Maxt, []*labels.Matcher{labels.MustNewMatcher(labels.MatchEqual, labels.MetricName, p.progressMetricName)})
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes we do require. I left this for the time. Will implement in the next push.

pkg/migration-tool/writer/writer.go
Harkishen-Singh: The response from fetch is promb.QueryResult. Promb.QueryResult has []*Timeseries. This means that by any means, we have to convert. But, converting at the last step (like here) is better, else will slow down the process due to transfering of hugeg data between cpu registers.

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ProgressSeries is the series that represents the progress metric series.
+type ProgressSeries struct {
+	*prompb.TimeSeries
+}
+
+// GetorGenerateProgressTimeseries returns the progress timeseries. It creates a new time-series based on input
+// if no time-series is present in the interner. To add a sample, call the append sample function.
+func GetorGenerateProgressTimeseries(metricName, migrationJobName string) (*ProgressSeries, error) {
+	if ps, ok := interner.Load(labels.MetricName); ok {
```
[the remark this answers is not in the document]
Harkishen-Singh: These are removed in the recent commits. Hence, resolving.

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return &prompb.Query{
+		StartTimestampMs: mint,
+		EndTimestampMs:   maxt,
+		Matchers:         ms,
+	}, nil
+}
+
+// ProgressSeries is the series that represents the progress metric series.
+type ProgressSeries struct {
```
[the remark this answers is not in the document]
Harkishen-Singh: I don't think there's any need for progress-series, and hence no progress_tracker, in accordance with the recent commits pushed.

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: Updated. How does it look now?

pkg/migration-tool/utils/utils.go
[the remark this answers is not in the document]
Harkishen-Singh: Now, the upstream functions have been localized. This is now in the line with what you expected here. Hence, resolving this.

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,96 @@
...
+		if len(result.Timeseries) == 0 {
+			rr.plan.DecrementBlockCount()
+			rr.plan.Update(0)
+			continue
+		}
+		qResultBytes, err := result.Marshal()
+		if err != nil {
+			return fmt.Errorf("marshalling prompb query result: %w", err)
+		}
+		numBytes := len(qResultBytes)
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes. Its implemented in recent commits. Hence, resolving.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Sorry, this `some issues with concurrency` went wrong. I meant to say that that lib has issues when two of its instance writes to the stdout, it shows weird behavior. However, I have figured out another library that handles multi-bars pretty well. I will update to that soon. Hence, not resolving this as it's still a TODO.

Harkishen-Singh: Also, each write on a progress-bar is concurrency safe. The issue is only when more than one bar are written simultaneously to the stdout.

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,114 @@
...
+		_, _ = fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+	log.Info("msg", fmt.Sprintf("%v+", conf))
+	if err := validateConf(conf); err != nil {
+		log.Error("msg", "could not parse flags", "error", err)
+		os.Exit(1)
+	}
+
+	planner, proceed, err := plan.CreatePlan(conf.mint, conf.maxt, conf.progressMetricName, conf.name, conf.writerReadURL, conf.progressEnabled, false)
```
blagojts, line 40: why not pass the config struct as an argument?

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,107 @@
...
+				break
+			default:
+			}
+			blockRef, err, timeRangeMinutesDelta = rr.plan.NextBlock()
+			if err != nil {
+				errChan <- fmt.Errorf("remote-run run: %w", err)
+				return
+			}
+			blockRef.SetDescription(fmt.Sprintf("fetching time-range: %d mins...", timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+			result, bytesCompressed, bytesUncompressed, err = blockRef.Fetch(rr.c, rr.client, blockRef.Mint(), blockRef.Maxt(), []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")})
```
blagojts, line 83: too long didn't read

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,114 @@
...
+		log.Error("msg", "could not create plan", "error", err)
+		os.Exit(2)
+	}
+	if !proceed {
+		os.Exit(0)
+	}
+
+	var (
+		sigBlockWrite = make(chan struct{})
+		sigBlockRead  = make(chan *plan.Block)
```
blagojts, line 51: I'm still not clear why sigBlockWrite is needed

Harkishen-Singh: Now this is removed. Meaning, we no longer need sigBlockRead. Thank you.

cmd/prom-migrator/main.go
Harkishen-Singh: Yeah, we can. Thank you for pointing out. Will implement in next commits.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Sorry, couldn't get this.

Harkishen-Singh:
Should each file in this PR contain the license header text?

pkg/migration-tool/reader/reader.go
blagojts: too long lines with multiple expressions in them are unreadable, we should avoid them

pkg/migration-tool/writer/writer.go
```diff
@@ -0,0 +1,168 @@
...
+			case <-rw.c.Done():
+				break
+			case blockRef, ok := <-rw.sigBlockRead:
+				if !ok {
+					return
+				}
+				blockRef.SetDescription(fmt.Sprintf("pushing %.2f...", float64(blockRef.BytesCompressed())/float64(utils.Megabyte)), 1)
+				rw.progressTimeSeries.Samples = []prompb.Sample{{Timestamp: blockRef.Maxt(), Value: 1}} // One sample per block, else it will lead to duplicate samples.
+				blockRef.Timeseries = append(blockRef.Timeseries, rw.progressTimeSeries)
+				ts = timeseriesRefToTimeseries(blockRef.Timeseries)
```
cevian, line 80: Can we encapsulate all this logic inside Block? Seems like there should be a function
```
ts := blockRef.GetTimeSeries(some arguments)
```
The arguments could be rw.ProgressTimeSeries. Then that function can set the description, set the sample in the progressTimeSeries, append the progress, convert to timeseries and return.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+			6,
+			progressbar.OptionOnCompletion(func() {
+				_, _ = fmt.Fprint(os.Stderr, "\n")
+			}),
+		),
+		mint:    mint,
+		maxt:    maxt,
+		pbarMux: p.pbarMux,
+	}
+	if p.IsTest {
```
cevian, line 177: why do we need this branch?

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// Block represents an in-memory storage for data that is fetched by the reader.
+type Block struct {
+	id                   int64
+	mint                 int64
+	maxt                 int64
+	done                 bool
+	percent              float64
+	pbarInitDetails      string
```
cevian, line 21:
```suggestion
	pbarDescriptionPrefix      string
```

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,129 @@
...
+	planner := &plan.Plan{
+		Mint:    conf.mint,
+		Maxt:    conf.maxt,
+		JobName: conf.name,
+		// Progress metric configs.
+		ProgressMetricName:        conf.progressMetricName,
+		ProgressEnabled:           conf.progressEnabled,
+		RemoteWriteStorageReadURL: conf.writerReadURL,
+	}
+	proceed, err := plan.Init(planner)
```
cevian, line 49: We should pass the config object here instead of a Plan. This will allow us to avoid field in Plan that aren't needed past init (e.g. RemoteWriteStorageReadURL, ProgressEnabled)

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+// Plan represents the plannings done by the planner.
+type Plan struct {
+	Mint                      int64
+	Maxt                      int64
+	JobName                   string
+	ProgressMetricName        string // Name for progress metric.
+	ProgressEnabled           bool
+	RemoteWriteStorageReadURL string
+	readClient                *utils.Client // Used to fetch last maxt pushed from write storage.
+	blockCounts               atomic.Int64  // Used in maintaining the ID of the in-memory blocks.
```
cevian, line 33: I can't find this being used anywhere anymore

pkg/migration-tool/utils/client.go
```diff
@@ -0,0 +1,239 @@
...
+	return &Client{
+		remoteName: remoteName,
+		url:        conf.URL,
+		Client:     httpClient,
+		timeout:    time.Duration(conf.Timeout),
+	}, nil
+}
+
+// Read reads from a remote endpoint. It returns the response size of compressed and uncompressed in bytes.
+func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, int, int, error) {
```
cevian, line 64: can we give names to return arguments so easier to read what's happening from declaration

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,94 @@
...
+	LabelJob = "job"
+)
+
+// CreateReadClient creates a new read client that can be used to fetch promb samples.
+func CreateReadClient(name, urlString string, readTimeout model.Duration) (*Client, error) {
+	parsedUrl, err := url.Parse(urlString)
+	if err != nil {
+		return nil, fmt.Errorf("parsing-url: %w", err)
+	}
+	readClient, err := NewClient(name, "read", &clientConfig{
```
cevian, line 24: why don't we change NewClient to have signature NewClient(remoteName, clientType, urlString, timeout) that way we can avoid having CreateReadClient/CreateWriteClient because NewClient will be easy to use. It's a bit weird to have CreateReadClient return a client that can both read and write.

cevian (changes requested):
This is GREAT work. Really love all the tests. Code looks mostly good. I suggested some structural changes:
- encapsulate a lot more logic within Block,
- for Plan, think carefully about what code/logic we can do in Init/New and avoid putting state related to that in Object once it's created.

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,106 @@
...
+const defaultReadTimeout = time.Minute * 5
+
+type RemoteRead struct {
+	c               context.Context
+	url             string
+	client          *utils.Client
+	plan            *plan.Plan
+	fetchersRunning atomic.Uint32
+	sigBlockRead    chan *plan.Block // To the writer.
+	SigForceStop    chan struct{}
```
blagojts, line 26: I can't find the code that feeds this channel

blagojts (approved):
Looks good, only left one question regarding the halting channel

pkg/migration-tool/planner/planner.go
Harkishen-Singh: It is actually used in the integration tests. Used to say how many blocks are sent by the read remote storage which actually means the number of samples in the progress metric.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: This is used inside the integration tests so as to introduce halts while migration. For introducing any halt, just stop the reader that that's where this channel finds its use-case.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+	lastNumBytes       int
+	lastTimeRangeDelta int64
+	pbarMux            *sync.Mutex
+	IsTest             bool
+}
+
+// InitPlan creates an in-memory planner nad initializes it. It is responsible for fetching the last pushed maxt and based on that, updates
+// the mint for the provided migration.
+func Init(config *Plan) (bool, error) {
+	rc, err := utils.CreateReadClient("reader-last-maxt-pushed", config.RemoteWriteStorageReadURL, model.Duration(time.Minute*2))
```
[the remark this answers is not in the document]
Harkishen-Singh: Thank you for pointing that out.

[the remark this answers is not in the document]
Harkishen-Singh: > That function could also close the client as it will no longer be using it.

i dont think the client has the capability of being closed.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+// NewBlock returns a new block after allocating the time-range for fetch.
+func (p *Plan) NextBlock() (reference *Block, err error, delta int64) {
+	timeDelta := determineTimeDelta(p.lastNumBytes, p.lastTimeRangeDelta)
+	var mint int64
+	if p.lastMaxT == math.MinInt64 {
+		mint = p.Mint
+	} else {
+		mint = p.lastMaxT + 1
+	}
+	maxt := mint + timeDelta
```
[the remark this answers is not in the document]
Harkishen-Singh: Right now I can't figure out a way to get rid of it. maybe sleepy. But, we do need some state to start the migrator with `p.Mint` and thereafter, things will be handled by `NextBlock()` itself.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+		pbarInitDetails: baseDescription,
+		pbar: progressbar.NewOptions(
+			6,
+			progressbar.OptionOnCompletion(func() {
+				_, _ = fmt.Fprint(os.Stderr, "\n")
+			}),
+		),
+		mint:    mint,
+		maxt:    maxt,
+		pbarMux: p.pbarMux,
```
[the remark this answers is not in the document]
Harkishen-Singh: We need to have a global mutex so that between progress-bars, we can communicate. We also need to make this golbal mux reachable to each block since the actual locking happens in the SetDescription(). Hence, the design.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,227 @@
...
+		// unchanged so that it can be reset back to the MaxTimeRangeDeltaLimit.
+	case numBytes < ResponseDataSizeHalfLimit:
+		// We increase the time-range linearly for the next fetch if the current time-range fetch resulted in size that is
+		// less than half the aimed maximum size of the in-memory block. This continues till we reach the
+		// maximum time-range delta.
+		prevTimeDelta += time.Minute.Milliseconds()
+	default:
+		// Decrease the time-range delta exponentially if the current fetch size exceeds the half aimed. This is done so that
+		// we reach an ideal time-range as the block numbers grow and successively reach the balanced scenario.
+		return prevTimeDelta / 2
```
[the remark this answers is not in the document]
Harkishen-Singh: The reason its half of it is that at any point, we have 2 blocks in memory. That means, 500MB is the assumed prom-migrator size as approx and hence we use the half of it to control the memory.

pkg/migration-tool/reader/reader.go
```diff
@@ -0,0 +1,106 @@
...
+				continue
+			}
+			blockRef.SetDescription(fmt.Sprintf("received %.2f MB with delta %d mins...", float64(bytesCompressed)/float64(utils.Megabyte), timeRangeMinutesDelta/time.Minute.Milliseconds()), 1)
+			blockRef.Timeseries = result.Timeseries
+			// We set compressed bytes in block since those are the bytes that will be pushed over the network to the write storage after snappy compression.
+			// The pushed bytes are not exactly the bytesCompressed since while pushing, we add the progress metric. But,
+			// the size of progress metric along with the sample is negligible. So, it is safe to consider bytesCompressed
+			// in such a scenario.
+			blockRef.SetBytesCompressed(bytesCompressed)
+			blockRef.SetBytesUncompressed(bytesUncompressed)
```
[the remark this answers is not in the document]
Harkishen-Singh: Ah, sure. We just need BytesUnCompressed here as thats whats the input to the planner.

pkg/migration-tool/utils/client.go
```diff
@@ -0,0 +1,239 @@
...
+	if len(resp.Results) != len(req.Queries) {
+		return nil, -1, -1, errors.Errorf("responses: want %d, got %d", len(req.Queries), len(resp.Results))
+	}
+
+	return resp.Results[0], len(compressed), len(uncompressed), nil
+}
+
+// ReadMultiple reads from a remote endpoint. It returns the response size of compressed and uncompressed in bytes.
+// ReadMultiple is an extension of Read() and will be helpful for those storages
+// that do not support ""=~".*". Example: Cortex.
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, this is for next version. I will delete it. Thank you.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: If we do not do this, then our logs will be filled with millions of lines of progress bars. You can visit the previous commits and see how the logs were :cold_sweat:

pkg/migration-tool/utils/utils.go
Harkishen-Singh: Sure. Updated and hence resolving

cmd/prom-migrator/main.go
Harkishen-Singh: The reason why I dont want to pass config object to plan is that auth (being added in upcoming commits) related stuff has nothing to do in the plan.

cmd/prom-migrator/main.go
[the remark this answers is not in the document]
Harkishen-Singh: Specified as `(in seconds)`

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: Now it is.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: `p.IsTest` = >`p.quite`

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: No `+1` as some remote storages leave the maxt for chunk formation.

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: Implemented the way we discussed.

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: tried to Implement as per discussion. Does it look good now?

cmd/prom-migrator/main.go
Harkishen-Singh: I hope the recent implementation is in line with what you wanted.

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
Harkishen-Singh: Changed this too.

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+type Config struct {
+	ProgressEnabled    bool
+	JobName            string
+	ProgressMetricURL  string
+	ProgressMetricName string // Name for progress metric.
+}
+
+// InitPlan creates an in-memory planner and initializes it. It is responsible for fetching the last pushed maxt and based on that, updates
+// the mint for the provided migration.
+func Init(plan *Plan) (bool, error) {
```
cevian, line 52: Config should include Mint and Maxt and Init should take the config object and return a *Plan
```suggestion
func Init(config *Config) (bool, error) {
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+		if err != nil {
+			return false, fmt.Errorf("create plan: %w", err)
+		}
+		if found && lastPushedMaxt > plan.Mint && lastPushedMaxt <= plan.Maxt {
+			plan.Mint = lastPushedMaxt
+			log.Warn("msg", fmt.Sprintf("Resuming from where we left off. Last push was on %d. "+
+				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, plan.Mint, plan.Maxt, (plan.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
+		}
+	}
+	plan.nextMint = plan.Mint
```
cevian, line 73: Let's collect all the progressEnabled code under one if
```suggestion
	if plan.Config.ProgressEnabled { 
	    if plan.Config.ProgressMetricURL == "" {
		return false, fmt.Errorf("read url for remote-write storage should be provided when progress metric is enabled"
             }
		log.Info("msg", "Resuming from where we left off is turned off. Starting at the beginning of the provided time-range.")
	    var found bool
	    lastPushedMaxt, found, err := plan.fetchLastPushedMaxt()
	    if err != nil {
	       return false, fmt.Errorf("create plan: %w", err)
	    }
	    if found && lastPushedMaxt > plan.Mint && lastPushedMaxt <= plan.Maxt {
		plan.Mint = lastPushedMaxt
		log.Warn("msg", fmt.Sprintf("Resuming from where we left off. Last push was on %d. "+
				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, plan.Mint, plan.Maxt, (plan.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
		
	}
	plan.nextMint = plan.Mint
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+
+func determineTimeDelta(numBytes, limit, deltaIncRegion int64, prevTimeDelta int64) int64 {
+	switch {
+	case numBytes <= deltaIncRegion:
+		// We increase the time-range linearly for the next fetch if the current time-range fetch resulted in size that is
+		// less than half the aimed maximum size of the in-memory block. This continues till we reach the
+		// maximum time-range delta.
+		if prevTimeDelta >= maxTimeRangeDeltaLimit {
+			return maxTimeRangeDeltaLimit
+		}
```
cevian, line 216: Can delete this because its handled by the clamp code
```suggestion

```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+		return fmt.Errorf("invalid maxt: %d: global-mint: %d and global-maxt: %d", mint, p.Mint, p.Maxt)
+	case mint > maxt:
+		return fmt.Errorf("mint cannot be greater than maxt: mint: %d and maxt: %d", mint, maxt)
+	}
+	return nil
+}
+
+func determineTimeDelta(numBytes, limit, deltaIncRegion int64, prevTimeDelta int64) int64 {
+	switch {
+	case numBytes <= deltaIncRegion:
```
cevian, line 210: I think defining the inc region inside this function rather than during init makes the logic more clear and self-contained. but it's a matter of taste
```suggestion
	case numBytes <= limit / 2:
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+
+// NewBlock returns a new block after allocating the time-range for fetch.
+func (p *Plan) NextBlock() (reference *Block, err error) {
+	timeDelta := determineTimeDelta(p.lastNumBytes.Load(), p.BlockSizeLimitBytes, p.deltaIncRegion, p.lastTimeRangeDelta)
+	mint := p.nextMint
+	maxt := mint + timeDelta
+	if maxt > p.Maxt {
+		maxt = p.Maxt
+	}
+	p.nextMint = maxt + 1
```
cevian, line 147: I don't think there should be a +1 here, if maxt is excusive

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (b *Block) SetDescription(description string, proceed int) {
+	b.pbarMux.Lock()
+	defer b.pbarMux.Unlock()
+	if b.pbarDescriptionPrefix == "" {
+		return
+	}
+	_ = b.pbar.Add(proceed)
+	b.pbar.Describe(fmt.Sprintf("progress: %.3f%% | %s | %s", b.percent, b.pbarDescriptionPrefix, description))
```
cevian, line 71: since b.percent never changes can we make this part of DescriptionPrefix and get rid of the percent field on Block?

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+	return prevTimeDelta
+}
+
+func clampTimeDelta(t int64, incr int64) int64 {
+	t += incr
+	if t > maxTimeRangeDeltaLimit {
+		return maxTimeRangeDeltaLimit
+	}
+	return t
+}
```
cevian, line 239: nit : a clamp function is traditionally written as the following and the math is written on input
```suggestion
func clampTimeDelta(t int64) int64 {
	if t > maxTimeRangeDeltaLimit {
		return maxTimeRangeDeltaLimit
	}
	return t
}
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,239 @@
...
+func determineTimeDelta(numBytes, limit, deltaIncRegion int64, prevTimeDelta int64) int64 {
+	switch {
+	case numBytes <= deltaIncRegion:
+		// We increase the time-range linearly for the next fetch if the current time-range fetch resulted in size that is
+		// less than half the aimed maximum size of the in-memory block. This continues till we reach the
+		// maximum time-range delta.
+		if prevTimeDelta >= maxTimeRangeDeltaLimit {
+			return maxTimeRangeDeltaLimit
+		}
+		return clampTimeDelta(prevTimeDelta, laIncrement)
```
cevian, line 217: nit: see comment on clamp function
```suggestion
		return clampTimeDelta(prevTimeDelta + laIncrement)
```

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	log.Info("msg", "migration successfully carried out")
+	log.Info("msg", "exiting!")
+}
+
+func parseFlags(conf *config, args []string) {
+	flag.StringVar(&conf.name, "progress-job-name", migrationJobName, "Name for the current migration that is to be carried out. It corresponds to the value of the label 'job' set inside the progress-metric-name.")
+	flag.Int64Var(&conf.mint, "mint", 0, "Minimum timestamp (in seconds) for carrying out data migration.")
```
cevian, line 128:
```suggestion
	flag.Int64Var(&conf.mint, "mint", 0, "Minimum timestamp (in seconds) for carrying out data migration. (inclusive)")
```

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,195 @@
...
+	flag.Int64Var(&conf.maxt, "maxt", time.Now().Unix(), "Maximum timestamp (in seconds) for carrying out data migration. Setting this value less than zero will indicate all data from mint upto now. "+
```
cevian, line 129:
```suggestion
	flag.Int64Var(&conf.maxt, "maxt", time.Now().Unix(), "Maximum timestamp (in seconds) for carrying out data migration (exclusive). Setting this value less than zero will indicate all data from mint upto now. "+
```

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,195 @@
...
+		"carry out migration with the same time-range. Without this, the migrator will resume the migration from the last time, where it as stopped.")
+	// Authentication.
+	// TODO: Auth/password via password_file and bearer_token via bearer_token_file.
+	flag.StringVar(&conf.readerAuth.Username, "read-auth-username", "", "Auth username for remote-read storage.")
+	flag.StringVar(&conf.readerAuth.Password, "read-auth-password", "", "Auth password for remote-read storage.")
+	flag.StringVar(&conf.readerAuth.BearerToken, "read-auth-bearer-token", "", "Auth username for remote-read storage. This should be mutually exclusive with username and password.")
+	flag.StringVar(&conf.writerAuth.Username, "write-auth-username", "", "Auth username for remote-write storage.")
+	flag.StringVar(&conf.writerAuth.Password, "write-auth-password", "", "Auth password for remote-write storage.")
+	flag.StringVar(&conf.writerAuth.BearerToken, "write-auth-bearer-token", "", "Auth username for remote-write storage. This should be mutually exclusive with username and password.")
```
cevian, line 146: where are we validating that this is mutually exclusive?

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,92 @@
...
+
+// CreatePrombRequest creates a new promb query based on the matchers.
+func CreatePrombQuery(mint, maxt int64, matchers []*labels.Matcher) (*prompb.Query, error) {
+	ms, err := toLabelMatchers(matchers)
+	if err != nil {
+		return nil, fmt.Errorf("fetch: %w", err)
+	}
+	return &prompb.Query{
+		StartTimestampMs: mint,
+		EndTimestampMs:   maxt,
```
cevian, line 56: if we are sticking to the idea that maxt is exclusive  then I think this should be 
```suggestion
		EndTimestampMs:   maxt-1,
```

cevian (changes requested):
Only major feedback is if Mint and Maxt are inclusive or exclusive. I've added suggestions with the idea that Mint is inclusive and MaxT is exclusive as that seems to be where prometheus is going. We should double check that all the code follows this convention

pkg/migration-tool/planner/planner.go
Harkishen-Singh: I think the reason for this is than, to progress the next block timestamps. If I don't add +1, then it wil not proceed.

cmd/prom-migrator/main.go
Harkishen-Singh: The validate() checks for mutual exclusion.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: Ah, this is much better. Thank you very much.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: I agree to this. It makes more cleanness and less vars at the plan level. Thank you :smile:

pkg/migration-tool/planner/block.go
Harkishen-Singh: Yes, that would be very good.

cmd/prom-migrator/main.go
Harkishen-Singh:
```
if err := httpConfig.Validate();
```
this line is responsible for it.

pkg/migration-tool/utils/utils.go
Harkishen-Singh: I am not sure, as the read-storage by itself will skip the maxt. If we do this, then the time to be considered as end will actually be `maxt-2` which is wrong.

cmd/prom-migrator/main.go
Harkishen-Singh: I have added tests for this in `main_test.go`. Please look for tests below the comment `// mutual exclusive tests` in that file.

pkg/migration-tool/utils/utils.go
Harkishen-Singh: Ah, i get the point. yes, it should be that way.

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,194 @@
...
+		"This should be mutually exclusive with username and password.")
+	flag.StringVar(&conf.writerAuth.Username, "write-auth-username", "", "Auth username for remote-write storage.")
+	flag.StringVar(&conf.writerAuth.Password, "write-auth-password", "", "Auth password for remote-write storage.")
+	flag.StringVar(&conf.writerAuth.BearerToken, "write-auth-bearer-token", "", "Auth username for remote-write storage. "+
+		"This should be mutually exclusive with username and password.")
+	_ = flag.CommandLine.Parse(args)
+	optimizeConf(conf)
+}
+
+func optimizeConf(conf *config) {
```
cevian, line 150:
```suggestion
func convertSecFlagToMs(conf *config) {
```

cmd/prom-migrator/main.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	log.Info("msg", "migration successfully carried out")
+	log.Info("msg", "exiting!")
+}
+
+func parseFlags(conf *config, args []string) {
+	flag.StringVar(&conf.name, "migration-name", migrationJobName, "Name for the current migration that is to be carried out. "+
+		"It corresponds to the value of the label 'job' set inside the progress-metric-name.")
+	flag.Int64Var(&conf.mint, "mint", 0, "Minimum timestamp (in seconds) for carrying out data migration. (inclusive)")
```
cevian, line 122: I would prefer these flags in Seconds be different conf variables than the mint fields and the convert code fill in the mint fields. Overwritting a var like that is a recipe for bugs down the line
```suggestion
	flag.Int64Var(&conf.mintSec, "mint", 0, "Minimum timestamp (in seconds) for carrying out data migration. (inclusive)")
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,242 @@
...
+		return clampTimeDelta(prevTimeDelta + laIncrement)
+	case numBytes > limit:
+		// The priority of size is more than that of time. That means, even if both bytes and time are greater than
+		// the respective limits, then we down size the time so that bytes size can be controlled (preventing OOM).
+		log.Info("msg", fmt.Sprintf("decreasing time-range delta to %d minute(s) since size beyond permittable limits", prevTimeDelta/(2*minute)))
+		return prevTimeDelta / 2
+	}
+	// Here, the numBytes is between the deltaIncRegion and the limit (i.e., deltaIncRegion < numBytes <= limit).
+	// This region is an ideal case of balance between migration speed and memory utilization.
+	// Example: If the limit is 500MB, then the deltaIncRegion will be 250MB. This means that till the numBytes is below
```
cevian, line 226: need to update the comments in this entire function

pkg/migration-tool/utils/utils.go
```diff
@@ -0,0 +1,98 @@
...
+
+// CreatePrombRequest creates a new promb query based on the matchers.
+func CreatePrombQuery(mint, maxt int64, matchers []*labels.Matcher) (*prompb.Query, error) {
+	ms, err := toLabelMatchers(matchers)
+	if err != nil {
+		return nil, fmt.Errorf("fetch: %w", err)
+	}
+	return &prompb.Query{
+		StartTimestampMs: mint,
+		EndTimestampMs:   maxt - 1,
```
cevian, line 62:
```suggestion
		EndTimestampMs:   maxt - 1, //maxt is exclusive while EndTimestampMs is inclusive
```

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/schollz/progressbar/v3"
+	"github.com/timescale/promscale/pkg/migration-tool/utils"
+)
+
+// Block represents an in-memory storage for data that is fetched by the reader.
+type Block struct {
+	id                    int64
+	mint                  int64
+	maxt                  int64
```
cevian, line 19:
```suggestion
	mint                  int64 //inclusive
	maxt                  int64 //exclusive
```

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,118 @@
...
+		1,
+	)
+	b.done = true
+	if err := b.pbar.Finish(); err != nil {
+		return fmt.Errorf("finish block-lifecycle: %w", err)
+	}
+	return nil
+}
+
+// Mint returns the mint of the block.
```
cevian, line 110:
```suggestion
// Mint returns the mint of the block (inclusive).
```

pkg/migration-tool/planner/block.go
```diff
@@ -0,0 +1,118 @@
...
+func (b *Block) Mint() int64 {
+	return b.mint
+}
+
+// Maxt returns the maxt of the block.
```
cevian, line 115:
```suggestion
// Maxt returns the maxt of the block (exclusive).
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,238 @@
...
+		if found && lastPushedMaxt > config.Mint && lastPushedMaxt <= config.Maxt {
+			config.Mint = lastPushedMaxt
+			log.Warn("msg", fmt.Sprintf("Resuming from where we left off. Last push was on %d. "+
+				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, config.Mint, config.Maxt, (config.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
+		}
+	} else {
+		log.Info("msg", "Resuming from where we left off is turned off. Starting at the beginning of the provided time-range.")
+	}
+	if config.Mint >= config.Maxt && found {
+		log.Info("msg", "mint greater than or equal to maxt. Migration has already been carried out.")
```
cevian, line 71:
```suggestion
		log.Info("msg", "mint greater than or equal to maxt. Migration is already complete.")
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,238 @@
...
+	id := p.blockCounts.Add(1)
+	timeRangeInMinutes := (maxt - mint) / time.Minute.Milliseconds()
+	percent := float64(maxt-p.config.Mint) * 100 / float64(p.config.Maxt-p.config.Mint)
+	if percent > 100 {
+		percent = 100
+	}
+	baseDescription := fmt.Sprintf("progress: %.3f%% | block-%d time-range: %d mins | mint: %d | maxt: %d", percent, id, timeRangeInMinutes, mint, maxt)
+	reference = &Block{
+		id:                    id,
+		percent:               percent,
```
cevian, line 172: I think we can delete this field
```suggestion
		
```

pkg/migration-tool/planner/planner.go
```diff
@@ -0,0 +1,238 @@
...
+			progressbar.OptionOnCompletion(func() {
+				_, _ = fmt.Fprint(os.Stderr, "\n")
+			}),
+		),
+		mint:    mint,
+		maxt:    maxt,
+		pbarMux: p.pbarMux,
+		plan:    p,
+	}
+	if p.Quite {
```
cevian, line 185: Misspelled variable name
```suggestion
	if p.Quiet {
```

cevian (approved):
Requesting some nits fixed, but nothing critical. Would also like to see the commit history cleaned up before merge

cevian (approved):

PR #350, opened 2020-11-16
Build EL8 RPM from spec file

blagojts:
The rpm build process is integrated with the goreleaser config
to make sure the same version is used and the built artifact
is uploaded with the others together


BuildRPMDockerfile
```diff
@@ -0,0 +1,16 @@
+# Go binary builder
+FROM promscale_builder AS builder
```
blagojts, line 2: `build_rpm.sh` creates the `promscale_builder` image from `Dockerfile` the same file used to build the release image

BuildRPMDockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+# RPM builder
+FROM centos:8
+ARG VERSION
+COPY --from=builder /promscale-$VERSION.tgz /root/rpmbuild/SOURCES/
+COPY promscale.spec promscale.spec
+RUN yum -y install rpm-build
+RUN sed -i "s/_VERSION_/$VERSION/g" promscale.spec
+RUN rpmbuild -ba promscale.spec
+ENTRYPOINT ["tail", "-f", "/dev/null"]
```
blagojts, line 16: built RPM will be extracted from the container by `build_rpm.sh`

.goreleaser.yml
```diff
@@ -4,13 +4,21 @@
...
+
 builds:
 - env:
   - CGO_ENABLED=0
   main: ./cmd/promscale/
+  hooks:
+    # build el8 rpm
+    # will be triggered multiple times but we need
+    # {{ .Tag }} to ensure same version is used
+    post: bash build_rpm.sh {{ .Tag }}
```
blagojts, line 21: Need to call `build_rpm.sh` from builds.hooks so we have access to {{ .Tag }} and make sure the same version is used

.goreleaser.yml
```diff
@@ -4,13 +4,21 @@
...
     #make sure generate up to date
     - go generate ./...
+
 builds:
 - env:
   - CGO_ENABLED=0
   main: ./cmd/promscale/
+  hooks:
+    # build el8 rpm
+    # will be triggered multiple times but we need
```
cevian, line 19: why is this triggered many tiimes?

promscale.spec
```diff
@@ -0,0 +1,132 @@
...
+
+cat > %{buildroot}%{_sysconfdir}/sysconfig/%{name} <<- EOT
+# TS_PROM_ASYNC_ACKS=""
+# TS_PROM_DB_CONNECT_RETRIES=""
+# TS_PROM_DB_CONNECTIONS_MAX="-1"
+# TS_PROM_DB_HOST="localhost"
+# TS_PROM_DB_NAME="timescale"
+# TS_PROM_DB_PASSWORD=""
+# TS_PROM_DB_PORT="5432"
+TS_PROM_DB_SSL_MODE="disable"
```
cevian, line 80:
```suggestion
TS_PROM_DB_SSL_MODE="prefer"
```

.goreleaser.yml
blagojts: each build for each os triggers the hooks

antekresic (approved):
This looks good to me. Maybe we could get somebody else that does this kind of builds like Rob?

Other than that, maybe we could organize these files into a separate `build` directory so we don't clutter the root folder?

BuildRPMDockerfile
cevian: let's add this as a comment in the file

BuildRPMDockerfile
cevian: let's add this as code comment in the code

cevian (approved):
Looks good but agree with Ante to put this in a /build/rpm/ sub directory.

PR #354, opened 2020-11-18
Add Makefile

VineethReddy02:
For easier local build, tests, go formatting & go linting


.gitignore
```diff
@@ -8,3 +8,4 @@
 /.idea
 .DS_Store
 .git
+/bin
```
cevian, line 11: is putting binaries under /bin common in go projects? Is this a known convention?

cevian (commented):
One question. Otherwise looks good

.gitignore
VineethReddy02: I personally prefer to have it under the bin directory, This makes running the built binary easier. i.e ```./bin/promscale```

As I added ```make build``` as ```go build -o bin/promscale ./cmd/promscale``` to avoid git level spam on bin directory added it to `.gitignore` 

But I am open to thoughts/suggestions.

cevian:
@blagojts can you also review this?

.gitignore
blagojts: The releaser we use puts everything in a dist folder, maybe we should use that so it's unified

blagojts (approved):
Looks ok, other than a preference to have the built artifacts be stored in /dist instead of /bin (because our other build config stores them there) the other changes look good.

PR #355, opened 2020-11-18
Enable compression for multinode

JLockerman:
This requires a custom compression job and some other workarounds for current multinode limitations.


pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -144,6 +141,26 @@ COMMENT ON PROCEDURE SCHEMA_CATALOG.finalize_metric_creation()
...
+    tables TEXT[];
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() AND SCHEMA_CATALOG.is_multinode() THEN
+        SELECT array_agg(format('%I', table_name))
+        FROM SCHEMA_CATALOG.metric
+        WHERE NOT creation_completed
+        INTO tables;
+    END IF;
+
+    RETURN tables;
```
cevian, line 160: is this guaranteed empty at initialization?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1078,9 +1097,27 @@ BEGIN
 
         --chunks where the end time is before now()-1 hour will be compressed
         IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
-            PERFORM add_compression_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+            IF SCHEMA_CATALOG.is_multinode() THEN
```
cevian, line 1100: should we only do this for multinode? My feeling is do this for all of 2.0

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1687,12 +1724,20 @@ DECLARE
     bgw_job_id int;
 BEGIN
     IF SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
-        SELECT p.id INTO bgw_job_id
-        FROM _timescaledb_config.bgw_job p
-        INNER JOIN _timescaledb_catalog.hypertable h ON (h.id = p.hypertable_id)
-        WHERE p.proc_name = 'policy_compression'
-          AND h.schema_name = 'SCHEMA_DATA'
-          AND h.table_name = ht_table;
+        IF SCHEMA_CATALOG.is_multinode() THEN
```
cevian, line 1727: again I don't think just multinode

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1712,48 +1757,102 @@ END
...
+AS $proc$
+BEGIN
+    -- in early versions of timescale multinode the access node catalog does not
+    -- store whether chunks were compressed, so we need to run the actual search
+    -- for nodes in need of decompression on the data nodes
+    IF SCHEMA_CATALOG.is_multinode() THEN
+        CALL distributed_exec(
+            format(
+                $dist$ CALL do_decompress_chunks_after(%s, %s) $dist$,
+                quote_literal(metric_table), quote_literal(min_time)),
```
cevian, line 1818:
```suggestion
                $dist$ CALL do_decompress_chunks_after(%L, %L) $dist$,
                metric_table, min_time),
```

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1712,48 +1757,102 @@ END
...
+            transactional => false);
+    ELSE
+        CALL do_decompress_chunks_after(metric_table, min_time);
+    END IF;
+END
+$proc$ LANGUAGE PLPGSQL;
+
+CREATE OR REPLACE PROCEDURE SCHEMA_CATALOG.compress_old_chunks(metric_table TEXT, last_time TIMESTAMPTZ)
```
cevian, line 1826:
```suggestion
CREATE OR REPLACE PROCEDURE SCHEMA_CATALOG.compress_old_chunks(metric_table TEXT, compress_before TIMESTAMPTZ)
```

cevian (changes requested):
Some issues. Also I think we are missing tests if things are passing with these issues

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: I had a specific reason I didn't want to do that, can I remember what it was...

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: let's merge this then throw the switch

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: I believe it should be NULL at init

pkg/pgmodel/migrations/sql/idempotent/base.sql
cevian: why are we waitting?

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1078,9 +1097,27 @@ BEGIN
...
+                    'SCHEMA_CATALOG.compression_job'::regproc,
+                    INTERVAL '1 hour',
+                    config => jsonb_build_object('metric_table', metric_table_name)
+                );
+            ELSE
+                PERFORM add_compression_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+            END IF;
         ELSE
-            PERFORM add_compress_chunks_policy(format('SCHEMA_DATA.%I', metric_table_name), INTERVAL '1 hour');
+            IF SCHEMA_CATALOG.is_multinode() THEN
```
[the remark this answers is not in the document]
cevian: Was a test added here?

pkg/pgmodel/migrations/sql/idempotent/base.sql
cevian: well let's change it then unless there is a reason

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1712,48 +1755,104 @@ END
...
-        COMMIT;
-    END LOOP;
-END;
-$$ LANGUAGE PLPGSQL;
+CREATE OR REPLACE PROCEDURE SCHEMA_CATALOG.compression_job(job_id int, config jsonb)
+AS $$
+    -- as of timescaledb-2.0-rc3 the is_compressed column of the chunks view is
+    -- not updated on the access node, therefore we need to one the compressor
+    -- on all the datanodes to search for uncompressed chunks
+    CALL distributed_exec(format($dist$
```
cevian, line 1855: does this work in a non-multinode setting. If it does, then why do we need the if on line 1812? I think the code should be congruent

pkg/pgmodel/migrations/sql/idempotent/base.sql
```diff
@@ -1712,48 +1755,104 @@ END
...
+        END LOOP;
+    END;
+    $$ LANGUAGE PLPGSQL;
+$ee$);
 
-        COMMIT;
-    END LOOP;
-END;
-$$ LANGUAGE PLPGSQL;
+CREATE OR REPLACE PROCEDURE SCHEMA_CATALOG.compression_job(job_id int, config jsonb)
```
cevian, line 1850: let's be a bit more explicit in the naming so we don't clash with something later.
```suggestion
CREATE OR REPLACE PROCEDURE SCHEMA_CATALOG.promscale_compression_job(job_id int, config jsonb)
```

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -621,14 +615,18 @@ func TestInsertCompressed(t *testing.T) {
 		}
 		var nextStartAfter time.Time
 		var statsQuery string
-		if *useTimescale2 {
-			statsQuery = "SELECT next_start FROM timescaledb_information.job_stats WHERE hypertable_schema=$1::text AND hypertable_name = $2::text"
-			err = db.QueryRow(context.Background(), statsQuery, "prom_data", "test").Scan(&nextStartAfter)
+		if *useTimescale2 || *useMultinode {
```
cevian, line 618:
```suggestion
		if *useTimescale2 {
```

pkg/pgmodel/end_to_end_tests/create_test.go
```diff
@@ -621,14 +615,18 @@ func TestInsertCompressed(t *testing.T) {
...
+				"WHERE proc_schema = '_prom_catalog' " +
+				"AND proc_name = 'compression_job' " +
+				"AND config->>'metric_table' = $1::text"
+			err = db.QueryRow(context.Background(), statsQuery, "test").Scan(&nextStartAfter)
 		} else {
 			statsQuery = "SELECT next_start FROM timescaledb_information.policy_stats WHERE hypertable = $1::text::regclass"
 			err = db.QueryRow(context.Background(), statsQuery, pgx.Identifier{"prom_data", "test"}.Sanitize()).Scan(&nextStartAfter)
 		}
 		if err != nil {
+			time.Sleep(1 * time.Hour)
```
cevian, line 629:
```suggestion
```

cevian (changes requested):
Some more nits. Mostly looks good.

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: @cevian it's a nop without multinode.
I should switch it it `exec_everywhere`

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: remember, it's in `_prom_catalog` so the names are
`_prom_catalog.compression_job()` versus
`_prom_catalog. promscale_compression_job()`

pkg/pgmodel/migrations/sql/idempotent/base.sql
JLockerman: nevermind, it's an _error_ without multinode, and 1812 is right; we shouldn't run things on the access node if it's going to be a lot of work for no benefit

JLockerman: (it wasn't caught because the new test only ran for multinode, not all of 2. fixed now)

cevian (approved):
One more nit from last time around? Otherwise LGTM

pkg/pgmodel/end_to_end_tests/create_test.go
JLockerman: huh, I thought I fixed that...

PR #359, opened 2020-11-26
Fix issue when ingesting batches that span multiple compressed chunks

antekresic:
When backfilling larged amounts of data, which span over multiple compressed
chunks, decompress_chunks_after function was not decompressing all the
chunks necessary. It would decompress only a single oldest chunks which causes
issues when trying to ingest data to the rest of the chunks. This commit changes
this to decompress all the chunks after the oldest chunk as well.


davidkohn88 (approved):
LGTM

cevian (approved):

PR #361, opened 2020-11-30
Update promb to recent commits till 19 Nov

Harkishen-Singh:

This commit updates the pkg/promb with all the upstream changes till
4eca4dffb89d9da927f1e7884efe85cbd6dc30f3 which was on 19 Nov, 2020.
This commit mainly makes promb to be able to accept metric metadata.


pkg/prompb/remote.pb.go
```diff
@@ -946,6 +975,40 @@ func (m *WriteRequest) Unmarshal(dAtA []byte) error {
...
+				return ErrInvalidLengthRemote
+			}
+			postIndex := iNdEx + msglen
+			if postIndex < 0 {
+				return ErrInvalidLengthRemote
+			}
+			if postIndex > l {
+				return io.ErrUnexpectedEOF
+			}
+			m.Metadata = append(m.Metadata, MetricMetadata{})
```
cevian, line 1007: We should implement the same memory optimization as in https://github.com/timescale/promscale/commit/a565b9270db7ccb6a65a7c262e689834a57a17d1
```suggestion
			if len(m.Metadata) < cap(m.Metadata) {
				m.Metadata = m.Metadata[:len(m.Metadata)+1]
				m.Metadata[len(m.Metadata)-1].Reset()
			} else {
				m.Metadata = append(m.Metadata, MetricMetadata{})
			}
```

pkg/prompb/remote.pb.go
```diff
@@ -62,10 +62,11 @@ func (ReadRequest_ResponseType) EnumDescriptor() ([]byte, []int) {
 }
 
 type WriteRequest struct {
-	Timeseries           []TimeSeries `protobuf:"bytes,1,rep,name=timeseries,proto3" json:"timeseries"`
-	XXX_NoUnkeyedLiteral struct{}     `json:"-"`
-	XXX_unrecognized     []byte       `json:"-"`
-	XXX_sizecache        int32        `json:"-"`
+	Timeseries           []TimeSeries     `protobuf:"bytes,1,rep,name=timeseries,proto3" json:"timeseries"`
+	Metadata             []MetricMetadata `protobuf:"bytes,3,rep,name=metadata,proto3" json:"metadata"`
```
cevian, line 66: with the addition of this field we have to change custom.ts.go  to have 
```
func (m *WriteRequest) Reset() { *m = WriteRequest{Timeseries: m.Timeseries[:0], Metadata: m.Metadata[:0]} }

cevian (changes requested):
Some changes required to implement the memory efficiency method for Metadata.

Also, we should file an issue to store the metadata in our catalog, so we don't forget.

cevian (approved):

PR #362, opened 2020-11-30
Promscale manual deploy docs using pre-built binary.

VineethReddy02:
This PR contains steps to install promscale & crontab (for compressing chunks in TimescaleDB)


docs/binary.md
```diff
@@ -5,6 +5,26 @@
...
+chmod +x promscale
+```
+
+To deploy Promscale, run the following command:
+```
+./promscale --db-name <DBNAME> --db-password <DB-Password> --db-ssl-mode allow
+```
+Note that the flags `db-name` and `db-password` refer to the name and password of your TimescaleDB database.
+
+Further note that the command above is to deploy Promscale with SSL mode disabled. To deploy Promscale with SSL mode enabled, configure your TimescaleDB instance with ssl certificates and drop the `--db-ssl-mode` flag.  Promscale will then authenticate via SSL by default.
```
cevian, line 26:
```suggestion
Further note that the command above is to deploy Promscale with SSL mode being allowed but not required. To deploy Promscale with SSL mode enabled, configure your TimescaleDB instance with ssl certificates and drop the `--db-ssl-mode` flag.  Promscale will then authenticate via SSL by default.
```

docs/binary.md
```diff
@@ -19,6 +39,34 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
+
+```
+#/bin/sh
+export PGPASSWORD=<PASSWORD>
+psql --host localhost --port 5432 --dbname postgres --user postgres -c 'CALL prom_api.execute_maintenance();'
+```
+
+Create an other script with the below code snippet to the file `install-crontab.sh` This script will create a crontab for the above defined task:
+
+```
```
cevian, line 52: can we add this script to the github repo instead of as just docs?

docs/binary.md
```diff
@@ -19,6 +39,34 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
 30 minutes. This is necessary to execute maintenance tasks such as enforcing
 data retention policies according to the configured policy.
 
+Add the below code snippet to the file `prom-execute-maintenance.sh` and add the database password in place of `<PASSWORD>`:
+
+```
```
cevian, line 44: can we add this script to the /scripts subdir?

docs/binary.md
```diff
@@ -19,6 +39,34 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
+#/bin/sh
+export PGPASSWORD=<PASSWORD>
+psql --host localhost --port 5432 --dbname postgres --user postgres -c 'CALL prom_api.execute_maintenance();'
+```
+
+Create an other script with the below code snippet to the file `install-crontab.sh` This script will create a crontab for the above defined task:
+
+```
+#/bin/sh
+echo "0,30 * * * * ./prom-execute-maintenance.sh" | crontab -
```
cevian, line 54: hmm not sure where the prom-execute-maintenance.sh should be but I think crontab can't use relative subdirectories (`./`) like this.

cevian (changes requested):
These are good changes, just had a few suggested improvements and I think we have to think about the right dir to put the prom-execute-maintenance file into.

docs/binary.md
VineethReddy02: We will let the user add the absolute path to the script.

docs/binary.md
```diff
@@ -5,6 +5,26 @@
...
+
+To deploy Promscale, run the following command:
+```
+./promscale --db-name <DBNAME> --db-password <DB-Password> --db-ssl-mode allow
+```
+Note that the flags `db-name` and `db-password` refer to the name and password of your TimescaleDB database.
+
+Further note that the command above is to deploy Promscale with SSL mode being allowed but not required. To deploy Promscale with SSL mode enabled, configure your TimescaleDB instance with ssl certificates and drop the `--db-ssl-mode` flag.  Promscale will then authenticate via SSL by default.
+
 We recommend installing the [promscale](https://github.com/timescale/promscale_extension/releases)
```
Harkishen-Singh, line 28: I wonder how useful this is, given that promscale does that itself.

Harkishen-Singh (changes requested):
Also, maybe we can rename these docs to `getting-started.md` or so. `binary.md` is not that helpful for the users in terms of the name.

docs/binary.md
```diff
@@ -5,6 +5,26 @@
...
+chmod +x promscale
+```
+
+To deploy Promscale, run the following command:
+```
+./promscale --db-name <DBNAME> --db-password <DB-Password> --db-ssl-mode allow
+```
+Note that the flags `db-name` and `db-password` refer to the name and password of your TimescaleDB database.
+
+Further note that the command above is to deploy Promscale with SSL mode being allowed but not required. To deploy Promscale with SSL mode enabled, configure your TimescaleDB instance with ssl certificates and drop the `--db-ssl-mode` flag.  Promscale will then authenticate via SSL by default.
```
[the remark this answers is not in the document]
VineethReddy02: I assume we have two modes of auth here 1. with `ssl` 2. without `ssl` by default promscale tries to authenticate in SSL mode. If the user has installed timescaleDB with SSL certificates then default promscale installation should work. As here we are skipping SSL configuration in timescaleDB we are explicitly mentioning the significance of the `db-ssl-mode` flag. And the cli doc also refers to the same flag with the same info.

docs/binary.md
VineethReddy02: I agree with this. Maybe @cevian can answer this.

cevian: Promscale needs to be available in the database installation.

docs/getting-started.md
```diff
@@ -19,6 +39,34 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
 
+Add the below code snippet to the file `prom-execute-maintenance.sh` and add the database password in place of `<PASSWORD>`:
+
+```
+#/bin/sh
+export PGPASSWORD=<PASSWORD>
+psql --host localhost --port 5432 --dbname postgres --user postgres -c 'CALL prom_api.execute_maintenance();'
+```
+
+Create an other script with the below code snippet to the file install-crontab.sh and configure the absolute path to the above script. This script will create a crontab for the above defined task:
```
cevian, line 50: make sure to tell people to change the <ABSOLUTE PATH>

cevian (changes requested):

cevian:
@Harkishen-Singh getting_started.md isn't a great name here because this does not apply to all environments. This should not be used with docker, k8, helm, etc.

docs/getting-started.md
VineethReddy02: To grant exe permissions? `chmod +x <>.sh` ?

Harkishen-Singh:
@cevian `getting_started_bare_metal.md`?

VineethReddy02:
@Harkishen-Singh  the filename is too long & we need to rename it accordingly to all. I don't see an actual need in renaming the files. In most of the cases, users land on this page via a hyperlink from `README.md`

docs/binary.md
```diff
@@ -19,6 +39,23 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
+Create an other script with the code snippet from the file [docs/scripts/install-crontab.sh](/docs/scripts/install-crontab.sh) and make sure to configure the absolute path to `prom-execute-maintenance.sh` in the script. This script will create a crontab for the above defined task.
+
+Then, grant executable privileges to both files:
+
+```
+chmod +x prom-execute-maintenance.sh
+chmod +x install-crontab.sh
+```
+
+Running compress crontab:
```
cevian, line 53:
```suggestion
Install the cron job:
```

cevian (approved):

Harkishen-Singh (approved):

PR #365, opened 2020-12-02
Add write endpoint API documentation

antekresic:


docs/writing_to_promscale.md
```diff
@@ -0,0 +1,173 @@
...
+
+As you can see, the labels object is pretty self explanatory. Fields represent the label name and value is the label value. Samples array consists of arrays with some strict rules:
+* There can be only two values in the array
+* First value is a UNIX timestamp in milliseconds which has to be an integer value
+* Second value is a floating point number which represents the corresponding value of the metric.
+
+
+In order to send a request to Promscale, you would need to send an HTTP POST request with the request body set to the JSON payload and set the required header values:
+* `Content-Type` header should be set to `application/json`
+* `Content-Encoding` header can optionally be set to `snappy` if you are using snappy compression on the JSON payload
```
Harkishen-Singh, line 74: I think this header is optional, but the way this sounds is, **it can be optionally set to snappy** (else set to what?).

How about:
```
* (OPTIONAL, if using snappy)`Content-Encoding` header should be set to `snappy` if you are using snappy compression on the JSON payload
```

docs/writing_to_promscale.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+Here is an example of a request sent using `curl` tool:
+
+```
+curl --header "Content-Type: application/json" \
+--request POST \
+--data '{"labels":{"__name__":"foo"},"samples":[[1577836800000, 100]]}' \
+http://localhost:3000/api/login
+```
```
Harkishen-Singh, line 84: Can we have any one-liner example for snappy encoding and sending with the optional header here too?

docs/writing_to_promscale.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Protobuf format
+
+This format is a little more complex and involved to use than the JSON format but generally is a bit more performant because the payload is smaller in size. It is recommended that you use this format if you are building a system which would ingest a lot of data continuously.
```
Harkishen-Singh, line 88: To be strictly correct, should we say *more performant* or *more efficient*? Because performance in proto-bufs can decreases due to the involved Unmarshalling/decoding mechanism. That means, it's only efficient (in space) and not performant.

docs/writing_to_promscale.md
```diff
@@ -0,0 +1,173 @@
...
+
+In order to use the endpoint, there are a few steps that need to be done:
+* Fetch the protocol buffer definition files [here](https://github.com/prometheus/prometheus/blob/master/prompb/)
+* [Compile them](https://developers.google.com/protocol-buffers/docs/tutorials) in into structures of the programming language you intend to use
+* Finally, use those structures to construct requests which you can then send to the Promscale write endpoint
+Next section will show a simple example of how to make a request to Promscale using the Go programming language.
+
+
+## Protobuf write request example in Go
+
```
Harkishen-Singh, line 98: Not sure if it is done or not, but having another example that explains JSON way too would be very helpful for the users.

Harkishen-Singh (changes requested):
Should we also reflect these in the README.md? Maybe, its time to have a good TOC in the readme, that highlights our new feature additions.

README.md
```diff
@@ -50,6 +50,7 @@ If you have any questions, please join the #promscale channel on
   * [Data Retention](docs/sql_schema.md#data-retention)
 * **[Quick Tips](#-quick-tips)**
 * **[High Availability](docs/high-avaliability/prometheus-HA.md)**
+* **[Writing data to Promscale](docs/writing_to_promscale.md)**
```
cevian, line 53: I would put this right after "Analyzing Data Using SQL". Also "data" should be "Data" (Capitalization)

cevian (approved):

docs/writing_to_promscale.md
cevian: I think the curl serves as that example

docs/writing_to_promscale.md
cevian: I don't think performant has a space and time definition. In general, I think performant and efficient are complete synonyms so it doesn't matter. We also explain `because the payload is smaller in size` explicitly

docs/writing_to_promscale.md
cevian: I'd use "If using Snappy compression set `Content-Encoding` header to `snappy`, otherwise leave unset."

docs/writing_to_promscale.md
```diff
@@ -0,0 +1,182 @@
...
+--data '{"labels":{"__name__":"foo"},"samples":[[1577836800000, 100]]}' \
+"http://localhost:9201/write"
+```
+Another example using snappy encoding:
+
+```
+curl --header "Content-Type: application/json" \
+--header "Content-Encoding: snappy" \
+--request POST \
+--data-binary "@snappy-payload.sz" \
```
Harkishen-Singh, line 91: Just to note, I hope we have unit-tests for these.

Harkishen-Singh (approved):
LGTM!

PR #375, opened 2020-12-08
Add support for HTTP Basic Authentication and TLS for web endpoints

antekresic:
This commit adds configuration flags for setting up HTTP Basic Authentication,
Bearer Token Authentication (JWT), and TLS certificate support for web server.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -22,6 +22,18 @@ You can also find information on flags with `promscale_<version> -help`
 | web-enable-admin-api | boolean | false | Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series. |
 | web-listen-address | string | `:9201` | Address to listen on for web endpoints. |
 | web-telemetry-path | string | `/metrics` | Web endpoint for exposing Promscale's Prometheus metrics. |
+| tls-cert-file | string | "" (disabled) | TLS certificate file path for web server. To disable TLS, leave this field as blank. |
+| tls-key-file | string | "" (disabled) | TLS key file path for web server. To disable TLS, leave this field as blank. |
+
+## Auth flags
+
+| Flag | Type | Default | Description |
+|------|:-----:|:-------:|:-----------:|
+| auth-username | string | "" | Authentication username used for web endpoint authentication. Disabled by default. |
+| auth-password | string | "" | Authentication password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password-file and bearer-token methods. |
+| auth-password-file | string | "" | Path for auth password file containing the actual password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password and bearer-token methods. |
+| bearer-token | string | "" (disabled) | Bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token-file and basic auth methods. |
+| bearer-token-file | string | "" (disabled) | Path of the file containing the bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token and basic auth methods. |
 
 ## Database flags
 
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -2,15 +2,17 @@ package api
 
 import (
 	"encoding/json"
+	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
@@ -24,12 +26,96 @@ var (
 
 	minTimeFormatted = minTime.Format(time.RFC3339Nano)
 	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
+
+	usernameAndTokenFlagsSetError = fmt.Errorf("at most one of basic-auth-username, bearer-token & bearer-token-file must be set")
+	noUsernameFlagSetError        = fmt.Errorf("invalid auth setup, cannot enable authorization with password only (username required)")
+	noPasswordFlagsSetError       = fmt.Errorf("one of basic-auth-password & basic-auth-password-file must be configured")
+	multiplePasswordFlagsSetError = fmt.Errorf("at most one of basic-auth-password & basic-auth-password-file must be configured")
+	multipleTokenFlagsSetError    = fmt.Errorf("at most one of bearer-token & bearer-token-file must be set")
 )
+
+type Auth struct {
+	BasicAuthUsername     string
+	BasicAuthPassword     string
+	BasicAuthPasswordFile string
+
+	BearerToken     string
+	BearerTokenFile string
+}
+
+func (a *Auth) Validate() error {
+	switch {
+	case a.BasicAuthUsername != "":
+		if a.BearerToken != "" || a.BearerTokenFile != "" {
+			return usernameAndTokenFlagsSetError
+		}
+		if a.BasicAuthPassword == "" && a.BasicAuthPasswordFile == "" {
+			return noPasswordFlagsSetError
+		}
+		if a.BasicAuthPassword != "" && a.BasicAuthPasswordFile != "" {
+			return multiplePasswordFlagsSetError
+		}
+		pwd, err := readFromFile(a.BasicAuthPasswordFile, a.BasicAuthPassword)
+		if err != nil {
+			return fmt.Errorf("error reading password file: %w", err)
+		}
+		a.BasicAuthPassword = pwd
+	case a.BasicAuthPassword != "" || a.BasicAuthPasswordFile != "":
+		// At this point, if we have password set with no username, throw
+		// error to warn the user this is an invalid auth setup.
+		return noUsernameFlagSetError
+	case a.BearerToken != "" || a.BearerTokenFile != "":
+		if a.BearerToken != "" && a.BearerTokenFile != "" {
+			return multipleTokenFlagsSetError
+		}
+		token, err := readFromFile(a.BearerTokenFile, a.BearerToken)
+		if err != nil {
+			return fmt.Errorf("error reading bearer token file: %w", err)
+		}
+		a.BearerToken = token
+	}
+
+	return nil
+}
 
 type Config struct {
 	AllowedOrigin   *regexp.Regexp
 	ReadOnly        bool
 	AdminAPIEnabled bool
+	TelemetryPath   string
+
+	Auth *Auth
+}
+
+func ParseFlags(cfg *Config) *Config {
+	cfg.Auth = &Auth{}
+
+	flag.BoolVar(&cfg.ReadOnly, "read-only", false, "Read-only mode for the connector. Operations related to writing or updating the database are disallowed. It is used when pointing the connector to a TimescaleDB read replica.")
+	flag.BoolVar(&cfg.AdminAPIEnabled, "web-enable-admin-api", false, "Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series.")
+	flag.StringVar(&cfg.TelemetryPath, "web-telemetry-path", "/metrics", "Web endpoint for exposing Promscale's Prometheus metrics.")
+
+	flag.StringVar(&cfg.Auth.BasicAuthUsername, "auth-username", "", "Authentication username used for web endpoint authentication. Disabled by default.")
+	flag.StringVar(&cfg.Auth.BasicAuthPassword, "auth-password", "", "Authentication password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password-file and bearer-token flags.")
+	flag.StringVar(&cfg.Auth.BasicAuthPasswordFile, "auth-password-file", "", "Path for auth password file containing the actual password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password and bearer-token methods.")
+	flag.StringVar(&cfg.Auth.BearerToken, "bearer-token", "", "Bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token-file and basic auth methods.")
+	flag.StringVar(&cfg.Auth.BearerTokenFile, "bearer-token-file", "", "Path of the file containing the bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token and basic auth methods.")
+	return cfg
+}
+
+func Validate(cfg *Config) error {
+	return cfg.Auth.Validate()
+}
+
+func readFromFile(path string, defaultValue string) (string, error) {
+	if path == "" {
+		return defaultValue, nil
+	}
+	bs, err := ioutil.ReadFile(path) // #nosec G304
+	if err != nil {
+		return "", fmt.Errorf("unable to read file %s: %w", path, err)
+	}
+
+	return strings.TrimSpace(string(bs)), nil
 }
 
 func corsWrapper(conf *Config, f http.HandlerFunc) http.HandlerFunc {
@@ -175,19 +261,19 @@ func parseTime(s string) (time.Time, error) {
 	case maxTimeFormatted:
 		return maxTime, nil
 	}
-	return time.Time{}, errors.Errorf("cannot parse %q to a valid timestamp", s)
+	return time.Time{}, fmt.Errorf("cannot parse %q to a valid timestamp", s)
 }
 
 func parseDuration(s string) (time.Duration, error) {
 	if d, err := strconv.ParseFloat(s, 64); err == nil {
 		ts := d * float64(time.Second)
 		if ts > float64(math.MaxInt64) || ts < float64(math.MinInt64) {
-			return 0, errors.Errorf("cannot parse %q to a valid duration. It overflows int64", s)
+			return 0, fmt.Errorf("cannot parse %q to a valid duration. It overflows int64", s)
 		}
 		return time.Duration(ts), nil
 	}
 	if d, err := model.ParseDuration(s); err == nil {
 		return time.Duration(d), nil
 	}
-	return 0, errors.Errorf("cannot parse %q to a valid duration", s)
+	return 0, fmt.Errorf("cannot parse %q to a valid duration", s)
 }
--- a/pkg/api/common_test.go
+++ b/pkg/api/common_test.go
@@ -2,10 +2,14 @@ package api
 
 import (
 	"context"
+	"errors"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"reflect"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/timescale/promscale/pkg/log"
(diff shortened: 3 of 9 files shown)


pkg/runner/runner.go
```diff
@@ -161,7 +167,12 @@ func Run(cfg *Config) error {
 	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
 	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
antekresic, line 169: Should we move these lines into the router as well?

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
 | web-enable-admin-api | boolean | false | Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series. |
 | web-listen-address | string | `:9201` | Address to listen on for web endpoints. |
 | web-telemetry-path | string | `/metrics` | Web endpoint for exposing Promscale's Prometheus metrics. |
+| tls-cert-file | string | "" | TLS Certificate file for web server, leave blank to disable TLS.
+| tls-key-file | string | "" | TLS Key file for web server, leave blank to disable TLS.
+| basic-auth-username | string | "" | HTTP Basic Authentication Username used for web endpoint authentication. Disabled by default.
+| basic-auth-password | string | "" | HTTP Basic Authentication Password used for web endpoint authentication. Password or password-file option should be set together with username.
+| basic-auth-password-file | string | "" | HTTP Basic Authentication Password file containing the actuall password used for web endpoint authentication. Password or password-file option should be set together with username.
```
Harkishen-Singh, line 29:
```suggestion
| auth-password-file | string | "" | Address for auth password file containing the actual password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password and bearer-token methods. |
```

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
 | web-enable-admin-api | boolean | false | Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series. |
 | web-listen-address | string | `:9201` | Address to listen on for web endpoints. |
 | web-telemetry-path | string | `/metrics` | Web endpoint for exposing Promscale's Prometheus metrics. |
+| tls-cert-file | string | "" | TLS Certificate file for web server, leave blank to disable TLS.
```
Harkishen-Singh, line 25:
```suggestion
| tls-cert-file | string | "" (disabled) | TLS certificate file path for web server. To disable TLS, leave this field as blank. |
```
We should give the ending pipe to main the order of all statements.

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
...
+| tls-key-file | string | "" | TLS Key file for web server, leave blank to disable TLS.
```
Harkishen-Singh, line 26:
```suggestion
| tls-key-file | string | "" (disabled) | TLS key file path for web server. To disable TLS, leave this field as blank. |
```

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
...
+| basic-auth-username | string | "" | HTTP Basic Authentication Username used for web endpoint authentication. Disabled by default.
```
Harkishen-Singh, line 27:
```suggestion
| auth-username | string | "" | Authentication username used for web endpoint authentication. Disabled by default. |
```

Do we need to mention `basic auth` everywhere? I think it is obvious that username and password are for basic auth itself.

pkg/runner/runner.go
blagojts: I would not add them at all to non-debug builds

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
...
+| basic-auth-password | string | "" | HTTP Basic Authentication Password used for web endpoint authentication. Password or password-file option should be set together with username.
```
Harkishen-Singh, line 28:
```suggestion
| auth-password | string | "" | Auth password used for web endpoint authentication. This flag should be set together with auth-username. It is mutually exclusive with auth-password-file and bearer-token methods.
```

blagojts (approved):
Great work :+1:

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
...
+| basic-auth-password-file | string | "" | HTTP Basic Authentication Password file containing the actuall password used for web endpoint authentication. Password or password-file option should be set together with username.
+| bearer-token | string | "" | Bearer token (JWT) used for web endpoint authentication. Disabled by default. Cannot be used together with bearer token file setting.
```
Harkishen-Singh, line 30:
```suggestion
| bearer-token | string | "" (disabled) | Bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token-file and basic auth methods. |
```

docs/cli.md
```diff
@@ -22,6 +22,13 @@ You can also find information on flags with `promscale_<version> -help`
...
+| bearer-token-file | string | "" | File path containing the bearer token (JWT) used for web endpoint authentication. Disabled by default. Cannot be used together with bearer token setting.
```
Harkishen-Singh, line 31:
```suggestion
| bearer-token-file | string | "" (disabled) | Address of the file containing the bearer token (JWT) used for web endpoint authentication. Disabled by default. Mutually exclusive with bearer-token and basic auth methods. |
```

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
...
+
+	flag.StringVar(&cfg.BasicAuthUsername, "basic-auth-username", "", "HTTP Basic Authentication Username used for web endpoint authentication. Disabled by default.")
+	flag.StringVar(&cfg.BasicAuthPassword, "basic-auth-password", "", "HTTP Basic Authentication Password used for web endpoint authentication. Password or password-file option should be set together with username.")
+	flag.StringVar(&cfg.BasicAuthPasswordFile, "basic-auth-password-file", "", "HTTP Basic Authentication Password file containing the actuall password used for web endpoint authentication. Password or password-file option should be set together with username.")
+	flag.StringVar(&cfg.BearerToken, "bearer-token", "", "Bearer token (JWT) used for web endpoint authentication. Disabled by default. Cannot be used together with bearer token file setting.")
+	flag.StringVar(&cfg.BearerTokenFile, "bearer-token-file", "", "File path containing the bearer token (JWT) used for web endpoint authentication. Disabled by default. Cannot be used together with bearer token setting.")
```
Harkishen-Singh, line 40: Lets update these with those in the docs.

pkg/api/common.go
```diff
@@ -2,12 +2,15 @@ package api
...
 	"io"
+	"io/ioutil"
 	"math"
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
```
Harkishen-Singh, line 16: We use this at line `252`. However, this adds to the build size as `fmt` package is enough for this. Lets get rid with this import.

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
 	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
 )
 
+func ParseFlags(cfg *Config) *Config {
```
Harkishen-Singh, line 32: nit: If we are accepting `Config` as reference, then why does `ParseFlags` need to return it? We can simply getrid of the return as its redundant here.

Harkishen-Singh (changes requested):
Pass one. Will give another pass to complete.

1. Lets make the cli changes in `common.go` as well.

pkg/api/router.go
```diff
@@ -15,49 +16,81 @@ import (
...
-	labelValuesHandler := timeHandler(metrics.HTTPRequestDuration, "label/:name/values", LabelValues(apiConf, queryable))
+	labelValuesHandler := authWrapper(timeHandler(metrics.HTTPRequestDuration, "label/:name/values", LabelValues(apiConf, queryable)))
 	router.Get("/api/v1/label/:name/values", labelValuesHandler)
 
 	router.Get("/healthz", Health(client))
 
 	return router
 }
 
+func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
```
Harkishen-Singh, line 66: not sure but i think there is a leak here if the user just provides password and no username & bearer-token.

pkg/api/router.go
```diff
@@ -15,49 +16,81 @@ import (
...
+	if cfg.BasicAuthUsername != "" {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			user, pass, ok := r.BasicAuth()
```
Harkishen-Singh, line 69: can we write complete names? 

`user` => `username`

`pass` => `password`

pkg/api/router.go
```diff
@@ -15,49 +16,81 @@ import (
...
+			if !ok || cfg.BasicAuthUsername != user || cfg.BasicAuthPassword != pass {
+				log.Error("msg", "Unauthorized access to endpoint, invalid username or password")
+				http.Error(w, "Unauthorized.", http.StatusUnauthorized)
```
Harkishen-Singh, line 72: also, I think we should pass a message as response to prometheus so that it is visible in prom-container logs, the exact issue if users report any bug or weird behaviour. Like, basic auth password not set; respectively for bearer-token.

pkg/api/router.go
```diff
@@ -15,49 +16,81 @@ import (
...
 
 	router.Post("/write", writeHandler)
 
-	readHandler := timeHandler(metrics.HTTPRequestDuration, "read", Read(client, metrics))
+	readHandler := authWrapper(timeHandler(metrics.HTTPRequestDuration, "read", Read(client, metrics)))
 	router.Get("/read", readHandler)
 	router.Post("/read", readHandler)
 
-	deleteHandler := timeHandler(metrics.HTTPRequestDuration, "delete_series", Delete(apiConf, client))
+	deleteHandler := authWrapper(timeHandler(metrics.HTTPRequestDuration, "delete_series", Delete(apiConf, client)))
```
Harkishen-Singh, line 36: can we have a unit-test for this? I mean, for any operation involving adminAPI.

Harkishen-Singh (changes requested):
Another pass required before completing :sweat_smile:

pkg/runner/runner.go
```diff
@@ -74,18 +75,28 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
...
-	flag.BoolVar(&cfg.ReadOnly, "read-only", false, "Read-only mode for the connector. Operations related to writing or updating the database are disallowed. It is used when pointing the connector to a TimescaleDB read replica.")
-	flag.BoolVar(&cfg.AdminAPIEnabled, "web-enable-admin-api", false, "Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series.")
+	flag.StringVar(&cfg.TLSCertFile, "tls-cert-file", "", "TLS Certificate file for web server, leave blank to disable TLS.")
+	flag.StringVar(&cfg.TLSKeyFile, "tls-key-file", "", "TLS Key file for web server, leave blank to disable TLS.")
 	envy.Parse("TS_PROM")
 	// Ignore errors; CommandLine is set for ExitOnError.
 	_ = flag.CommandLine.Parse(args)
 
+	// Checking if TLS files are not both set or both empty.
+	if (cfg.TLSCertFile != "") != (cfg.TLSKeyFile != "") {
```
Harkishen-Singh, line 85: shouldn't we use `len()` here instead of `""`?

pkg/runner/runner.go
```diff
@@ -74,18 +75,28 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
...
 	if err != nil {
 		err = fmt.Errorf("could not compile CORS regex string %v: %w", corsOriginFlag, err)
 		return nil, err
 	}
-	cfg.CorsOrigin = corsOriginRegex
+	cfg.APICfg.AllowedOrigin = corsOriginRegex
+
+	if err := api.ValidateConfig(&cfg.APICfg); err != nil {
+		err = fmt.Errorf("error validating API configuration: %w", err)
+		return nil, err
```
Harkishen-Singh, line 98:
```suggestion
		return nil, fmt.Errorf("error validating API configuration: %w", err)
```

Harkishen-Singh (changes requested):

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
...
+	case cfg.BasicAuthUsername != "":
+		if cfg.BearerToken != "" || cfg.BearerTokenFile != "" {
+			return fmt.Errorf("at most one of basic-auth-username, bearer-token & bearer-token-file must be set")
+		}
+		if cfg.BasicAuthPassword == "" && cfg.BasicAuthPasswordFile == "" {
+			return fmt.Errorf("one of basic-auth-password & basic-auth-password-file must be configured")
+		}
+		if cfg.BasicAuthPassword != "" && cfg.BasicAuthPasswordFile != "" {
+			return fmt.Errorf("at most one of basic-auth-password & basic-auth-password-file must be configured")
+		}
```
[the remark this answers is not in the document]
cevian: I think the way it is may be cleaner

docs/cli.md
cevian: Should start "Path to auth password file."

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
...
+	if path == "" {
+		return defaultValue, nil
+	}
+	bs, err := ioutil.ReadFile(path)
+
+	if err != nil {
+		return "", fmt.Errorf("unable to read file %s: %w", path, err)
+	}
+
+	return strings.TrimSpace(string(bs)), nil
```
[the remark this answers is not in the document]
cevian: I don't know of an authentication method that should preserver trailing spaces. Maybe bearer token?

pkg/api/router.go
```diff
@@ -15,49 +16,81 @@ import (
...
 
 	return router
 }
 
+func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
+	if cfg.BasicAuthUsername != "" {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			user, pass, ok := r.BasicAuth()
+			if !ok || cfg.BasicAuthUsername != user || cfg.BasicAuthPassword != pass {
+				log.Error("msg", "Unauthorized access to endpoint, invalid username or password")
```
cevian, line 71: we probably want to limit the frequency with which these messages are printed and/or just print the first one. I do think we need to add a Prometheus metric to track the number of unauthorized accesses (counter)

pkg/runner/runner.go
cevian: I would put them behind the auth handler, keep them on by default, but add a cli flag to disable

Harkishen-Singh: I agree with @cevian but I think adding a cli flag for this maybe be overdo.

pkg/api/common.go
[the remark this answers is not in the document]
Harkishen-Singh: Do we really need to trim space?

pkg/api/common.go
antekresic: Replicated the same convention from other places, I don't see the problem there.

pkg/api/common.go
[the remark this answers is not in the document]
antekresic: I don't see an issue with trimming space. The only thing we can do there is fix file contents if somebody were to store some unintentional whitespace in the file. 

I guess its possible to use whitespace in a password, in that case this could cause an issue.

pkg/api/common_test.go
```diff
@@ -109,3 +111,122 @@ func doCORSWrapperRequest(t *testing.T, queryHandler http.Handler, url, origin s
 	queryHandler.ServeHTTP(w, req)
 	return w
 }
+
+func TestValidateConfig(t *testing.T) {
+	fileContents, err := ioutil.ReadFile("common_test.go")
+	if err != nil {
+		t.Fatal("error reading file contents common_test.go")
+	}
+	fileContentsString := strings.TrimSpace(string(fileContents))
```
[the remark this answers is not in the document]
antekresic: If you reuse the function you are testing in the test, you wouldn't be actually testing it. While this can seem redundant, its the best way I know of actually testing the functionality.

pkg/api/router.go
antekresic: Can you explain this a bit better?

pkg/api/router.go
antekresic: I don't see the point in doing this, since the variables are only used one line after. This way its less text to read and its very clear.

pkg/api/router.go
antekresic: I'm OK with adding the same message we are logging. I would not like to specify what is wrong specifically just because of security issues. Saying only `invalid username` opens up to brute-forcing the username, same with other fields.

pkg/runner/runner.go
antekresic: Why `len`? We are checking against the default value here.

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
...
+
+func readFromFile(path string, defaultValue string) (string, error) {
+	if path == "" {
+		return defaultValue, nil
+	}
+	bs, err := ioutil.ReadFile(path)
+
+	if err != nil {
+		return "", fmt.Errorf("unable to read file %s: %w", path, err)
+	}
```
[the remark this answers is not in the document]
antekresic: I don't see the difference here?

pkg/api/common.go
```diff
@@ -26,10 +29,74 @@ var (
...
 	AllowedOrigin   *regexp.Regexp
 	ReadOnly        bool
 	AdminAPIEnabled bool
+
+	BasicAuthUsername     string
+	BasicAuthPassword     string
+	BasicAuthPasswordFile string
+
+	BearerToken     string
+	BearerTokenFile string
```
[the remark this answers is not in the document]
antekresic: Added this change but the code is now worse IMO. Increases artificial complexity.

pkg/runner/runner.go
antekresic: Putting them behind auth handler, leaving out cli flag since this is outside the scope of this PR.

pkg/api/common.go
Harkishen-Singh: The main point here is of returning an object which wasnt used. Name isnt much of an issue.

pkg/api/common.go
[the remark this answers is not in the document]
Harkishen-Singh: >  don't see an issue with trimming space

Why do we need to trim space? Is it that secrets lead to unintended padding while marshalling? 
I am fine with trimming but there has to be a need of it which I do not know. Can you please mention the reason of trimming.

pkg/api/common.go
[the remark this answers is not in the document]
Harkishen-Singh: earlier, the trailing line is before err. Now, the line separation is before ReadFile() which makes actual difference in readability. Giving a line between a function that returns an error and the check of an error, i dont think helps much in readability.

pkg/runner/runner.go
Harkishen-Singh: i think if we want to check if the string is empty or not, i think we should do with `len()`. I recall reading source code of other big projects and thats how they did when things related to checking empty strings came into picture. So thought, we could do the same here.

 Also, it makes the code look cleaner IMO.

```suggestion
	if len(cfg.TLSCertFile) > 0 != len(cfg.TLSKeyFile) > 0 {
```

pkg/api/common.go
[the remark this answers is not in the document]
antekresic: > The only thing we can do there is fix file contents if somebody were to store some unintentional whitespace in the file.

That is the reason. You can see similar functionality in other projects.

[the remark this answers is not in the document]
Harkishen-Singh: Ah, got it. We can go ahead this way but I still think that if someone is having white spaces, then its his/her fault and we doing this silently is not good.

pkg/api/common.go
antekresic: No, the main point is that you are complaining about something which already present in the codebase. 
I prefer to adhere to the codebase standards. 
If you have problems with this, please open a separate PR details why its worthwhile to change all these instances across the whole project.

Please keep in line to codebase idioms.

Examples:
https://github.com/timescale/promscale/blob/master/pkg/log/log.go#L36
https://github.com/timescale/promscale/blob/master/pkg/pgclient/config.go#L35
https://github.com/timescale/promscale/blob/master/pkg/runner/runner.go#L61

pkg/runner/runner.go
antekresic: I don't think your suggestion looks cleaner. In fact, looks harder to read. I need a better reason to change this.

Harkishen-Singh (approved):

PR #381, opened 2020-12-11
Improve delete end-to-end test times

antekresic:
Times before:
```
--- PASS: TestDeleteWithMetricNameEQLRegex (57.40s)
--- PASS: TestDeleteMixins (141.55s)
```

Times after fix:
```
--- PASS: TestDeleteWithMetricNameEQLRegex (8.77s)
--- PASS: TestDeleteMixins (9.12s)
```


Harkishen-Singh (approved):
lgtm. I think the reason i did that way is deleting without a fresh db lead to almost the entire database going empty, hence not being accurate. But, with these changes, since the matchers are not changed, I was wrong. So, it should be good to go ahead.

PR #382, opened 2020-12-11
Provide real-time speed updates for fetching data in prom-migrator

Harkishen-Singh:

This commit decouples the progress bars from writer and reader and
implements real-time speed updates on the fetching of block(s) side.

### Screenshot (Note the spinner to the left of the last line)

Screenshot from 2020-12-11 22-53-01

### DIFF
--- a/pkg/migration-tool/planner/planner.go
+++ b/pkg/migration-tool/planner/planner.go
@@ -16,6 +16,7 @@ import (
 )
 
 var (
+	second                 = time.Second.Milliseconds()
 	minute                 = time.Minute.Milliseconds()
 	laIncrement            = minute
 	maxTimeRangeDeltaLimit = minute * 120
@@ -62,7 +63,7 @@ func Init(config *Config) (*Plan, bool, error) {
 		if found && lastPushedMaxt > config.Mint && lastPushedMaxt <= config.Maxt {
 			config.Mint = lastPushedMaxt
 			log.Warn("msg", fmt.Sprintf("Resuming from where we left off. Last push was on %d. "+
-				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, config.Mint, config.Maxt, (config.Maxt-lastPushedMaxt+1)/time.Minute.Milliseconds()))
+				"Resuming from mint: %d to maxt: %d time-range: %d mins", lastPushedMaxt, config.Mint, config.Maxt, (config.Maxt-lastPushedMaxt+1)/minute))
 		}
 	} else {
 		log.Info("msg", "Resuming from where we left off is turned off. Starting at the beginning of the provided time-range.")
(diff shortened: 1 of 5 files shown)


pkg/migration-tool/utils/client.go
```diff
@@ -117,7 +118,18 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (result *prompb.
...
+	var reader bytes.Buffer
+	if desc != "" {
+		bar := progressbar.DefaultBytes(
+			httpResp.ContentLength,
+			desc,
+		)
+		_, _ = io.Copy(io.MultiWriter(&reader, bar), httpResp.Body)
+	} else {
+		// This case belongs to when the client is used to fetch the progress metric from progress-metric url.
+		_, _ = io.Copy(io.MultiWriter(&reader), httpResp.Body)
```
cevian, line 130:
```suggestion
		_, _ = io.Copy(&reader, httpResp.Body)
```

pkg/migration-tool/planner/planner.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 var (
+	second                 = time.Second.Milliseconds()
 	minute                 = time.Minute.Milliseconds()
```
cevian, line 20: should these be const and not var?

pkg/migration-tool/planner/planner.go
```diff
@@ -161,19 +162,19 @@ func (p *Plan) createBlock(mint, maxt int64) (reference *Block, err error) {
...
 	percent := float64(maxt-p.config.Mint) * 100 / float64(p.config.Maxt-p.config.Mint)
 	if percent > 100 {
 		percent = 100
 	}
-	baseDescription := fmt.Sprintf("progress: %.3f%% | block-%d time-range: %d mins | mint: %d | maxt: %d", percent, id, timeRangeInMinutes, mint, maxt)
+	baseDescription := fmt.Sprintf("progress: %.3f%% | block-%d time-range: %d mins | mint: %d | maxt: %d", percent, id, timeRangeInMinutes, mint/second, maxt/second)
 	reference = &Block{
 		id:                    id,
 		pbarDescriptionPrefix: baseDescription,
 		pbar: progressbar.NewOptions(
```
cevian, line 174: How do these 2 progressbars interact? The one here and the one in client.go? Do they simply override one another? How will this work when we have multiple blocks in-flight at the same time?

cevian (approved):
This is ok to merge now (several nits). But I do worry how this will work when we have concurrent blocks being transferred (see comment below).

pkg/migration-tool/planner/planner.go
Harkishen-Singh: They override each other. But I do fear that if the reader is puling multiple blocks simulatenously, then it can be an issue as I don't think that `io.Copy` can do copying in a concurrent state. So maybe, if the user chooses to have concurrent fetches then we disable the transfer-rate and only show transfer-rate if fetching concurrent blocks is disabled. 
(as if concurrent fetches and concurrent flushes are enabled, then the progress bar itself becomes more interactive).
WDYT?

pkg/migration-tool/planner/planner.go
Harkishen-Singh: Sorry, but any functional call over a const (like here time.Second.Milliseconds() over time.Second) is considered as var by the compiler.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: @cevian does it look good?

cevian: for now, yes

cevian (approved):

PR #383, opened 2020-12-14
Add config file support for cli

antekresic:


pkg/runner/runner.go
```diff
@@ -80,7 +82,10 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
 	envy.Parse("PROMSCALE")
 
 	// Ignore errors; CommandLine is set for ExitOnError.
-	_ = flag.CommandLine.Parse(args)
+	_ = ff.Parse(flag.CommandLine, args,
+		ff.WithConfigFile("config.yml"),
```
Harkishen-Singh, line 86: Shouldn't this name be supplied from CLI?

cevian: Yeah I think using `WithConfigFileFlag` is better and define a new flag with `config.yml` as the default value

cevian (changes requested):
Looks good. But, I'd like some test of the fact that this parses the config file(lowest priority), env, and cli flags (highest priority) in the right priority.

pkg/runner/runner.go
antekresic: Yes, I simplified it for the draft, will add a flag for the filename.

cevian (approved):
LGTM

PR #385, opened 2020-12-15
Adds support for concurrent flushing in prom-migrator to remote write storage by writer

Harkishen-Singh:
This PR should be merged after #382 .

TODO:

- [x] Duplicate sample issue. Possible that the previous maxt fetched is not being respected. 
- [x] Perform an actual migration test before merging


pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,189 @@
...
+	errChan chan error
+}
+
+func newShardsSet(writerCtx context.Context, writeURL string, capacity int) (*shardsSet, error) {
+	var (
+		set         = make([]*shard, capacity)
+		cancelFuncs = make([]context.CancelFunc, capacity)
+	)
+	ss := &shardsSet{
+		num:     capacity,
```
blagojts, line 34: maybe rename num to capcity or capacity to num?

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,189 @@
...
+type shardsSet struct {
+	num         int
+	set         []*shard
+	cancelFuncs []context.CancelFunc
+
+	done    chan struct{}
+	errChan chan error
+}
+
+func newShardsSet(writerCtx context.Context, writeURL string, capacity int) (*shardsSet, error) {
```
blagojts, line 28: should add a comment that this func starts goroutines

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,189 @@
...
+	for i := tsIndex; i < s.num; i++ {
+		var batch []prompb.TimeSeries
+		if i == s.num-1 {
+			batch = ts[tsIndex:]
+		} else {
+			batch = ts[tsIndex : tsIndex+tsEachShard]
+			tsIndex += tsEachShard
+		}
+		batches[i] = &batch
+	}
```
blagojts, line 79: I think we can avoid the allocation of the batches slice. As you create a batch in the cycle just enqueue it at s.set[tsIndex].queue and no need to call assign. Or I'm missing something

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,189 @@
...
+
+// assign assigns the batches to the queues.
+func (s *shardsSet) assign(batches []*[]prompb.TimeSeries) {
+	for shardIndex := 0; shardIndex < len(batches); shardIndex++ {
+		s.set[shardIndex].queue <- batches[shardIndex]
+	}
+}
+
+// shard is an atomic unit of writer. Each shard accepts a sharded time-series set and
+// pushes the respective time-series via a POST request. More the number of shards, more is the throughput.
```
blagojts, line 91:
```suggestion
// shard is an atomic unit of writing. Each shard accepts a sharded time-series set and
// pushes the respective time-series via a POST request. More shards should lead to bigger throughput if the target system can handle it.
```

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,189 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		err = attemptStore()
+
+		if err != nil {
```
blagojts, line 151:
```suggestion
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
		}

		if client.Store(ctx, *buf); err != nil {
```

pkg/migration-tool/writer/writer.go
```diff
@@ -24,22 +20,26 @@ const (
...
 	if err != nil {
-		return nil, fmt.Errorf("creating write-client: %w", err)
+		return nil, fmt.Errorf("creating shards: %w", err)
 	}
 	write := &RemoteWrite{
 		c:                  c,
-		client:             wc,
+		shardsSet:          ss,
+		url:                remoteWriteUrl,
+		numShards:          numShards,
```
blagojts, line 42: is numShards required if we have access to shardsSet.num?

cmd/prom-migrator/main.go
```diff
@@ -130,6 +131,8 @@ func parseFlags(conf *config, args []string) {
 		"Setting this value less than zero will indicate all data from mint upto now. ")
 	flag.StringVar(&conf.maxBlockSize, "max-read-size", "500MB", "(units: B, KB, MB, GB, TB, PB) the maximum size of data that should be read at a single time. "+
 		"More the read size, faster will be the migration but higher will be the memory usage. Example: 250MB.")
+	flag.IntVar(&conf.numShards, "num-shards", 4, "Number of shards involved in pushing a block. More the number of shards, more will be the number of"+
+		"parallel flushers involved per in-memory block and hence greater will be the through-put for any in-memory block.")
```
blagojts, line 135:
```suggestion
	flag.IntVar(&conf.numShards, "num-shards", 4, "Number of shards to split a fetched block in. Each shard is written concurrently")
```

Reduce implementation details from the flag desc

blagojts (commented):
I like the implementation of the block sharding and concurrent writes. Left some minor comments for some simplifications here and there. 

The tl;dr comments are for lines that are too long and/or contain calculations that can be split in several expressions

pkg/migration-tool/planner/planner.go
```diff
@@ -161,19 +162,19 @@ func (p *Plan) createBlock(mint, maxt int64) (reference *Block, err error) {
...
 		percent = 100
 	}
-	baseDescription := fmt.Sprintf("progress: %.3f%% | block-%d time-range: %d mins | mint: %d | maxt: %d", percent, id, timeRangeInMinutes, mint, maxt)
+	baseDescription := fmt.Sprintf("progress: %.3f%% | block-%d time-range: %d mins | mint: %d | maxt: %d", percent, id, timeRangeInMinutes, mint/second, maxt/second)
 	reference = &Block{
 		id:                    id,
 		pbarDescriptionPrefix: baseDescription,
 		pbar: progressbar.NewOptions(
-			6,
+			4,
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, the 4 represents the number of steps taken by a progressbar to get to the full. The 2 steps decreased here is because this progress bar is now independent of the reader's functionalities (or steps) (handled by bytes bar in `client.go`), thus decreasing its journey to full.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: Oh yes. Thank you for the advice. That will help save a lot of very useful memory.

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Yeah, we do not require `numShards` anymore. Thank you for pointing it out.

cmd/prom-migrator/main.go
Harkishen-Singh: Definitely. However, along with this suggestion, I am adding another thing as, `Note: Larger shards count will lead to significant memory usages.`. I hope it is fine with the sentence.

pkg/migration-tool/planner/planner.go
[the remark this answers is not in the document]
blagojts: create a named const so it's self documenting

cmd/prom-migrator/main.go
blagojts: Yup, now that you mention it, Prometheus uses the same kind of warning on their remote_write config 👍

blagojts (approved):
Looks good 👍

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,178 @@
...
+	// Run the shards.
+	for i := 0; i < ss.num; i++ {
+		go ss.set[i].run(i)
+	}
+	return ss, nil
+}
+
+// scheduleTS batches the time-series based on the available shards. These batches are then
+// fed to the shards in a single go which then consume and push to the respective clients.
+func (s *shardsSet) scheduleTS(ts []prompb.TimeSeries) {
```
cevian, line 67: I think there needs to be special handling for the progress metric. Right now the progress metric is written by just one of the shards. But what happens if that shard succeeds and updates the progress metric, while the process crashes before the other shards finished? That means some data will be missing on restart, which is not acceptable.

You can look at how we respond to writeRequests in promscale for one idea of how to fix this using waitgroups.

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,178 @@
...
+	var (
+		tsIndex     = 0
+		tsEachShard = len(ts) / s.num
+	)
+	for shardIndex := tsIndex; shardIndex < s.num; shardIndex++ {
+		var batch []prompb.TimeSeries
+		if shardIndex == s.num-1 {
+			batch = ts[tsIndex:]
+		} else {
+			batch = ts[tsIndex : tsIndex+tsEachShard]
```
cevian, line 77: Prometheus shards items using a consistent hash on the series. This seems to round-robin. I think it's better to use consistent hash like Prometheus since long term stores may optimize for that.

cevian (changes requested):
Some blockers. Especially important is that we write the progress metric only after the other shards have succeeded.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: Won't that just add to the cost, as I am unable to understand the importance of it than just dividing among shards for which here we use indexes.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: I see one way, to send the progress metric in a different POST request after the shards have completed pushing the data.

blagojts (approved):
Looks good

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,181 @@
...
+	log.Info("msg", fmt.Sprintf("shard-%d is active", shardIndex))
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case refTs, ok := <-s.queue:
+			if !ok {
+				return
+			}
+			if len(*refTs) == 0 {
```
cevian, line 105: important: don't we need to `s.copyShardSet.done <- struct{}{}` here too? Otherwise we'll end up waiting for a done for forever?

pkg/migration-tool/planner/block.go
```diff
@@ -49,15 +50,31 @@ func (b *Block) Fetch(context context.Context, client *utils.Client, mint, maxt
...
-	b.SetDescription(fmt.Sprintf("pushing %.2f...", float64(b.numBytesCompressed)/float64(utils.Megabyte)), 1)
-	ts.Samples = []prompb.Sample{{Timestamp: b.Maxt(), Value: 1}} // One sample per block.
-	b.timeseries = append(b.timeseries, ts)
+// Series returns the time-series in the block.
+func (b *Block) Series() []*prompb.TimeSeries {
 	return b.timeseries
 }
 
+// GetProgressSeries returns a time-series after appending a sample to the progress-metric.
+func (b *Block) GetProgressSeries(ts *prompb.TimeSeries) *prompb.TimeSeries {
```
cevian, line 59:
```suggestion
func (b *Block) UpdateProgressSeries(ts *prompb.TimeSeries) *prompb.TimeSeries {
```

pkg/migration-tool/writer/writer.go
```diff
@@ -74,12 +88,24 @@ func (rw *RemoteWrite) Run(errChan chan<- error) {
 				if !ok {
 					return
 				}
-				blockRef.SetDescription("preparing to push", 1)
-				ts = timeseriesRefToTimeseries(blockRef.MergeProgressSeries(rw.progressTimeSeries))
-				buf = []byte{}
-				if err = rw.sendSamplesWithBackoff(context.Background(), ts, &buf); err != nil {
-					errChan <- fmt.Errorf("remote-write run: %w", err)
-					return
+				blockRef.UpdatePBarMax(blockRef.PBarMax() + rw.shardsSet.num)
```
cevian, line 91: nit: shouldn't this be `blockRef.UpdatePBarMax(blockRef.PBarMax() + rw.shardsSet.num+1)` isn't there one extra step needed for the progressTimeSeries?

pkg/migration-tool/writer/writer.go
```diff
@@ -74,12 +88,24 @@ func (rw *RemoteWrite) Run(errChan chan<- error) {
...
+					// Pushing data to remote-write storage.
+					blockRef.SetDescription("preparing to push", 1)
+					shards.scheduleTS(timeseriesRefToTimeseries(blockRef.Series()))
+					blockRef.SetDescription("pushing ...", 1)
+					var pushed int
+					for i := 0; i < shards.num; i++ {
+						pushed = waitForSig(blockRef, pushed, shards.num)
+					}
+				}
+				{
```
cevian, line 102: why these extra paranthesis?

pkg/migration-tool/writer/writer.go
```diff
@@ -74,12 +88,24 @@ func (rw *RemoteWrite) Run(errChan chan<- error) {
...
 					return
 				}
-				blockRef.SetDescription("preparing to push", 1)
-				ts = timeseriesRefToTimeseries(blockRef.MergeProgressSeries(rw.progressTimeSeries))
-				buf = []byte{}
-				if err = rw.sendSamplesWithBackoff(context.Background(), ts, &buf); err != nil {
-					errChan <- fmt.Errorf("remote-write run: %w", err)
-					return
+				blockRef.UpdatePBarMax(blockRef.PBarMax() + rw.shardsSet.num)
+				{
```
cevian, line 92: why this block?

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+// shardsSet represents a set of shards. It consists of configuration related
+// to handling and management of shards.
+type shardsSet struct {
+	num         int
+	set         []*shard
+	cancelFuncs []context.CancelFunc
+
+	done    chan struct{}
```
cevian, line 24: optional: a more idiomatic way to do this is to use a sync.WaitGroup, I'll leave it up to you if you want to change it.

cevian (changes requested):
Only one of my comments is blocking. It is marked "important"

pkg/migration-tool/writer/shards.go
Harkishen-Singh: Oh yes. Thank you for pointing that out.

pkg/migration-tool/writer/writer.go
Harkishen-Singh: It is just for easier reading. Each block represents a (pushed + waitForSig). So, in order to make it easier to understand the two push operations and differentiate them, I kept them as different blocks (of same nature).

Do you think this is not that beneficial?

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Same reason as https://github.com/timescale/promscale/pull/385#discussion_r551721431.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: The reason why wait group is not used is that using wait-group will not give the signal on the writer goroutine when a shard is complete in order to proceed the progress-bar per shard level, rather just give a whole done, making the progress-bar fewer steps (hence less interactive).

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Oh yes. Sorry, I missed that.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: For some reason, then I do this, the shards at this place get stuck. It's as if the writer.Run() does not expect a done signal when the `len(ts)` is 0. This should not happen but I gave an attempt to figure out why it's happening, but in vain. Would be happy if I can get some help here.

pkg/migration-tool/writer/shards.go
```diff
@@ -0,0 +1,186 @@
...
+	log.Info("msg", fmt.Sprintf("shard-%d is active", shardIndex))
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case refTs, ok := <-s.queue:
+			if !ok {
+				return
+			}
+			if len(*refTs) != 0 {
```
cevian, line 109: I'd actually consider throwing a panic if len ==0, that represents and irrecoverable logic error.

pkg/migration-tool/writer/writer.go
cevian: Personally, I don't like this style. Any time a block like this comes up, it usually implies it's better as a factored-out function and/or without the block. But I'll leave this up to you.

cevian (approved):
Minor comments. Feel free to change or not. Can be merged without further review. BTW I like you solution with scheduleTS returning number of shards to wait for.

pkg/migration-tool/writer/shards.go
Harkishen-Singh: I feel like panic is something severe and may silently stall the system given the future plans of streaming. An error logging with details would be better without exiting the shards.

PR #388, opened 2020-12-16
Add service-url flag

VineethReddy02:
The service-url helps to connect with timescaleDB with a single flag. To set pool_max_conns, pool_min_conns use respective flags.

```
go run cmd/promscale/main.go --db-uri postgres://postgres:password@localhost:5432/postgres?sslmode=allow
```


pkg/pgclient/config.go
```diff
@@ -46,13 +48,17 @@ func ParseFlags(cfg *Config) *Config {
 	flag.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", pgmodel.DefaultMetricCacheSize, "Maximum number of metric names to cache.")
 	flag.IntVar(&cfg.WriteConnectionsPerProc, "db-writer-connection-concurrency", 4, "Maximum number of database connections for writing per go process.")
 	flag.IntVar(&cfg.MaxConnections, "db-connections-max", -1, "Maximum number of connections to the database that should be opened at once. It defaults to 80% of the maximum connections that the database can handle.")
+	flag.StringVar(&cfg.ServiceUrl, "service-url", DefaultServiceUrlFromDefaultFlags, "TimescaleDB/Vanilla Postgres service URL.")
```
cevian, line 51: 2 things:
1. Postgres calls this a URI not a URL, so lets change it to URI. Also, should we call this `db-uri` not service... This would fall in-line with the other db- parameters.
2. As mentioned before, I think an empty-string default is better. But we should move the current default into the description string as an example.

pkg/util/util.go
```diff
@@ -71,5 +75,10 @@ func (dt *ThroughputCalc) Start() {
...
 func MaskPassword(s string) string {
 	s = maskPasswordRegex1.ReplaceAllString(s, maskPasswordReplaceString1)
-	return maskPasswordRegex2.ReplaceAllString(s, maskPasswordReplaceString2)
+	s = maskPasswordRegex2.ReplaceAllString(s, maskPasswordReplaceString2)
+	return maskPasswordRegex3.ReplaceAllString(s, maskPasswordReplaceString3)
 }
+
+func MaskPasswordinServiceURL(s string) string {
+	return maskPasswordRegex4.ReplaceAllString(s, maskPasswordReplaceString4)
+}
```
cevian, line 84: Let's keep all the variants in one function `MaskPassword`. These are so seldom used performance doesn't matter and its safer if there is just one function.

pkg/pgclient/config.go
```diff
@@ -46,13 +48,17 @@ func ParseFlags(cfg *Config) *Config {
...
 	return cfg
 }
 
 // GetConnectionStr returns a Postgres connection string
 func (cfg *Config) GetConnectionStr() string {
-	return fmt.Sprintf("host=%v port=%v user=%v dbname=%v password='%v' sslmode=%v connect_timeout=10",
-		cfg.Host, cfg.Port, cfg.User, cfg.Database, cfg.Password, cfg.SslMode)
+	if cfg.ServiceUrl == DefaultServiceUrlFromDefaultFlags {
+		return fmt.Sprintf("host=%v port=%v user=%v dbname=%v password='%v' sslmode=%v connect_timeout=10",
+			cfg.Host, cfg.Port, cfg.User, cfg.Database, cfg.Password, cfg.SslMode)
```
cevian, line 59: We should validate that we are only setting either the URI or these parameters but not both. We should throw an error if setting a combination. The reason is that otherwise users may think we'll combine the URI with these in some way.

cevian (changes requested):
Overall looks good but some changes requested.

docs/cli.md
```diff
@@ -48,4 +48,5 @@ You can also find information on flags with `promscale_<version> -help`
 | db-connections-max | integer | 80% of possible connections db | Maximum number of connections to the database that should be opened at once. It defaults to 80% of the maximum connections that the database can handle. |
 | db-ssl-mode | string | require | TimescaleDB/Vanilla Postgres connection ssl mode. If you do not want to use ssl, pass `allow` as value. |
 | db-writer-connection-concurrency | int | 4 | Maximum number of database connections for writing per go process. |
+| db-uri | string | | TimescaleDB/Vanilla Postgres db uri. |
```
antekresic, line 51: Minor nit: URI should be capitalized :)

helm-chart/README.md
```diff
@@ -71,6 +102,8 @@ helm install --name my-release -f myvalues.yaml .
 | `replicaCount`                    | Number of pods for the connector            | `1`                                |
 | `connection.user`                 | Username to connect to TimescaleDB with     | `postgres`                         |
 | `connection.password.secretTemplate`| The template for generating the name of a secret object which will hold the db password | `{{ .Release.Name }}-timescaledb-passwords` |
+| `connection.dburi.name`| DB uri name is used as a key for secret which will hold the db uri value |  |
+| `connection.dburi.secretTemplate`| The template for generating the name of a secret object which will hold the db uri | `{{ .Release.Name }}-timescaledb-uri` |
```
antekresic, line 106: URI capitalization in description.

pkg/pgclient/client_test.go
```diff
@@ -183,3 +184,57 @@ func TestHealthCheck(t *testing.T) {
...
+					WriteConnectionsPerProc: 1,
+					MaxConnections:          0,
+					UsesHA:                  false,
+					DbUri:                   "postgres://postgres:password@localhost:5432/postgres?sslmode=allow",
+				},
+				schemaLocker: nil,
+			},
+			nil,
+			true,
+			errors.New("failed to build DB credentials with provided flags. Please use either db flags or db-uri not both"),
```
antekresic, line 225: If we are testing for this exact error, it should be made into a variable and reused in the test. Also, no need to use `errors` package, `fmt.Errorf` is used everywhere to make new errors and its already imported.

pkg/pgclient/client_test.go
```diff
@@ -183,3 +184,57 @@ func TestHealthCheck(t *testing.T) {
 		t.Fatal("health check method not called when expected")
 	}
 }
+
+func TestNewClient(t *testing.T) {
```
antekresic, line 188: Could use a bit more test cases, to at least cover `GetNumConnections` and `ParseConfig` errors.

pkg/pgclient/config.go
```diff
@@ -2,14 +2,14 @@ package pgclient
...
+	"errors"
 	"flag"
 	"fmt"
-	"runtime"
-	"strconv"
-
 	"github.com/jackc/pgx/v4"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel"
+	"runtime"
```
antekresic, line 11: Its a minor thing but imports need to be grouped by standard and 3rd party packages, like it was before this change.

antekresic (changes requested):
Some minor stuff, once that's addressed, should be good IMO :+1:

pkg/pgclient/client_test.go
VineethReddy02: In `CreateNewClient` we are creating` connectionPool` adding more test cases to validate ParseFlags config is failing the func call as DB calls will fail. Any thoughts/workaround to handle this?

pkg/pgclient/config.go
```diff
@@ -29,30 +29,52 @@ type Config struct {
...
+	if cfg.DbUri == DefaultDBUri {
+		return fmt.Sprintf("host=%v port=%v user=%v dbname=%v password='%v' sslmode=%v connect_timeout=10",
+			cfg.Host, cfg.Port, cfg.User, cfg.Database, cfg.Password, cfg.SslMode), nil
+	} else if cfg.Database != DefaultDBName || cfg.Host != DefaultDBHost || cfg.Port != DefaultDBPort || cfg.User != DefaultDBUser ||
+		cfg.Password != DefaultDBPassword || cfg.SslMode != DefaultSSLMode {
+		return "", excessDBFlagsError
+	}
+
+	uri := cfg.DbUri + "&connect_timeout=10"
+	return uri, nil
```
antekresic, line 77:
```suggestion
func (cfg *Config) GetConnectionStr() (string, error) {
    if cfg.Database != DefaultDBName ||
        cfg.Host != DefaultDBHost ||
        cfg.Port != DefaultDBPort ||
        cfg.User != DefaultDBUser ||
        cfg.Password != DefaultDBPassword ||
        cfg.SslMode != DefaultSSLMode {

        if cfg.DbUri != DefaultDBUri {
            return "", excessDBFlagsError
        }
        return fmt.Sprintf("host=%v port=%v user=%v dbname=%v password='%v' sslmode=%v connect_timeout=10",
            cfg.Host, cfg.Port, cfg.User, cfg.Database, cfg.Password, cfg.SslMode), nil
    }
    return cfg.DbUri + "&connect_timeout=10", nil
```

antekresic: This needs some unit tests with the increased complexity.

pkg/pgclient/client.go
```diff
@@ -35,24 +35,34 @@ type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
...
+		if err != nil {
+			log.Error("msg", "configuring connection", "err", util.MaskPassword(err.Error()))
+			return nil, err
+		}
+	} else {
+		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf("&pool_max_conns=%d&pool_min_conns=%d", maxConnections, minConnections))
+		if err != nil {
+			log.Error("msg", "configuring connection", "err", util.MaskPassword(err.Error()))
+			return nil, err
+		}
```
antekresic, line 60: Would lessen the code duplication if you were to only construct the connection string in the conditional path. You might even want to just switch between the concatenation strings ` ` or `&` depending whether its a DB URI or not.

Should make the code simpler and easier to read.

pkg/pgclient/client_test.go
antekresic: I just want to make sure we covered all the cases logically using the unit test. If you feel we achieved that, you can leave the test as is.

antekresic (changes requested):
Some more tweaks :smile:

helm-chart/README.md
```diff
@@ -34,13 +34,36 @@ connection:
...
 
+OR 
+
+You can use db uri to connect to TimescaleDB is stored in a
+Kubernetes Secret created before the chart is deployed.
+You can set the secret name by modifying the  `connection.dburi.secretTemplate` value.
+Templating is supported and you can use:
+```yaml
+connection:
+  dburi:
```
cevian, line 54: can this be `db-uri`? or are - not allowed?

helm-chart/README.md
```diff
@@ -34,13 +34,36 @@ connection:
...
+    name: db-uri
```
cevian, line 55: do we really need this configurable?

helm-chart/README.md
cevian: otherwise maybe dbUri or dbURI?

cevian (commented):
some comments on ux. Will let @antekresic take lead on PR

pkg/pgclient/config.go
VineethReddy02: Yeah, I will add unit tests for this. But the code you suggested with fail in one edge case. i.e if user wants to connect to DB with the default flags, in that case, nothing will be changed with the above validation we cannot get into the line 78 build `("host=%v port=%v....)` this case will return by default `   &connect_timeout=10` which isn't expected.

pkg/pgclient/client.go
VineethReddy02: I will move to check for error redundant code out of if-else for more code singularity.

helm-chart/README.md
VineethReddy02: I will make it as dbURI

helm-chart/README.md
VineethReddy02: I don't see a real need. Should we hardcode it? As we definitely need some key in secret to refer to the value as the secret is a key/vale pair.

pkg/pgclient/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    string
+		wantErr bool
+		err     error
+	}{
+		{
+			name: "test1",
```
antekresic, line 32: Love the new tests, would like to see them named a bit better so we know what the test case it is. That way, it's easier to see what we are testing and is the test setup properly.

antekresic (approved):
Looks good, I believe you have addressed all the things. 

I have one thing about test case naming, would like them to be a bit more descriptive.

Once that is addressed, should be ready to merge the PR.

docs/cli.md
```diff
@@ -48,4 +48,5 @@ You can also find information on flags with `promscale_<version> -help`
 | db-connections-max | integer | 80% of possible connections db | Maximum number of connections to the database that should be opened at once. It defaults to 80% of the maximum connections that the database can handle. |
 | db-ssl-mode | string | require | TimescaleDB/Vanilla Postgres connection ssl mode. If you do not want to use ssl, pass `allow` as value. |
 | db-writer-connection-concurrency | int | 4 | Maximum number of database connections for writing per go process. |
+| db-uri | string | | TimescaleDB/Vanilla Postgres db URI. |
```
cevian, line 51:
```suggestion
| db-uri | string | | TimescaleDB/Vanilla PostgresSQL URI. Example:`postgres://postgres:password@localhost:5432/timescale?sslmode=require` |
```

pkg/pgclient/config.go
```diff
@@ -29,30 +29,55 @@ type Config struct {
...
 	flag.BoolVar(&cfg.AsyncAcks, "async-acks", false, "Acknowledge asynchronous inserts. If this is true, the inserter will not wait after insertion of metric data in the database. This increases throughput at the cost of a small chance of data loss.")
 	flag.IntVar(&cfg.ReportInterval, "tput-report", 0, "Interval in seconds at which throughput should be reported.")
 	flag.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "Maximum number of labels to cache.")
 	flag.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", pgmodel.DefaultMetricCacheSize, "Maximum number of metric names to cache.")
 	flag.IntVar(&cfg.WriteConnectionsPerProc, "db-writer-connection-concurrency", 4, "Maximum number of database connections for writing per go process.")
 	flag.IntVar(&cfg.MaxConnections, "db-connections-max", -1, "Maximum number of connections to the database that should be opened at once. It defaults to 80% of the maximum connections that the database can handle.")
+	flag.StringVar(&cfg.DbUri, "db-uri", DefaultDBUri, "TimescaleDB/Vanilla Postgres DB URI. Example DB URI `postgres://postgres:password@localhost:5432/timescale?sslmode=require`")
```
cevian, line 60:
```suggestion
	flag.StringVar(&cfg.DbUri, "db-uri", DefaultDBUri, "TimescaleDB/Vanilla PostgresSQL URI. Example:`postgres://postgres:password@localhost:5432/timescale?sslmode=require`")
```

cevian (approved):

PR #390, opened 2020-12-17
Refactor: SQL write and read path, split SQL tests

Harkishen-Singh:
This PR does refactoring on:

- [x]  Splits sql_ingest.go into multiple files (1st commit)
- [x] Shift the ingestor component into its respective dir (2nd commit)
- [x]  Split SQL reader and writer tests
- [x]  Shift querier component into its respective dir
- [x]  License headers across files in the project
- [x]  Shift minTime and maxTime to utils so as to keep a common source for api/common.go and delete.go


pkg/api/delete.go
```diff
@@ -2,13 +2,14 @@ package api
 
 import (
 	"fmt"
+	delete2 "github.com/timescale/promscale/pkg/pgmodel/delete"
```
antekresic, line 5: This name `delete2` sticks out like a sore thumb. 
Makes the code confusing, personally I would need to try to find out what the `2` stands for.

```suggestion
	deletePkg "github.com/timescale/promscale/pkg/pgmodel/delete"
```
Also, imports need to be properly grouped and sorted.

pkg/api/query_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/timescale/promscale/pkg/pgmodel/querier"
```
antekresic, line 8: Import needs to be grouped with other Promscale packages.

pkg/pgclient/client.go
```diff
@@ -3,6 +3,10 @@ package pgclient
 import (
 	"context"
 	"fmt"
+	"github.com/timescale/promscale/pkg/pgmodel/cache"
+	"github.com/timescale/promscale/pkg/pgmodel/ingestor"
+	querier2 "github.com/timescale/promscale/pkg/pgmodel/querier"
```
antekresic, line 8: Is there a good reason for `querier2` naming? I'd rather change the name of the local variable its clashing with instead of the import.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -1,56 +1,42 @@
-// This file and its contents are licensed under the Apache License 2.0.
-// Please see the included NOTICE for copyright information and
-// LICENSE for a copy of the license.
-
-package pgmodel
+package ingestor
```
antekresic, line 1: Missing copyright header.

pkg/query/queryable.go
```diff
@@ -2,21 +2,22 @@ package query
 
 import (
 	"context"
+	querier2 "github.com/timescale/promscale/pkg/pgmodel/querier"
```
antekresic, line 5: Reason for name `querier2`?

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -15,7 +15,8 @@ import (
 	pgx "github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 	_ "github.com/jackc/pgx/v4/stdlib"
-	. "github.com/timescale/promscale/pkg/pgmodel"
+	igstr "github.com/timescale/promscale/pkg/pgmodel/ingestor"
```
antekresic, line 18: Again, I prefer to change variable names if they are clashing to package names. 
Also, we are using `ingr`, `ingstr`,  and here `igstr` for a short version of the word `ingestor`. This needs to be consistent across the project.

antekresic (changes requested):
Lots of things are pushed into `utils` package as a catch-all. That doesn't improve the situation from before where we had all those things in `pgmodel`, just adds more imports everywhere. 

These things need to be scoped, grouped and moved into their logical packages IMO.

Lots of missing copyright headers in new files (one or two places in existing files too). 
Some questionable import renaming that need to be addressed.

pkg/api/delete.go
Harkishen-Singh: Sorry, that was by the code editor. I will fix all these.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: BTW, do we need to have the copyright header across each and every file in the project? (Just for my info)

pkg/query/queryable.go
Harkishen-Singh: Sorry, that's by the editor. Will change these in the next commit.

pkg/pgmodel/migrate.go
```diff
@@ -390,14 +391,14 @@ func replaceSchemaNames(r io.ReadCloser) (string, error) {
...
 	s := buf.String()
-	s = strings.ReplaceAll(s, "SCHEMA_CATALOG", catalogSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_EXT", extSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_PROM", promSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_SERIES", seriesViewSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_METRIC", metricViewSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_DATA_SERIES", dataSeriesSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_DATA", dataSchema)
-	s = strings.ReplaceAll(s, "SCHEMA_INFO", infoSchema)
+	s = strings.ReplaceAll(s, "SCHEMA_CATALOG", utils.CatalogSchema)
```
antekresic, line 394: These things don't look like they should be in `utils` package... 
How about we make a new package called `common`? Might make more sense IMO.

pkg/pgmodel/utils/utils.go
```diff
@@ -0,0 +1,49 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
antekresic, line 1: This might be another candidate for a `common` package instead of `utils`.

antekresic (changes requested):
Looks better. Still some minor things to sort out, after that should be good IMO.

pkg/pgmodel/ingestor/ingestor_test.go
```diff
@@ -1,95 +1,16 @@
...
+
+package ingestor
 
 import (
 	"fmt"
-	"testing"
-
+	"github.com/timescale/promscale/pkg/pgmodel/utils"
 	"github.com/timescale/promscale/pkg/prompb"
+	"testing"
```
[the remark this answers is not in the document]
Harkishen-Singh: Agreed. I keep missing this for some reason because the auto import in goland is kind of causing difficulties.

pkg/api/health.go
```diff
@@ -1,13 +1,17 @@
...
 import (
+	"github.com/timescale/promscale/pkg/pgmodel/heath"
 	"net/http"
 
 	"github.com/timescale/promscale/pkg/log"
-	"github.com/timescale/promscale/pkg/pgmodel"
 )
 
-func Health(hc pgmodel.HealthCheckerFn) http.HandlerFunc {
+func Health(hc heath.HealthCheckerFn) http.HandlerFunc {
```
blagojts, line 14:
```suggestion
func Health(hc health.HealthCheckerFn) http.HandlerFunc {
```

pkg/pgmodel/ingestor/buffer.go
```diff
@@ -0,0 +1,88 @@
...
+	// was chosen arbitrarily.
+	flushSize                = 2000
+	getCreateMetricsTableSQL = "SELECT table_name FROM " + schema.Catalog + ".get_or_create_metric_table_name($1)"
+	finalizeMetricCreation   = "CALL " + schema.Catalog + ".finalize_metric_creation()"
+	getSeriesIDForLabelSQL   = "SELECT * FROM " + schema.Catalog + ".get_or_create_series_id_for_kv_array($1, $2, $3)"
+	getEpochSQL              = "SELECT current_epoch FROM " + schema.Catalog + ".ids_epoch LIMIT 1"
+	maxCopyRequestsPerTxn    = 100
+)
+
+// epoch for the ID caches, -1 means that the epoch was not set
```
blagojts, line 27: Do you think it's good to have a `const EpochNotSet = Epoch(-1)` so we can use it as a const throughout the code?

pkg/pgmodel/querier/series_set.go
```diff
@@ -1,14 +1,20 @@
-package pgmodel
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package querier
 
 import (
 	"fmt"
+	"github.com/timescale/promscale/pkg/pgmodel/common/errors"
```
blagojts, line 9: auto import

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -25,7 +26,7 @@ func TestSQLRetentionPeriod(t *testing.T) {
 		ts := []prompb.TimeSeries{
 			{
 				Labels: []prompb.Label{
-					{Name: MetricNameLabelName, Value: "TEST"},
+					{Name: utils.MetricNameLabelName, Value: "test"},
```
blagojts, line 29: is the "TEST" reduced to lower case on purpose?

pkg/tests/end_to_end_tests/migrate_test.go
```diff
@@ -137,15 +138,15 @@ func TestMigrateTwice(t *testing.T) {
...
 	testhelpers.WithDB(t, *testDatabase, testhelpers.NoSuperuser, func(db *pgxpool.Pool, t testing.TB, connectURL string) {
 		performMigrate(t, connectURL, testhelpers.PgConnectURL(*testDatabase, testhelpers.Superuser))
-		if *useExtension && !pgmodel.ExtensionIsInstalled {
+		if *useExtension && !utils.ExtensionIsInstalled {
 			t.Errorf("extension is not installed, expected it to be installed")
 		}
 
 		//reset the flag to make sure it's set correctly again.
-		pgmodel.ExtensionIsInstalled = false
+		utils.ExtensionIsInstalled = false
```
blagojts, line 146: man this feels wrong. I really really don't like global state. But I would leave this for another PR

blagojts (changes requested):
I love the new changes, especially that the pgmodel 
package is no longer huge. 🔥 

But I dislike the utils package. 'Utils' in general is 
a code smell, since it provides no real information. 
I would suggest we move some of the things currently there 
in the `pgmodel/common` package
or a new `pgmodel/model` package (and rename `pgmodel`). 
More specifically the model package should contain:
 - labels.go
 - labels_reader.go
 - samples.go
 - MetricNameLabelName, MinTime, MaxTime consts
 - SeriesId as a type 
since these all are very data model related, they are
not just utilities.
`utils/extension` should go to `common` and I agree with
Ante that utils.go should go into `common`

cevian:
I am taking myself and Josh off this review. Will let Ante and Blagoj shepherd this one.

pkg/tests/end_to_end_tests/drop_test.go
Harkishen-Singh: I guess, it happened while resolving conflicts.

pkg/pgmodel/ingestor/buffer.go
Harkishen-Singh: I am not sure. The changes as such are huge. Maybe, a follow on PR (by me or someone else who has more used to epoch code) for this and the comment below for global state?

pkg/pgmodel/common/extension/extension.go
```diff
@@ -2,7 +2,7 @@
 // Please see the included NOTICE for copyright information and
 // LICENSE for a copy of the license.
 
-package pgmodel
+package extension
```
blagojts, line 5: ❤️

pkg/pgmodel/ingestor/buffer.go
blagojts: 👍

blagojts (approved):

antekresic (approved):
👍

PR #394, opened 2020-12-21
add flags for upgrade-extensions

VineethReddy02:
**--install-extensions**: It installs the extenion if not present. By default set to true.  Formerly this flag was **install-timescaledb**

**--upgrade-extensions**: Upgrade extension to available next version. By default set to true.

**--upgrade-prerelease-extensions**: Upgrade to available highest prerelease if available. By default set to false.

1. Install TimescaleDB & Promscale extensions only if ``--install-extensions`` is enabled.
2. Upgrade TimescaleDB & Promscale to available latest version if ``--upgrade-extensions`` is enabled.
3. Upgrade TimescaleDB & Promscale to available latest `rc` version if --upgrade-prerelease-extensions`` is enabled.

Note:

1. If the current installed version in timescaleDB is an `rc` version we ignore `--upgrade-prerelease-extensions` flag that is set to `false` & will let promscale run with the already installed `rc` version.  

### DIFF
--- a/pkg/pgmodel/common/extension/extension.go
+++ b/pkg/pgmodel/common/extension/extension.go
@@ -113,7 +113,7 @@ func checkPromscaleExtensionVersion(conn *pgx.Conn) error {
 	return nil
 }
 
-func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, validRange semver.Range, rangeString string) error {
+func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, validRange semver.Range, rangeString string, upgradeExt, upgradePreRelease bool) error {
 	availableVersions, err := fetchAvailableExtensionVersions(conn, extName)
 	if err != nil {
 		return err
@@ -132,9 +132,17 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		return fmt.Errorf("Could not get the installed extension version: %w", err)
 	}
 
-	newVersion, ok := getNewExtensionVersion(extName, availableVersions, defaultVersion, validRange, isInstalled, currentVersion)
+	if isInstalled && !upgradeExt {
+		log.Info("msg", "skipping "+extName+" extension upgrade as upgrade extension is disabled. The current extension version is "+currentVersion.String())
+		if !validRange(currentVersion) {
+			return fmt.Errorf("The %v extension is not installed at the right version and upgrades are disabled, need version: %v and the installed version is %s ", extName, rangeString, currentVersion)
+		}
+		return nil
+	}
+
+	newVersion, ok := getNewExtensionVersion(extName, availableVersions, defaultVersion, validRange, isInstalled, upgradePreRelease, currentVersion)
 	if !ok {
-		return fmt.Errorf("the extension is not available at the right version, need version: %v", rangeString)
+		return fmt.Errorf("The %v extension is not available at the right version, need version: %v and the default version is %s ", extName, rangeString, defaultVersion)
 	}
 
 	if !isInstalled {
@@ -144,6 +152,7 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		if extErr != nil {
 			return extErr
 		}
+		log.Info("msg", "successfully installed "+extName+" extension of version "+newVersion.String())
 		return nil
 	}
 
@@ -169,8 +178,12 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		_, err := connAlter.Exec(context.Background(),
 			fmt.Sprintf("ALTER EXTENSION %s UPDATE TO '%s'", extName,
 				getSqlVersion(newVersion, extName)))
+		// if migration fails, Do not crash just log an error. As there is an extension already present.
 		if err != nil {
-			return err
+			if !validRange(currentVersion) {
+				return fmt.Errorf("The %v extension is not installed at the right version and the upgrades failed, need version: %v and the installed version is %s and the upgrade failed with %w ", extName, rangeString, currentVersion, err)
+			}
+			log.Error("msg", fmt.Sprintf("Failed to migrate extension %v from %v to %v: %v", extName, currentVersion, newVersion, err))
 		}
 	}
 
@@ -261,7 +274,7 @@ func getNewExtensionVersion(extName string,
 	availableVersions semver.Versions,
 	defaultVersion semver.Version,
 	validRange semver.Range,
-	validCurrentVersion bool,
+	validCurrentVersion, upgradePreRelease bool,
 	currentVersion semver.Version) (semver.Version, bool) {
 	//sort higher extensions first
 	sort.Sort(sort.Reverse(availableVersions))
@@ -271,6 +284,14 @@ func getNewExtensionVersion(extName string,
 		if availableVersions[i].GT(defaultVersion) {
 			continue
 		}
+
+		// if upgradePreRelease is false skip the pre-releases.
+		// if the current installed version is an rc version return it as an valid available version.
+		if len(availableVersions[i].Pre) > 0 && !upgradePreRelease && availableVersions[i].String() != currentVersion.String() {
+			log.Warn("msg", "skipping upgrade to prerelease version "+availableVersions[i].String()+" as --upgrade-prerelease-extensions is disabled.")
+			continue
+		}
+
 		/* Do not auto-upgrade across Major versions of extensions */
 		if validCurrentVersion && currentVersion.Major != availableVersions[i].Major {
 			/* Print a warning if there is a a non-prerelease newer major version available */
--- a/pkg/pgmodel/migrate.go
+++ b/pkg/pgmodel/migrate.go
@@ -93,23 +93,39 @@ func (p prefixedNames) getNames() []string {
 // Also this takes a connection string not a connection because for
 // updates the ALTER has to be the first command on the connection
 // thus we cannot reuse existing connections
-func MigrateTimescaleDBExtension(connstr string) error {
+func InstallUpgradeTimescaleDBExtensions(connstr string, upgradeExt, upgradePre bool) error {
 	db, err := pgx.Connect(context.Background(), connstr)
 	if err != nil {
 		return err
 	}
 	defer func() { _ = db.Close(context.Background()) }()
 
-	err = extension.MigrateExtension(db, "timescaledb", "public", version.TimescaleVersionRange, version.TimescaleVersionRangeFullString)
+	err = extension.MigrateExtension(db, "timescaledb", "public", version.TimescaleVersionRange, version.TimescaleVersionRangeFullString, upgradeExt, upgradePre)
 	if err != nil {
 		return fmt.Errorf("could not install timescaledb: %w", err)
 	}
+
+	return nil
+
+}
+
+func installUpgradePromscaleExtensions(db *pgx.Conn, upgradeExt, upgradePre bool) error {
+	extension.ExtensionIsInstalled = false
+	err := extension.MigrateExtension(db, "promscale", schema.Ext, version.ExtVersionRange, version.ExtVersionRangeString, upgradeExt, upgradePre)
+	if err != nil {
+		log.Warn("msg", fmt.Sprintf("could not install promscale: %v. continuing without extension", err))
+	}
+
+	if err = extension.CheckExtensionsVersion(db); err != nil {
+		return fmt.Errorf("error encountered while migrating extension: %w", err)
+	}
+
 	return nil
 
 }
 
 // Migrate performs a database migration to the latest version
-func Migrate(db *pgx.Conn, versionInfo VersionInfo) (err error) {
+func Migrate(db *pgx.Conn, versionInfo VersionInfo, upgradeExt, upgradePre bool) (err error) {
 	migrateMutex.Lock()
 	defer migrateMutex.Unlock()
 	extension.ExtensionIsInstalled = false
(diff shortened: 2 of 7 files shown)


pkg/pgmodel/migrate.go
```diff
@@ -123,16 +145,6 @@ func Migrate(db *pgx.Conn, versionInfo VersionInfo) (err error) {
...
-	err = migrateExtension(db, "promscale", extSchema, version.ExtVersionRange, version.ExtVersionRangeString)
-	if err != nil {
-		log.Warn("msg", fmt.Sprintf("could not install promscale: %v. continuing without extension", err))
-	}
-
-	if err = checkExtensionsVersion(db); err != nil {
-		return fmt.Errorf("Error encountered while migrating extension: %w", err)
-	}
-
 	metadataUpdate(db, ExtensionIsInstalled, "version", versionInfo.Version)
```
cevian, line 148: This line requires ExtensionIsInstalled to be set correctly. Right now, it's only set after this function returns, when `InstallUpgradePromscaleExtensions` is called. Which won't work. Maybe this should be split out into a separate function called from CreateClient or just move the `InstallUpgradePromscaleExtensions` call back in here ?

pkg/pgmodel/extension.go
```diff
@@ -270,6 +276,13 @@ func getNewExtensionVersion(extName string,
 		if availableVersions[i].GT(defaultVersion) {
 			continue
 		}
+
+		// if upgradePreRelease is false skip the pre-releases.
+		// if the current installed version is an rc version return it as an valid available version.
+		if len(availableVersions[i].Pre) > 0 && !upgradePreRelease && availableVersions[i].String() != currentVersion.String() {
```
cevian, line 282: I would log a warning here like we do in the major version update check to let people know a pre-release is available but we won't be updating to it.

cevian (changes requested):
Some minor fixes necessary

pkg/pgmodel/extension.go
VineethReddy02: But this logs multiple times If there are multiple `rc` versions available.

cevian (approved):

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +635,213 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+			t.Fatal(err)
+		}
+		err = db.QueryRow(ctx, `SELECT extversion FROM pg_extension where extname='timescaledb'`).Scan(&version)
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+		db.Close(ctx)
+		// Note this needs to be fixed the expected version is 1.7.4 but
+		// db is failing to update with an error ERROR: extension "timescaledb" cannot be updated after the old version has already been loaded
+		if version != "1.7.3" {
```
VineethReddy02, line 790: @cevian I noticed an issue when writing the e2e test. By default, if timescaleDB installed version is 1.7.3 & the timescaleDB image has now 1.7.4 & 2.0.0 version available.

The expected behavior is promscale on startup should upgrade timescaleDB version from `1.7.3` to `1.7.4` . But while performing this DB returns an error  `ERROR: extension "timescaledb" cannot be updated after the old version has already been loaded` not sure about the exact reason as this is from tsdb. I am also verified promscale is creating a new conn before peforming ext upgrade. 

This isn't a blocker for PR to merge. But the e2e case can be fulfilled with the workaround.

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +635,213 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+	ctx := context.Background()
+
+	db, dbContainer, closer := startDB(t, ctx)
+	defer testhelpers.StopContainer(ctx, dbContainer, false)
+
+	defer closer.Close()
+
+	// start promscale & test upgrade-extensions as false
+	func() {
+		connectorImage := "vineeth97/promscale:t1"
```
VineethReddy02, line 731: Currently, I am using image with changes from this branch to make tests succeed but after the next release. I can update to the official promscale latest image.

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +635,213 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dataDir, err := testhelpers.TempDir("update_test_data")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dbContainer, closer, err := testhelpers.StartDatabaseImage(ctx, "vineeth97/timescaledb-ext-test", tmpDir, dataDir, *printLogs, extensionState)
```
VineethReddy02, line 810: The timescaleDB image default & installed version are frequently changed as the new extensions are released. I am maintaining a custom image with ext versions required for the tests. 

Is that fine with the image is from my public docker repo?

cevian: I can push this no problem but first I'd like to know how you generated this image.

pkg/tests/upgrade_tests/upgrade_test.go
cevian: This seems like a big issue I'll have to look into it

pkg/tests/upgrade_tests/upgrade_test.go
cevian: what we should do here is use the tag timescale/promscale:latest and have the ci execute something like https://github.com/timescale/promscale/blob/master/docker-compose/high-availability/test.sh#L11 before this test

pkg/tests/upgrade_tests/upgrade_test.go
cevian: hmm when I run this into this it seems like the error is actually "ERROR:  extension "timescaledb" has no update path from version "1.7.3" to version "1.7.4"  on the db side. And that's because the only upgrade path in the docker image `vineeth97/timescaledb-ext-test` is TO 2.0.0-rc4 so 1.7.3-> 2.0.0-rc4 not 1.7.3->1.7.4 (we usually only include upgrade scripts to one version in our docker images). Perhaps we should use another image for this test?

pkg/pgmodel/common/extension/extension.go
```diff
@@ -169,8 +175,9 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		_, err := connAlter.Exec(context.Background(),
 			fmt.Sprintf("ALTER EXTENSION %s UPDATE TO '%s'", extName,
 				getSqlVersion(newVersion, extName)))
+		// if migration fails, Do not crash just log an error. As there is an extension already present.
 		if err != nil {
-			return err
+			log.Error("failed to migrate extension for newer version i.e. ", newVersion, err)
```
cevian, line 180:
```suggestion
			log.Errorf("Failed to migrate extension %v from %v to %v: %v", extName, currentVersion, newVersion, err)
```

pkg/pgmodel/common/extension/extension.go
```diff
@@ -132,9 +132,14 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		return fmt.Errorf("Could not get the installed extension version: %w", err)
 	}
 
-	newVersion, ok := getNewExtensionVersion(extName, availableVersions, defaultVersion, validRange, isInstalled, currentVersion)
+	if isInstalled && !upgradeExt {
+		log.Info("msg", "skipping "+extName+" extension upgrade as upgrade extension is disabled. The current extension version is "+currentVersion.String())
```
cevian, line 136: Important: Sorry missed this before. We still have to validate the extension version if not upgrading
```suggestion
		log.Info("msg", "skipping "+extName+" extension upgrade as upgrade extension is disabled. The current extension version is "+currentVersion.String())
		if !validRange(currentVersion) {
		     return fmt.Errorf("The %v extension is not installed at the right version and upgrades are disabled, need version: %v and the installed version is %s", extName, rangeString, currentVersion)
		}
```

pkg/pgmodel/common/extension/extension.go
```diff
@@ -132,9 +132,14 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
...
+		return nil
+	}
+
+	newVersion, ok := getNewExtensionVersion(extName, availableVersions, defaultVersion, validRange, isInstalled, upgradePreRelease, currentVersion)
 	if !ok {
-		return fmt.Errorf("the extension is not available at the right version, need version: %v", rangeString)
+		return fmt.Errorf("the extension is not available at the right version, need version: %v and the default version is %s", rangeString, defaultVersion)
```
cevian, line 142: need to add extension name to log line
```suggestion
		return fmt.Errorf("The %v extension is not available at the right version, need version: %v and the default version is %s", extName, rangeString, defaultVersion)
```

pkg/pgmodel/common/extension/extension.go
```diff
@@ -271,6 +278,14 @@ func getNewExtensionVersion(extName string,
 		if availableVersions[i].GT(defaultVersion) {
 			continue
 		}
+
+		// if upgradePreRelease is false skip the pre-releases.
+		// if the current installed version is an rc version return it as an valid available version.
+		if len(availableVersions[i].Pre) > 0 && !upgradePreRelease && availableVersions[i].String() != currentVersion.String() {
+			log.Warn("msg", "skipping upgrade to rc version "+availableVersions[i].String()+" as --upgrade-prerelease-extensions is disabled.")
```
cevian, line 285:
```suggestion
			log.Warn("msg", "skipping upgrade to prerelease version "+availableVersions[i].String()+" as --upgrade-prerelease-extensions is disabled.")
```

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +636,234 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+			t.Fatal(err.Error())
+		}
+		defer testhelpers.StopContainer(ctx, connector, *printLogs)
+
+		err = db.QueryRow(ctx, `SELECT extversion FROM pg_extension where extname='timescaledb'`).Scan(&version)
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+
+		if version != "2.0.0-rc4" {
```
cevian, line 669: why is the result an rc if -upgrade-prerelease-extensions = false. This needs some comments to explain.

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +636,234 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+	err = db.QueryRow(ctx, `SELECT extversion FROM pg_extension where extname='timescaledb'`).Scan(&version)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	db.Close(ctx)
+	if version != extVersion {
+		t.Fatal("failed to verify upgrade extension with -upgrade-prerelease-extension false")
+	}
+
```
cevian, line 691: I would maybe run  --upgrade-prerelease-extensions = false once here on -rc2 to make sure that it does not upgrade to rc4. then run with true

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -635,3 +636,234 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
...
+		if exitCode != 0 || err != nil {
+			t.Fatal("promscale failed to run extension migration failure", err)
+		}
+	}()
+
+	os.Exit(0)
+}
+
+func buildPromscaleImageFromRepo(t *testing.T) {
+	t.Log("building promscle image from the codebase")
```
cevian, line 812: I would maybe make sure this runs only once (for efficiency)

cevian (changes requested):
Sorry I missed some issues on last review. But I think we are getting close.

franck102:
Thanks cevian and others for addressing this. So I assume that

1. This is not merged yet and so not in the 0.1.4 release?
2. by setting all 3 flags to false, we will ensure that promscale never tries to install or upgrade any extension at runtime?

VineethReddy02:
yes, @franck102 you will have full control over extensions with these flags.

pkg/pgmodel/common/extension/extension.go
```diff
@@ -169,8 +178,9 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 		_, err := connAlter.Exec(context.Background(),
 			fmt.Sprintf("ALTER EXTENSION %s UPDATE TO '%s'", extName,
 				getSqlVersion(newVersion, extName)))
+		// if migration fails, Do not crash just log an error. As there is an extension already present.
 		if err != nil {
-			return err
+			log.Error("msg", fmt.Sprintf("Failed to migrate extension %v from %v to %v: %v", extName, currentVersion, newVersion, err))
```
cevian, line 183: ugh still not quite right I'm sorry. we can ignore upgrade failures only if the current version is still valid
```suggestion
                         if !validRange(currentVersion) {
			    return fmt.Errorf("The %v extension is not installed at the right version and the upgrades failed, need version: %v and the installed version is %s and the upgrade failed with %w", extName, rangeString, currentVersion, err)
			}
			log.Error("msg", fmt.Sprintf("Failed to migrate extension %v from %v to %v: %v", extName, currentVersion, newVersion, err))
```

cevian (approved):
One important change to make, but I don't need to re-review unless you have questions

cevian:
@franck102 yes these flags are not in 0.1.4 yet.

cevian (approved):

PR #398, opened 2020-12-23
Add retry on 422 HTTP status code from Prometheus in end-to-end tests

antekresic:
Prometheus can run background processes on its internal database to compact
the dataset, if a query is ran at this time, it is possible to get a 422
HTTP response. In this case, we just retry. This commit also adds some custom
Prometheus configuration to preserve the dataset intact.


Harkishen-Singh (approved):
LGTM! Finally :1st_place_medal:

PR #399, opened 2020-12-24
Fix sample throughput report interval

### LINKED ISSUE #367: tput-report does respect the interval
```
./promscale -db-ssl-mode=allow -db-password=*** -log-format=logfmt -tput-report=100
```

Is expected to report the through-put (samples/sec) every 100 seconds. But, the reporting happens every second.

@cevian

antekresic:
Applied the `-tput-report` flag value to the write throughput calculator in
the exposed Prometheus metrics.

Fixes #367

### DIFF
--- a/pkg/api/metrics.go
+++ b/pkg/api/metrics.go
@@ -7,7 +7,7 @@ import (
 	"github.com/timescale/promscale/pkg/util"
 )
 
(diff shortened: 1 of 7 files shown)


Harkishen-Singh (approved):
LGTM!

PR #400, opened 2020-12-24
Prom-migrator: adds support for concurrent pulling of data

Harkishen-Singh:

PR must be merged after merging #385.

TODO (dependent on other PRs)

- [x] Update the flag `num-shards` to `concurrent-pushes` so as to keep the names in line with `concurrent-pulls`
- [x] Verify concurrent pulls working after merging with writer bars
- [x] Update the docs with new CLI flags (concurrent pullings and ensure dependent PR's flags are mentioned)
- [x] Perform an actual migration test before merging


cmd/prom-migrator/main.go
```diff
@@ -130,6 +131,8 @@ func parseFlags(conf *config, args []string) {
 		"Setting this value less than zero will indicate all data from mint upto now. ")
 	flag.StringVar(&conf.maxBlockSize, "max-read-size", "500MB", "(units: B, KB, MB, GB, TB, PB) the maximum size of data that should be read at a single time. "+
 		"More the read size, faster will be the migration but higher will be the memory usage. Example: 250MB.")
+	flag.IntVar(&conf.concurrentPulls, "concurrent-pulls", 1, "Concurrent pulls enables fetching of data parallelly. "+
```
blagojts, line 134:
```suggestion
	flag.IntVar(&conf.concurrentPulls, "concurrent-pulls", 1, "Concurrent pulls enables fetching of data concurrently. "+
```

cmd/prom-migrator/main.go
```diff
@@ -130,6 +131,8 @@ func parseFlags(conf *config, args []string) {
...
+		"This enables higher throughput for significantly faster pulling of data remote-read storages.")
```
blagojts, line 135:
```suggestion
		"May enable higher throughput by pulling data faster from remote-read stores.")
```

pkg/migration-tool/planner/block.go
```diff
@@ -21,24 +21,89 @@ type Block struct {
...
+		contexts[i] = cctx
+		cancelFuncs[i] = cancelFunc
+	}
+	var (
+		doneChan = make(chan *utils.PrompbResponse, totalRequests)
+		errChan  = make(chan error, totalRequests)
+	)
+	for i := 0; i < totalRequests; i++ {
+		// Initiate concurrent fetching of data.
+		go client.Read(contexts[i], readRequests[i], i, fmt.Sprintf("%s | time-range: %d mins", b.pbarDescriptionPrefix, timeRangeMinutesDelta), doneChan, errChan)
```
blagojts, line 80:
```suggestion
                descriptionMsg := fmt.Sprintf("%s | time-range: %d mins", b.pbarDescriptionPrefix, timeRangeMinutesDelta)
		go client.Read(contexts[i], readRequests[i], i, descriptionMsg, doneChan, errChan)
```

pkg/migration-tool/planner/block.go
```diff
@@ -21,24 +21,89 @@ type Block struct {
...
-		return nil, fmt.Errorf("executing client-read: %w", err)
+	close(doneChan)
+	close(errChan)
+	// Combine the stores.
+	var result []*prompb.TimeSeries
+	for storeIndex := 0; storeIndex < b.numStores; storeIndex++ {
+		result = append(result, readResponses[storeIndex].Timeseries...)
 	}
-	b.timeseries = result.Timeseries
+	b.timeseries = result
```
blagojts, line 106: Was there a reason why the result is cached in the block? (since I see it also returned from this func)

pkg/migration-tool/planner/planner.go
```diff
@@ -99,11 +100,18 @@ func (c *Config) fetchLastPushedMaxt() (lastPushedMaxt int64, found bool, err er
 	if err != nil {
 		return -1, false, fmt.Errorf("create fetch-last-pushed-maxt reader: %w", err)
 	}
-	result, _, _, err := readClient.Read(context.Background(), query, "")
-	if err != nil {
+	var (
+		doneChan = make(chan *utils.PrompbResponse, 1)
+		errChan  = make(chan error, 1)
+	)
+	go readClient.Read(context.Background(), query, 0, "", doneChan, errChan)
```
blagojts, line 107: any special need to have `readClient.Read` be a goroutine now? since it spawns the concurrent routines, and here we just wait for that one routine to finish?
if there is an explicit reason why we need this to be a routine, consider using a wait group here instead of the two channels

blagojts (approved):
Looks good 👍 
I have some small comments/suggestions and some questions throughout the code. No need for re-review after applying/ignoring the suggestions 😃

pkg/migration-tool/planner/planner.go
```diff
@@ -140,7 +148,7 @@ func (p *Plan) LastMemoryFootprint() int64 {
 }
 
 // NewBlock returns a new block after allocating the time-range for fetch.
-func (p *Plan) NextBlock() (reference *Block, err error) {
+func (p *Plan) NextBlock(numStores int) (reference *Block, err error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: We do not expect it to change during migration. So, yes it can be part of Plan.

cmd/prom-migrator/main.go
Harkishen-Singh: Sorry, we are definite about this. Parallel fetching will always increase the throughput since requests sent are concurrent.

pkg/migration-tool/planner/block.go
Harkishen-Singh: This is to return the combined stores (as blocks) to the writer. But yes, we should make the stores `nil` after this operation to save memory.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: This is to fetch parallely. Imagine a request of 10 mins. If suppose `-concurrent-pulls=5`, then the request gets divided into:
1. first routine handles 1-2 mins.
2. second routine handles 3-4 mins.
3. third routine handles 5-6 mins.
4. ...

And after each go-routine has completed the fetching process, we combine these to form a block and sent to the writer.

The reason why I didn't use wait group is since we anyways need a channel to get the response, so why take another parameter for waitgroup. Plus, two channels acn be decreased to one by using interface channel. But this change is to be made in the writer as well, which can be done only once other PRs are merged. hence, I will do this in a different commit.

pkg/migration-tool/planner/block.go
blagojts: but what about b.timeseries?
result is assigned to it here, and result is returned from the method, why is b.timeseries needed?

pkg/migration-tool/planner/planner.go
blagojts: I wouldn't use a single channel for the done and error.

And do we need readClient.Read to go in to a separate routine?

cmd/prom-migrator/main.go
blagojts: If remote-read system is slow (like really slow), issuing more requests can lead to congestion

pkg/migration-tool/planner/block.go
Harkishen-Singh: `b.timeseries` is used to return the timeseries in the writer (https://github.com/timescale/promscale/blob/5f87ccd9f09a8aaa2549e441d79120e4f28f1e98/pkg/migration-tool/planner/block.go#L62). The result is returned from here only to make sure that this result is not empty.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: > And do we need readClient.Read to go in to a separate routine?

How would we do parallel fetching without it?

blagojts: isn't readClient.Read the one starting the goroutines that fetch the data concurrently?

pkg/migration-tool/planner/block.go
Harkishen-Singh: But yes, this does points out an important point. The function does not need to return such hefty values that will slow the registers. Since this returns only determins whether its empty or not, we can return `true` or `false`, for `containsTs` and this would be far efficient. Thank you for pointing this out.

pkg/migration-tool/planner/planner.go
Harkishen-Singh: We want the func calls (to a single function) happen concurrently, as through the same method is called, but each call does a different purpose (similar in nature i.e., fetching of data). And to make this faster, we can allow this to run parallely.

Harkishen-Singh: Ah, thank you very much for pointing this out. I called this async way while dev, but it was not at all required.

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,103 @@ type Block struct {
...
 	// We set compressed bytes in block since those are the bytes that will be pushed over the network to the write storage after snappy compression.
 	// The pushed bytes are not exactly the bytesCompressed since while pushing, we add the progress metric. But,
 	// the size of progress metric along with the sample is negligible. So, it is safe to consider bytesCompressed
 	// in such a scenario.
 	b.numBytesCompressed = bytesCompressed
 	b.numBytesUncompressed = bytesUncompressed
 	b.plan.update(bytesUncompressed)
-	return result, nil
+	return len(result) != 0, nil
```
cevian, line 124: how is this different from len(b.timeseries) != 0 and if it's not why do we need a separate return?

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,103 @@ type Block struct {
...
+		cancelFuncs[i] = cancelFunc
+	}
+	var (
+		doneChan = make(chan *utils.PrompbResponse, totalRequests)
+		errChan  = make(chan error, totalRequests)
+	)
+	for i := 0; i < totalRequests; i++ {
+		// Initiate concurrent fetching of data.
+		descriptionMsg := fmt.Sprintf("%s | time-range: %d mins", b.pbarDescriptionPrefix, timeRangeMinutesDelta)
+		go client.Read(contexts[i], readRequests[i], i, descriptionMsg, doneChan, errChan)
```
cevian, line 90: why don't we simply move these lines up to line 81 in the previous loop. That way we don't need all these slices for contexts, readRequests etc.

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,103 @@ type Block struct {
...
+			return false, fmt.Errorf("create promb query: %w", err)
+		}
+		readRequests[i] = readRequest
+		cctx, cancelFunc := context.WithCancel(ctx)
+		contexts[i] = cctx
+		cancelFuncs[i] = cancelFunc
+	}
+	var (
+		doneChan = make(chan *utils.PrompbResponse, totalRequests)
+		errChan  = make(chan error, totalRequests)
```
cevian, line 85: This should move above the previous loop

pkg/migration-tool/planner/planner.go
```diff
@@ -104,11 +106,18 @@ func (c *Config) fetchLastPushedMaxt() (lastPushedMaxt int64, found bool, err er
...
+	var (
+		doneChan = make(chan *utils.PrompbResponse, 1)
+		errChan  = make(chan error, 1)
+	)
+	readClient.Read(context.Background(), query, 0, "", doneChan, errChan)
+	var ts []*prompb.TimeSeries
+	select {
+	case resp := <-doneChan:
+		ts = resp.Result.Timeseries
+	case err := <-errChan:
```
cevian, line 118: This seems silly. Why dont we have the client have two functions Read and ReadChannels where the latter responds on channels while the former has regular returns. In fact it's easy to write ReadChannels by calling Read.

pkg/migration-tool/utils/client.go
```diff
@@ -85,22 +85,32 @@ func NewClient(remoteName, urlString string, clientType uint, timeout model.Dura
...
+type PrompbResponse struct {
+	ID                   int
+	Result               *prompb.QueryResult
+	NumBytesCompressed   int
+	NumBytesUncompressed int
+}
+
 // Read reads from a remote endpoint. It returns the response size of compressed and uncompressed in bytes.
-func (c *Client) Read(ctx context.Context, query *prompb.Query, desc string) (result *prompb.QueryResult, numBytesCompressed int, numBytesUncompressed int, err error) {
+func (c *Client) Read(ctx context.Context, query *prompb.Query, shardID int, desc string, doneChan chan<- *PrompbResponse, errChan chan<- error) {
```
cevian, line 97: As I mentioned before, I'd keep Read as it was and just add a ReadChannels that responds on channels.

pkg/migration-tool/reader/reader.go
```diff
@@ -81,12 +79,12 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
 			}
 			ms := []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")}
-			result, err = blockRef.Fetch(rr.c, rr.client, blockRef.Mint(), blockRef.Maxt(), ms)
+			containsTs, err = blockRef.Fetch(rr.c, rr.client, blockRef.Mint(), blockRef.Maxt(), ms)
 			if err != nil {
 				errChan <- fmt.Errorf("remote-run run: %w", err)
 				return
 			}
-			if len(result.Timeseries) == 0 {
+			if !containsTs {
```
cevian, line 87: why not `len(blockRef.Timeseries) == 0`

cevian (changes requested):

pkg/migration-tool/planner/block.go
Harkishen-Singh: The only reason why earlier we used to return result complete is it was being used to check the length of the result. If the length is 0, decrement the block count. That's it. This simple logical operation can be done here itself and hence make this more efficient by decreasing the functional return value.

pkg/migration-tool/utils/client.go
Harkishen-Singh: Agreed. That would be much better.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Same as https://github.com/timescale/promscale/pull/400#discussion_r551758175.

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,100 @@ type Block struct {
...
+			return false, fmt.Errorf("create promb query: %w", err)
+		}
+		readRequests[i] = readRequest
+		cctx, cancelFunc := context.WithCancel(ctx)
+		contexts[i] = cctx
+		cancelFuncs[i] = cancelFunc
+		// Initiate concurrent fetching of data.
+		descriptionMsg := fmt.Sprintf("%s | time-range: %d mins", b.pbarDescriptionPrefix, timeRangeMinutesDelta)
+		go client.ReadChannels(contexts[i], readRequests[i], i, descriptionMsg, doneChan, errChan)
+	}
```
cevian, line 88: You don't need the ReadRequests, ReadResponses or Context slices
```suggestion
		totalRequests         = b.numStores
		cancelFuncs           = make([]context.CancelFunc, totalRequests)
		timeRangeMinutesDelta = (maxt - mint) / time.Minute.Milliseconds()

		doneChan = make(chan *utils.PrompbResponse, totalRequests)
		errChan  = make(chan error, totalRequests)
	)
	for i := 0; i < totalRequests; i++ {
		readRequest, err := utils.CreatePrombQuery(b.stores[i].mint, b.stores[i].maxt, matchers)
		if err != nil {
			return false, fmt.Errorf("create promb query: %w", err)
		}
		cctx, cancelFunc := context.WithCancel(ctx)
		cancelFuncs[i] = cancelFunc
		// Initiate concurrent fetching of data.
		descriptionMsg := fmt.Sprintf("%s | time-range: %d mins", b.pbarDescriptionPrefix, timeRangeMinutesDelta)
		go client.ReadChannels(cctx, readRequests, i, descriptionMsg, doneChan, errChan)
	}
```

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,100 @@ type Block struct {
...
+	var (
+		bytesCompressed   int
+		bytesUncompressed int
+	)
+	for i := 0; i < totalRequests; i++ {
+		select {
+		case resp := <-doneChan:
+			bytesCompressed += resp.NumBytesCompressed
+			bytesUncompressed += resp.NumBytesUncompressed
+			readResponses[resp.ID] = resp.Result
```
cevian, line 98:
```suggestion
			b.timeseries = append(b.timeseries, resp.Result.Timeseries...)
```

pkg/migration-tool/planner/block.go
```diff
@@ -25,32 +25,100 @@ type Block struct {
...
-		return nil, fmt.Errorf("executing client-read: %w", err)
+	close(doneChan)
+	close(errChan)
+	// Combine the stores.
+	var result []*prompb.TimeSeries
+	for storeIndex := 0; storeIndex < b.numStores; storeIndex++ {
+		result = append(result, readResponses[storeIndex].Timeseries...)
 	}
-	b.timeseries = result.Timeseries
+	b.timeseries = result
```
cevian, line 113: This can be deleted if we just append to b.timeseries above.

pkg/migration-tool/planner/block.go
cevian: sorry but we already have b.timeseries which is what we need for the check. We don't actually need neither result not len(result)!=0.

pkg/migration-tool/reader/reader.go
cevian: again we don't need a result here. we need `len(blockRef.timeseries) == 0` or am I missing something?

pkg/migration-tool/utils/client.go
```diff
@@ -160,6 +159,28 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query, desc string) (re
...
+	Result               *prompb.QueryResult
+	NumBytesCompressed   int
+	NumBytesUncompressed int
+}
+
+// ReadChannels calls the Read and responds on the channels.
+func (c *Client) ReadChannels(ctx context.Context, query *prompb.Query, shardID int, desc string, doneChan chan<- *PrompbResponse, errChan chan<- error) {
+	result, numBytesCompressed, numBytesUncompressed, err := c.Read(ctx, query, desc)
+	if err != nil {
+		errChan <- fmt.Errorf("read-channels: %w", err)
```
cevian, line 174: important: I believe the user of this call expects either an error response or a done response but not both
```suggestion
		errChan <- fmt.Errorf("read-channels: %w", err)
		return
```

cevian (changes requested):
Some changes requested for cleanup.

pkg/migration-tool/planner/block.go
Harkishen-Singh: Ah, thank you very much for this. This will save significant memory.

pkg/migration-tool/utils/client.go
Harkishen-Singh: Thank you for pointing this out.

pkg/migration-tool/planner/block.go
Harkishen-Singh: Yeah I agree, that would be very efficient. The b.timeseries thing. I have updated accordingly. Thank you.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Yes you are right. That way, it would be efficient. I have updated it accordingly. Thank you very much.

cmd/prom-migrator/main.go
```diff
@@ -135,8 +137,13 @@ func parseFlags(conf *config, args []string) {
 		"Setting this value less than zero will indicate all data from mint upto now. ")
 	flag.StringVar(&conf.maxBlockSize, "max-read-size", "500MB", "(units: B, KB, MB, GB, TB, PB) the maximum size of data that should be read at a single time. "+
 		"More the read size, faster will be the migration but higher will be the memory usage. Example: 250MB.")
-	flag.IntVar(&conf.numShards, "num-shards", 4, "Number of shards to split a fetched block in. Each shard is written concurrently. "+
+	flag.IntVar(&conf.concurrentPush, "num-shards", 4, "Number of shards to split a fetched block in. Each shard is written concurrently. "+
 		"Note: Larger shards count will lead to significant memory usage.")
+	flag.IntVar(&conf.concurrentPulls, "concurrent-pulls", 1, "Concurrent pulls enables fetching of data concurrently. "+
+		"This may enable higher throughput by pulling data faster from remote-read stores. "+
+		"Note: Setting concurrent-pulls > 1 will disable real-time transfer rate speed while fetching data. Moreover, setting this value too high may cause "+
```
cevian, line 144: Not sure what this means.

cmd/prom-migrator/main.go
```diff
@@ -135,8 +137,13 @@ func parseFlags(conf *config, args []string) {
...
+		"TLS handshake error on the read storage side or may lead to starvation of fetch requests, depending on the bandwidth. "+
```
cevian, line 145: how can this lead to starvation or handshake errors?

cevian (approved):
The code looks ok but have some comments about docs. Also seems like you have a few incomplete TODOs in the PR description before merging

cmd/prom-migrator/main.go
Harkishen-Singh: I have noticed that if you send too many concurrent request to the host, it gives back a `TLS handshake error`, so added it here.

cmd/prom-migrator/main.go
Harkishen-Singh: That is because I didn't update the code of this part earlier (as it was dependent on the previous PR). Now, it is updated and after testing it out, it looks pretty good (ui-ux) as explained here (explanation updated).

PR #403, opened 2020-12-29
HA-implementation

VineethReddy02:
This PR contains the support for prometheus HA by supporting the write load across the promscale instances.  


pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -115,6 +126,85 @@ func (i *DBIngestor) parseData(tts []prompb.TimeSeries, req *prompb.WriteRequest
...
+
+	s, _, err := model.LabelProtosToLabels(tts[0].Labels)
+	if err != nil {
+		return nil, rows, err
+	}
+
+	replicaName := s.GetReplicaName()
+	clusterName := s.GetClusterName()
+
+	if replicaName == "" && clusterName == "" {
```
blagojts, line 149:
```suggestion
	if replicaName == "" || clusterName == "" {
```
also, I think we should use constants for " __replica__ " and "__cluster__"

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -115,6 +126,85 @@ func (i *DBIngestor) parseData(tts []prompb.TimeSeries, req *prompb.WriteRequest
...
+			continue
+		}
+
+		if t.Samples[0].Timestamp < minT || minT == 0 {
+			minT = t.Samples[0].Timestamp
+		}
+
+		if t.Samples[len(t.Samples)-1].Timestamp > maxT {
+			maxT = t.Samples[len(t.Samples)-1].Timestamp
+		}
```
blagojts, line 166: are we sure that the samples in a a single prompb.TimeSeries are ordered by time?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -115,6 +126,85 @@ func (i *DBIngestor) parseData(tts []prompb.TimeSeries, req *prompb.WriteRequest
...
+	// memory blowup, and to allow faster deserializing, we recycle the WriteRequest
+	// here, allowing it to be either garbage collected or reused for a new request.
+	// In order for this to work correctly, any data we wish to keep using (e.g.
+	// samples) must no longer be reachable from req.
+	FinishWriteRequest(req)
+
+	minTUnix := time.Unix(minT, 0)
+	maxTUnix := time.Unix(maxT, 0)
+	if !ha.CheckInsert(minTUnix, maxTUnix, clusterName, replicaName) {
+		log.Warn("the samples aren't from the leader prom instance. skipping the insert")
```
blagojts, line 201: I would reduce this from Warn to Info (or maybe even Debug), since it is expected that in HA mode there will be samples from the non-leader

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -49,10 +52,18 @@ func NewPgxIngestor(conn pgxconn.PgxConn) (*DBIngestor, error) {
 //     tts the []Timeseries to insert
 //     req the WriteRequest backing tts. It will be added to our WriteRequest
 //         pool when it is no longer needed.
-func (i *DBIngestor) Ingest(tts []prompb.TimeSeries, req *prompb.WriteRequest) (uint64, error) {
-	data, totalRows, err := i.parseData(tts, req)
+func (i *DBIngestor) Ingest(tts []prompb.TimeSeries, req *prompb.WriteRequest, ha *ha.HA) (uint64, error) {
```
blagojts, line 55: since the ha state is not modified throughout the lifetime of the ingestor, we're always working with the same instance, it's better to store it as a unexported field of DBIngestor, no need to pass it around

pkg/pgmodel/model/labels.go
```diff
@@ -179,3 +179,21 @@ func (l *Labels) Swap(i, j int) {
 	l.Names[j], l.Names[i] = l.Names[i], l.Names[j]
 	l.Values[j], l.Values[i] = l.Values[i], l.Values[j]
 }
+
+func (l *Labels) GetReplicaName() string {
+	for index, key := range l.Names {
+		if key == "__replica__" {
```
blagojts, line 185: use a const

pkg/runner/runner.go
```diff
@@ -56,6 +60,7 @@ var (
 	appVersion         = pgmodel.VersionInfo{Version: version.Version, CommitHash: version.CommitHash}
 	migrationLockError = fmt.Errorf("Could not acquire migration lock. Ensure there are no other connectors running and try again.")
 	startupError       = fmt.Errorf("startup error")
+	HAState            = &ha.HA{}
```
blagojts, line 63: should be moved in the Run function, can be a local var init'ed there and passed to the components that need it

pkg/pgmodel/ingestor/ingestor.go
blagojts: would also remove the requirement to modify all the tests that call Ingest

pkg/version/version.go
```diff
@@ -47,7 +47,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.0.0",
+		Safe: ">=1.7.3 <=2.0.0",
```
blagojts, line 50: should be separate pr

blagojts (changes requested):
I added some comments throughout the files on how we can make things better structured and easier to test. 
I still have an itch that the parse function doing the write check is not right, but I don't have a better proposal at the moment, hopefully when we re-org the code a bit it will be clearer how to separate those two concerns

pkg/pgmodel/ha/HA.go
```diff
@@ -0,0 +1,73 @@
...
+type HA struct {
+	Cluster         string
+	Leader          string
+	LeaseStart      time.Time
+	LeaseUntil      time.Time
+	MaxTimeSeen     time.Time //max data time seen by any instance
+	MaxTimeInstance string    // the instance name that’s seen the maxtime
+	DBClient        pgxconn.PgxConn
+	LeaseTimeout    string
+	LeaseRefresh    string
```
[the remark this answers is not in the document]
VineethReddy02: The `leaseTimeOut` & `leaseRefresh` are exposed as flags so I made the exportable fields. During Promscale startup the custom values can be assigned. 

I know we thought of having them as mutable in DB level. I will update accordingly.

pkg/pgmodel/ingestor/ingestor.go
VineethReddy02: I checked this on write request we get `[]promb.Timeseries` in which each series contains set of samples sorted based on timestamps.

pkg/pgmodel/ha/HA.go
```diff
@@ -0,0 +1,73 @@
...
+	Enabled         bool
+	Ha              HAWriter
+	mu              sync.Mutex
+}
+
+type HAWriter interface {
```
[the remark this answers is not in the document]
VineethReddy02: Yup, I enhanced the writer part to plug mock writer & actual writer on HAState creation in my recent changes.

pkg/pgmodel/ha/HA.go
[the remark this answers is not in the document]
blagojts: even if configurable by flags, since they are not supposed to change at runtime, it's better to have them set at init (through a constructor func, together with the passed PgxConn instance) and be unexported

blagojts:
Will close this PR because I pushed the same code to this repo and opened up #407 . Additional development will continue as PRs against that branch, which we'll ultimately merge into master

PR #404, opened 2020-12-29
Bump timescaleDB supported range to <2.1.0

VineethReddy02:
Update the supported version range of promscale from `Safe: ">=1.7.3 <2.0.0"` to `Safe: ">=1.7.3 <2.1.0"`

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -47,7 +47,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.0.0",
+		Safe: ">=1.7.3 <2.1.0",
 		Warn: ">=1.7.0 <1.7.3",
 	}
 	timescaleVersionSafeRange = semver.MustParseRange(TimescaleVersionRangeString.Safe)


pkg/version/version.go
```diff
@@ -47,7 +47,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.0.0",
+		Safe: ">=1.7.3 <=2.0.0",
```
VineethReddy02, line 50: @cevian What is support matrix for timescaleDB? Is it latest (i.e. 2.0.0) -1 (i.e. 1.7.4) or -2 (i.e. 1.7.3) ?

pkg/version/version.go
```diff
@@ -37,7 +37,7 @@ var (
 	// since an app version must uniquely determine the state of the schema.
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
-	Version                    = "0.1.4-dev.1"
+	Version                    = "0.1.4"
```
cevian, line 40: No need to change this. I'll do it during release

pkg/version/version.go
cevian, line 50: Let's do >=1.7.3 <2.1.0 so we include any patches 2.0.x

cevian (changes requested):
A few changes requested. No need to bump the app version now. That belongs in the release commit

cevian (approved):

PR #407, opened 2020-12-30
External Labels based HA

atanasovskib:
This feature branch will contain the implementation of 
external labels based HA mode. Each Prometheus 
instance will be able to identify itself with a cluster and 
instance external label. Promscale will make sure to
let only one instance per cluster write to the database.


pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,144 @@
...
+        RAISE EXCEPTION 'LEADER_HAS_CHANGED';
+    END IF;
+
+    SELECT value::INTERVAL
+    INTO lease_refresh
+    FROM SCHEMA_CATALOG.default
+    WHERE key = 'ha_lease_refresh';
+
+    new_lease_timeout = max_time + lease_timeout;
+    IF new_lease_timeout > lease_until + lease_refresh THEN
```
cevian, line 97: note to self: makes this more random so we don't have stampeding herd effect and avoid concurrent updates.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,144 @@
...
+                RAISE EXCEPTION 'LEADER_HAS_CHANGED';
+            END IF;
+        END IF;
+    END IF;
+    RETURN QUERY SELECT cluster, leader, lease_start, lease_until;
+END;
+$func$ LANGUAGE plpgsql VOLATILE ;
+
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.try_change_leader(cluster TEXT, new_leader TEXT,
+                                                            max_time TIMESTAMPTZ) RETURNS SETOF ha_locks
```
cevian, line 119: this should not return setof it only returns one row.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+func (h *Service) tryChangeLeader() {
+	ok := h._leaderChangeLock.TryAcquire(1)
+	if !ok {
+		// change leader already in progress
+		return
+	}
+	defer h._leaderChangeLock.Release(1)
+
+	for {
+		stateView := h.state.clone()
```
cevian, line 90: I don't think we need to readLockState here. I think the early-exit criteria is `if stateView.leaseUntil.After(stateView.maxTimeSeen) then do nothing (exit)`. Remember stateView.leaseUntil should be updated by checkInsert()

if there is no early-exit we can go right to calling tryChangeLeader.

cevian (changes requested):
Some requests for changes. The sql looks mostly good but we need to change when checkInsert is called and make some other changes.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+	// leader changed
+	if err == leaderHasChanged {
+		// read latest lock state
+		lockState, err = h.lockClient.readLockState(context.Background(), clusterName)
+		// couldn't get latest lock state
+		if err != nil {
+			return false, fmt.Errorf("could not check ha lock state: %#v", err)
+		}
+		// update state asynchronously
+		go h.state.update(lockState, replicaName, maxT)
```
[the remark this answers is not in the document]
VineethReddy02: We can definitely change this to be synchronous. But in the current flow, the conditions we are evaluating aren't dependent on the updation because we are explicitly using `lockState`, `replicaName`, & `maxT` in the next checks.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+		go h.tryChangeLeader()
```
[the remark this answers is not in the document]
VineethReddy02: The `checkInsert` either returns the existing `lockState` or returns an error `leaderHasChanged` when the leader has changed we attempt to take over the lock by trying to change leader. Before changing the leader we read the current lockstate and validate 

```
		if stateView.leader == lockState.leader || !lockState.leaseUntil.After(stateView.maxTimeSeen) {
			// no need to change leader
			return
		}  
```

If the condition fails we try to acquire the lock else just skip acquiring the leader task. Maybe we should change the name to be more descriptive to make sense of the action.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+			return true, nil
+		} else {
+
+			return false, nil
+		}
+	}
+
+	// requesting replica is leader, allow
+	if err == nil && replicaName == lockState.leader && !minT.Before(lockState.leaseStart) {
+		go h.state.update(lockState, replicaName, maxT)
```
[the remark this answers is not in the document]
VineethReddy02: As we are not dependent on state updation in the next steps this is async. In this case yes we are just updating the `maxT` as the lockState & samples received prom instance are the same and fall in the expected lease interval.

pkg/ha/ha_service.go
VineethReddy02: Yeah, This avoids a redundant DB call of `readLockState`

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		err := h.validateClusterState(minT, maxT, clusterName, replicaName)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	currentState := h.state[clusterName]
+	allow := false
+	if replicaName == currentState.leader && maxT.Before(currentState.leaseUntil) && minT.After(currentState.leaseStart) {
```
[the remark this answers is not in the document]
cevian: Finally if minT is older than leaseStart but the leader is correct we need to throw away everything before leaseStart but we have to keep everything after. So there needs to be a way for this function to communicate that to the caller.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+		// read latest lock state
+		lockState, err = h.lockClient.readLockState(context.Background(), clusterName)
+		// couldn't get latest lock state
+		if err != nil {
+			log.Error("could not check ha lock state: %#v", err)
+		}
+
+		// update state asynchronously
+		h.state[clusterName].updateStateFromDB(lockState, maxT, replicaName)
+		go h.tryChangeLeader(clusterName)
```
cevian, line 105: do we need this here? why?

cevian: I think this should be called from `haStateSyncer` if `stateView.leaseUntil.After(stateView.maxTimeSeen)`

pkg/ha/state.go
```diff
@@ -0,0 +1,58 @@
...
+	leaseUntil      time.Time
+	maxTimeSeen     time.Time // max data time seen by any instance
+	maxTimeInstance string    // the instance name that’s seen the maxtime
+	_mu             sync.RWMutex
+}
+
+type StateView struct {
+	leader          string
+	leaseStart      time.Time
+	leaseUntil      time.Time
```
cevian, line 21: I think we need to add a maxTimeSeenLeader here. This will help us advance leaseUntil in the heartbeater.

cevian (changes requested):
High-level I think we need to figure out what's responsible for what. This is my proposal:
- The heartbeater is responsible for 1) updating the go's state from the db & telling the db to advance the leaseUntil based on the max time seen from the leader. 2) kicking off tryChangeLeader if the maxSeen (from anyone) > leaseUntil.
- The checkInsert (fast path) should just verify that everything is ok and update all the maxSeen stuff. If the maxTime it sees is after leaseUntil but the leader and everything else is right it needs to advance leaseUntil synchronously (with a call to checkInsert). 


There are also a bunch of problems with concurrency in go that need to be resolved.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+			if err != nil {
+				log.Error("failed to validate cluster %s state %v", clusterName, err)
+			}
+		}
+	}
+}
+
+// checkInsert verifies the samples are from prom leader & in expected time range.
+// 	returns true if the insert is allowed
+func (h *Service) checkInsert(minT, maxT time.Time, clusterName, replicaName string) (bool, error) {
```
Harkishen-Singh, line 64: I would suggest calling this as `areSamplesFromLeader` which would make the func self-explanatory and returning true would mean **yes, it is from leader**

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+		h.state[clusterName] = &State{
+			_mu: sync.RWMutex{},
+		}
+		err := h.validateClusterState(minT, maxT, clusterName, replicaName)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	currentState := h.state[clusterName]
```
Harkishen-Singh, line 76:
```suggestion
	currentLease := h.state[clusterName]
```
Is this appropriate?

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (h *Service) validateClusterState(minT, maxT time.Time, clusterName, replicaName string) error {
+	fmt.Println("minT:", minT, "maxT", maxT)
+	lockState, err := h.lockClient.checkInsert(context.Background(), clusterName, replicaName, minT, maxT)
+	if err != nil && err.Error() != leaderHasChanged.Error() {
+		return fmt.Errorf("could not check ha lock state: %#v", err)
+	}
+
+	// leader changed
+	if err != nil && err.Error() == leaderHasChanged.Error() {
```
Harkishen-Singh, line 95: We can merge this two blocks to one.

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,100 @@
+package ha
```
Harkishen-Singh, line 1: The file name is misleading. We do not implement any language parser. We just check validity of labels and parsing data. `parse_ts.go` seems easier.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
+CREATE TABLE SCHEMA_CATALOG.ha_locks
```
Harkishen-Singh, line 1: I know this is from the design doc. But, I think changing this name to `ha_lease_state` is more descriptive and better understanding. Frankly, `ha_locks` does not describe anything.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
...
+(
+    cluster_name TEXT PRIMARY KEY,
+    leader       TEXT,
+    lease_start  TIMESTAMPTZ,
+    lease_until  TIMESTAMPTZ
+);
+
+CREATE TABLE SCHEMA_CATALOG.ha_locks_log
```
Harkishen-Singh, line 9: `ha_lease_state_logs` here. or, should this be `_logs` to `_history`?

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,144 @@
...
+-- function that trigger to automatically keep the log calls
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.ha_locks_audit_fn()
+    RETURNS TRIGGER
+AS
+$func$
+BEGIN
+    INSERT INTO SCHEMA_CATALOG.ha_locks_log (cluster_name, leader, lease_start, lease_until)
+    VALUES (NEW.cluster_name, NEW.leader, NEW.lease_start, NEW.lease_until)
+           -- if lease is extended, but leader didn't change
+    ON CONFLICT (cluster_name, leader, lease_start) DO UPDATE SET lease_until=EXCLUDED.lease_until;
```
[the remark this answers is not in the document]
Harkishen-Singh: I do feel the same.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
...
+CREATE TRIGGER ha_locks_audit
+    AFTER INSERT OR UPDATE
+    ON SCHEMA_CATALOG.ha_locks
+    FOR EACH ROW
+EXECUTE PROCEDURE SCHEMA_CATALOG.ha_locks_audit_fn();
+
+-- default values for lease
+INSERT INTO SCHEMA_CATALOG.default(key, value)
+VALUES ('ha_lease_timeout', '1m'),
+       ('ha_lease_refresh', '10s')
```
Harkishen-Singh, line 46: Shouldn't these be values from CLI flags? This should really be configurable as timeouts are user dependent.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
+CREATE TABLE SCHEMA_CATALOG.ha_locks
+(
+    cluster_name TEXT PRIMARY KEY,
+    leader       TEXT,
```
Harkishen-Singh, line 4: For better understanding, we should rename `leader` => `leader_label` or `leader_label_name`, so it becomes cluster_name, leader_name.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,142 @@
...
+		allow = true
+	}
+
+	currentState.updateMaxSeenTime(replicaName, maxT)
+	return allow, nil
+}
+
+func (h *Service) validateClusterState(minT, maxT time.Time, clusterName, replicaName string) error {
+	fmt.Println("minT:", minT, "maxT", maxT)
+	lockState, err := h.lockClient.checkInsert(context.Background(), clusterName, replicaName, minT, maxT)
```
Harkishen-Singh, line 89: as per name, checkInsert should report true or false. So, seeing the use case, we should rename that.

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
...
+EXECUTE PROCEDURE SCHEMA_CATALOG.ha_locks_audit_fn();
+
+-- default values for lease
+INSERT INTO SCHEMA_CATALOG.default(key, value)
+VALUES ('ha_lease_timeout', '1m'),
+       ('ha_lease_refresh', '10s')
+ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value;
+
+-- ha api functions
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.check_insert(cluster TEXT, writer TEXT, min_time TIMESTAMPTZ,
```
Harkishen-Singh, line 50: we should give a better name than check_insert. I am not sure to what as I have a significant amount of code to read.

pkg/ha/ha_service.go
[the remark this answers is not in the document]
Harkishen-Singh: also, i have some concerns. What if the remote write that is sending is the leader. But, it has not sent frequently, so it sends a big batch of samples which exceeds our permittable lease duration and hence we throw it completely. This should not happen. I think, we should ingest for all cases after `minT.After(currentState.leaseStart)` if its from leader.

Harkishen-Singh (changes requested):
First pass. I have some doubts but in general, its looks good.
Some points to note:

1. The comments would be better if proper sentences are used. (Start with capitals, end with full-stops. We shouldn't miss such simple things).
2. We should not re-use package names in func names. I see this often in `ha` package. since the `ha` is already the package, there's no need for `haLockState`, `haLockClient`, then `NewHAService` (which can be `CreateService()` and would be called as `ha.CreateService()`).
3. If we want to return a pointer, then lets initialize with a pointer than first initializing with non-pointer and then sending the address of it. (this leads to first creating the variable in stack memory region and then again in the heap).

The code is a bit huge, so would need few more passes to get confident. But, awesome work :100: .

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
...
+        VALUES (cluster, writer, min_time, max_time + lease_timeout)
+        ON CONFLICT DO NOTHING;
+        -- needed due to on-conflict clause;
+        SELECT h.leader, h.lease_start, h.lease_until
+        INTO leader, lease_start, lease_until
+        FROM SCHEMA_CATALOG.ha_locks as h
+        WHERE cluster_name = cluster;
+    END IF;
+
+    IF leader <> writer OR lease_start > min_time THEN
```
cevian, line 87:
```suggestion
    IF leader <> writer  THEN
```

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,146 @@
...
+        SET lease_until = new_lease_timeout
+        WHERE h.cluster_name = cluster
+          AND h.leader = writer
+          AND h.lease_until + lease_refresh < new_lease_timeout;
+        IF NOT FOUND THEN -- concurrent update
+            SELECT h.leader, h.lease_start, h.lease_until
+            INTO leader, lease_start, lease_until
+            FROM SCHEMA_CATALOG.ha_locks as h
+            WHERE cluster_name = cluster;
+            IF leader <> writer OR lease_start > min_time OR lease_until < max_time
```
cevian, line 108:
```suggestion
            IF leader <> writer OR lease_until < max_time
```

pkg/migrations/sql/preinstall/006-set_up_ha.sql
cevian: Still needs to be fixed

pkg/ha/ha_lock_client.go
```diff
@@ -0,0 +1,127 @@
...
+	dbLock := haLockState{cluster: cluster}
+	row := h.dbConn.QueryRow(ctx, latestLockStateSql, cluster)
+	if err := row.Scan(&dbLock.leader, &dbLock.leaseStart, &dbLock.leaseUntil); err != nil {
+		return nil, err
+	}
+	return &dbLock, nil
+}
+
+func (h *haLockClientDB) readLeaseSettings(ctx context.Context) (timeout, refresh time.Duration, err error) {
+	rows, err := h.dbConn.Query(ctx, readLeaseSettingsSql)
```
cevian, line 101: I would just simplify the code here by running two queries, one for each setting. This is not performance critical code

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+
+	// requesting replica is leader, allow
+	return true, acceptedMinT, nil
+}
+
+func (h *Service) syncLockStateFromDB(minT, maxT time.Time, clusterName, replicaName string) error {
+	// TODO use proper context
+	lockState, err := h.lockClient.checkInsert(context.Background(), clusterName, replicaName, minT, maxT)
```
cevian, line 108: Yes I think checkInsert->updateLease would make it more clear what's going on. Better name

cevian (changes requested):
Overall code looks better. Still have a bunch of nits

pkg/ha/state.go
```diff
@@ -0,0 +1,64 @@
+package ha
```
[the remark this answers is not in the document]
VineethReddy02: Agree, I will split the current files into `state`, `client` & `service` packages as there are too many files under `ha` package.

pkg/ha/ha_lock_client.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := row.Scan(&value); err != nil {
+		return -1, -1, err
+	}
+	if timeout, err = parseTimestampToDuration(value); err != nil {
+		return -1, -1, err
+	}
+
+	// get leaseRefresh
+	row = h.dbConn.QueryRow(ctx,
+		"SELECT value FROM "+schema.Catalog+".default where key IN('"+leaseRefreshKey+"')")
```
cevian, line 107:
```suggestion
		"SELECT value FROM "+schema.Catalog+".default where key IN($1)", leaseRefreshKey)
```

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert")
+		return nil, 0, nil
+	}
+
+	// insert allowed -> parse samples
+	acceptedMinTUnix := acceptedMinT.UnixNano() / 1000000
+	for i := range tts {
+		t := &tts[i]
+
+		if t.Samples[0].Timestamp < acceptedMinTUnix {
```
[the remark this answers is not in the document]
cevian: This still needs fixing

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,121 @@
...
+			continue
+		}
+
+		// Drop __replica__ labelSet from samples
+		for ind, value := range t.Labels {
+			if value.Name == model.ReplicaNameLabel {
```
cevian, line 91: let's add a sanity check that value.Value == replicaName

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		if e, ok := err.(*pgconn.PgError); ok && e.Code == "PS010" {
+			leaderHasChanged = true
+		} else {
+			return fmt.Errorf("could not check ha lock state: %w", err)
+		}
+	}
+
+	// leader changed
+	if err != nil && leaderHasChanged {
```
cevian, line 122:
```suggestion
	if leaderHasChanged {
```

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,191 @@
...
+		return false, time.Time{}, nil
+	}
+
+	acceptedMinT := minT
+	if minT.Before(stateView.LeaseStart) {
+		acceptedMinT = stateView.LeaseStart
+	}
+
+	if maxT.After(stateView.LeaseUntil) {
+		err := h.syncLockStateFromDB(minT, maxT, clusterName, replicaName)
```
cevian, line 99: hmm what happens if there was a leader change instead of a successful update here? In that case there won't be an error but the current return would be wrong. I think we have to re-check who is the leader after the call to  h.syncLockStateFromDB.

pkg/ha/state/state.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (h *State) UpdateStateFromDB(latestState *HALockState, maxT time.Time, replicaName string) {
+	fmt.Println("latest state: ", latestState)
+	h._mu.Lock()
+	defer h._mu.Unlock()
+	h.leader = latestState.Leader
+	h.leaseStart = latestState.LeaseStart
+	h.leaseUntil = latestState.LeaseUntil
+	h.updateMaxTimeOnZero(maxT, replicaName)
```
cevian, line 44: instead of calling this here why don't we set the initial maxTimeSeen&maxTimeInstance in a  NewState function that sets this upon creation of the state. And call this from ha_service.go:78

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,143 @@
...
+    IF OLD IS NOT NULL AND OLD.leader_name = NEW.leader_name THEN
+        RETURN NEW;
+    END IF;
+
+    -- leader changed, set lease until to existing log line
+    IF OLD IS NOT NULL AND OLD.leader_name <> NEW.leader_name THEN
+        UPDATE ha_leases_logs
+        SET lease_until = OLD.lease_until
+        WHERE cluster_name = OLD.cluster_name
+          AND leader_name = OLD.leader_name
```
cevian, line 35:
```suggestion
          AND leader_name = OLD.leader_name
          AND lease_start = OLD.lease_start
```

pkg/migrations/sql/preinstall/006-set_up_ha.sql
cevian: This should not be a cli flag because if we have multiple connectors running, they cannot have different values. But, we should have DB functions to change these values

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,143 @@
...
+    UPDATE SCHEMA_CATALOG.ha_leases
+    SET leader_name = new_leader,
+        lease_start = lease_until,
+        lease_until = max_time + lease_timeout
+    WHERE cluster_name = cluster
+      AND lease_until < max_time;
+
+
+    SELECT *
+    INTO lease_state
```
cevian, line 137:
```suggestion
    INTO STRICT lease_state
```

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,143 @@
...
+            INTO leader, lease_start, lease_until
+            FROM SCHEMA_CATALOG.ha_leases as h
+            WHERE cluster_name = cluster;
+            IF leader <> writer OR lease_until < max_time
+            THEN
+                RAISE EXCEPTION 'LEADER_HAS_CHANGED' USING ERRCODE = 'PS010';
+            END IF;
+        END IF;
+    END IF;
+    SELECT * INTO lease_state FROM ha_leases WHERE cluster_name = cluster;
```
cevian, line 115:
```suggestion
    SELECT * INTO STRICT lease_state FROM ha_leases WHERE cluster_name = cluster;
```

pkg/migrations/sql/preinstall/006-set_up_ha.sql
```diff
@@ -0,0 +1,143 @@
...
+    RETURN lease_state;
+END;
+$func$ LANGUAGE plpgsql VOLATILE;
+
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.try_change_leader(cluster TEXT, new_leader TEXT,
+                                                            max_time TIMESTAMPTZ,
+                                                            lease_timeout INTERVAL DEFAULT '1m') RETURNS ha_leases
```
cevian, line 122: shouldn't lease_timeout be set from the default setting in the db as opposed to a parameter?

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,191 @@
...
+	for range h._syncTicker.C {
+		h.state.Range(func(c, s interface{}) bool {
+			cluster := fmt.Sprint(c)
+			state := castToState(s)
+			stateView := state.Clone()
+			err := h.syncLockStateFromDB(stateView.LeaseStart, stateView.MaxTimeSeenLeader, cluster, stateView.Leader)
+			if err != nil {
+				log.Error("failed to validate cluster %s state %v", cluster, err)
+			}
+			if stateView.MaxTimeSeen.After(stateView.LeaseUntil) {
```
cevian, line 61: can we factor this boolean expression into a function since it used in 2 places. here and on line 156. Plus we may want to adjust these.

cevian (changes requested):
Partial review

pkg/ha/ha_service.go
VineethReddy02: If I am getting it right the check here is

we do this check before calling tryChangeLeader `if stateView.MaxTimeSeen.After(stateView.LeaseUntil)` and check 156 is 
`stateView.LeaseUntil.After(stateView.MaxTimeSeen)` where this check is to re-verify before changing the leader. 

DO you mean we should create another generic func for this comparison? Also, can you elaborate on adjusting this?

pkg/migrations/sql/preinstall/006-set_up_ha.sql
Harkishen-Singh: But if you edit via DB functions, the moment the next fresh promscale instance is connected, it will update again?

Harkishen-Singh:
Before we merge this PR, I hope we can write some benchmark that shows the allocations of the existing non-ha code and then compare with this. Just want to make sure that we are not slowing down the expensive (high cpu) write path.

pkg/tests/end_to_end_tests/ha_check_insert_sql_test.go
```diff
@@ -0,0 +1,204 @@
...
+		wg := sync.WaitGroup{}
+		wg.Add(len(clusters) * len(writers))
+		for _, c := range clusters {
+			for _, w := range writers {
+				writer := w
+				cluster := c
+				go func() {
+					minT := time.Unix(0, 0)
+					maxT := time.Unix(2, 0)
+					firstLockRes, firstErr := callCheckInsert(db, cluster, writer, minT, maxT)
```
cevian, line 180: can we make sure only three items got a not-nil firstLockRes

cevian (changes requested):
Really getting into the nits now. The only thing to really look at is making sure leaseUntil is treated exclusive and leaseStart as inclusive. Did only a partial review of tests. Will do a full test review tomorrow.

Harkishen-Singh (changes requested):
Great work :+1: 

Few points, doubts:

1. Do we have logic for handling either leader-election or the new-ha mode, and not both (detect that and report the scenario)?
2. [Discussion] Sometimes I wonder if we should skip calculation of metrics that involves loops, since we already have our write path heavy. This should be controllable from CLI, so that those users with high ingestion can turn off this.
3. In CheckInsert(), what happens if a prom instance goes forward & start sending samples from future (with respect to Promscale and current lease)?
4. What happens if you do get samples form non-HA Prom instances? (We should be accepting them, since HA decision should just be on __replica__, and if sample does not have that, then its a normal insertion and just use the normal parser than using the ha_parser (my thoughts)).

TODO (Harkishen): Review tests.

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,249 @@
...
+		stateRef  *state.State
+		stateView *state.StateView
+		err       error
+	)
+
+	s, ok := h.state.Load(clusterName)
+	if !ok {
+		stateRef = &state.State{}
+		stateRef.UpdateMaxTimeOnZero(maxT, replicaName)
+		h.state.LoadOrStore(clusterName, stateRef)
```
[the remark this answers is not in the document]
blagojts: Double lock mechanism, in case some other routine also got a !ok on the Load at line 98

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,249 @@
...
+		stateView, err = h.syncLockStateFromDB(minT, maxT, clusterName, replicaName)
+		if err != nil {
+			return false, time.Time{}, err
+		}
+		go exposeHAStateToMetrics(clusterName, "", stateView.Leader)
```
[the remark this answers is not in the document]
blagojts: This just updates the prom metrics for the HA state of this instance

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,249 @@
...
+		return false, time.Time{}, nil
+	}
+
+	acceptedMinT := minT
+	if minT.Before(stateView.LeaseStart) {
+		acceptedMinT = stateView.LeaseStart
+	}
+
+	if maxT.After(stateView.LeaseUntil) {
+		stateView, err = h.syncLockStateFromDB(minT, maxT, clusterName, replicaName)
```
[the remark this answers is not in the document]
blagojts: when checking a lease minT is not updated, just maxT

pkg/ha/ha_service.go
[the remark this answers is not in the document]
Harkishen-Singh: Yes, but since its in HA, there are other Promscale attached. Do the other Promscale update their metrics (irrespective of this line)?

pkg/api/write.go
```diff
@@ -60,16 +60,21 @@ func Write(writer ingestor.DBInserter, elector *util.Elector, metrics *Metrics)
...
 		metrics.ReceivedSamples.Add(float64(receivedBatchCount))
 		begin := time.Now()
 
+		// If samples in write request are empty, then we do not need to
+		// proceed further.
+		if len(req.GetTimeseries()) == 0 {
+			return
+		}
+
 		numSamples, err := writer.Ingest(req.GetTimeseries(), req)
```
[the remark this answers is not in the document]
blagojts: I think with async acks we may end up null the pointers too soon

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,160 @@
...
+
+	dataSamples := make(map[string][]model.SamplesInfo)
+	rows := 0
+
+	s, _, err := model.LabelProtosToLabels(tts[0].Labels)
+	if err != nil {
+		return nil, rows, err
+	}
+	replicaName := s.GetReplicaName()
+	clusterName := s.GetClusterName()
```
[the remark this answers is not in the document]
blagojts: LabelProtosToLabels is also used in non-ha code, I'd rather keep it clean

plus, cyclomatic complexity of 
for{
  cmd1
  cmd2
}

and for{cmd1} for{cmd2} is the same

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, rows, err
+	}
+
+	// find samples time range
+	var (
+		minTUnix, maxTUnix int64
+		minTWasSet         = false
+	)
+	for i := range tts {
+		t := &tts[i]
```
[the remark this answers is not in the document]
blagojts: Local vars should get erased on compilation, and everything ends up being a register address. 
On the other hand it may help with not having to de-reference the pointer on each access if the compiler doesn't optimize that also

pkg/ha/ha_parser.go
```diff
@@ -0,0 +1,160 @@
...
+		log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert")
+		return nil, 0, nil
+	}
+
+	// insert allowed -> parse samples
+	acceptedMinTUnix := int64(promModel.TimeFromUnixNano(acceptedMinT.UnixNano()))
+	for i := range tts {
+		t := &tts[i]
+
+		t.Samples = filterSamples(t.Samples, acceptedMinTUnix)
```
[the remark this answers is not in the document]
blagojts: The db won't handle it better, on conflict is not a cheap operation to have for a significant amount of rows. 

Simple example - insert 1mil points in table with 2 columns, took my local db 2.3 seconds (avg), inserting 1 million points with ~30% conflicts -> 3 seconds

pkg/ha/ha_service.go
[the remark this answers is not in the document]
blagojts: If properly implemented, yes :)

pkg/ha/state/state.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type State struct {
+	leader                string
+	leaseStart            time.Time
+	leaseUntil            time.Time
+	maxTimeSeen           time.Time // max data time seen by any instance
+	maxTimeInstance       string    // the instance name that’s seen the maxtime
+	maxTimeSeenLeader     time.Time
+	_mu                   sync.RWMutex
```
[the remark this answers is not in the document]
blagojts: I use them to signify a field that shouldn't be assigned anywhere else other than init functions

pkg/ha/state/state.go
```diff
@@ -0,0 +1,90 @@
...
+	recentLeaderWriteTime time.Time
+}
+
+type StateView struct {
```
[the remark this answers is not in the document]
blagojts: State (now renamed to Lease) is modified concurrently, and any access needs to be through the RWMutex. We use the view to get a representation of the data by only acquiring the mutex with a read lock once. The obtained clone is to be used in a single goroutine that needs to access multiple fields, or access the same fields multiple times, but it is acceptable that the underlying state may change in the meantime.

pkg/pgmodel/ingestor/parser.go
```diff
@@ -0,0 +1,58 @@
...
+type DataParser struct{}
+
+// Parse data into a set of samplesInfo infos per-metric.
+// returns: map[metric name][]SamplesInfo, total rows to insert
+func (i *DataParser) ParseData(tts []prompb.TimeSeries) (map[string][]model.SamplesInfo, int, error) {
+	if len(tts) == 0 {
+		return nil, 0, nil
+	}
+
+	dataSamples := make(map[string][]model.SamplesInfo)
```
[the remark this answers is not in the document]
blagojts: map is indexed by metric name, not len(tts), and we don't know the metric names apriori. Regarding the wg.Done() I don't know which one you're referring to.

pkg/pgmodel/model/labels.go
```diff
@@ -16,12 +16,17 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+const (
+	ReplicaNameLabel = "__replica__"
+	ClusterNameLabel = "__cluster__"
```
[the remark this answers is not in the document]
blagojts: Ultimately it would be configurable, if users are actually requesting it.

pkg/pgmodel/model/labels.go
```diff
@@ -16,12 +16,17 @@ import (
...
+)
+
 // Labels stores a labels.Labels in its canonical string representation
 type Labels struct {
 	Names      []string
 	Values     []string
 	MetricName string
-	str        string
+	Str        string
```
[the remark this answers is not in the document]
blagojts: Will revert. Needed this in the tests only

pkg/pgmodel/model/labels.go
```diff
@@ -179,3 +184,21 @@ func (l *Labels) Swap(i, j int) {
...
+}
+
+func (l *Labels) GetClusterName() string {
+	for index, key := range l.Names {
+		if key == ClusterNameLabel {
+			return l.Values[index]
+		}
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
blagojts: It's called only once per batch. And depending on the values the insert might not be allowed, which results in LabelProtosToLabels not being called at all.

cevian (commented):
Partial review

pkg/tests/end_to_end_tests/ha_single_promscale_test.go
```diff
@@ -0,0 +1,579 @@
...
+					expectedLeaseStateInDb: leaseState{
+						cluster:    "cluster",
+						leader:     "1",
+						leaseStart: time.Unix(0, 0),
+						leaseUntil: time.Unix(60, 0),
+					},
+				},
+				tickSyncRoutine: true,
+			}, {
+				desc: "2 sends data after lease, no insert -> triggers leader change",
```
cevian, line 154: This isn't great. I shouldTryToChangeLeader and the new lease will include this point. But, it's not written because it needs to trigger a tick first. Does this mean we will loose some points for any leader change? Should we introduce a non-async code path for try leader change here to avoid this?

pkg/tests/end_to_end_tests/ha_multiple_promscales_test.go
```diff
@@ -0,0 +1,238 @@
...
+				},
+				tickSyncRoutine: false,
+				explicitLease: &leaseState{
+					cluster:    "cluster",
+					leader:     "2",
+					leaseStart: time.Unix(60, 0),
+					leaseUntil: time.Unix(120, 0),
+				},
+			}, {
+				desc: "2 tries to send data in lease, lease is cached from before-> rejected; sync",
```
cevian, line 182: this is the same less than Ideal case here. I guess we should try to do something clever if the input is for a point past the cached leaseUntil. Let's think about this case a bit more

cevian (changes requested):
Ok reviewed the rest. The tests look wonderful. Have some comments about the test but mostly minor.

The only real question I have is whether we should introduce a synchronous version of change leader if a follow is sending data after leaseUntil and shouldChangeLeader is true. Not sure, but may be worth it. 

I think the next steps are to resolve my last few comments and rebase. I will then take a final look.

antekresic (changes requested):
Need to fix the failing tests (update a function name refactor) and some minor changes.

Should be in a good place after that.

pkg/ha/state/lease.go
```diff
@@ -0,0 +1,187 @@
...
+// 	Access to any field should be controlled through the _mu RW Mutex.
+type Lease struct {
+	cluster               string
+	leader                string
+	leaseStart            time.Time
+	leaseUntil            time.Time
+	maxTimeSeen           time.Time // max data time seen from any prometheus replica
+	maxTimeInstance       string    // the replica name that’s seen the maxtime
+	maxTimeSeenLeader     time.Time // max data time seen by current leader
+	recentLeaderWriteTime time.Time // real time when leader last wrote data
```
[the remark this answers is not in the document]
atanasovskib: will reset maxTimeSeenLeader and recentLeaderWriteTime, I'm not sure we should reset the maxTimeSeen (total) especially do math like leaseUntil - refreshTime

pkg/tests/end_to_end_tests/ha_multiple_promscales_test.go
atanasovskib: You mean to try to sync with the db if we have data after leaseUntil from a non-leader instance?

pkg/tests/end_to_end_tests/ha_single_promscale_test.go
atanasovskib: Yes, points will be lost until the sync routine ticks and updates the lease. Latest commit will sync with the db to see if leader changed from outside, but will not try to change leader

cevian (approved):

antekresic (approved):

PR #408, opened 2020-12-30
Create HA Client connected to DB

blagojts:
* Create migration with HA tables and api
* Create an HA client that reads ha locks in db
  * The client only issues queries to the database.
  * HaState has been split into a Service and State.
    * ha.Service checks whether a write should be
       permitted
    * ha.State only keeps the state and makes sure
       it is modified by one routine at a time
   * tryChangeLeader is moved to an async routine
      semaphore is added to check that only one of
      these routines runs at a time
   * No tests have been written and so far hasn't been
      tested in any way whatsoever


pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+	// leader changed
+	if err == leaderHasChanged {
+		// read latest lock state
+		lockState, err = h.lockClient.readLockState(context.Background(), clusterName)
+		// couldn't get latest lock state
+		if err != nil {
+			return false, fmt.Errorf("could not check ha lock state: %#v", err)
+		}
+		// update state asynchronously
+		go h.state.update(lockState, replicaName, maxT)
```
blagojts, line 60: we have lockState as a local copy of what the db returned, no need to update state synchronously

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+		go h.tryChangeLeader()
```
blagojts, line 61: also changing the leader can take some time, no need to wait for it

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+	if err == nil && replicaName == lockState.leader && !minT.Before(lockState.leaseStart) {
+		go h.state.update(lockState, replicaName, maxT)
+		return true, nil
+	}
+
+	// replica is not leader or timestamps out of lease range, ignore them
+	return false, nil
+}
+
+func (h *Service) tryChangeLeader() {
```
blagojts, line 81: should rename to checkForLeaderChange or something since the check whether the leader needs changing is here

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+	ok := h._leaderChangeLock.TryAcquire(1)
+	if !ok {
+		// change leader already in progress
+		return
+	}
+	defer h._leaderChangeLock.Release(1)
+
+	for {
+		stateView := h.state.clone()
```
blagojts, line 90: clone state so we don't hold a lock while reading the various fields

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+		lockState, err := h.lockClient.readLockState(context.Background(), stateView.cluster)
+		if err != nil {
+			log.Error("msg", "Couldn't check lock state", "err", err)
+			return
+		}
+		if stateView.leader == lockState.leader || !lockState.leaseUntil.After(stateView.maxTimeSeen) {
```
blagojts, line 96:
```suggestion
		if stateView.leader == lockState.leader || stateView.maxTimeSeen.Before(lockState.leaseUntil) {
```

pkg/ha/ha_service.go
```diff
@@ -0,0 +1,115 @@
...
+		if lockState.leader == replicaName && !minT.Before(lockState.leaseStart) {
+			return true, nil
+		} else {
+
+			return false, nil
+		}
+	}
+
+	// requesting replica is leader, allow
+	if err == nil && replicaName == lockState.leader && !minT.Before(lockState.leaseStart) {
```
blagojts, line 72:
```suggestion
	if err == nil && replicaName == lockState.leader && lockState.leaseStart.After(minT) {
```

go.mod
```diff
@@ -1,9 +1,6 @@
 module github.com/timescale/promscale
 
-//TODO remove after repo rename
-replace github.com/timescale/promscale => ./
-
-go 1.14
+go 1.15
```
VineethReddy02, line 3: Any specific reason for upgrading the go version in this PR?

VineethReddy02 (approved):

PR #410, opened 2020-12-31
prom-migrator: Adds support for reading prom tsdb blocks directly by the reader

Harkishen-Singh:
This PR enables the reader to take the dir path of prometheus blocks and read them directly, than accessing the data through the `/api/v1/read` endpoint. This enables reading much faster than the endpoint way.

TODO

- [ ] Test in actual migration before merging


pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+			iteratorStore.Range(callBack)
+			timeseries, bytesUncompressed := tsMapToTs(ts)
+			ts = nil
+			blockRef.Fill(timeseries, bytesUncompressed)
+			//fmt.Println("Sending")
+			rr.sigBlockRead <- blockRef
+			blockRef = nil
+		}
+	}
+	switch rr.readType {
```
cevian, line 210: This seems really strange. I would have expected usage of different readers instead. e.g. RemoteReader->RemoteReaderURL, PathReader

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+				if itr.Seek(mint) {
+					appender()
+					for itr.Next() {
+						if !appender() {
+							break
+						}
+					}
+				}
+				if len(samples) > 0 {
+					ts[cus.labels.String()] = struct {
```
cevian, line 194: why are we not just creating a *prompb.Timeseries here?

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+						labels  []prompb.Label
+						samples []prompb.Sample
+					}{labels: labelsToPrombLabels(cus.labels), samples: samples}
```
cevian, line 197: this will overwrite whatever was in the map before, is that right?

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
 			rr.sigBlockRead <- blockRef
 			blockRef = nil
 		}
-	}()
+	}
+
+	readPromBlocks := func() {
+		// Comment.
```
cevian, line 123: what is this?

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+			err = tsdb_errors.NewMulti(err, db.Close()).Err()
+		}()
+		querier, err := db.Querier(rr.c, rr.plan.Mint(), rr.plan.Maxt())
+		if err != nil {
+			errChan <- fmt.Errorf("remote-read run: prom tsdb querier: %w", err)
+			return
+		}
+		defer querier.Close()
+		ss := querier.Select(false, nil, labels.MustNewMatcher(labels.MatchRegexp, "", ".*"))
+		var iteratorStore = new(sync.Map)
```
cevian, line 140: why is this a sync.Map and not a regular map?

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+							Timestamp: t,
+							Value:     v,
+						})
+						return true
+					}
+					return false
+				}
+				if itr.Seek(mint) {
+					appender()
+					for itr.Next() {
```
cevian, line 187: why aren't we stopping at maxt?

pkg/migration-tool/reader/reader.go
```diff
@@ -93,5 +117,133 @@ func (rr *RemoteRead) Run(errChan chan<- error) {
...
+				cus, ok := value.(cusMap)
+				if !ok {
+					errChan <- fmt.Errorf("remote-read run: value not found in iteratorStore")
+					return false
+				}
+				itr := cus.itr
+				var samples []prompb.Sample
+				appender := func() bool {
+					t, v := itr.At()
+					if t >= mint && t < maxt {
```
cevian, line 176: shouldn't this be guaranteed by the iterator?

cevian (changes requested):

pkg/migration-tool/reader/reader.go
Harkishen-Singh: We do that in 234 line.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Yep, but I don't think that's an issue as each series would be written to the `tsMap` only once, so I don't think we should worry about overwriting.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: Sorry, this is to add comments for what the respect funcs do. But now, since we want different readers, so this should no longer be a requirement.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: No specific reason. I should convert to a regular map. I think, while implementing i though of a concurrent safe map, but i dont think its a requirement now.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: We stop at `maxt`. Its in 174 line.

pkg/migration-tool/reader/reader.go
Harkishen-Singh: I don't think so. The iterator will be per series basis, but the samples iteration isn't within the time-range. It happens by `Seek()` which goes to the time required and iterates till the last `Next()` (stopped by break here) and stays there as cache in the iteratorStore (so that for the next time-range delta, it can resume from there)

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
...
+		errChan <- fmt.Errorf("path-read run: opening prom tsdb: %w", err)
+		return
+	}
+	querier, err := db.Querier(rr.c, rr.plan.Mint(), rr.plan.Maxt())
+	if err != nil {
+		errChan <- fmt.Errorf("path-read run: prom tsdb querier: %w", err)
+		return
+	}
+	ss := querier.Select(false, nil, labels.MustNewMatcher(labels.MatchRegexp, "", ".*"))
+	type itrMap struct {
```
cevian, line 56: This is not a map
```suggestion
	type labelIterator struct {
```

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
...
+			default:
+			}
+			blockRef, err := rr.plan.NextBlock()
+			if err != nil {
+				errChan <- fmt.Errorf("path-run run: %w", err)
+				return
+			}
+			ts := make(tsMap)
+			mint, maxt := blockRef.Mint(), blockRef.Maxt()
+			populateTimeseries := func(m itrMap) bool {
```
cevian, line 95: this is way too big to be a closure, it should be a top level function instead

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
...
+	size := 0
+	i := 0
+	for _, tsm := range m {
+		pts := &prompb.TimeSeries{
+			Labels:  tsm.labels,
+			Samples: tsm.samples,
+		}
+		b, _ := pts.Marshal()
+		size += len(b)
+		ts[i] = pts
```
cevian, line 156: we are marshalling just for a size? That can't be right. Its soo much wasted work.

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
...
+				if itr.Seek(mint) {
+					appender()
+					for itr.Next() {
+						if !appender() {
+							break
+						}
+					}
+				}
+				if len(samples) > 0 {
+					ts[m.labels.String()] = struct {
```
cevian, line 121: This can be a slice of []*prompb.Timeseries or [][]byte(see below). That way we avoid a tsMapToTs and and intermediate representation.

pkg/migration-tool/reader/path_reader.go
cevian: Can we save the byte slice instead of the []prompb.TimeSeries slice? then the work is not wasted.

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return reader, nil
+}
+
+func (rr *PathRead) SigStop() {
+	rr.sigForceStop <- struct{}{}
+}
+
+// Run runs the remote read and starts fetching the samples from the read storage.
+func (rr *PathRead) Run(errChan chan<- error) {
```
cevian, line 43: I think this function need a refactoring/cleanup. Its very hard to follow. I suggest moving some functions define in here out to the toplevel as a start.

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,169 @@
+package reader
+
```
cevian, line 2: factoring out the PathReader into s separate thing helps a lot. good job

pkg/migration-tool/reader/remote_reader.go
```diff
@@ -18,34 +16,38 @@ import (
...
-	read.fetchersRunning.Store(0)
-	return read, nil
+	if needsStopChan {
+		reader.sigForceStop = make(chan struct{})
+	}
+	return reader, nil
+}
+
+func (rr *RemoteRead) SigStop() {
+	rr.sigForceStop <- struct{}{}
```
cevian, line 50: I think you need a if not nil check here? Otherwise this will block? or do we want that?

cmd/prom-migrator/main.go
```diff
@@ -136,6 +142,8 @@ func parseFlags(conf *config, args []string) {
 		"More the read size, faster will be the migration but higher will be the memory usage. Example: 250MB.")
 	flag.StringVar(&conf.readURL, "read-url", "", "URL address for the storage where the data is to be read from.")
 	flag.StringVar(&conf.writeURL, "write-url", "", "URL address for the storage where the data migration is to be written.")
+	flag.StringVar(&conf.promTSDBPath, "prom-tsdb-path", "", "Address for prometheus tsdb dir. This enables direct fetching of blocks "+
```
cevian, line 145: should mention this replaces the read-url

cevian (changes requested):
I think there is some cleanup and effeciency fixes that need to happen. Left the review of the testing to a future review.

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: I don't think that is possible, as we read blocks which are in form of iterators than byte slices. Any suggestions?

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: Could you re-look this comment now as the recent commits have made some funcs top-level, so if this holds true even now.

pkg/migration-tool/reader/remote_reader.go
Harkishen-Singh: This is just used during testing. But, I agree as it will save time during debugging in-case something goes wrong near this place (as non-initialized channels never strike the mind during debugging).

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,161 @@
...
+			samples = append(samples, prompb.Sample{
+				Timestamp: t,
+				Value:     v,
+			})
+			return true
+		}
+		return false
+	}
+	if itr.Seek(li.mint) {
+		appender()
```
cevian, line 139: don't we need to check the return value here too and break if need be?

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,161 @@
...
+				break
+			}
+		}
+	}
+	if len(samples) > 0 {
+		containsTs = true
+		ts.Labels = labelsToPrombLabels(li.labels)
+		ts.Samples = samples
+	}
+	return ts, containsTs
```
cevian, line 151: isnt containTs == (len(ts.Samples) > 0), in which case we don't need a separate return.

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			mint, maxt := blockRef.Mint(), blockRef.Maxt()
+			var timeseries []*prompb.TimeSeries
+			for _, iterator := range iteratorStore {
+				iterator.setTimeRange(mint, maxt)
+				if ts, containsTs := iterator.populateTimeseries(); containsTs {
+					timeseries = append(timeseries, ts)
+				}
+			}
+			blockRef.Fill(timeseries, 0)
```
cevian, line 100: so we're not doing sizes here? How do you plan to address this?

pkg/migration-tool/reader/path_reader.go
cevian: Looks much better now

cevian (changes requested):
Everything looks good except for the size/adaptation issue. Let'a think about that.

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: Yes we can, but it will anyways be rejected by the next appender(). Still, I will break here too.

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: We anyways need to return `ts` so that it can be appended at line 96. A single bool will not do this.

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: I was planning for `timeseries.Marshal()` and consider its length, but I think we should have a discussion on this as you said.

pkg/pgmodel/ingestor/inserter.go
```diff
@@ -135,7 +135,6 @@ func (p *pgxInserter) InsertData(rows map[string][]model.SamplesInfo) (uint64, e
 		workFinished.Wait()
 		select {
 		case err = <-errChan:
-			fmt.Println("here", err)
```
blagojts, line 138: I know "here" is not proper, but is the error logged somewhere else?

blagojts (commented):
Partial review. Left some nits and questions. But might have to checkout the code for a more complete review

cmd/prom-migrator/main.go
```diff
@@ -185,16 +193,18 @@ func validateConf(conf *config) error {
...
 	case conf.progressMetricName != progressMetricName:
 		if !regexp.MustCompile(validMetricNameRegex).MatchString(conf.progressMetricName) {
 			return fmt.Errorf("invalid metric-name regex match: prom metric must match %s: recieved: %s", validMetricNameRegex, conf.progressMetricName)
 		}
 	case strings.TrimSpace(conf.readURL) == "" && strings.TrimSpace(conf.writeURL) == "":
 		return fmt.Errorf("remote read storage url and remote write storage url must be specified. Without these, data migration cannot begin")
-	case strings.TrimSpace(conf.readURL) == "":
-		return fmt.Errorf("remote read storage url needs to be specified. Without read storage url, data migration cannot begin")
+	case strings.TrimSpace(conf.readURL) == "" && strings.TrimSpace(conf.promTSDBPath) == "":
+		return fmt.Errorf("either remote-read storage url or prom-tsdb-path must be specified. Without these, data migration cannot begin")
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, I didn't understand how this is same as 196. The one at 196 says that if both are set, then error out.

pkg/migration-tool/reader/interface.go
```diff
@@ -0,0 +1,13 @@
...
+import "time"
+
+const defaultReadTimeout = time.Minute * 5
+
+type Reader interface {
+	// Run runs the reader and starts fetching the samples from the storage.
+	Run(chan<- error)
+	// SigStop signals a stop signal to the reader. This is mainly used to introduce halts
+	// while doing integration tests.
+	SigStop()
```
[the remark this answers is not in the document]
Harkishen-Singh: I am not sure if I should remove this. It may be useful when we plan to integrate it in Promscale.

pkg/migration-tool/reader/interface.go
```diff
@@ -0,0 +1,13 @@
+package reader
+
+import "time"
+
+const defaultReadTimeout = time.Minute * 5
+
+type Reader interface {
+	// Run runs the reader and starts fetching the samples from the storage.
+	Run(chan<- error)
```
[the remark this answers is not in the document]
Harkishen-Singh: data chan is passed at init, since it has to be common with the reader. But I agree,the error channel should be returned, than actually passing it.

pkg/migration-tool/reader/path_reader.go
```diff
@@ -0,0 +1,164 @@
...
+			log.Info("msg", "reader is down")
+			close(errChan)
+		}()
+		log.Info("msg", "reader is up")
+		// Start creating blocks and filling them with samples.
+		for rr.plan.ShouldProceed() {
+			select {
+			case <-rr.c.Done():
+				return
+			case <-rr.sigForceStop:
```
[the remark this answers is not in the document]
Harkishen-Singh: I don't think we can use ShouldProceed() that way. ShuoldProceed() is built for the smooth exit of migration after the migration range is met. sigForceStop is for a non-graceful exit so as to mimic an actual crash, just like what happens in real-world, to test the migration resumption. Hence, it is useful in tests.

pkg/pgmodel/ingestor/inserter.go
Harkishen-Singh: Yes, we are already logging it. https://github.com/timescale/promscale/blob/724c48f858c92e3d319f7e1ad1ff0870e9abf8ed/pkg/api/write.go#L74

pkg/migration-tool/reader/path_reader.go
Harkishen-Singh: Ah got it. Yes, `containTs` is not required. Thank you for pointing out.

PR #415, opened 2021-01-05
Log once per metric table in case of a repetitive duplicate sample.

Harkishen-Singh:

### DIFF
--- a/pkg/pgmodel/ingestor/duplicates.go
+++ b/pkg/pgmodel/ingestor/duplicates.go
@@ -1,0 +1,42 @@
+package ingestor
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+)
+
+const reportDuplicatesInterval = time.Minute
+
+var (
+	launchReporterOnce    sync.Once
+	duplicateMetricsTotal uint64
+)
+
+func init() {
+	atomic.StoreUint64(&duplicateMetricsTotal, 0)
+}
+
+func registerDuplicates(duplicateSamples int64) {
+	metrics.DuplicateSamples.Add(float64(duplicateSamples))
+	metrics.DuplicateWrites.Inc()
+}
+
+func reportDuplicates(duplicateMetrics uint64) {
+	atomic.AddUint64(&duplicateMetricsTotal, duplicateMetrics)
+	metrics.DuplicateMetrics.Add(float64(duplicateMetrics))
+	launchReporterOnce.Do(func() {
+		go func() {
+			report := time.NewTicker(reportDuplicatesInterval)
+			for range report.C {
+				if atomic.LoadUint64(&duplicateMetricsTotal) != 0 {
+					log.Warn("msg", "duplicate data in sample", "total-duplicate-metrics", atomic.LoadUint64(&duplicateMetricsTotal))
+					atomic.StoreUint64(&duplicateMetricsTotal, 0)
+				}
+			}
+		}()
+	})
+}
--- a/pkg/pgmodel/ingestor/insert.go
+++ b/pkg/pgmodel/ingestor/insert.go
@@ -381,18 +381,18 @@ func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
 		return err
 	}
 
-	for i, numRows := range numRowsPerInsert {
+	var affectedMetrics uint64
+	for _, numRows := range numRowsPerInsert {
 		ct, err := results.Exec()
 		if err != nil {
 			return err
 		}
 		if int64(numRows) != ct.RowsAffected() {
-			log.Warn("msg", "duplicate data in sample", "table", reqs[i].table, "duplicate_count", int64(numRows)-ct.RowsAffected(), "row_count", numRows)
-			metrics.DuplicateSamples.Add(float64(int64(numRows) - ct.RowsAffected()))
-			metrics.DuplicateWrites.Inc()
+			affectedMetrics++
+			registerDuplicates(int64(numRows) - ct.RowsAffected())
 		}
 	}
+	reportDuplicates(affectedMetrics)
 	metrics.DbBatchInsertDuration.Observe(time.Since(start).Seconds())
-
 	return nil
 }
--- a/pkg/pgmodel/metrics/metrics.go
+++ b/pkg/pgmodel/metrics/metrics.go
@@ -24,6 +24,13 @@ var (
 			Help:      "Total number of writes that contained duplicates",
 		},
 	)
+	DuplicateMetrics = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: util.PromNamespace,
+			Name:      "duplicate_metrics_total",
+			Help:      "Total number of affected metrics due to duplicate samples.",
+		},
+	)
 	DecompressCalls = prometheus.NewCounter(
 		prometheus.CounterOpts{
 			Namespace: util.PromNamespace,
@@ -69,6 +76,7 @@ func init() {
 		DuplicateWrites,
 		DecompressCalls,
 		DecompressEarliest,
+		DuplicateMetrics,
 		NumInsertsPerBatch,
 		NumRowsPerBatch,
 		DbBatchInsertDuration,
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	PromNamespace              = "ts_prom"
+	PromNamespace              = "promscale"
 	maskPasswordReplaceString1 = "password=$1'****'"
 	maskPasswordReplaceString2 = "password:$1****$3"
 	/* #nosec */


pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,31 @@
...
+		previousCount := previousCountInf.(int64)
+		netDuplicates += previousCount
+	} else {
+		// We log the duplicates once per table name. Detailed logging affects performance and hence must be avoided.
+		// As such, details related to counts, etc, can be visualised as prometheus metrics.
+		log.Warn("msg", "duplicate data in sample", "table", tableName, "duplicate_count", count)
+	}
+	metrics.DuplicateSamples.Add(float64(count))
+	metrics.DuplicateWrites.Inc()
+	duplicatesStore.Insert(tableName, netDuplicates)
```
cevian, line 30: we never actually use netDeplicates and Insert is a much more expensive operation than Get, so I'd prefer to store a constant but avoid Inserts in all cases other than cache miss.

cevian (changes requested):

pkg/pgmodel/ingestor/duplicates.go
Harkishen-Singh: > Insert is a much more expensive operation than Get

Insert internally first call the gets (via checking `self.elements[]`). So, isn't it the same?

Harkishen-Singh: The recent commit pushed is in line with what you expect.

Harkishen-Singh: Also, I feel a bug in `insert()` of `cache.go`. The insert expects key and value. That means, it should store the value. If someone inserts a pair with the key being already in the cache, but with an updated value, the updated value will never be accepted as, if the key is present, the `insert()` is just returning the previous value. But in actuality, it should return only if the key is present and the incoming value is the same as the present value in the cache.

Harkishen-Singh: @cevian

pkg/util/util.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	PromNamespace              = "ts_prom"
+	PromNamespace              = "promscale"
```
Harkishen-Singh, line 18: This is a breaking change. We should mention during the release. @cevian

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,49 @@
...
+
+func reportDuplicates(duplicates uint64) {
+	effectedMetrics.Store(duplicates)
+	metrics.EffectedMetrics.Set(float64(duplicates))
+	if duplicates > 0 {
+		pendingReportableDuplicates.Store(true)
+	}
+	if !isReportDuplicatesActive.Load() {
+		isReportDuplicatesActive.Store(true)
+		report := time.NewTicker(reportDuplicatesInterval)
```
cevian, line 38: this should go inside the goroutine

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+)
+
+const reportDuplicatesInterval = time.Second * 10
+
+var (
+	effectedMetrics             atomic.Uint64
+	isReportDuplicatesActive    atomic.Bool
```
cevian, line 16: note name change and should also be a sync.Once
```suggestion
	LaunchReporterOnce    sync.Once
```

pkg/pgmodel/ingestor/insert.go
```diff
@@ -381,18 +381,18 @@ func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
...
 		}
 		if int64(numRows) != ct.RowsAffected() {
-			log.Warn("msg", "duplicate data in sample", "table", reqs[i].table, "duplicate_count", int64(numRows)-ct.RowsAffected(), "row_count", numRows)
-			metrics.DuplicateSamples.Add(float64(int64(numRows) - ct.RowsAffected()))
-			metrics.DuplicateWrites.Inc()
+			affectedMetrics++
+			registerDuplicates(int64(numRows) - ct.RowsAffected())
 		}
 	}
+	reportDuplicates(affectedMetrics)
```
cevian, line 395: I am actually not sure how to interpret `affectedMetrics` This is the number of metrics that had duplicates in a batch, but I'm really not sure what this tells me especially since I dont get the sum, just one number.

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,49 @@
...
+	pendingReportableDuplicates.Store(false)
+}
+
+func registerDuplicates(count int64) {
+	metrics.DuplicateSamples.Add(float64(count))
+	metrics.DuplicateWrites.Inc()
+}
+
+func reportDuplicates(duplicates uint64) {
+	effectedMetrics.Store(duplicates)
```
cevian, line 31: again not sure what this number says or how to interpret it. since this is a store not a sum. I'd actually expect this to be a sum of the /samples/ seen since last reporting. Or a sum of the table inserts (metrics) that had this since last reporting.

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,49 @@
...
+	metrics.EffectedMetrics.Set(float64(duplicates))
+	if duplicates > 0 {
+		pendingReportableDuplicates.Store(true)
+	}
+	if !isReportDuplicatesActive.Load() {
+		isReportDuplicatesActive.Store(true)
+		report := time.NewTicker(reportDuplicatesInterval)
+		go func() {
+			for range report.C {
+				if pendingReportableDuplicates.Load() {
```
cevian, line 41: why aren't we simply checking that effectedMetrics != 0

pkg/util/util.go
cevian: Please add to notes here https://github.com/timescale/promscale/releases/tag/untagged-5b8262593f32177fe63d

cevian (changes requested):
Looks better but I don't know how to interpret the effectedMetrics metric

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,26 @@
...
+
+func init() {
+	duplicatesStore = clockcache.WithMax(maxDuplicateStoreSize)
+}
+
+func registerDuplicates(tableName string, count int64) {
+	if _, found := duplicatesStore.Get(tableName); !found {
+		// We log the duplicates once per table name. Detailed logging affects performance and hence must be avoided.
+		// As such, details related to counts, etc, can be visualised as prometheus metrics.
+		log.Warn("msg", "duplicate data in sample", "table", tableName, "duplicate_count", count)
```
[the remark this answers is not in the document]
Harkishen-Singh: This is now changed. We only log once in every 10 secs with the information of num of metrics affected.

pkg/pgmodel/ingestor/insert.go
Harkishen-Singh: It represents the number of metrics that have got duplicate samples. Do you think this needs to be changed to "sum of all duplicate rows count?"

pkg/pgmodel/ingestor/duplicates.go
Harkishen-Singh: Yeah, better to store this as a counter (i.e., `metrics.AffectedMetrics`. But, the `effectedMetrics` needs to be Store since it's used in the comparison.

pkg/pgmodel/metrics/metrics.go
```diff
@@ -37,6 +37,13 @@ var (
 			Name:      "decompress_min_unix_time",
 			Help:      "Earliest decdompression time",
 		}, []string{"table"})
+	AffectedMetrics = prometheus.NewCounter(
```
cevian, line 40: Using the same convention as above this should be called DuplicateMetrics

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+const reportDuplicatesInterval = time.Second * 10
+
+var (
+	affectedMetrics    atomic.Uint64
+	launchReporterOnce sync.Once
+)
+
+func registerDuplicates(count int64) {
```
cevian, line 20:
```suggestion
func registerDuplicates(duplicateSamples int64) {
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
...
+	metrics.DuplicateSamples.Add(float64(count))
+	metrics.DuplicateWrites.Inc()
+}
+
+func reportDuplicates(duplicates uint64) {
```
cevian, line 25:
```suggestion
func reportDuplicates(duplicateMetrics uint64) {
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
...
+	affectedMetrics.Store(duplicates)
+	metrics.AffectedMetrics.Add(float64(duplicates))
+	launchReporterOnce.Do(func() {
+		go func() {
+			report := time.NewTicker(reportDuplicatesInterval)
+			for range report.C {
+				if affectedMetrics.Load() != 0 {
+					log.Warn("msg", "duplicate data in sample", "num-metrics-affected", affectedMetrics.Load())
```
cevian, line 33:
```suggestion
					log.Warn("msg", "duplicate data in sample", "num-duplicate-metrics", affectedMetrics.Load())
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+)
+
+const reportDuplicatesInterval = time.Second * 10
+
+var (
+	affectedMetrics    atomic.Uint64
```
cevian, line 16:
```suggestion
              duplicateMetrics    atomic.Uint64
```

cevian (changes requested):
Looks good except for naming stuff

pkg/pgmodel/metrics/metrics.go
```diff
@@ -37,6 +37,13 @@ var (
 			Name:      "decompress_min_unix_time",
 			Help:      "Earliest decdompression time",
 		}, []string{"table"})
+	DuplicateMetrics = prometheus.NewCounter(
```
cevian, line 40: I would prefer this go under the DuplicateWrites declaration. Keep all the duplicate stuff together

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
+package ingestor
+
+import (
+	"sync"
+	"time"
+
+	"go.uber.org/atomic"
```
cevian, line 7: blocking: why are we using uber's atomic instead of the standard golang one? We shouldn't introduce a new dependency like this without discussion

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,39 @@
...
+	duplicateMetricsTotal atomic.Uint64
+)
+
+func registerDuplicates(duplicateSamples int64) {
+	metrics.DuplicateSamples.Add(float64(duplicateSamples))
+	metrics.DuplicateWrites.Inc()
+}
+
+func reportDuplicates(duplicateMetrics uint64) {
+	duplicateMetricsTotal.Store(duplicateMetrics)
```
cevian, line 26: shouldn't this be an Add?
```suggestion
	duplicateMetricsTotal.Add(duplicateMetrics)
```

cevian (changes requested):

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,42 @@
...
+	duplicateMetricsTotal.Store(uint64(0))
+}
+
+func registerDuplicates(duplicateSamples int64) {
+	metrics.DuplicateSamples.Add(float64(duplicateSamples))
+	metrics.DuplicateWrites.Inc()
+}
+
+func reportDuplicates(duplicateMetrics uint64) {
+	duplicateMetricsTotal.Store(duplicateMetricsTotal.Load().(uint64) + duplicateMetrics)
```
cevian, line 29: this is the wrong way to use atomics. This isn't concurrency safe (if two goroutines call this at the same time, the last one wins, ignoring the first addition),  Instead you should be using
```suggestion
	atomic.AddUint64(&duplicateMetricsTotal, duplicateMetrics)
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+)
+
+const reportDuplicatesInterval = time.Minute
+
+var (
+	launchReporterOnce    sync.Once
+	duplicateMetricsTotal atomic.Value
```
cevian, line 16:
```suggestion
	duplicateMetricsTotal uint64
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,42 @@
...
+	duplicateMetricsTotal.Store(duplicateMetricsTotal.Load().(uint64) + duplicateMetrics)
+	metrics.DuplicateMetrics.Add(float64(duplicateMetrics))
+	launchReporterOnce.Do(func() {
+		go func() {
+			report := time.NewTicker(reportDuplicatesInterval)
+			for range report.C {
+				if duplicateMetricsTotal.Load().(uint64) != 0 {
+					log.Warn("msg", "duplicate data in sample", "total-duplicate-metrics", duplicateMetricsTotal.Load())
+					duplicateMetricsTotal.Store(uint64(0))
+				}
```
cevian, line 38:
```suggestion
				if atomic. LoadUint64(&duplicateMetricsTotal) != 0 {
					log.Warn("msg", "duplicate data in sample", "total-duplicate-metrics", duplicateMetricsTotal.Load())
					atomic.StoreUint64(&duplicateMetricsTotal, uint64(0))
				}
```

cevian (changes requested):
Need a correction for atomic usage

pkg/pgmodel/ingestor/duplicates.go
Harkishen-Singh: Oh yes, sorry I took this as the uber's atomic package usage way.

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func reportDuplicates(duplicateMetrics uint64) {
+	atomic.AddUint64(&duplicateMetricsTotal, duplicateMetrics)
+	metrics.DuplicateMetrics.Add(float64(duplicateMetrics))
+	launchReporterOnce.Do(func() {
+		go func() {
+			report := time.NewTicker(reportDuplicatesInterval)
+			for range report.C {
+				if duplicateMetricsTotal != 0 {
```
cevian, line 35:
```suggestion
				if atomic.LoadUint64(&duplicateMetricsTotal) != 0 {
```

pkg/pgmodel/ingestor/duplicates.go
```diff
@@ -0,0 +1,42 @@
...
+					log.Warn("msg", "duplicate data in sample", "total-duplicate-metrics", duplicateMetricsTotal)
```
cevian, line 36:
```suggestion
					log.Warn("msg", "duplicate data in sample", "total-duplicate-metrics", atomic.LoadUint64(&duplicateMetricsTotal))
```

cevian (approved):
Two more atomic fixes necessary, but I don't need to review again.

PR #416, opened 2021-01-05
Update docker.md with more details and examples.

Harkishen-Singh:

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -5,9 +5,46 @@
 A docker image for the Promscale Connector is available
 on Docker Hub at [timescale/promscale](https://hub.docker.com/r/timescale/promscale/).
 
+To fetch the latest Promscale image, please refer to the [releases](https://github.com/timescale/promscale/releases)
+section of Promscale github repository.
+
 A docker image of TimescaleDB with the `promscale`
 extension is available at on Docker Hub at
-[`timescaledev/promscale-extension:latest-pg12`](https://hub.docker.com/r/timescaledev/promscale-extension).
+[`timescaledev/promscale-extension:latest-pg12`](https://hub.docker.com/r/timescaledev/promscale-extension). You can
+install this via `docker pull timescaledev/promscale-extension:latest-pg12`.
+
+**Example using docker:**
+
+First, let's create a network specific to Promscale-TimescaleDB.
+
+```dockerfile
+docker network create --driver bridge promscale-timescaledb
+```
+
+Install and run TimescaleDB with Promscale extension:
+```dockerfile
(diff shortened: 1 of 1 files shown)


docs/docker.md
```diff
@@ -16,17 +42,28 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
 
 You must tell prometheus to use this remote storage connector by adding
 the following lines to `prometheus.yml`:
-```
+```yaml
 remote_write:
   - url: "http://<connector-address>:9201/write"
 remote_read:
   - url: "http://<connector-address>:9201/read"
+  - read_recent: true
```
VineethReddy02, line 61: This should be `read_recent: true` without `-`

docs/docker.md
```diff
@@ -16,17 +42,28 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
 30 minutes. This is necessary to execute maintenance tasks such as enforcing
 data retention policies according to the configured policy.
 
+**Example:**
+
+When using timescale as the database name:
```
VineethReddy02, line 47: Does the database name matter here? It might be a bit confusing as dbname isn't used in cmd at all. Maybe just add the command that needs to be used.

VineethReddy02 (changes requested):
As the reason behind updating this doc is user has faced an issue while upgrading to the latest version. Maybe a section on upgrade flow would be helpful.

VineethReddy02 (approved):

VineethReddy02:
Left a few comments. Feel free to merge after updating them. No need for another review from my end.

docs/docker.md
```diff
@@ -5,9 +5,41 @@
...
+install this via `docker pull timescaledev/promscale-extension:latest-pg12`.
+
+**Example using plain docker:**
+
+Install and run TimescaleDB with Promscale extension:
+```dockerfile
+docker run --name timescaledb -e POSTGRES_PASSWORD=<password> -d -p 5432:5432 timescaledev/promscale-extension:latest-pg12 postgres -csynchronous_commit=off
+```
+
+Get IP address of the above container which would be supplied as `db-host` flag to Promscale:
```
cevian, line 23: I don't think this is the right way to connect two containers in docker. I think using a bridge network that supports name-based resolution is the way to go https://docs.docker.com/network/network-tutorial-standalone/#use-user-defined-bridge-networks (look for "automatic service discovery")

docs/docker.md
```diff
@@ -16,17 +48,27 @@ procedure on a regular basis (e.g. via cron). We recommend executing it every
...
-```
+```yaml
 remote_write:
   - url: "http://<connector-address>:9201/write"
 remote_read:
   - url: "http://<connector-address>:9201/read"
+    read_recent: true
 ```
 
+**Note:** If you do not want Prometheus to read data from Promscale for those queries beyond Prometheus's tsdb,
```
cevian, line 69: I would remove this, we don't suggest read_recent: false

cevian (approved):

PR #422, opened 2021-01-11
CI flaky issue fix

VineethReddy02:
Having two different test functions is creating flakiness in containers creation. So merging two different tests into one supporting test containers are spun up once on test invocation.

### DIFF
--- a/pkg/tests/upgrade_tests/upgrade_test.go
+++ b/pkg/tests/upgrade_tests/upgrade_test.go
@@ -637,18 +637,11 @@ func copyMetrics(metrics []prompb.TimeSeries) []prompb.TimeSeries {
 	return out
 }
 
-func TestUpgradeExtensions(t *testing.T) {
-	buildPromscaleImageFromRepo(t)
-	testPrereleaseExtensionUpgrade(t)
-	testExtMigrationFailure(t)
-	os.Exit(0)
-}
-
-func testPrereleaseExtensionUpgrade(t *testing.T) {
+func TestExtensionUpgrade(t *testing.T) {
 	var err error
 	var version string
 	ctx := context.Background()
-
+	buildPromscaleImageFromRepo(t)
 	db, dbContainer, closer := startDB(t, ctx)
 	defer closer.Close()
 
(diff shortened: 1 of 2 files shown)


pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -723,9 +716,10 @@ func testPrereleaseExtensionUpgrade(t *testing.T) {
 		}
 		t.Logf("successfully tested extension upgrade flow with --upgrade-prereleases-extensions true")
 	}()
+	os.Exit(0)
 }
 
-func testExtMigrationFailure(t *testing.T) {
+func TestExtMigrationFailure(t *testing.T) {
```
cevian, line 722: do we not need a new image here? I think we do so we need to do a buildPromscaleImageFromRepo(t) in here too.

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -723,9 +716,10 @@ func testPrereleaseExtensionUpgrade(t *testing.T) {
 		}
 		t.Logf("successfully tested extension upgrade flow with --upgrade-prereleases-extensions true")
 	}()
+	os.Exit(0)
```
cevian, line 719: why do we need these exits?

cevian (changes requested):

pkg/tests/upgrade_tests/upgrade_test.go
VineethReddy02: We need but Test functions are executed in sequence the previous Test func is already building the image & it's available for this test as well. So we are skipping the build process which saves 2-3 mins in running the tests. 

The tests are succeeding because the previously built image is available for the current Test func as well.

pkg/tests/upgrade_tests/upgrade_test.go
VineethReddy02: While performing the tests locally to test flakiness. I added them for faster iteration. I noticed after logging the last success log message the exit of test func takes 20-30 secs. So adding it made the test execution faster. 

But removing them in the PR as we don't need them here.

cevian (approved):

PR #423, opened 2021-01-12
Multinode support add node

cevian:
This commit support adding nodes in mutlinode deployments.
To make this work we have to remember the distributed
commands we run on existing nodes (to do things like
set up schema and permissions) and replay them
on new nodes.

We save the commands in a table called `remote_commands`
and use a `add_prom_node` procedure to replay the commands
for new nodes. The remote_commands table has a textual key for easier
manipulation during migrations (e.g. it allows us to update the
table idempotently).

Note that `add_prom_node` needs to be run manually by
the user after a node is added to TimescaleDB with `add_data_node`.


blagojts (approved):

blagojts (approved):

docs/multinode.md
```diff
@@ -0,0 +1,24 @@
...
+## Expanding the cluster
+
+When adding nodes to a TimescaleDB cluster that is already being written to by
+Promscale, you should run the `add_prom_node(node_name)` function
+after running the standard `add_data_node()` function. For example:
+
+```sql
+SELECT add_data_node('example_node_name', host => 'example_host_address')
+SELECT add_prom_node('example_node_name');
+```
```
JLockerman, line 21: can we have `add_prom_node()` just call `add_data_node()`? Would that actually be helpful?

pkg/migrations/sql/preinstall/000-utils.sql
```diff
@@ -1,9 +1,24 @@
-CREATE OR REPLACE PROCEDURE execute_everywhere(command TEXT, transactional BOOLEAN = true)
+ --perms for schema will be addressed later;
+ CREATE SCHEMA IF NOT EXISTS SCHEMA_CATALOG;
```
JLockerman, line 2: kinda sad we have to do this here, but I don't think there's a better way to do this...

pkg/migrations/sql/preinstall/000-utils.sql
```diff
@@ -1,9 +1,24 @@
...
+
+CREATE OR REPLACE PROCEDURE execute_everywhere(command_key text, command TEXT, transactional BOOLEAN = true)
 AS $func$
 BEGIN
     EXECUTE command;
 
     BEGIN
+        INSERT INTO SCHEMA_CATALOG.remote_commands(key, command, transactional) VALUES(command_key, command, transactional)
+        ON CONFLICT (key) DO UPDATE SET command = excluded.command, transactional = excluded.transactional;
+
```
JLockerman: I have a concern about non-idempotent commands, or things like temporarily creating tables, where executing them twice is the correct thing to do; can you expand a bit on why we do it like this instead of recording the command executed twice? The execute-multiple-times alternative _would_ require different functions for recording and replay.

JLockerman: Oh wait, I misunderstood, the current one requires the user to create a unique key for each time `execute_everywhere()` is called. Can you comment on why that instead of just recording the sequence of commands we execute everywhere on the first go, then replaying all of them when we add a new node? (I could see it play _interestingly_ with update scripts, and with enough manual intervention the current version could be better...)

pkg/migrations/sql/versions/dev/0.1.5-dev/1-update_execute_everywhere.sql
```diff
@@ -0,0 +1,30 @@
+--table to save commands so they can be run when adding new nodes
+ CREATE TABLE SCHEMA_CATALOG.remote_commands(
+    key TEXT PRIMARY KEY,
+    seq SERIAL,
+    transactional BOOLEAN,
+    command TEXT
+);
```
JLockerman, line 7: Do we have to pre-populate this table with the non-idempotent stuff? I don't see how it'll end up in the table otherwise...

JLockerman: ahh, I think that's [below](https://github.com/timescale/promscale/pull/423/files#diff-e992ae989c80b1f77be0a5f601764a61035072e169adbc8f931f77d3c8397b0cR22-R116l), and just wasn't in the commit I was reviewing

JLockerman (approved):
Assuming the tests are good, this seems like a reasonable approach. I don't love the amount of work needed to create unique names for each `execute_everywhere()` but I suspect that it may play better with the eventual needs of upgrade scripts.

docs/multinode.md
cevian: that would require putting in host and port and all the other add_data_node parameters in here just to pass down and having things break if we aren't in sync. I think the current approach is better

Issue #351, opened 2020-11-16
Non-leaders should not fill the logs


Harkishen-Singh:
Currently the non-leaders in the leader-election method do not stop logging and hence fill the logs which are enough to be logged just once until a change. This needs to be fixed. Log only if the situation compared to previous is changed.

- [ ] Lesser logging by non-leaders

cevian:
Clarrification: This is only the debug log level. The default log level already doesn't pollute log

PR #425, opened 2021-01-14
Fewer logging of non-leaders in debug log-level

Harkishen-Singh:
Fixes: #351

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -24,6 +25,9 @@ var (
 		func() time.Time { return time.Now().UTC() },
 		"2006-01-02T15:04:05.000Z07:00",
 	)
+
+	logStoreMux sync.Mutex
+	logStore    = make(map[string][]interface{})
 )
 
 // Config represents a logger configuration used upon initialization.
@@ -107,4 +111,37 @@ func parseLogLevel(logLevel string) (level.Option, error) {
 	default:
 		return nil, fmt.Errorf("unrecognized log level %q", logLevel)
 	}
+}
+
+const logOnceTimedDuration = time.Minute
+
+// timedLogger logs from logStore every logOnceTimedDuration. It deletes the log entry from the store
+// after it has logged once.
+func timedLogger() {
+	time.Sleep(logOnceTimedDuration)
+	applyKind := func(logMsg []interface{}) (newLogMsg []interface{}) {
+		newLogMsg = append(newLogMsg, "kind", "rate-limited", "duration", logOnceTimedDuration.String())
+		newLogMsg = append(newLogMsg, logMsg...)
+		return
+	}
+	logStoreMux.Lock()
+	defer logStoreMux.Unlock()
+	for _, logMsg := range logStore {
+		Debug(applyKind(logMsg)...)
+	}
+	logStore = make(map[string][]interface{})
+}
+
+// DebugRateLimited logs Debug level logs once in every logOnceTimedDuration.
+func DebugRateLimited(keyvals ...interface{}) {
+	logStoreMux.Lock()
+	defer logStoreMux.Unlock()
+	val := fmt.Sprintf("%v", keyvals)
+	if len(logStore) == 0 {
+		go timedLogger()
+	}
+	if _, toBeLogged := logStore[val]; toBeLogged {
+		return
+	}
+	logStore[val] = keyvals
 }
--- a/pkg/api/write.go
+++ b/pkg/api/write.go
@@ -47,7 +47,7 @@ func Write(writer ingestor.DBInserter, elector *util.Elector, metrics *Metrics) 
 		}
 		if !shouldWrite {
 			metrics.LeaderGauge.Set(0)
-			log.Debug("msg", fmt.Sprintf("Election id %v: Instance is not a leader. Can't write data", elector.ID()))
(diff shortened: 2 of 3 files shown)


cevian (changes requested):
For debug printing I think it would be better to rate-limit these messages (say print every 1 minute) rather than just once. The reason is in debug mode you want logging to be verbose.

pkg/log/log.go
```diff
@@ -108,3 +112,38 @@ func parseLogLevel(logLevel string) (level.Option, error) {
...
+func timedLogger() {
+	t := time.NewTicker(logOnceTimedDuration)
+	for range t.C {
+		logStoreMux.Lock()
+		for k, logMsg := range logStore {
+			Debug("msg", "timed-debug")
+			Debug(logMsg...)
+			delete(logStore, k)
+		}
+		if len(logStore) == 0 {
```
cevian, line 129: This will always, be true no?

pkg/log/log.go
```diff
@@ -108,3 +112,38 @@ func parseLogLevel(logLevel string) (level.Option, error) {
...
+
+// timedLogger logs from logStore every logOnceTimedDuration. It deletes the log entry from the store
+// after it has logged once.
+func timedLogger() {
+	t := time.NewTicker(logOnceTimedDuration)
+	for range t.C {
+		logStoreMux.Lock()
+		for k, logMsg := range logStore {
+			Debug("msg", "timed-debug")
+			Debug(logMsg...)
```
cevian, line 126:
```suggestion
			Debug("kind", "rate-limited", logMsg...)
```

pkg/log/log.go
cevian: in fact I'm ok with a panic if this isn't true

pkg/log/log.go
```diff
@@ -108,3 +112,38 @@ func parseLogLevel(logLevel string) (level.Option, error) {
...
+		if len(logStore) == 0 {
+			logStoreMux.Unlock()
+			return
+		}
+		logStoreMux.Unlock()
+	}
+}
+
+// DebugTimed logs Debug level logs once in every logOnceTimedDuration.
+func DebugTimed(keyvals ...interface{}) {
```
cevian, line 138:
```suggestion
func DebugRateLimited(keyvals ...interface{}) {
```

pkg/log/log.go
```diff
@@ -108,3 +112,38 @@ func parseLogLevel(logLevel string) (level.Option, error) {
 		return nil, fmt.Errorf("unrecognized log level %q", logLevel)
 	}
 }
+
+const logOnceTimedDuration = time.Second * 10
+
+// timedLogger logs from logStore every logOnceTimedDuration. It deletes the log entry from the store
+// after it has logged once.
+func timedLogger() {
+	t := time.NewTicker(logOnceTimedDuration)
```
cevian, line 121: I think this only needs to fire once, so not a ticker?

cevian (changes requested):

pkg/log/log.go
Harkishen-Singh: Yeah, this should always be true. Here, just for sanitary checks.

pkg/log/log.go
Harkishen-Singh: That will lead to compilation errors. However, I will implement a work-around here.

pkg/log/log.go
```diff
@@ -108,3 +112,42 @@ func parseLogLevel(logLevel string) (level.Option, error) {
...
+	for k, logMsg := range logStore {
+		Debug(applyKind(logMsg)...)
+		delete(logStore, k)
+	}
+	if len(logStore) != 0 {
+		panic(fmt.Sprintf("logStore not empty after flushing rate-limited logs: expected size to be 0 but received %d", len(logStore)))
+	}
+	// delete() does not remove the value from the map. It just sets the flag, meaning memory will increase even though less (at first).
+	// Hence, better to re-initialize it within the lock.
+	logStore = make(map[string][]interface{})
```
cevian, line 138: if we do this then we don't need the delete above (delete can get expensive), Or the panic. I like that more.

pkg/log/log.go
```diff
@@ -108,3 +112,42 @@ func parseLogLevel(logLevel string) (level.Option, error) {
...
+
+// timedLogger logs from logStore every logOnceTimedDuration. It deletes the log entry from the store
+// after it has logged once.
+func timedLogger() {
+	applyKind := func(logMsg []interface{}) (newLogMsg []interface{}) {
+		newLogMsg = append(newLogMsg, "kind", "rate-limited", "duration", logOnceTimedDuration.String())
+		newLogMsg = append(newLogMsg, logMsg...)
+		return
+	}
+	time.Sleep(logOnceTimedDuration)
```
cevian, line 126: nit: let's make this the first line in the function

pkg/log/log.go
Harkishen-Singh: >  I like that more

If we just do the delete, then we aren't clearing the memory used by the map for storing log. That's the behaviour of delete() in map. Hence, I reinitialized it.

Harkishen-Singh: I think, going with just initialization should be better way here in terms of efficiency and memory. Do you agree?

cevian: sorry I meant that I like the initialization more - we should remove the delete and the panic

cevian (approved):

pkg/log/log.go
Harkishen-Singh: oh, got it. The current was in line with that method. No prob.

PR #429, opened 2021-01-15
Update PromQL from upstream upto 7 Jan 

Harkishen-Singh:

This commit updates the promql dir from upstream upto
34656bbf8eb8f45422bf24a9720ea087ab4e6434 commit which is till jan 7.
This update involves fetching latest promtheus commit so that indirect
parser related optimizations can also be uesd by our promql engine.


cevian (approved):
Thanks for doing this

PR #433, opened 2021-01-20
Adjust autovacuum parameters

cevian:
There is an issue in Postgres the stats collector is doing a lot of writes.
Doing some experiments, this seems to get  worse with the number of tables
(with 4k metrics I can get it to do 177M/s!!!).

What happens is that the  autovacuum worker issues a command to get updated
stats every time it checks whether to do a vacuum/analyze operation. This
command rewrites all the stats files in the tmp  dir (4.6M of stuff in my
experiment). In experiments I’ve seen  70 analyzes/s and (70*4.6M and we start
seeing the issue).

This PR fixes the problem by bumping the autovacuum threshold settings. High
threshold settings guarantee that every autovacuum runs takes a longer amount
of time, (e.g. slowing down the 70 analyzes/s number in the example above).

More info about why this problem happens inside PostgreSQL:

The trace inside the autovacuum worker is:
do_autovacuum->table_recheck_autovac->pgstat_fetch_stat_dbentry->backend_read_statsfile->pgstat_send_inquiry

Apparently the stats collector /does/ rate limit the writes of the stats file
normally to once every 500ms but for Autovacuum jobs it rates limits to once
every 10ms! Supposedly this is because autovaccum needs up to date statistics
to prevent it from vacuuming just-vacuumed tables. It’s a poor man’s race
preventor - which obvs causes many other issues.


cevian:
hmm should we be adding some randomness here?

PR #434, opened 2021-01-20
Adds support to make db-connection timeout configurable via CLI.

Harkishen-Singh:

### DIFF
--- a/pkg/util/lock.go
+++ b/pkg/util/lock.go
@@ -16,9 +16,7 @@ import (
 	"github.com/timescale/promscale/pkg/log"
 )
 
-const (
-	waitForConnectionTimeout = time.Second
-)
+const defaultConnectionTimeout = time.Minute
 
 var (
 	SharedLeaseFailure = fmt.Errorf("failed to acquire shared lease")
@@ -202,12 +200,18 @@ func (l *PgAdvisoryLock) getConn(connStr string, cur, maxRetries int) (*pgx.Conn
 		return nil, fmt.Errorf("max attempts reached. giving up on getting a db connection")
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), waitForConnectionTimeout)
-	defer cancel()
-
 	cfg, err := pgx.ParseConfig(connStr)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing config connection: %w", err)
+	}
+
+	ctx := context.Background()
+	if cfg.ConnectTimeout.Seconds() == 0 {
+		// Set the defaultConnectionTimeout if the connection string does not contain the
+		// the connection timeout information.
+		cctx, cancel := context.WithTimeout(context.Background(), defaultConnectionTimeout)
+		defer cancel()
+		ctx = cctx
 	}
 
 	lockConn, err := pgx.ConnectConfig(ctx, cfg)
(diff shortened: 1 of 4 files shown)


pkg/util/lock.go
```diff
@@ -202,7 +200,7 @@ func (l *PgAdvisoryLock) getConn(connStr string, cur, maxRetries int) (*pgx.Conn
 		return nil, fmt.Errorf("max attempts reached. giving up on getting a db connection")
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), waitForConnectionTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), l.connTimeout)
```
cevian, line 203: I think the code should only impose a timeout if one isn't defined by the connection string. Remember a connection string can be define by the -db-* parameters (including -db-connection-timeout) OR by the -db-uri 
parameter. We should only impose a timeout here if the connection string doesn't have one. We can use something like:

```
      ctx := context.Background()
      cfg, err := pgx.ParseConfig(connStr)
      if cfg.ConnectTimeout == 0 {
            ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
            defer cancel()
      }

```

cevian (changes requested):

pkg/util/lock.go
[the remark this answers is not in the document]
Harkishen-Singh: This has been changed in the recent commits, so resolving this.

pkg/util/lock.go
```diff
@@ -202,14 +200,22 @@ func (l *PgAdvisoryLock) getConn(connStr string, cur, maxRetries int) (*pgx.Conn
...
-	defer cancel()
-
 	cfg, err := pgx.ParseConfig(connStr)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing config connection: %w", err)
 	}
 
+	var (
+		ctx    = context.Background()
+		cancel context.CancelFunc
```
cevian, line 210: nit: why declare this outside the if statement, it's only used inside. Just use the `:=` operator to create the cancel inside the if.  

The reason the current code is worse is that currently cancel survive to after line 217 but would be uninitiated if the if block isn't hit. Uninitiated variables are bad and should be avoided if possible.

cevian (approved):
One code quality issue to fix but otherwise looks good.

pkg/util/lock.go
Harkishen-Singh: yep, that's my mistake. I thought it's being used somewhere else which is not local to that block. Anyways, thanks for pointing it out.

Harkishen-Singh: Well, the reason is if i do not declare it here then the `:=` operator will make ctx too local in the block which is a compilation error. So, in order to allow ctx accessible to the below block and that goes into the ConnectConfig, I need to declare both of them outside.

> The reason the current code is worse is that currently cancel survive to after line 217 but would be uninitiated if the if block isn't hit.

If I am not wrong, I read somewhere that the difference between initializing `:=` vs `var` is that the memory is immediately used in `:=` but in `var`, its not used unless the variable is used somewhere. Do you think this is wrong? @cevian

Harkishen-Singh: but still, it looks weird. I agree. I have a workaround that fits you suggestion. Will implement that and push. Thank you.

VineethReddy02 (approved):

PR #450, opened 2021-01-25
Drop __replica__ labelset from samples in HA

VineethReddy02:


pkg/ha/ha_parser.go
```diff
@@ -86,6 +86,14 @@ func (h *haParser) ParseData(tts []prompb.TimeSeries) (map[string][]model.Sample
 			continue
 		}
 
+		// Drop __replica__ labelSet from samples
+		for ind, value := range t.Labels {
+			if value.Name == model.ReplicaNameLabel {
+				t.Labels = append(t.Labels[:ind], t.Labels[ind+1:]...)
```
blagojts, line 92: we have no guarantee that the replica name label will always be in the same position between samples right?

blagojts (approved):

pkg/ha/ha_parser.go
VineethReddy02: In my observation, I haven't noticed it. That's the reason we are iterating over labels to find `cluster` & `replica` labels too for HA logic.

PR #451, opened 2021-01-27
Restructure HA package

VineethReddy02:
Restructure HA package + iterate over review comments.


pkg/ha/ha_parser.go
```diff
@@ -78,7 +78,7 @@ func (h *haParser) ParseData(tts []prompb.TimeSeries) (map[string][]model.Sample
 	}
 
 	// insert allowed -> parse samples
-	acceptedMinTUnix := acceptedMinT.UnixNano() / 1000000
+	acceptedMinTUnix := int64(promModel.TimeFromUnixNano(acceptedMinT.UnixNano()))
```
blagojts, line 81: why are we converting a time.Time to unix nano, and then parsing the unix nano to a time.Time?

blagojts (approved):
looks good, one nit only

pkg/ha/ha_parser.go
VineethReddy02: `UnixNano()` converts the time to nano secs & `TimeFromUnixNano()` converts time in nano secs to milli-seconds. As Prometheus timestamps are in milliseconds we are converting them to milli-sec for comparison with timestamps.

Btw these funcs are imported from Prometheus common module.

Issue #439, opened 2021-01-20
Fix query metric emitted by promscale.


cevian:
Some metrics may only reflect the remote_read path but not the promql endpoint path. We should fix this.

example metric: ReceivedQueries 

Originally reported by @jnadler

PR #453, opened 2021-01-27
Consider metric evaluations in query endpoints & update PromQL engine namespace

Harkishen-Singh:

Fixes: #439

### DIFF
--- a/pkg/api/query.go
+++ b/pkg/api/query.go
@@ -14,12 +14,12 @@ import (
 	"github.com/timescale/promscale/pkg/promql"
 )
 
-func Query(conf *Config, queryEngine *promql.Engine, queryable promql.Queryable) http.Handler {
-	hf := corsWrapper(conf, queryHandler(queryEngine, queryable))
+func Query(conf *Config, queryEngine *promql.Engine, queryable promql.Queryable, metrics *Metrics) http.Handler {
+	hf := corsWrapper(conf, queryHandler(queryEngine, queryable, metrics))
 	return gziphandler.GzipHandler(hf)
 }
 
-func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable) http.HandlerFunc {
+func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable, metrics *Metrics) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		var ts time.Time
 		var err error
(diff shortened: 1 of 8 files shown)


pkg/promql/engine.go
```diff
@@ -69,7 +71,7 @@ type Querier interface {
 }
 
 const (
-	namespace            = "prometheus"
+	namespace            = util.PromNamespace
```
Harkishen-Singh, line 74: Breaking change but necessary. @cevian

Harkishen-Singh: We do not want our PromQL engine stats to have names that match exactly with that of Prometheus's PromQL engine.

cevian: Please make a note on the release notes draft

pkg/api/query.go
```diff
@@ -14,12 +14,12 @@ import (
...
 
-func Query(conf *Config, queryEngine *promql.Engine, queryable promql.Queryable) http.Handler {
-	hf := corsWrapper(conf, queryHandler(queryEngine, queryable))
+func Query(conf *Config, queryEngine *promql.Engine, queryable promql.Queryable, metrics *Metrics) http.Handler {
+	hf := corsWrapper(conf, queryHandler(queryEngine, queryable, metrics))
 	return gziphandler.GzipHandler(hf)
 }
 
-func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable) http.HandlerFunc {
+func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable, metrics *Metrics) http.HandlerFunc {
```
cevian, line 22: This is good. But I think we need to add two metrics: the equivalent of QueryBatchDuration (call it QueryDuration) that times the queries. And the equivalent of InvalidReadReqs (call it InvalidQueryReqs) which traces the queries with incorrect parameters.

pkg/api/query_range.go
```diff
@@ -14,12 +14,12 @@ import (
 	"github.com/timescale/promscale/pkg/promql"
 )
 
-func QueryRange(conf *Config, queryEngine *promql.Engine, queriable promql.Queryable) http.Handler {
-	hf := corsWrapper(conf, queryRange(queryEngine, queriable))
+func QueryRange(conf *Config, queryEngine *promql.Engine, queryable promql.Queryable, metrics *Metrics) http.Handler {
```
cevian, line 17: similar to above we should add: QueryRangeDuration and InvalidQueryRangeReqs

cevian (changes requested):

pkg/promql/engine.go
Harkishen-Singh: Updated the release. I hope the sentence is not too long.

cevian (approved):

Issue #437, opened 2021-01-20
Cache Metrics Should Be Gauges


jnadler:
Promscale exposes metrics for the number of cached metric names and labels.   These are currently counters, but should probably be gauges:

https://github.com/timescale/promscale/blob/f6a700b19f78be5f1499567a7d98028475e7a64f/pkg/pgclient/metrics.go#L25

https://github.com/timescale/promscale/blob/f6a700b19f78be5f1499567a7d98028475e7a64f/pkg/pgclient/metrics.go#L41

jnadler:
i'm happy to PR this if you all agree

Harkishen-Singh:
@jnadler thank you for pointing this out. We would be happy to receive PR from you. Thank you very much.

cevian:
Thanks!

PR #456, opened 2021-01-27
Change cache size metrics to prom gauges

jnadler:
Because these functions return the current number of cached items, they should be gauges not counters.   #437


cevian (approved):
Thanks a bunch @jnadler

PR #458, opened 2021-01-28
Update PromQL upto Jan-20 (@ <timestamp> addition)

blagojts:


Harkishen-Singh (approved):
LGTM! We can merge this after a final look from @cevian

cevian (approved):
Would like to see some end to end test that the enable features flag is correctly passed down, but won't insist on it. You can merge without another review

PR #461, opened 2021-01-28
Fix promQL integration tests

cevian:
Previously the PromQL tests on the real dataset didn't run because
the start and end variables in runPromQLQueryTests were incorrectly
shadowed.

Once that was fixed it was revealed that some of our queried didn't
match Prometheus - this was also a testing (and not real) issue. Turns
out our way of generating prom data was incorrect. Namely, to start up
prom correctly on the data the data should be snapshoted and not simply
left in plain DB mode.

Besides that we have some fixes to equality checking and the like.

### DIFF
--- a/pkg/tests/end_to_end_tests/promql_endpoint_integration_test.go
+++ b/pkg/tests/end_to_end_tests/promql_endpoint_integration_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
+	"math"
 	"math/rand"
 	"net/http"
 	"reflect"
@@ -35,6 +36,7 @@ type sample struct {
 	Raw       string
 	Metric    model.Metric       `json:"metric"`
 	Values    []model.SamplePair `json:"values"`
+	Value     model.SamplePair   `json:"value"`
 }
 
 type dummySample sample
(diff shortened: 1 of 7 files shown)


blagojts (approved):

pkg/tests/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -68,17 +69,30 @@ func (s sample) Equal(other sample) bool {
 		return false
 	}
 
+	checkValues := func(left model.SamplePair, right model.SamplePair) bool {
+		if left.Value.Equal(right.Value) {
+			return true
+		}
+
+		diff := left.Value - right.Value
+		return diff < valueDiffThreshold && diff > -valueDiffThreshold
```
Harkishen-Singh, line 78:
```suggestion
		return math.Abs(diff) < math.Abs(valueDiffThreshold)
```

pkg/tests/end_to_end_tests/promql_endpoint_integration_test.go
```diff
@@ -68,17 +69,30 @@ func (s sample) Equal(other sample) bool {
 		return false
 	}
 
+	checkValues := func(left model.SamplePair, right model.SamplePair) bool {
```
Harkishen-Singh, line 72: How about `checkValues` => `areValuesWithinThreshold`?

pkg/tests/end_to_end_tests/promql_endpoint_integration_test.go
Harkishen-Singh: I also suggest renaming `valueDiffThreshold` to `permittableThreshold`

Harkishen-Singh (approved):

PR #463, opened 2021-01-29
Filter out all samples outside of lease range

blagojts:


pkg/ha/ha_parser.go
```diff
@@ -119,3 +122,15 @@ func (h *haParser) ParseData(tts []prompb.TimeSeries) (map[string][]model.Sample
 
 	return dataSamples, rows, nil
 }
+
+func filterSamples(samples []prompb.Sample, acceptedMinTUnix int64) []prompb.Sample {
```
VineethReddy02, line 126: As the samples are many here can we use `*[]promb.Sample` ?

VineethReddy02 (approved):

PR #465, opened 2021-02-01
Add timeout for tests in go-scheduled.yaml

VineethReddy02:
Add timeout to go schedule job. As it's failing due to timeout  https://github.com/timescale/promscale/runs/1803289305?check_suite_focus=true

We added the timeout in `go.yml` and missed adding to `go-scheduled.yaml` for running the same tests.

### DIFF
--- a/.github/workflows/go-scheduled.yml
+++ b/.github/workflows/go-scheduled.yml
@@ -82,7 +82,7 @@ jobs:
         fi
 
     - name: Test
-      run: go test -v -race ./...
+      run: go test -v -race ./... -timeout 15m
 
     - name: Test extended dataset
       run: go test -v -race -timeout=30m ./pkg/tests/end_to_end_tests/ -extended-test


.github/workflows/go.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         fi
 
     - name: Test
-      run: go test -v -race ./... -timeout 15m
+      run: go test -v -race ./... -timeout 30m
```
Harkishen-Singh, line 87: This will increase the timeout in normal CI which is not failing (and hence the current 10m is fine). The one that is failing is the scheduled CI for which `go-scheduled.yml` is responsible.

If you see that file, you can see the line `go test -v -race -timeout=30m ./pkg/tests/end_to_end_tests/ -extended-test` is responsible for running all tests that need an increased timeout of 30m. This is the one that is failing. But, if you note, this test already has increased timeout, which mean something else if causing the hault.

Harkishen-Singh (changes requested):

Harkishen-Singh:
I think, the reason may be that those tests are unable to pull the rc4 version of timescaledb which is why it is stuck infinitely.

Harkishen-Singh (approved):
:+1: . A look from Mat before we merge.

Issue #443, opened 2021-01-21
Need to support TS_PROM_INSTALL_TIMESCALEDB = false in read only mode


cevian:
Now only supports TS_PROM_INSTALL_TIMESCALEDB being unset

https://timescaledb.slack.com/archives/C011FC0PPC5/p1609876325060700

PR #469, opened 2021-02-01
Add additional parseflag checks when env variables are used

VineethReddy02:
Add additional checks to parseFlags and change `---install-timescaledb` to `--install-extensions` in the checks

Resolves: #443


pkg/runner/runner.go
```diff
@@ -134,9 +134,6 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
 		if flagset["migrate"] || flagset["use-schema-version-lease"] {
 			return nil, fmt.Errorf("Migration flags not supported in read-only mode")
 		}
-		if flagset["install-timescaledb"] {
```
cevian, line 137: we should not remove this completely; there should be an error if someone sets install extension to true. the if should be `if flagset["install-extension"] && cfg.InstallExtensions { return error }

cevian (changes requested):

cevian (approved):

cevian (approved):
Please update the commit&pr title &description

PR #470, opened 2021-02-01
Mask passwords with spaces

### LINKED ISSUE #413: util.MaskPassword doesn't handle passwords with spaces
Hey,

I was looking at my logs from promscale and was pleased to see the PG database password was masked, but then surprised to see only the first word was masked.

Looking at the regexes, they only match until a space is seen. Looks to me like they should match until "( \w+:|$)" is seen. 

Cheers,
Andrew

cevian:
fixes #413.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -24,7 +24,7 @@ const (
 
 var (
 	maskPasswordRegex1 = regexp.MustCompile(`[p|P]assword=(\s*?)'([^']+?)'`)
-	maskPasswordRegex2 = regexp.MustCompile(`[p|P]assword:(\s*?)([^\s]+?)( |$)`)
+	maskPasswordRegex2 = regexp.MustCompile(`[p|P]assword:(\s*)(.*?)(\s*\w+:|$)`)
 	maskPasswordRegex3 = regexp.MustCompile(`postgres:(([^:]*\:){1})([^@]*)`)
 )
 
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -24,7 +24,7 @@ const (
 
 var (
 	maskPasswordRegex1 = regexp.MustCompile(`[p|P]assword=(\s*?)'([^']+?)'`)
```
Harkishen-Singh, line 26: Shouldn't this be space consumable as well?

pkg/util/util_test.go
```diff
@@ -73,9 +73,12 @@ func TestMaskPassword(t *testing.T) {
 		"pass='foobar' host='localhost'":                                     "pass='foobar' host='localhost'",
 		"host='localhost' password='foobar'":                                 "host='localhost' password='****'",
 		"password:foobar  host: localhost":                                   "password:****  host: localhost",
-		"password:foo bar  host: localhost":                                  "password:**** bar  host: localhost",
+		"password:foo bar  host: localhost":                                  "password:****  host: localhost",
```
Harkishen-Singh, line 76: Can we have more cases like `Password: "foo bar"`, `password="foo bar"`, `Password="foo bar"` ?

Harkishen-Singh (approved):
:+1: , some cases if applicable?

pkg/util/util.go
cevian: This works because the class is `NOT '` which includes a space

Issue #129, opened 2020-06-22
Add the ability to print the connector version


JLockerman:
Add a `--version` flag which prints out which version of the connector is running. This is useful when debugging.

VineethReddy02:
Will a cmd like version work? i.e 
```
$ ./timescale-prometheus version
using patched pgconn
Version:  0.1.0-alpha.4.1
```

A simple way of achieving this would be to add a block like this in main.go 

```
	versionFlag := os.Args[1]
	if versionFlag == "version" {
		fmt.Println("Version: ", Version)
		os.Exit(1)
	}
```

OR 

we can use https://github.com/spf13/cobra (CLI library) for adding more cmds in timescale-prometheus like run, version, etc...

I'm open to suggestions & thoughts.

Harkishen-Singh:
if we do `promscale -h`, we already get to know the version at the last line. So, not sure why we need this. Anyways, a `version` flag is common among binaries.

PR #474, opened 2021-02-02
Refactor runner and add support for printing Promscale version.

Harkishen-Singh:
Fixes: #129


pkg/runner/args.go
```diff
@@ -0,0 +1,25 @@
...
+// ParseArgs parses the provided args and prints accordingly. This function should be called before ParseFlags in order
+// to process the non-input flags like "-version".
+func ParseArgs(args []string) (shouldProceed bool) {
+	shouldProceed = true
+	for _, flag := range args {
+		flag = flag[1:]
+		switch flag {
+		case "version":
+			shouldProceed = false
+			fmt.Println("Version: ", version.Version, "; Commit Hash: ", version.CommitHash)
```
cevian, line 21: Let's just print the plain version; This will make it easier for users to integrate this command into automation scripts
```suggestion
			fmt.Println(version.Version)
```

cevian (approved):
One change requested

PR #477, opened 2021-02-02
Add docker-compose file

VineethReddy02:
add docker-compose file, so that we can refer to this compose file in promscale docs in website.

We do have a docker-compose file in `build/` & `docker-compose/high-availability` but this file will as a reference for docs purpose.


cevian (changes requested):
Can we add a test in CI that tests this? should be a much simpler version of https://github.com/timescale/promscale/blob/master/docker-compose/high-availability/test.sh. All we need to check is that the components come up and communicate with each other.

cevian:
Also let's remove the old docker compose file here https://github.com/timescale/promscale/blob/master/build/docker-compose.yml

cevian (approved):

PR #478, opened 2021-02-02
Add ARM builds for docker

### LINKED ISSUE #460: ARM docker image
I'm using promscale in a AMD machine and wanted to run in in a ARM machine too but I couldn't find an ARM docker image,

cevian:
Fixes #460

### DIFF
--- a/build/.goreleaser.yml
+++ b/build/.goreleaser.yml
@@ -14,6 +14,7 @@ builds:
 - env:
   - CGO_ENABLED=0
   main: ./cmd/promscale/
+  id: promscale
   hooks:
     # build el8 rpm
     # will be triggered multiple times but we need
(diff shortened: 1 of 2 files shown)


cevian:
@blagojts I don't really like how much copy/paste goreleaser now has. One alternative could be to use https://github.com/docker/build-push-action What do you think?

build/.goreleaser.yml
```diff
@@ -66,23 +67,39 @@ changelog:
...
+  - "timescale/promscale:{{ .Tag }}-amd64"
 -
-  binaries:
-  - prom-migrator
+  ids: []
+  skip_push: false
+  dockerfile: ./build/Dockerfile
+  use_buildx: true
+  build_flag_templates:
+  - "--pull"
```
blagojts, line 91: maybe we can use yaml anchoring here to reduce the duplicates

blagojts (approved):
I think we can reduce the copy-pasting with yaml anchors, and should go with .goreleaser for this release, but maybe replace it with build-push-action after we've tested it. 
The build-push-action job seems easier to configure, and I'm a bit torn between having smaller/faster CI steps, or having a single CI step for all deliverables

PR #480, opened 2021-02-03
add ha instrumentation + last write timestamp

VineethReddy02:


pkg/ha/ha_service.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+
```
blagojts, line 10: organize imports

blagojts (approved):

PR #490, opened 2021-02-11
Add pprof profiles

wilfriedroset:
While I was investigating promscale memory consumption I've found that heap profiling was missing, therefore let's add it and the rest of them.

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -77,6 +77,11 @@ func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client, 
 	router.Get("/debug/pprof/profile", pprof.Profile)
 	router.Get("/debug/pprof/symbol", pprof.Symbol)
 	router.Get("/debug/pprof/trace", pprof.Trace)
+	router.Get("/debug/pprof/heap", pprof.Handler("heap").ServeHTTP)
+	router.Get("/debug/pprof/goroutine", pprof.Handler("goroutine").ServeHTTP)
+	router.Get("/debug/pprof/threadcreate", pprof.Handler("threadcreate").ServeHTTP)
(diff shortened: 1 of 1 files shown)


cevian (approved):
Thanks @wilfriedroset this LGTM. Want @blagojts to take a look as well.

pkg/api/router.go
```diff
@@ -77,6 +77,21 @@ func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client,
 	router.Get("/debug/pprof/profile", pprof.Profile)
 	router.Get("/debug/pprof/symbol", pprof.Symbol)
 	router.Get("/debug/pprof/trace", pprof.Trace)
+	router.Get("/debug/pprof/heap", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		pprof.Handler("heap").ServeHTTP(w, r)
+	}))
```
blagojts, line 82:
```suggestion
	router.Get("/debug/pprof/heap", pprof.Handler("heap").ServeHTTP)
```
I think we can simplify it a bit

blagojts (changes requested):
Looks good, I think we can just simplify the code a bit since pprof.Handler already creates a http.HandlerFunc

pkg/api/router.go
wilfriedroset: Thank you for your review. this is now taken into account.

blagojts (approved):

pkg/api/router.go
blagojts: Accepted, just rebase the branch on top of master so we can merge

wilfriedroset: done, ready to be merged at your convenience.

PR #491, opened 2021-02-11
remote write benchmarker

cevian:


pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (qmi *qmInfo) run() {
+	ticker := time.NewTicker(shardUpdateDuration)
+	defer ticker.Stop()
+	for range ticker.C {
+		qmi.samplesIn.Tick()
+	}
+}
+
+func getQM(conf *BenchConfig) (*qmInfo, error) {
```
Harkishen-Singh, line 28:
```suggestion
func getQMInfo(conf *BenchConfig) (*qmInfo, error) {
```

pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	common_config "github.com/prometheus/common/config"
+	"github.com/prometheus/common/promlog"
+	"github.com/prometheus/prometheus/config"
+	"github.com/timescale/promscale/pkg/remote"
+)
+
+type qmInfo struct {
+	qm        *remote.QueueManager
+	samplesIn *remote.EwmaRate
```
Harkishen-Singh, line 16:
```suggestion
	samplesInRate *remote.EwmaRate
```

pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+	highestTs *remote.MaxTimestamp
```
Harkishen-Singh, line 17: Comment for this, as per purpose.

pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+func getQM(conf *BenchConfig) (*qmInfo, error) {
+	url, err := url.Parse(conf.WriteEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	rwConf := config.DefaultRemoteWriteConfig
```
Harkishen-Singh, line 34: I havent read evreything, but Prometheus uses [this](https://github.com/prometheus/prometheus/blob/cdc71a1c2ac23d0ed2f7284c2a667fbfc1296414/config/config.go#L115). I wonder if we should use the same, unless some other reason.

pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	qmi.qm = remote.NewQueueManager(
+		remote.NewQueueManagerMetrics(nil, "", ""),
+		nil,
+		nil,
+		promlog.New(&promlog.Config{}),
+		"",
+		qmi.samplesIn,
+		rwConf.QueueConfig,
+		rwConf.MetadataConfig,
```
Harkishen-Singh, line 68: Is this required?

pkg/bench/queue_manager_info.go
```diff
@@ -0,0 +1,80 @@
...
+		nil,
```
Harkishen-Singh, line 69: we might wanna attach external labels which would be the benchmark job name, just like what prom-migrator does.

pkg/bench/runner.go
```diff
@@ -0,0 +1,126 @@
...
+	defer q.Close()
+
+	ss := q.Select(false, nil, labels.MustNewMatcher(labels.MatchRegexp, "", ".*"))
+
+	qmi, err := getQM(conf)
+	if err != nil {
+		return err
+	}
+
+	sth, err := NewSeriesTimeHeap(conf, ss, qmi)
```
Harkishen-Singh, line 66: a more descriptive name for `sth`?

Harkishen-Singh: sounds like `something`

Harkishen-Singh (changes requested):
Partial review. Will do a more detailed in the morning. But, the implementation was quick, so a :+1: 

Points:
1. should the package `bench` be `remote_bench`? bench would block something in future for go benchmarks for promscale.
2. License headers missing for new files.

pkg/bench/queue_manager_info.go
Harkishen-Singh: I think, having this as `*BenchConfig.QueueManager()` would be more suitable.

pkg/bench/series_time_heap.go
```diff
@@ -0,0 +1,110 @@
...
+
+type SeriesTimeHeap []*SeriesItem
+
+func NewSeriesTimeHeap(conf *BenchConfig, ss storage.SeriesSet, qmi *qmInfo) (SeriesTimeHeap, error) {
+	sth := make(SeriesTimeHeap, 0, 100)
+	refSeries := make([]record.RefSeries, 10000)
+	for ss.Next() {
+		series := ss.At()
+		it := series.Iterator()
+		it.Next()
```
Harkishen-Singh, line 30: Are we missing the first time-series due to this?

Harkishen-Singh:
Another thing, should the number of shards be constant? Otherwise, the test results on one storage might not be same for another storage, due to more number of shards, so even if the other storage is slow, due to increased shards, it performs well in samples delivery.

Btw, do we want the benchmarked to mimic the performance that would be by Prometheus remote write (non-constant shards required), or to measure the performance of a storage system (constant shards required)?

blagojts (approved):
Great and straightforward implementation.

1 question, why is the remote package copied from prometheus?

Harkishen-Singh:
Do we plan to restart work on this?

cevian:
@Harkishen-Singh Yes, I've added a few commits recently towards that end goal

PR #492, opened 2021-02-12
Support for warning user on slow data ingestion.

Harkishen-Singh:

This commit adds support for warning user if:
1. The ratio between incoming samples and outgoing samples is more than
   the threshold ratio value.
2. The time taken by a batch is above the threshold time alloted.

### DIFF
--- a/pkg/ewma/ewma.go
+++ b/pkg/ewma/ewma.go
@@ -1,0 +1,61 @@
+// This code has been borrowed from https://github.com/prometheus/prometheus/blob/main/storage/remote/ewma.go
+//
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package ewma
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Rate tracks an exponentially weighted moving average of a per-second rate.
+type Rate struct {
+	newEvents int64
+
+	alpha    float64
+	interval time.Duration
+	lastRate float64
+	init     bool
+	mutex    sync.Mutex
+}
+
+// NewEWMARate always allocates a new ewmaRate, as this guarantees the atomically
+// accessed int64 will be aligned on ARM.  See prometheus#2666.
+func NewEWMARate(alpha float64, interval time.Duration) *Rate {
+	return &Rate{
+		alpha:    alpha,
+		interval: interval,
+	}
+}
+
+// Rate returns the per-second rate.
+func (r *Rate) Rate() float64 {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	return r.lastRate
+}
+
+// Tick assumes to be called every r.interval.
+func (r *Rate) Tick() {
+	newEvents := atomic.SwapInt64(&r.newEvents, 0)
+	instantRate := float64(newEvents) / r.interval.Seconds()
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if r.init {
+		r.lastRate += r.alpha * (instantRate - r.lastRate)
+	} else if newEvents > 0 {
+		r.init = true
+		r.lastRate = instantRate
+	}
+}
+
+// Incr counts incr events.
+func (r *Rate) Incr(incr int64) {
+	atomic.AddInt64(&r.newEvents, incr)
+}
--- a/pkg/pgmodel/ingestor/dispatcher.go
+++ b/pkg/pgmodel/ingestor/dispatcher.go
@@ -36,7 +36,7 @@ type pgxDispatcher struct {
 	insertedDatapoints     *int64
 	toCopiers              chan copyRequest
 	seriesEpochRefresh     *time.Ticker
-	doneChannel            chan bool
+	doneChannel            chan struct{}
 	doneWG                 sync.WaitGroup
 	labelArrayOID          uint32
 }
@@ -76,8 +76,9 @@ func newPgxDispatcher(conn pgxconn.PgxConn, cache cache.MetricCache, scache cach
 		toCopiers:              toCopiers,
 		// set to run at half our deletion interval
 		seriesEpochRefresh: time.NewTicker(30 * time.Minute),
-		doneChannel:        make(chan bool),
+		doneChannel:        make(chan struct{}),
 	}
+	runThroughputWatcher(inserter.doneChannel)
 	if cfg.AsyncAcks && cfg.ReportInterval > 0 {
 		inserter.insertedDatapoints = new(int64)
 		reportInterval := int64(cfg.ReportInterval)
@@ -194,7 +195,8 @@ func (p *pgxDispatcher) Close() {
 // actually inserted) and any error.
 // Though we may insert data to multiple tables concurrently, if asyncAcks is
 // unset this function will wait until _all_ the insert attempts have completed.
-func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, error) {
+func (p *pgxDispatcher) InsertData(dataTS model.Data) (uint64, error) {
+	rows := dataTS.Rows
 	var numRows uint64
 	workFinished := &sync.WaitGroup{}
 	workFinished.Add(len(rows))
@@ -213,6 +215,8 @@ func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, err
 	var err error
 	if !p.asyncAcks {
 		workFinished.Wait()
+		decBatch(numRows)
+		registerIngestionDuration(dataTS.InTime)
 		select {
 		case err = <-errChan:
 		default:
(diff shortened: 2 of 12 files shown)


pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+		"less resources allocated, or few concurrent writes (shards) from Prometheus. Please tune your resource to improve performance. ",
+		"Throughput ratio", fmt.Sprintf("%.2f", ratio), "Ratio threshold", reportRatioThreshold)
+}
+
+func (w *throughtputWatcher) watchTime() {
+	t := time.NewTicker(checkIngestionDurationInterval)
+	for range t.C {
+		if w.shouldReportLastDuration.Load().(bool) {
+			w.shouldReportLastDuration.Store(false)
+			log.Warn("msb", "[WARNING] Samples ingestion duration too high: Please make sure you have low network latency and "+
```
blagojts, line 86: tl;dr

pkg/pgmodel/model/samples.go
```diff
@@ -13,6 +13,13 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+// DataWithTS wraps incoming data with its in-timestamp. It is used to warn if the rate
+// of incoming samples vs outgoing samples is too low, based on time.
+type DataWithTS struct {
```
blagojts, line 18: We need to benchmark the memory impact of this (with an overview on GC activity)

blagojts: I have a feeling that we'll need at least an object pool

pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func incSamples(size uint64) {
+	atomic.AddUint64(&watcher.incomingBatches, size)
+}
+
+func decSamples(size uint64) {
+	atomic.AddUint64(&watcher.outgoingBatches, size)
+}
```
blagojts, line 99: can we use a prometheus gauge? I think we can handle an approximate value for this number no need to go atomic

blagojts (changes requested):
I think we need to confirm several things with this change
1. The impact of the new struct that holds the samples and incoming time on memory (and GC) 
2. Can we use approximate values for the counters
3. Can we use prometheus metrics to monitor this
4. Instead of using the new data struct can we just track the request time in the http handlers upon request received and before a response is sent?

pkg/pgmodel/model/interface.go
```diff
@@ -27,7 +27,7 @@ var (
 
 // inserter is responsible for inserting label, series and data into the storage.
 type Inserter interface {
-	InsertNewData(rows map[string][]SamplesInfo) (uint64, error)
+	InsertNewData(rows DataWithTS) (uint64, error)
```
cevian, line 30: why don't we just have this as two parameters instead?

pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+var watcher *throughtputWatcher
+
+func init() {
+	watcher = &throughtputWatcher{
+		incomingBatches: 0,
+		outgoingBatches: 0,
+	}
+	watcher.shouldReportLastDuration.Store(false)
+	go watcher.watch()
+	go watcher.watchTime()
```
cevian, line 53: These two goroutines need to shut down cleanly. See https://github.com/timescale/promscale/commit/dad1c6f660e30c602bd64e0a2186a0ec3f3f8434 for an example

pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+// throughtputWatcher is a light weight samples batch watching type, that serves as a watching
+// routine and keeps the track of incoming write batches. It keeps a ratio of
+// incoming samples vs outgoing samples and warn about low throughput to the user
+// which might be due to external causes like network latency, resources allocated, etc.
+//
+// We watch the batches than the samples in total, since watching batching is fast
+// and does the same purpose of watching samples, but on a higher-level.
+type throughtputWatcher struct {
+	// Warn based on ratio.
+	incomingBatches uint64
```
cevian, line 36: should we use https://github.com/prometheus/prometheus/blob/main/storage/remote/ewma.go here instead?

pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+		itr uint8
+		t   = time.NewTicker(checkRatioInterval)
+	)
+	for range t.C {
+		if w.outgoingBatches > 1 {
+			if ratio := float64(atomic.LoadUint64(&w.incomingBatches)) / float64(atomic.LoadUint64(&w.outgoingBatches)); ratio > reportRatioThreshold {
+				w.warn(ratio)
+			}
+		}
+		if itr > refreshIteration {
```
cevian, line 68: I don't get this. itr is always 0

pkg/pgmodel/ingestor/watcher.go
cevian: It seems like we want to compare smoothed rates. instead of total batches.

pkg/pgmodel/ingestor/watcher.go
cevian, line 86: We don't know if network latency is the culprit
```suggestion
			log.Warn("msg", "[WARNING] Ingestion is taking too long", "ingest duration",  w.lastIngestionDuration.Load().(time.Duration).String(), "threshold", reportDurationThreshold.String())
```

pkg/pgmodel/ingestor/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+			atomic.StoreUint64(&watcher.incomingBatches, 0)
+			atomic.StoreUint64(&watcher.outgoingBatches, 0)
+		}
+	}
+}
+
+func (w *throughtputWatcher) warn(ratio float64) {
+	log.Warn("msg", "[WARNING] Incoming samples rate higher than outgoing samples. This may happen due to poor network latency, "+
```
cevian, line 76: let's see how this code looks after we switch to ewma and then adjust this message.

cevian (changes requested):

pkg/pgmodel/model/interface.go
Harkishen-Singh: Having a different param for time may look trivial as the job of `InsertNewData` is to insert. So an abstraction here should be good. BTW, `DataWithTS` should be `Data` and then it makes proper sense.

pkg/pgmodel/ingestor/watcher.go
Harkishen-Singh: Sorry but I don't understand what's wrong with the current approach? The current way is simple and easier to maintain.

pkg/pgmodel/ingestor/watcher.go
Harkishen-Singh: Why do we need to switch from the current method?

Note: switching from the current method will have to reimplement what prom does, as they are unexported funcs. If we have to reimplement in any way, why not continue with what we have.

pkg/pgmodel/ingestor/watcher.go
Harkishen-Singh: TODO

Harkishen-Singh:
@cevian ewma implementation done.

pkg/ewma/ewma.go
```diff
@@ -0,0 +1,59 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
cevian, line 1: Let's leave an attribution to the prometheus code.

cevian (changes requested):
Looks much better. some naming nits

Harkishen-Singh:
@cevian ping.

cevian (changes requested):

pkg/pgmodel/model/samples.go
```diff
@@ -18,6 +18,13 @@ type Samples interface {
 	getSample(int) *prompb.Sample
 }
 
+// Data wraps incoming data with its in-timestamp. It is used to warn if the rate
+// of incoming samples vs outgoing samples is too low, based on time.
+type Data struct {
```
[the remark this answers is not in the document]
Harkishen-Singh: It is being used in model itself. If we place this in ingestor, I fear for cyclic imports.

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -204,10 +208,16 @@ func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, err
 		// the following is usually non-blocking, just a channel insert
 		p.getMetricBatcher(metricName) <- &insertDataRequest{metric: metricName, data: data, finished: workFinished, errChan: errChan}
 	}
+	reportIncomingBatch(numRows)
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes, this was doubt count, thank you for catching this.

cevian (approved):
LGTM

Harkishen-Singh:
@cevian can you please do a manual merge here?

PR #501, opened 2021-02-19
Remove a layer of caching for series

cevian:
Previously, series were stored in two separate caches:
1) the label interner
2) the handlers series->series id cache

This commit unifies that into 1 cache (the interner) which now
stores both the series and it's id.

The listener for epoch changes has also now moved from a per-handler
thing to a more singleton ingestor proc. There is no reason to
have many epoch change listeners since there is only one epoch.

### DIFF
--- a/pkg/pgmodel/cache/series_cache.go
+++ b/pkg/pgmodel/cache/series_cache.go
@@ -17,7 +17,7 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
-const DefaultSeriesCacheSize = 100000
+const DefaultSeriesCacheSize = 500000
 
 type SeriesCache interface {
 	Reset()
--- a/pkg/pgmodel/cache/series_cache.go
+++ b/pkg/pgmodel/cache/series_cache.go
@@ -17,6 +17,7 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+//this seems like a good upper bound on default /active/ series. Takes about 64MB
 const DefaultSeriesCacheSize = 500000
 
 type SeriesCache interface {
--- a/pkg/pgmodel/cache/series_cache.go
+++ b/pkg/pgmodel/cache/series_cache.go
@@ -2,7 +2,7 @@
 // Please see the included NOTICE for copyright information and
 // LICENSE for a copy of the license.
 
-package scache
+package cache
 
 import (
 	"encoding/binary"
@@ -10,43 +10,64 @@ import (
 	"math"
 	"sort"
 	"strings"
-	"sync"
 
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/timescale/promscale/pkg/clockcache"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
-var LabelsInterner = sync.Map{}
+const DefaultSeriesCacheSize = 100000
 
-//ResetStoredLabels should be concurrency-safe
-func ResetStoredLabels() {
-	//TODO change this when switching to proper cache
-	LabelsInterner.Range(func(key interface{}, value interface{}) bool {
-		LabelsInterner.Delete(key)
-		return true
-	})
+type SeriesCache interface {
+	Reset()
+	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)
+	Len() int
+	Cap() int
 }
 
-// Get the canonical version of a Labels if one exists.
-// input: the string representation of a Labels as defined by getStr()
+type SeriesCacheImpl struct {
+	cache *clockcache.Cache
+}
+
+func NewSeriesCache(max uint64) *SeriesCacheImpl {
+	return &SeriesCacheImpl{
+		clockcache.WithMax(max),
+	}
+}
+
+func (t *SeriesCacheImpl) Len() int {
+	return t.cache.Len()
+}
+
+func (t *SeriesCacheImpl) Cap() int {
+	return t.cache.Cap()
+}
+
+//ResetStoredLabels should be concurrency-safe
+func (t *SeriesCacheImpl) Reset() {
+	t.cache.Reset()
+}
+
+// Get the canonical version of a series if one exists.
+// input: the string representation of a Labels as defined by generateKey()
 // This function should not be called directly, use labelProtosToLabels() or
 // LabelsFromSlice() instead.
-func loadSeries(str string) (l *model.Series) {
-	val, ok := LabelsInterner.Load(str)
+func (t *SeriesCacheImpl) loadSeries(str string) (l *model.Series) {
+	val, ok := t.cache.Get(str)
 	if !ok {
 		return nil
 	}
 	return val.(*model.Series)
 }
 
-// Try to set a Labels as the canonical Labels for a given string
+// Try to set a series as the canonical Series for a given string
 // representation, returning the canonical version (which can be different in
 // the even of multiple goroutines setting labels concurrently).
 // This function should not be called directly, use labelProtosToLabels() or
 // LabelsFromSlice() instead.
-func setSeries(str string, lset *model.Series) *model.Series {
-	val, _ := LabelsInterner.LoadOrStore(str, lset)
+func (t *SeriesCacheImpl) setSeries(str string, lset *model.Series) *model.Series {
+	val, _ := t.cache.Insert(str, lset)
 	return val.(*model.Series)
 }
 
@@ -116,26 +137,26 @@ func generateKey(labels []prompb.Label) (key string, metricName string, error er
 }
 
 // GetSeriesFromLabels converts a labels.Labels to a canonical Labels object
-func GetSeriesFromLabels(ls labels.Labels) (*model.Series, error) {
+func (t *SeriesCacheImpl) GetSeriesFromLabels(ls labels.Labels) (*model.Series, error) {
 	ll := make([]prompb.Label, len(ls))
 	for i := range ls {
 		ll[i].Name = ls[i].Name
 		ll[i].Value = ls[i].Value
 	}
-	l, _, err := GetSeriesFromProtos(ll)
+	l, _, err := t.GetSeriesFromProtos(ll)
 	return l, err
 }
 
 // GetSeriesFromProtos converts a prompb.Label to a canonical Labels object
-func GetSeriesFromProtos(labelPairs []prompb.Label) (*model.Series, string, error) {
+func (t *SeriesCacheImpl) GetSeriesFromProtos(labelPairs []prompb.Label) (*model.Series, string, error) {
 	key, metricName, err := generateKey(labelPairs)
 	if err != nil {
 		return nil, "", err
 	}
-	series := loadSeries(key)
+	series := t.loadSeries(key)
 	if series == nil {
 		series = model.NewSeries(key, labelPairs)
-		series = setSeries(key, series)
+		series = t.setSeries(key, series)
 	}
 
 	return series, metricName, nil
--- a/pkg/pgmodel/ingestor/handler.go
+++ b/pkg/pgmodel/ingestor/handler.go
@@ -8,22 +8,17 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"time"
 
-	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
 	"github.com/timescale/promscale/pkg/pgxconn"
 )
 
 type insertHandler struct {
-	conn               pgxconn.PgxConn
-	input              chan insertDataRequest
-	pending            *pendingBuffer
-	seriesCache        map[string]model.SeriesID
-	seriesCacheEpoch   Epoch
-	seriesCacheRefresh *time.Ticker
-	metricTableName    string
-	toCopiers          chan copyRequest
+	conn            pgxconn.PgxConn
+	input           chan insertDataRequest
+	pending         *pendingBuffer
+	metricTableName string
+	toCopiers       chan copyRequest
 }
 
 func (h *insertHandler) hasPendingReqs() bool {
@@ -31,20 +26,14 @@ func (h *insertHandler) hasPendingReqs() bool {
 }
 
 func (h *insertHandler) blockingHandleReq() bool {
-	for {
-		select {
-		case req, ok := <-h.input:
-			if !ok {
-				return false
-			}
-
-			h.handleReq(req)
-
-			return true
-		case <-h.seriesCacheRefresh.C:
-			h.refreshSeriesCache()
-		}
+	req, ok := <-h.input
+	if !ok {
+		return false
 	}
+
+	h.handleReq(req)
+
+	return true
 }
 
 func (h *insertHandler) nonblockingHandleReq() bool {
@@ -58,39 +47,12 @@ func (h *insertHandler) nonblockingHandleReq() bool {
 }
 
 func (h *insertHandler) handleReq(req insertDataRequest) bool {
-	// we fill in any SeriesIds we have in cache now so we can free any Labels
-	// that are no longer needed, and because the SeriesIds might get flushed.
-	// (neither of these are that critical at the moment)
-	_, epoch := h.fillKnownSeriesIds(req.data)
-	needsFlush := h.pending.addReq(req, epoch)
+	needsFlush := h.pending.addReq(req)
 	if needsFlush {
 		h.flushPending()
 		return true
 	}
 	return false
-}
-
-// Fill in any SeriesIds we already have in cache.
-// This must be idempotent: if called a second time it should not affect any
-// sampleInfo whose series was already set.
-func (h *insertHandler) fillKnownSeriesIds(sampleInfos []model.SamplesInfo) (numMissingSeries int, epoch Epoch) {
-	epoch = h.seriesCacheEpoch
-	for i, series := range sampleInfos {
-		// When we first create the sampleInfos we should have set the seriesID
-		// to -1 for any series whose labels field is nil. Real seriesIds must
-		// always be greater than 0.
-		if series.SeriesID > -1 {
-			continue
-		}
-		id, ok := h.seriesCache[series.Labels.String()]
-		if ok {
-			sampleInfos[i].SeriesID = id
-			series.Labels = nil
-		} else {
-			numMissingSeries++
-		}
-	}
-	return
 }
 
 func (h *insertHandler) flush() {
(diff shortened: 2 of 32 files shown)


blagojts (approved):
Looks a lot simpler now

pkg/pgmodel/model/labels.go
```diff
@@ -9,23 +9,50 @@ import (
...
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+// SeriesID represents a globally unique id for the series. This should be equivalent
+// to the PostgreSQL type in the series table (currently BIGINT).
+type SeriesID int64
+
+const unsetSeriesID = -1
```
Harkishen-Singh, line 24: can we have this as `defaultSeriesID`? unsetSeriesID makes me sound like a function that unsets the seriesID, which is unrelated.

pkg/pgmodel/model/labels.go
```diff
@@ -9,23 +9,50 @@ import (
...
+
+func (s SeriesID) String() string {
+	return strconv.FormatInt(int64(s), 10)
+}
+
+//Epoch represents the series epoch
+type SeriesEpoch int64
```
Harkishen-Singh, line 31: Sorry, but whenever I read the epoch thing, I still get confused even today. Can we please add a good explanation somewhere here, on why is epoch needed and what's it all about? The importance of this would increase more as we have more contributors.

pkg/pgmodel/model/labels.go
```diff
@@ -9,23 +9,50 @@ import (
...
+
+const UnsetSeriesEpoch = -1
```
Harkishen-Singh, line 33: `defaultSeriesEpoch`?

pkg/pgmodel/model/labels.go
```diff
@@ -130,10 +157,13 @@ func LabelProtosToLabels(labelPairs []prompb.Label) (*Labels, string, error) {
...
 	if labels == nil {
-		labels = new(Labels)
-		labels.str = str
-		labels.Names = make([]string, len(labelPairs))
-		labels.Values = make([]string, len(labelPairs))
+		labels = &Labels{
+			Names:    make([]string, len(labelPairs)),
+			Values:   make([]string, len(labelPairs)),
+			str:      str,
+			seriesID: unsetSeriesID,
```
Harkishen-Singh, line 164: A thing for discussion. We create a new Labels object and then store in the cache with the seriesID being `-1`. Then, every time before insertion, we find out that seriesID (from table if its new) and replace -1 with the ID it deserves. Do we make sure that the cache is updated with that finding? If not, we are wasting a lot a CPU and also trips to DB for no reason.

Harkishen-Singh: If we do not update the cache with new series ID, then can we do the findings of a seriesID here and insert directly in the cache with a non-default (non `-1`) seriesID, and it again saves a complete loop from the cache which we would do later on to fill the seriesID to this `-1`.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -92,9 +93,8 @@ func (i *DBIngestor) parseData(tts []prompb.TimeSeries, req *prompb.WriteRequest
 			return nil, rows, errors.ErrNoMetricName
 		}
 		sample := model.SamplesInfo{
-			Labels:   seriesLabels,
-			SeriesID: -1, // sentinel marking the seriesId as unset
-			Samples:  t.Samples,
+			Labels:  seriesLabels,
+			Samples: t.Samples,
```
Harkishen-Singh, line 97: (not in this PR) Instead of making a new local allocation, should SamplesInfo struct just extend the `promb.Timeseries` with just our `Labels`? This will save from local allocation, especially in high ingestion with more samples count.

pkg/pgmodel/model/labels.go
```diff
@@ -164,18 +194,24 @@ func (l *Labels) Equal(b *Labels) bool {
...
+	return l.seriesID != unsetSeriesID
 }
 
-func (l *Labels) Less(i, j int) bool {
-	return l.Names[i] < l.Names[j]
+func (l *Labels) GetSeriesID() (SeriesID, SeriesEpoch, error) {
+	switch l.seriesID {
+	case unsetSeriesID:
+		return 0, 0, fmt.Errorf("Series id not set")
+	case 0:
```
Harkishen-Singh, line 205: Is this possible? (for sake of knowing more)

Harkishen-Singh (changes requested):

pkg/pgmodel/model/labels.go
blagojts: It's an invalid seriesID, shouldn't be called default

pkg/pgmodel/model/labels.go
blagojts: 1. Yes cache is updated, we most likely have an integration test for that
2. If you set the series id here as reported by the db, you waste the batching done currently where all the missing series id being retrieved in one query

pkg/pgmodel/ingestor/ingestor.go
blagojts: I think extensions are just syntactic sugar for writing 
```
type myType{
  typeImExtending
  somethingElse OfType
}
```
The alloc will happen

pkg/pgmodel/model/labels.go
blagojts: There is a readme https://github.com/timescale/promscale/blob/master/pkg/pgmodel/epochs.readme.md but maybe we need to add more details to it

pkg/pgmodel/model/labels.go
Harkishen-Singh: I agree, but something like initSeriesID then?

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: I mean, since the incoming ts are already allocated, so something like

```
type PSSeries struct {
  *promb.Samples
  labels *Labels
}
```

then using it as

```
&PSSeries{&t.Samples, seriesLabels}
```

will this be any good?

Harkishen-Singh: I wrote this assuming it was all pointers coming, but now I see them as not. So, just in case.

pkg/pgmodel/model/labels.go
cevian: ^ what blagoj said

pkg/pgmodel/ingestor/ingestor.go
cevian: Also t.Samples is a slice so what we are doing is just copyng a pointer

pkg/pgmodel/model/labels.go
cevian: Not unless it's unset(programming error)

pkg/pgmodel/model/labels.go
cevian: Yes this way is done for batching. It is updated when set because this is the same object in the cache

pkg/pgmodel/model/labels.go
cevian: renaming to invalidSeriesID

pkg/pgmodel/model/labels.go
Harkishen-Singh: If that has to be programming error, then we should have a explicit check for this in tests, so that it can be caught before merging.

cevian:
Changes up to and including "Optimize memory: unset series labels and names when setting id" save memory by 33%:
Representative heap profile before:
```
(pprof) top
Showing nodes accounting for 422.28MB, 82.02% of 514.87MB total
Dropped 56 nodes (cum <= 2.57MB)
Showing top 10 nodes out of 89
      flat  flat%   sum%        cum   cum%
   89.01MB 17.29% 17.29%   214.66MB 41.69%  github.com/timescale/promscale/pkg/pgmodel/model.LabelProtosToLabels
   82.53MB 16.03% 33.32%    82.53MB 16.03%  strings.(*Builder).grow (inline)
      66MB 12.82% 46.14%       66MB 12.82%  github.com/timescale/promscale/pkg/prompb.(*Label).Unmarshal
   52.52MB 10.20% 56.34%   118.52MB 23.02%  github.com/timescale/promscale/pkg/prompb.(*TimeSeries).Unmarshal
   46.09MB  8.95% 65.29%    46.09MB  8.95%  github.com/timescale/promscale/pkg/pgmodel/ingestor.(*pendingBuffer).addReq
   24.28MB  4.72% 70.00%    24.28MB  4.72%  github.com/timescale/promscale/pkg/pgmodel/ingestor.(*pgxInserter).getMetricInserter
   19.81MB  3.85% 73.85%    36.63MB  7.11%  sync.(*Map).LoadOrStore
   15.13MB  2.94% 76.79%    26.24MB  5.10%  github.com/timescale/promscale/pkg/pgmodel/ingestor.(*insertHandler).setSeriesIds
   13.81MB  2.68% 79.47%    13.81MB  2.68%  sync.(*Map).dirtyLocked
   13.10MB  2.54% 82.02%    13.10MB  2.54%  github.com/jackc/chunkreader/v2.(*ChunkReader).newBuf (inline)
```
and after: 
```
Showing nodes accounting for 294.52MB, 86.59% of 340.13MB total
Dropped 57 nodes (cum <= 1.70MB)
Showing top 10 nodes out of 80
      flat  flat%   sum%        cum   cum%
   66.52MB 19.56% 19.56%    66.52MB 19.56%  strings.(*Builder).grow
   58.15MB 17.10% 36.65%    58.15MB 17.10%  github.com/timescale/promscale/pkg/clockcache.(*Cache).Reset
   45.39MB 13.34% 50.00%    45.39MB 13.34%  github.com/timescale/promscale/pkg/pgmodel/ingestor.(*pendingBuffer).addReq
   44.51MB 13.08% 63.08%    44.51MB 13.08%  github.com/timescale/promscale/pkg/pgmodel/model.NewSeries
   26.39MB  7.76% 70.84%    26.39MB  7.76%  github.com/timescale/promscale/pkg/pgmodel/ingestor.(*pgxInserter).getMetricInserter
   17.01MB  5.00% 75.84%    32.51MB  9.56%  github.com/timescale/promscale/pkg/prompb.(*TimeSeries).Unmarshal
   15.50MB  4.56% 80.40%    15.50MB  4.56%  github.com/timescale/promscale/pkg/prompb.(*Label).Unmarshal
       9MB  2.65% 83.04%        9MB  2.65%  fmt.Sprintf
    6.55MB  1.93% 84.97%     6.55MB  1.93%  github.com/jackc/chunkreader/v2.(*ChunkReader).newBuf (inline)
    5.51MB  1.62% 86.59%    31.88MB  9.37%  github.com/timescale/promscale/pkg/pgmodel/ingestor.doInsert
```
```

File: pkg/pgmodel/cache/series_cache.go [hunk 7]
```diff
@@ -2,70 +2,81 @@
...
-	MetricName string
-	str        string
+const DefaultSeriesCacheSize = 500000
+
+type SeriesCache interface {
+	Reset()
+	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)
+	Len() int
+	Cap() int
+}
```
Harkishen-Singh, line 27: This interface is used only once (which does not justify its need).
Can we not have a generic interface, something like 

```
type Cache interface {
  Len() int
  Cap() int
  Reset()
}
```
And this should be inside `pkg/clockcache` and then imported from there to every other place. This would be awesome for the cache implementation. Else, I see 3 different kinds of interfaces - `SeriesCache`, `MetricCache`, `LabelsCache` and all have a singlar usage (except MetricCache, which can be inside the Cache interface as I said), which is not great.

pkg/pgmodel/ingestor/inserter.go
```diff
@@ -125,9 +166,8 @@ func (p *pgxInserter) InsertData(rows map[string][]model.SamplesInfo) (uint64, e
 		for _, si := range data {
 			numRows += uint64(len(si.Samples))
 		}
-		// insertMetricData() is expected to be non-blocking,
-		// just a channel insert
-		p.insertMetricData(metricName, data, workFinished, errChan)
+		// the following is usually non-blocking, just a channel insert
+		p.getMetricInserter(metricName) <- insertDataRequest{metric: metricName, data: data, finished: workFinished, errChan: errChan}
```
Harkishen-Singh, line 172: Should we be inserting a pointer? given that channels have limitations in speed for large data, which is why we pass address in prom-migrator as well.

pkg/pgmodel/ingestor/inserter.go
```diff
@@ -107,11 +108,11 @@ func (p *pgxInserter) refreshSeriesEpoch(existingEpoch model.SeriesEpoch) model.
 		// and continue execution
 		log.Error("msg", "error refreshing the series cache", "err", err)
```
Harkishen-Singh, line 110: (optional) I think we should always do bare logging from the parent, and from a child. So, this should return an error and `runSeriesEpochSync` would be responsible for logging it out.

pkg/pgmodel/model/series.go
```diff
@@ -89,6 +115,8 @@ func (l *Series) GetSeriesID() (SeriesID, SeriesEpoch, error) {
 
 //note this has to be idempotent
 func (l *Series) SetSeriesID(sid SeriesID, eid SeriesEpoch) {
+	l.lock.Lock()
```
Harkishen-Singh, line 115: This is good.

pkg/pgmodel/cache/series_cache.go
```diff
@@ -2,70 +2,81 @@
...
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
-// Labels stores a labels.Labels in its canonical string representation
-type Labels struct {
-	Names      []string
-	Values     []string
-	MetricName string
-	str        string
+const DefaultSeriesCacheSize = 500000
```
Harkishen-Singh, line 20: Can we please have a comment as to how setting this high helps us, based on the recent benchmarks?

Harkishen-Singh (changes requested):
Few comments, rest all looks good.

pkg/pgmodel/ingestor/handler.go
```diff
@@ -147,118 +104,85 @@ func (h *insertHandler) setSeriesIds(sampleInfos []model.SamplesInfo) (string, E
...
 		batchSeries = append(batchSeries, []*model.SamplesInfo{curr})
 
-		lastSeenLabel = curr.Labels
+		lastSeenLabel = curr.Series
 	}
 
 	if numSQLFunctionCalls != len(batchSeries) {
-		return "", epoch, fmt.Errorf("unexpected difference in numQueries and batchSeries")
+		return fmt.Errorf("unexpected difference in numQueries and batchSeries")
 	}
```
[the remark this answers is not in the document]
cevian: I don't think so. batchSeries is only appended to in line 127 so the continue in those two blocks should not effect the length of batchSeries.

pkg/pgmodel/model/series.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Equal returns true if two Labels are equal
+func (l *Series) Equal(b *Series) bool {
+	return l.str == b.str
+}
+
+func (l *Series) isSeriesIDSetNoLock() bool {
+	return l.seriesID != invalidSeriesID
+}
```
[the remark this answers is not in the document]
cevian: we need this for the NameValues function which already takes it's own lock. Note this is a private function

pkg/pgmodel/cache/series_cache.go
cevian: We have already seperate interfaces for MetricCache and and LabelCache, so I'm just following an established convention. If we want to change, let's discuss but not for this PR

pkg/pgmodel/ingestor/inserter.go
cevian: that's changed in PR 502

pkg/pgmodel/ingestor/handler.go
```diff
@@ -120,21 +80,18 @@ func (h *insertHandler) flushPending() {
...
 // TODO move up to the rest of insertHandler
-func (h *insertHandler) setSeriesIds(sampleInfos []model.SamplesInfo) (string, Epoch, error) {
-	numMissingSeries, epoch := h.fillKnownSeriesIds(sampleInfos)
-
-	if numMissingSeries == 0 {
-		return "", epoch, nil
-	}
-
-	seriesToInsert := make([]*model.SamplesInfo, 0, numMissingSeries)
+func (h *insertHandler) setSeriesIds(sampleInfos []model.SamplesInfo) error {
```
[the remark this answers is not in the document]
cevian: done. (in a more user-friendly error)

pkg/pgmodel/cache/series_cache.go
Harkishen-Singh: We should change that, sure for the other PR, but then we should add a note in the project board.

Harkishen-Singh (approved):
Looks good, but I am not sure if we should go with the merge, unless results from prom-migrator is clear.

Harkishen-Singh:
After careful evaluation of the changes in relation with the perf testing, I don't see any changes over here would slow down in the perf tests. Rather, things like getStr()/parseData() have actually speed up, which is obvious due to change in GetProtoToPRotoLabels.

However, for some reason, in each prom-migrator test, I find this branch lagging behind by at least 1 second. I don't have any proof from profiling, so for now, I think will have to take it as noise and move ahead.

Thank you.

pkg/pgmodel/cache/series_cache.go
blagojts: The MetricCache interface has methods that simplify calling the underlying clock cache,
so we don't have to convert string to []interface{}
```
type MetricCache interface {
	Get(metric string) (string, error)
	Set(metric string, tableName string) error
	Len() int
	Cap() int
}

type LabelsCache interface {
	GetValues(keys []interface{}, valuesOut []interface{}) (numFound int)
	InsertBatch(keys []interface{}, values []interface{}) (numInserted int)
	Len() int
	Cap() int
}
```
And the SeriesCache will have a `GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)` that does not apply to the other caches, we can put the Len, Cap and Reset methods to an interface in the clockcache package, but we either keep these 3 (while extending the clockcache.Cache interface) , or define a new SuperCacheInterface that has methods for all of them, or write new wrappers around the single interface that will hold the logic for the now distinct methods

blagojts: We can end up with 
```go
type SeriesCache interface {
	clockcache.CacheInterface
	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)
}

type seriesCacheImpl struct {
	*clockcache.CacheStruct
}

func GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error){
  panic("implement me")
}
```

pkg/pgmodel/ingestor/inserter.go
```diff
@@ -86,6 +94,42 @@ func (p *pgxInserter) runCompleteMetricCreationWorker() {
...
+		epoch, err = p.refreshSeriesEpoch(epoch)
+		log.Error("msg", "error refreshing the series cache", "err", err)
+	}
+}
+
+func (p *pgxInserter) refreshSeriesEpoch(existingEpoch model.SeriesEpoch) (model.SeriesEpoch, error) {
+	dbEpoch, err := p.getServerEpoch()
+	if err != nil {
+		// Trash the cache just in case an epoch change occurred, seems safer
+		p.scache.Reset()
```
blagojts, line 113: Is it a good idea to store the epoch with each series so we can purge only data from an earlier epoch?

blagojts (approved):
Looks great. Left one question re: cache resets, which if it makes sense might go in a separate PR

pkg/pgmodel/ingestor/inserter.go
cevian: Good idea but I think separate PR

PR #503, opened 2021-02-22
Update PromQL till Feb 20

Harkishen-Singh:

Upto aff3c702abbf13010ce4bd5688938b2ae5d14914, Feb 20.

This commit implements various new PromQL functions that help in @
modifier and some generic functions in PromQL. It also fixes flaky tests
in PromQL's @ modifier.


pkg/promql/functions_test.go
```diff
@@ -42,8 +42,8 @@ func TestDeriv(t *testing.T) {
 	a := storage.Appender(context.Background())
 
 	metric := labels.FromStrings("__name__", "foo")
-	a.Add(metric, 1493712816939, 1.0)
-	a.Add(metric, 1493712846939, 1.0)
+	a.Append(0, metric, 1493712816939, 1.0)
```
cevian, line 45: there are golang-ci lint failures here

cevian (approved):
Looks good if you fix the lint errors

pkg/promql/functions_test.go
Harkishen-Singh: that's from upstream and we planned not to change them according to us. So left it.

PR #509, opened 2021-02-23
Improve state update

blagojts:


pkg/ha/state/state.go
```diff
@@ -88,3 +164,18 @@ func (h *State) Clone() *StateView {
 		RecentLeaderWriteTime: h.recentLeaderWriteTime,
 	}
 }
+
+func exposeHAStateToMetrics(cluster, oldLeader, newLeader string) {
+	if oldLeader != "" {
+		metrics.HAClusterLeaderDetails.WithLabelValues(cluster, oldLeader).Set(0)
+	}
+	metrics.HAClusterLeaderDetails.WithLabelValues(cluster, newLeader).Set(1)
+	if oldLeader == newLeader {
```
cevian, line 173: shouldn't this block be at the very top of the function?

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
 }
 
-func (h *State) UpdateLastLeaderWriteTime() {
-	h._mu.Lock()
-	defer h._mu.Unlock()
-	h.recentLeaderWriteTime = time.Now()
+// SafeGetLeader returns the current leader. To be used
+// when only the current leader is required, so we can avoid
+// a complete state copy with Clone().
+func (h *Lease) SafeGetLeader() string {
```
cevian, line 143: Why Safe?
```suggestion
func (h *Lease) GetLeader() string {
```

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+	h.leaseUntil = stateFromDB.LeaseUntil
+	go exposeHAStateToMetrics(stateFromDB.Cluster, oldLeader, stateFromDB.Leader)
+	return nil
 }
 
-func (h *State) UpdateMaxSeenTime(currentReplica string, currentMaxT time.Time) {
+// UpdateMaxSeenTime updates the maximum data time seen by the current leader,
+// the maximum data time seen by any Prometheus instance/replica, and writes the
+// current real time of when the current leader last sent data samples.
+func (h *Lease) UpdateMaxSeenTime(currentReplica string, currentMaxT, currentTime time.Time) {
```
cevian, line 122:
```suggestion
func (h *Lease) UpdateMaxSeenTime(currentReplica string, currentDataTime, currentWallTime time.Time) {
```

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+// An error is returned as a safecheck if the stateFromDB is for a different cluster.
+func (h *Lease) SetUpdateFromDB(stateFromDB *client.LeaseDBState) error {
+	h._mu.RLock()
+	if h.cluster != stateFromDB.Cluster {
+		errMsg := `Attempt to set lease state for wrong cluster. 
+Lease belongs to cluster [%s], attempted to update with lease for cluster [%s]`
+		return fmt.Errorf(errMsg, h.cluster, stateFromDB.Cluster)
 	}
+	oldLeader := h.leader
+	h._mu.RUnlock()
```
cevian, line 109: yeah I wouldn't do this. Releasing the read lock before grabbing the write is a recipe for deadlocks. Let's just grab the write lock to begin with. If there is an error so be it, it's still ok to grab the write lock to begin with.

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+	h._mu.Lock()
+	defer h._mu.Unlock()
+	h.leader = stateFromDB.Leader
+	h.leaseStart = stateFromDB.LeaseStart
+	h.leaseUntil = stateFromDB.LeaseUntil
+	go exposeHAStateToMetrics(stateFromDB.Cluster, oldLeader, stateFromDB.Leader)
```
cevian, line 115: exposeHAstate seems like a very simple non-blocking function. I'd run it inline 
```suggestion
	exposeHAStateToMetrics(stateFromDB.Cluster, oldLeader, stateFromDB.Leader)
```

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+// signified by the database as the source of truth.
+func (h *Lease) UpdateFromDB(c client.LeaseClient, potentialLeader string, minT, maxT time.Time) error {
+	h._mu.RLock()
+	cluster := h.cluster
+	oldLeader := h.leader
+	h._mu.RUnlock()
+	stateFromDB, err := c.UpdateLease(context.Background(), cluster, potentialLeader, minT, maxT)
+	if err != nil {
+		return fmt.Errorf("could not update lease from db: %#v", err)
+	}
```
cevian, line 89: why not just call SetUpdateFromDB here?
```suggestion
	}
	h.SetUpdateFromDB(stateFromDB)
	return nil
```

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+		recentLeaderWriteTime: currentTime,
+	}, nil
 }
 
-func (h *State) UpdateStateFromDB(latestState *HALockState) {
+// UpdateFromDB uses the supplied client to attempt to update the lease for the potentialLeader.
+// It either updates the lease, or a new leader with the assigned lease interval is set, as
+// signified by the database as the source of truth.
+func (h *Lease) UpdateFromDB(c client.LeaseClient, potentialLeader string, minT, maxT time.Time) error {
+	h._mu.RLock()
```
cevian, line 82: this is good. I think the read lock makes more sense here because we want to run a db command before taking the write lock. (unlike in SetUpdateFromDB)

pkg/ha/state/state.go
```diff
@@ -27,58 +45,116 @@ type StateView struct {
...
+	dbState, err := c.UpdateLease(context.Background(), cluster, potentialLeader, minT, maxT)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new lease: %#v", err)
+	}
+	var maxTimeInstance string
+	if maxT.After(dbState.LeaseUntil) {
+		maxTimeInstance = potentialLeader
+	} else {
+		maxTimeInstance = dbState.Leader
+	}
```
cevian, line 64: Why is it wrong to always use potentialLeade for maxTimeInstance? I mean the maxTimeSeen will be whatever the potentialLeader sent. If there is a good reason let's add a comment.

pkg/ha/client/client.go
```diff
@@ -0,0 +1,103 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
cevian, line 1: I like this, seems very clean.

pkg/ha/ha_service.go
```diff
@@ -179,71 +154,43 @@ func (h *Service) tryChangeLeader(cluster, currentLeader string) {
...
+			context.Background(), cluster, leaseView.MaxTimeInstance, leaseView.MaxTimeSeen,
 		)
 		if err != nil {
 			log.Error("msg", "Couldn't change leader", "err", err)
 			return
 		}
 
-		ss.UpdateStateFromDB(lockState)
-		if lockState.Leader != currentLeader {
+		err = currentLease.SetUpdateFromDB(leaseState)
```
cevian, line 169: so the lease UpdateFromDB path has the service pass the client to the lease, which does the operation on the client and internally sets the state. For changing leader we have the service use the client to do the operation and then change the lease state with setUpdateFromDB. I think we should choose one approach or the other but not both. Either the lease executes the operation on the client or the service does. My hunch is the service uses the client and updates the lease but I am not sure, will leave up to you.

cevian (changes requested):
Looks better. still need some changes. Could be in the PR or a new one

pkg/ha/ha_service.go
```diff
@@ -25,152 +24,128 @@ const (
...
+	newLease, err := state.NewLease(s.leaseClient, clusterName, replicaName, minT, maxT, currentTime)
 	if err != nil {
 		return nil, err
 	}
-
-	ss.UpdateStateFromDB(stateFromDB)
-	return ss.Clone(), nil
+	l, _ = s.state.LoadOrStore(clusterName, newLease)
+	newLease = l.(*state.Lease)
+	return newLease, nil
```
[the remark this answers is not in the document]
blagojts: We use the returned lease so we can call UpdateFromDB without an additional load when `!maxT.Before(leaseView.LeaseUntil)`

pkg/ha/state/state.go
blagojts: RWMutex is not aware that the same routine is holding the RLock, so Lock without the RUnlock will block

pkg/ha/state/state.go
blagojts: I forgot that if leader && maxT > leaseUntil the lease would be extended in the db call

pkg/ha/state/state.go
blagojts: because of the explicit call from outside, i needed a check for cluster!=h.cluster, no longer needed

pkg/ha/state/state.go
```diff
@@ -56,20 +56,14 @@ func NewLease(c client.LeaseClient, cluster, potentialLeader string, minT, maxT,
...
-	}
 
 	return &Lease{
 		cluster:               dbState.Cluster,
 		leader:                dbState.Leader,
 		leaseStart:            dbState.LeaseStart,
 		leaseUntil:            dbState.LeaseUntil,
 		maxTimeSeen:           maxT,
-		maxTimeInstance:       maxTimeInstance,
+		maxTimeInstance:       potentialLeader,
```
VineethReddy02, line 66: I like this new name, For more clarity. :)

VineethReddy02: May be should update the struct field from `maxTimeInstance` to `potentialLeader`

pkg/ha/ha_service.go
```diff
@@ -5,245 +5,170 @@
...
-		h.state.Range(func(c, s interface{}) bool {
+		s.syncTicker.Wait()
+		s.state.Range(func(c, l interface{}) bool {
 			cluster := fmt.Sprint(c)
-			ss := castToState(s)
-			stateView := ss.Clone()
-			latestState, err := h.syncLockStateFromDB(stateView.LeaseStart, stateView.MaxTimeSeenLeader, cluster, stateView.Leader)
+			lease := l.(*state.Lease)
+			stateBeforeUpdate := lease.Clone()
+			err := lease.UpdateFromDB(s.leaseClient, stateBeforeUpdate.Leader, stateBeforeUpdate.LeaseStart, stateBeforeUpdate.MaxTimeSeenLeader)
```
cevian, line 74: Totally optional (and not sure it's a good idea): the fact that the function UpdateFromDB takes so much states that's part of lease itself might imply it's signature should be
`lease.UpdateFromDB(s.leaseClient)`  or rather `lease.UpdateLease(s.leaseClient)`

for CheckLease below we change to:
`lease.CheckLease(s.leaseClient))` where the code block starting with `if !maxT.Before(leaseView.LeaseUntil) {` below moves into lease.CheckLease and uses the times set by `UpdateMaxSeenTime`. That does imply lease should track minT for leader in addition to maxT (so minT should be passed into and set in `UpdateMaxSeenTime`). 

Again, maybe this is a bad idea.

pkg/ha/state/state.go
```diff
@@ -1,23 +1,41 @@
 package state
```
cevian, line 1: should this file be in pkg/ha/state/lease.go or  pkg/ha/lease/lease.go?

pkg/ha/state/state.go
```diff
@@ -27,58 +45,103 @@ type StateView struct {
...
+// TryChangeLeader uses the supplied client to attempt to change the leader
+// of the cluster based on the maximum observed data time and the instance
+// that had it. If updates the lease with the latest state from the database.
+// An error is returned if the db can't be reached.
+func (h *Lease) TryChangeLeader(c client.LeaseClient) error {
+	h._mu.RLock()
+	leaseState, err := c.TryChangeLeader(
+		context.Background(), h.cluster, h.maxTimeInstance, h.maxTimeSeen,
+	)
+	h._mu.RUnlock()
```
cevian, line 97:
```suggestion
	h._mu.RLock()
	cluster := h.cluster
	maxTimeInstance := h.maxTimeInstance
	maxTimeSeen := h.maxTimeSeen
	h._mu.RUnlock()
	leaseState, err := c.TryChangeLeader(
		context.Background(), cluster, maxTimeInstance, maxTimeSeen,
	)
	
```

pkg/ha/state/state.go
```diff
@@ -27,58 +45,103 @@ type StateView struct {
...
 		h.maxTimeSeenLeader = currentMaxT
 	}
+	h.recentLeaderWriteTime = currentWallTime
 }
 
-func (h *State) UpdateLastLeaderWriteTime() {
-	h._mu.Lock()
-	defer h._mu.Unlock()
-	h.recentLeaderWriteTime = time.Now()
+// SafeGetLeader returns the current leader. To be used
```
cevian, line 127:
```suggestion
// GetLeader returns the current leader. To be used
```

cevian (approved):
Some changes requested, but I don't need to review again before merge into ha branch

pkg/ha/state/state.go
blagojts: will leave it as state/lease so we don't end up with an import of lease.Lease

pkg/ha/ha_service.go
blagojts: I think it's not ok, 
1. We need to track minT, and there may be edge cases for when it needs/doesn't need to be updated
2. Sending the leader as an argument provides a good check that the state was not modified from another routine between Clone() and UpdateFromDB

PR #510, opened 2021-02-23
fix CI

VineethReddy02:


cevian (changes requested):
I think this should be replaced by #511

PR #511, opened 2021-02-23
Update accepted TS version

cevian:
Support any 2.x version. By idea the 2.x line should not
break API and is safe to use.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -48,7 +48,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.1.0",
+		Safe: ">=1.7.3 <2.99.0",
 		Warn: ">=1.7.0 <1.7.3",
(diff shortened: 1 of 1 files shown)


pkg/version/version.go
```diff
@@ -48,7 +48,7 @@ var (
 	TimescaleVersionRangeString = struct {
 		Safe, Warn string
 	}{
-		Safe: ">=1.7.3 <2.1.0",
+		Safe: ">=1.7.3 <2.99.0",
```
Harkishen-Singh, line 51: Why not do

```suggestion
		Safe: ">=1.7.3 <2.0.x",
```

as we are not sure if 2.1.x+ might bring some changes, which can be breaking.

Harkishen-Singh (changes requested):

Harkishen-Singh (approved):

Harkishen-Singh:
**Note:** Make a stricter hold on version 2.x incase we break in future.

PR #515, opened 2021-03-03
Make db driver statement cache be configurable

atanasovskib:
By default pgx uses prepared statements implicitly.
But this makes it incompatible with PGBouncer. 

The statements used are cached in each pgx connection
separately, and the size of the statement cache is
configurable. This statement cache helps with query 
performance, disabling it (by preferring the simple
protocol) reduces performance by more than 20%. But 
not having a large enough statement cache can hurt 
performance by up to 50%. 

This PR adds an option for the user to select whether they
want to disable the statement cache by setting the 
`--db-statements-cache` flag to false. The PR also makes 
the statement cache depend on the size of the configured
metrics cache, as the expected number of metrics is a good
estimator for the required statements to be cached.

### DIFF
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -77,14 +77,16 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 		return nil, numCopiers, err
 	}
 
-	var pgConfig *pgxpool.Config
-	var connectionArgsFmt string
+	var (
+		pgConfig          *pgxpool.Config
+		connectionArgsFmt string
+	)
 	if cfg.DbUri == defaultDBUri {
-		connectionArgsFmt = "%s pool_max_conns=%d pool_min_conns=%d"
+		connectionArgsFmt = "%s pool_max_conns=%d pool_min_conns=%d statement_cache_capacity=%d"
 	} else {
-		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d"
+		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d&statement_cache_capacity=%d"
 	}
-	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections)
+	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize)
 	pgConfig, err = pgxpool.ParseConfig(connectionStringWithArgs)
 	if err != nil {
 		log.Error("msg", "configuring connection", "err", util.MaskPassword(err.Error()))
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -86,17 +86,22 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 	} else {
 		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d&statement_cache_capacity=%d"
 	}
-	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize)
+	statementCacheCapacity := cfg.CacheConfig.MetricsCacheSize * 2
+	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, statementCacheCapacity)
 	pgConfig, err = pgxpool.ParseConfig(connectionStringWithArgs)
 	if err != nil {
 		log.Error("msg", "configuring connection", "err", util.MaskPassword(err.Error()))
 		return nil, numCopiers, err
 	}
-	if cfg.PreferSimpleProtocol {
-		log.Info("msg", "Using simple protocol for database connections")
+	if cfg.EnableStatementsCache {
+		pgConfig.AfterRelease = observeStatementCacheState
+		statementCacheEnabled.Set(1)
+		statementCacheCap.Set(float64(statementCacheCapacity))
+	} else {
+		log.Info("msg", "Statements cached disabled, using simple protocol for database connections.")
 		pgConfig.ConnConfig.PreferSimpleProtocol = true
 	}
-	log.Info("msg", util.MaskPassword(connectionStr), "numCopiers", numCopiers, "pool_max_conns", maxConnections, "pool_min_conns", minConnections)
+	log.Info("msg", util.MaskPassword(connectionStr), "numCopiers", numCopiers, "pool_max_conns", maxConnections, "pool_min_conns", minConnections, "statement_cache", cfg.EnableStatementsCache)
 	return pgConfig, numCopiers, nil
 }
 
@@ -134,7 +139,6 @@ func NewClientWithPool(cfg *Config, numCopiers int, dbConn pgxconn.PgxConn) (*Cl
 	}
 
 	InitClientMetrics(client)
-
 	return client, nil
 }
 
@@ -201,4 +205,23 @@ func (c *Client) HealthCheck() error {
 // with the same underlying Querier as the Client.
 func (c *Client) Queryable() promql.Queryable {
 	return c.queryable
+}
+
+func observeStatementCacheState(conn *pgx.Conn) bool {
+	// connections have been opened and are released already
+	// but the Client metrics have not been initialized yet
+	if statementCacheLen == nil {
+		return true
+	}
+	statementCache := conn.StatementCache()
+	if statementCache == nil {
+		statementCacheEnabled.Set(0.0)
+		return true
+	}
+
+	statementCacheEnabled.Set(1.0)
+	statementCacheCap.Set(float64(statementCache.Cap()))
+	statementCacheSize := statementCache.Len()
+	statementCacheLen.Observe(float64(statementCacheSize))
+	return true
 }
--- a/pkg/pgclient/client.go
+++ b/pkg/pgclient/client.go
@@ -78,17 +78,22 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 	}
 
 	var pgConfig *pgxpool.Config
+	var connectionArgsFmt string
 	if cfg.DbUri == defaultDBUri {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s pool_max_conns=%d pool_min_conns=%d"
 	} else {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf("&pool_max_conns=%d&pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d"
 	}
-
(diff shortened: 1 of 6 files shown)


VineethReddy02:
LGTM! Can we add details on this option in FAQ or in docs? It will be easy for users to find this option.

pkg/pgclient/config.go
```diff
@@ -75,6 +77,7 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	fs.IntVar(&cfg.WriteConnectionsPerProc, "db-writer-connection-concurrency", 4, "Maximum number of database connections for writing per go process.")
 	fs.IntVar(&cfg.MaxConnections, "db-connections-max", -1, "Maximum number of connections to the database that should be opened at once. It defaults to 80% of the maximum connections that the database can handle.")
 	fs.StringVar(&cfg.DbUri, "db-uri", defaultDBUri, "TimescaleDB/Vanilla Postgres DB URI. Example DB URI `postgres://postgres:password@localhost:5432/timescale?sslmode=require`")
+	fs.BoolVar(&cfg.PreferSimpleProtocol, "simple-protocol", defaultUseSimpleProtocol, "simple-protocol disables implicit prepared statement usage by the db driver. Set to true if using PGBouncer")
```
Harkishen-Singh, line 80: Please update this in the `docs/cli.md`.

(feel free to ignore the suggestion)
```suggestion
	fs.BoolVar(&cfg.PreferSimpleProtocol, "simple-protocol", defaultUseSimpleProtocol, "Setting 'simple-protocol' to true disables the implicit prepared statement usage by the db driver. This can be done for using PGBouncer.")
```

pkg/pgclient/client.go
```diff
@@ -78,17 +78,22 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 	}
 
 	var pgConfig *pgxpool.Config
+	var connectionArgsFmt string
```
Harkishen-Singh, line 81: Should we combine this and the above statement into a single var block?

Harkishen-Singh (changes requested):

pkg/pgmodel/ingestor/inserter.go
```diff
@@ -106,12 +106,16 @@ func (p *pgxInserter) runSeriesEpochSync() {
...
+		log.Error("msg", "error refreshing the series cache", "err", err)
+	}
 	for {
 		select {
 		case <-p.seriesEpochRefresh.C:
 			epoch, err = p.refreshSeriesEpoch(epoch)
-			log.Error("msg", "error refreshing the series cache", "err", err)
+			if err != nil {
+				log.Error("msg", "error refreshing the series cache", "err", err)
+			}
```
Harkishen-Singh, line 118: If this and the line on 109 are exactly the same, shouldn't we just combine them into a local func(), and reduce more redundancy.

Harkishen-Singh (approved):
Good to go after this nit. (Free to merge after the change)

pkg/pgclient/client.go
```diff
@@ -77,18 +77,25 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
...
 	}
-
+	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize)
+	pgConfig, err = pgxpool.ParseConfig(connectionStringWithArgs)
 	if err != nil {
 		log.Error("msg", "configuring connection", "err", util.MaskPassword(err.Error()))
 		return nil, numCopiers, err
 	}
-
+	if cfg.PreferSimpleProtocol {
```
cevian, line 95: Let's add a gauge that exposes this in telemetry. 

Also, can we expose the StatementCache Cap() and Len() through telemetry. Cap() should probably be a single gauge but for Len() we need a histogram of measurements for all the connections we use. We can probably fill this histogram in the pool's AfterRelease callback.

pkg/pgclient/client.go
```diff
@@ -77,18 +77,25 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
...
+	)
 	if cfg.DbUri == defaultDBUri {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s pool_max_conns=%d pool_min_conns=%d statement_cache_capacity=%d"
 	} else {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf("&pool_max_conns=%d&pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d&statement_cache_capacity=%d"
 	}
-
+	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize)
```
cevian, line 89: There are 2 statements we may prepare per metric one for the data and one for the series.
```suggestion
	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize*2)
```

cevian (changes requested):
Changes look good. 2 things:
1) I think we should expose telemetry about the state of the statement cache's
2) can you put in the PR and/or commit messages your findings about simple protocol vs describe?

pkg/pgclient/client.go
```diff
@@ -193,3 +201,22 @@ func (c *Client) HealthCheck() error {
...
+	statementCache := x.StatementCache()
+	if statementCache == nil {
+		statementCacheEnabled.Set(0.0)
+		return true
+	}
+
+	statementCacheEnabled.Set(1.0)
+	statementCacheCap.Set(float64(statementCache.Cap()))
+	statementCacheSize := statementCache.Len()
+	statementCacheLen.Observe(float64(statementCacheSize))
```
atanasovskib, line 220: statement cache is per connection, and each observation is the cache size per connection

pkg/pgclient/client.go
```diff
@@ -193,3 +201,22 @@ func (c *Client) HealthCheck() error {
...
+		return true
+	}
+	statementCache := conn.StatementCache()
+	if statementCache == nil {
+		statementCacheEnabled.Set(0.0)
+		return true
+	}
+
+	statementCacheEnabled.Set(1.0)
+	statementCacheCap.Set(float64(statementCache.Cap()))
```
cevian, line 218: These metrics are static. so can we move these to be set in `getPgConfig`

pkg/pgclient/client.go
```diff
@@ -77,19 +77,28 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
...
+	)
 	if cfg.DbUri == defaultDBUri {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf(" pool_max_conns=%d pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s pool_max_conns=%d pool_min_conns=%d statement_cache_capacity=%d"
 	} else {
-		pgConfig, err = pgxpool.ParseConfig(connectionStr + fmt.Sprintf("&pool_max_conns=%d&pool_min_conns=%d", maxConnections, minConnections))
+		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d&statement_cache_capacity=%d"
 	}
-
+	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize)
```
cevian, line 89: Didn't see this change implemented?
```suggestion
	connectionStringWithArgs := fmt.Sprintf(connectionArgsFmt, connectionStr, maxConnections, minConnections, cfg.MetricsCacheSize*2)
```

pkg/pgclient/client.go
cevian: looks great

cevian (changes requested):
Looks good. Some minor changes requested. Also can we `put in the PR and/or commit messages your findings about simple protocol vs describe?`

pkg/pgclient/client.go
```diff
@@ -77,19 +77,31 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
...
 	}
-
-	log.Info("msg", util.MaskPassword(connectionStr), "numCopiers", numCopiers, "pool_max_conns", maxConnections, "pool_min_conns", minConnections)
+	if cfg.EnableStatementsCache {
+		pgConfig.AfterRelease = observeStatementCacheState
+		statementCacheEnabled.Set(1)
+		statementCacheCap.Set(float64(statementCacheCapacity))
+	} else {
+		log.Info("msg", "Statements cached disabled, using simple protocol for database connections.")
+		pgConfig.ConnConfig.PreferSimpleProtocol = true
```
cevian, line 102:
```suggestion
		pgConfig.ConnConfig.PreferSimpleProtocol = true
		statementCacheEnabled.Set(0)
		statementCacheCap.Set(0)
```

pkg/pgclient/client.go
```diff
@@ -193,3 +204,22 @@ func (c *Client) HealthCheck() error {
...
+
+func observeStatementCacheState(conn *pgx.Conn) bool {
+	// connections have been opened and are released already
+	// but the Client metrics have not been initialized yet
+	if statementCacheLen == nil {
+		return true
+	}
+	statementCache := conn.StatementCache()
+	if statementCache == nil {
+		statementCacheEnabled.Set(0.0)
```
cevian, line 216:
```suggestion
```

pkg/pgclient/client.go
```diff
@@ -193,3 +204,22 @@ func (c *Client) HealthCheck() error {
...
+		return true
+	}
+
+	statementCacheEnabled.Set(1.0)
+	statementCacheCap.Set(float64(statementCache.Cap()))
```
cevian, line 221:
```suggestion

```

pkg/pgclient/metrics.go
```diff
@@ -72,12 +88,45 @@ func InitClientMetrics(client *Client) {
...
+		statementCacheEnabled,
+		statementCacheCap,
+		statementCacheLen,
+	)
+}
+
+func createStatementCacheLengthHistogramMetric(client *Client) prometheus.Histogram {
+	// we know the upper bound of the cache, so we want
+	// to make that the last bucket of the histogram
+	statementCacheUpperBound := client.metricCache.Cap()
```
cevian, line 108:
```suggestion
	statementCacheUpperBound := client.metricCache.Cap()*2
```

cevian (changes requested):
Some more changes needed

pkg/pgclient/client.go
```diff
@@ -77,19 +77,38 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
...
+		log.Info("msg", "Statements cached disabled, using simple protocol for database connections.")
+		pgConfig.ConnConfig.PreferSimpleProtocol = true
+		statementCacheEnabled.Set(0)
+		statementCacheCap.Set(0)
+	}
+	log.Info("msg", util.MaskPassword(connectionStr),
+		"numCopiers", numCopiers,
+		"pool_max_conns", maxConnections,
+		"pool_min_conns", minConnections,
+		"statement_cache", cfg.EnableStatementsCache,
```
cevian, line 110: should we log the statement cache cap instead (0 if disabled)?

cevian (approved):

PR #516, opened 2021-03-03
Improve series creation and fetching 

cevian:
This commit changes the method we use to fetch and create series.
Before we created series by passing down the array of names and
values directly. This necessited for us to create a series in
it's own transaction because otherwise we could get a deadlock
between label creation ordering of different series.

Now we switched to doing this in 2 parts:
1) create all the labels (sorted beforehand)
2) create series by passing down the label arrays constructed
   from the label ids fetched in step 1.

This has two advantages:
1) in micro-benchmarks this reduced series creation/fetching time by 3x
2) it significantly reduces the number of txns

The one disadvantage is that this necessitates 2 round-trips,
to the db per batch instead of one. C'est la vie.


pkg/pgmodel/ingestor/handler.go
```diff
@@ -72,113 +76,177 @@ func (h *insertHandler) flushPending() {
...
 		return fmt.Errorf("Error setting series ids: %w", err)
 	}
-	defer br.Close()
+	if len(labelArraySet) == 0 {
+		return nil
+	}
 
-	// BEGIN;
-	_, err = br.Exec()
+	labelArrayArray := pgtype.NewArrayType("prom_api.label_array[]", h.labelArrayOID, func() pgtype.ValueTranscoder { return &pgtype.Int4Array{} })
```
blagojts, line 143: can we create this once per insert handler?

blagojts (approved):

cevian:
Some performance results. 
On an overfilled cache (thus fetching mostly already created series): 
insert rate with this change: 40-30k
insert rate before this change: 18k at best

to fill the cache with changes; 2 min and a bit
before change 3 min and a bit

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -740,6 +740,33 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.get_or_create_label_array(text, text[], text[
...
+            LEFT JOIN SCHEMA_CATALOG.label_key_position lkp
+               ON
+               (
+                  lkp.metric_name = get_or_create_label_ids.metric_name AND
+                  lkp.key = kv.key
+               )
+$$
+LANGUAGE SQL VOLATILE;
+COMMENT ON FUNCTION SCHEMA_CATALOG.get_or_create_label_ids(text, text[], text[])
+IS 'converts a metric name, array of keys, and array of values label ids';
```
Harkishen-Singh, line 766: I think the comment isn't explanatory. Converts to what?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -740,6 +740,33 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.get_or_create_label_array(text, text[], text[
...
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_or_create_label_ids(metric_name TEXT, label_keys text[], label_values text[])
+RETURNS TABLE(pos int, id int, label_key text, label_value text) AS $$
+        SELECT
+            -- only call the functions to create new key positions
+            -- and label ids if they don't exist (for performance reasons)
+            coalesce(lkp.pos,
+              SCHEMA_CATALOG.get_or_create_label_key_pos(get_or_create_label_ids.metric_name, kv.key)) idx,
+            coalesce(l.id,
+              SCHEMA_CATALOG.get_or_create_label_id(kv.key, kv.value)) val,
```
Harkishen-Singh, line 751: Haven't gone into details, but is 749 and 751 dependent on each other? (if yes, then I hope they are in proper sync).

pkg/pgmodel/ingestor/handler.go
```diff
@@ -72,113 +76,179 @@ func (h *insertHandler) flushPending() {
...
+				if !ok {
+					labelMap[key] = labelInfo{}
+					labelList.Add(names[i], values[i])
+				}
+			}
 		}
-
-		batch.Queue("BEGIN;")
-		batch.Queue(getSeriesIDForLabelSQL, curr.GetSeries().MetricName(), names, values)
-		batch.Queue("COMMIT;")
```
Harkishen-Singh, line 121: In my cpu profilings, i did get issues here. Removing these should be very helpful.

pkg/pgmodel/ingestor/handler.go
```diff
@@ -72,113 +76,179 @@ func (h *insertHandler) flushPending() {
...
-			//was already set
-			continue
-		}
-		if lastSeenLabel != nil && lastSeenLabel.Equal(curr.GetSeries()) {
-			batchSeries[len(batchSeries)-1] = append(batchSeries[len(batchSeries)-1], curr)
-			continue
+	metricName := seriesToInsert[0].MetricName()
+	labelMap := make(map[labels.Label]labelInfo, len(seriesToInsert))
+	labelList := model.NewLabelList(len(seriesToInsert))
+	{
```
Harkishen-Singh, line 104: Not sure but is this required? (We were against this in prom-migrator)

pkg/pgmodel/ingestor/handler.go
```diff
@@ -72,113 +76,179 @@ func (h *insertHandler) flushPending() {
...
+	}
+	res, err := h.conn.Query(context.Background(), "SELECT r.series_id, l.nr FROM unnest($2::prom_api.label_array[]) WITH ORDINALITY l(elem, nr) INNER JOIN LATERAL _prom_catalog.get_or_create_series_id_for_label_array($1, l.elem) r ON (TRUE)", metricName, labelArrayArray)
+	if err != nil {
+		return fmt.Errorf("Error setting series_id: cannot query for series_id: %w", err)
+	}
+	defer res.Close()
+
+	for res.Next() {
+		var id model.SeriesID
+		var ordinality int64
```
Harkishen-Singh, line 158: we should merge into single var block.

pkg/pgmodel/ingestor/insert.go
```diff
@@ -60,7 +60,7 @@ func initializeInserterRoutine(conn pgxconn.PgxConn, metricName string, complete
 	return tableName, err
 }
 
-func runInserterRoutine(conn pgxconn.PgxConn, input chan *insertDataRequest, metricName string, completeMetricCreationSignal chan struct{}, metricTableNames cache.MetricCache, toCopiers chan copyRequest) {
+func runInserterRoutine(conn pgxconn.PgxConn, input chan *insertDataRequest, metricName string, completeMetricCreationSignal chan struct{}, metricTableNames cache.MetricCache, toCopiers chan copyRequest, labelArrayOID uint32) {
```
Harkishen-Singh, line 63: (optional) This line is already too long. I think we should switch to a struct for this.

Harkishen-Singh (changes requested):
DIdn't go much deep, but should be good to go except for a few nits.

pkg/migrations/sql/idempotent/base.sql
cevian: There should be no dependency here

pkg/pgmodel/ingestor/handler.go
cevian: This is inline, purely for performance reasons. added a comment

pkg/pgmodel/ingestor/insert.go
cevian: fixed formatting; leaving change to struct later

Harkishen-Singh (approved):

PR #521, opened 2021-03-10
Fix potential deadlocks

cevian:
Fixes some potential deadlocks in the new series creation stuff due to the possibility of sql picking unexpected plans. The fixes force a particular ordering for both label and series creation at the SQL layer.


pkg/pgmodel/ingestor/handler.go
```diff
@@ -186,6 +187,10 @@ func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelLi
 	var dbEpoch model.SeriesEpoch
 	maxPos := 0
 
+	items := len(labelList.Names)
+	if items != len(labelMap) {
+		return dbEpoch, 0, fmt.Errorf("Error filling labels: map size doesn't match")
```
atanasovskib, line 192:
```suggestion
		return dbEpoch, 0, fmt.Errorf("Error filling labels: number of items in labelList and LabelMap doesn't match")
```

pkg/pgmodel/ingestor/ingestor_sql_test.go
```diff
@@ -65,7 +65,7 @@ func TestPGXInserterInsertSeries(t *testing.T) {
 					Err: error(nil),
 				},
 				{
-					Sql: "SELECT r.series_id, l.nr FROM unnest($2::prom_api.label_array[]) WITH ORDINALITY l(elem, nr) INNER JOIN LATERAL _prom_catalog.get_or_create_series_id_for_label_array($1, l.elem) r ON (TRUE)",
+					Sql: "SELECT (_prom_catalog.get_or_create_series_id_for_label_array($1, l.elem)).series_id, l.nr FROM unnest($2::prom_api.label_array[]) WITH ORDINALITY l(elem, nr) ORDER BY l.elem",
```
atanasovskib, line 68: I see this repeated multiple times, can we move it to a const?

atanasovskib (approved):
Looks like a good simplification of the query. I left two comments on small nits

PR #529, opened 2021-03-12
Make series cache size auto expand to memory limit

cevian:
Allows us to track approx. how much memory any given cache is
using.

### DIFF
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -15,11 +15,10 @@ type Cache struct {
 	// stores indexes into storage
 	elements map[interface{}]*element
 	storage  []element
-
-	//Information elements protected by elementsLock
-	//size of everything stored by storage, does not include size of the cache structure itself
+	// Information elements:
+	// size of everything stored by storage, does not include size of the cache structure itself
 	dataSize uint64
-	//number of evictions
+	// number of evictions
 	evictions uint64
 
 	// guards next, and len(storage) and ensures that at most one eviction
@@ -49,23 +48,25 @@ func WithMax(max uint64) *Cache {
 	}
 }
 
-// Insert a key/value mapping into the cache if the key is not already present
+// Insert a key/value mapping into the cache if the key is not already present,
+// The sizeBytes represents the in-memory size of the key and value (used to estimate cache size).
 // returns the canonical version of the value
 // and if the value is in the map
-func (self *Cache) Insert(key interface{}, value interface{}, size uint64) (canonicalValue interface{}, in_cache bool) {
+func (self *Cache) Insert(key interface{}, value interface{}, sizeBytes uint64) (canonicalValue interface{}, in_cache bool) {
 	self.insertLock.Lock()
 	defer self.insertLock.Unlock()
 
-	_, canonicalValue, in_cache = self.insert(key, value, size)
+	_, canonicalValue, in_cache = self.insert(key, value, sizeBytes)
 	return
 }
 
 // Insert a batch of keys with their corresponding values.
 // This function will _overwrite_ the keys and values slices with their
 // canonical versions.
+// sizesBytes is the in-memory size of the key+value of each element.
 // returns the number of elements inserted, is lower than len(keys) if insertion
 // starved
-func (self *Cache) InsertBatch(keys []interface{}, values []interface{}, sizes []uint64) int {
+func (self *Cache) InsertBatch(keys []interface{}, values []interface{}, sizesBytes []uint64) int {
 	if len(keys) != len(values) {
 		panic(fmt.Sprintf("keys and values are not the same len. %d keys, %d values", len(keys), len(values)))
 	}
@@ -75,7 +76,7 @@ func (self *Cache) InsertBatch(keys []interface{}, values []interface{}, sizes [
 
 	for idx := range keys {
 		var inserted bool
-		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx], sizes[idx])
+		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx], sizesBytes[idx])
 		if !inserted {
 			return idx
 		}
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -16,6 +16,12 @@ type Cache struct {
 	elements map[interface{}]*element
 	storage  []element
 
+	//Information elements protected by elementsLock
+	//size of everything stored by storage, does not include size of the cache structure itself
+	dataSize uint64
+	//number of evictions
+	evictions uint64
+
 	// guards next, and len(storage) and ensures that at most one eviction
 	// occurs at a time, always grabbed _before_ elementsLock
 	insertLock sync.Mutex
@@ -30,9 +36,10 @@ type element struct {
 
 	// CLOCK marker if this is recently used
 	used uint32
+	size uint64
 
 	// pad Elements out to be cache aligned
-	_ [24]byte
+	_ [16]byte
 }
 
 func WithMax(max uint64) *Cache {
@@ -45,11 +52,11 @@ func WithMax(max uint64) *Cache {
 // Insert a key/value mapping into the cache if the key is not already present
 // returns the canonical version of the value
 // and if the value is in the map
-func (self *Cache) Insert(key interface{}, value interface{}) (canonicalValue interface{}, in_cache bool) {
+func (self *Cache) Insert(key interface{}, value interface{}, size uint64) (canonicalValue interface{}, in_cache bool) {
 	self.insertLock.Lock()
 	defer self.insertLock.Unlock()
 
-	_, canonicalValue, in_cache = self.insert(key, value)
+	_, canonicalValue, in_cache = self.insert(key, value, size)
 	return
 }
 
@@ -58,7 +65,7 @@ func (self *Cache) Insert(key interface{}, value interface{}) (canonicalValue in
 // canonical versions.
 // returns the number of elements inserted, is lower than len(keys) if insertion
 // starved
-func (self *Cache) InsertBatch(keys []interface{}, values []interface{}) int {
+func (self *Cache) InsertBatch(keys []interface{}, values []interface{}, sizes []uint64) int {
 	if len(keys) != len(values) {
 		panic(fmt.Sprintf("keys and values are not the same len. %d keys, %d values", len(keys), len(values)))
 	}
@@ -68,7 +75,7 @@ func (self *Cache) InsertBatch(keys []interface{}, values []interface{}) int {
 
 	for idx := range keys {
 		var inserted bool
-		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx])
+		keys[idx], values[idx], inserted = self.insert(keys[idx], values[idx], sizes[idx])
 		if !inserted {
 			return idx
 		}
@@ -76,7 +83,7 @@ func (self *Cache) InsertBatch(keys []interface{}, values []interface{}) int {
 	return len(keys)
 }
 
-func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey interface{}, canonicalValue interface{}, inserted bool) {
+func (self *Cache) insert(key interface{}, value interface{}, size uint64) (canonicalKey interface{}, canonicalValue interface{}, inserted bool) {
 	elem, present := self.elements[key]
 	if present {
 		// we'll count a double-insert as a hit. See the comment in get
@@ -95,11 +102,15 @@ func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey inte
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
 		delete(self.elements, insertLocation.key)
-		*insertLocation = element{key: key, value: value}
+		self.dataSize -= insertLocation.size
+		self.dataSize += size
+		self.evictions++
+		*insertLocation = element{key: key, value: value, size: size}
 	} else {
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
-		self.storage = append(self.storage, element{key: key, value: value})
+		self.storage = append(self.storage, element{key: key, value: value, size: size})
+		self.dataSize += size
 		insertLocation = &self.storage[len(self.storage)-1]
 	}
 
@@ -286,6 +297,20 @@ func (self *Cache) Len() int {
 	self.elementsLock.RLock()
 	defer self.elementsLock.RUnlock()
 	return len(self.storage)
+}
+
+func (self *Cache) Evictions() uint64 {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	return self.evictions
+}
+
+func (self *Cache) SizeBytes() uint64 {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	//derived from BenchmarkMemoryEmptyCache 120 bytes per element
+	cacheSize := cap(self.storage) * 120
+	return uint64(cacheSize) + self.dataSize
 }
 
 func (self *Cache) Cap() int {
--- a/pkg/clockcache/cache_bench_test.go
+++ b/pkg/clockcache/cache_bench_test.go
@@ -17,7 +17,7 @@ func BenchmarkIntCache(b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		cache.Insert(keys[i], keys[i])
+		cache.Insert(keys[i], keys[i], 32)
 	}
 }
 
@@ -32,7 +32,7 @@ func BenchmarkStringCache(b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		cache.Insert(keys[i], keys[i])
+		cache.Insert(keys[i], keys[i], 32)
 	}
 }
 
@@ -64,7 +64,7 @@ func BenchmarkEviction(b *testing.B) {
 			b.ResetTimer()
 
 			for i := 0; i < b.N; i++ {
-				bval, _ = cache.Insert(insertKeys[i%n], insertKeys[i%n])
+				bval, _ = cache.Insert(insertKeys[i%n], insertKeys[i%n], 16)
 				cache.markAll()
 			}
 		})
@@ -85,9 +85,10 @@ func BenchmarkMembership(b *testing.B) {
 
 			keys := make([]interface{}, n)
 			vals := make([]interface{}, n)
+			sizes := make([]uint64, n)
 
 			for i := 0; i < n; i++ {
-				keys[i], vals[i] = i, i
+				keys[i], vals[i], sizes[i] = i, i, 16
 			}
 
 			rng.Shuffle(len(keys), func(i, j int) {
@@ -95,7 +96,7 @@ func BenchmarkMembership(b *testing.B) {
 			})
 
 			cache := WithMax(uint64(n))
-			cache.InsertBatch(keys, vals)
+			cache.InsertBatch(keys, vals, sizes)
 			b.ReportAllocs()
 			b.ResetTimer()
 
@@ -114,10 +115,11 @@ func BenchmarkNotFound(b *testing.B) {
 
 			keys := make([]interface{}, n)
 			vals := make([]interface{}, n)
+			sizes := make([]uint64, n)
 			gets := make([]interface{}, n)
 
 			for i := 0; i < n; i++ {
-				keys[i], vals[i] = i, i
+				keys[i], vals[i], sizes[i] = i, i, 16
 				gets[i] = n + i
 			}
 
@@ -126,7 +128,7 @@ func BenchmarkNotFound(b *testing.B) {
 			})
 
 			cache := WithMax(uint64(n))
-			cache.InsertBatch(keys, vals)
+			cache.InsertBatch(keys, vals, sizes)
 			b.ReportAllocs()
 			b.ResetTimer()
 
@@ -161,7 +163,7 @@ func BenchmarkInsertUnderCapacity(b *testing.B) {
 			inserts := 0
 			var inserted bool
 			for i := 0; i < b.N; i++ {
-				bval, inserted = cache.Insert(keys[i%n], vals[i%n])
+				bval, inserted = cache.Insert(keys[i%n], vals[i%n], 16)
 				if inserted {
 					inserts++
 				}
@@ -192,7 +194,7 @@ func BenchmarkInsertOverCapacity(b *testing.B) {
 			inserts := 0
 			var inserted bool
 			for i := 0; i < b.N; i++ {
-				bval, inserted = cache.Insert(keys[i%n], vals[i%n])
+				bval, inserted = cache.Insert(keys[i%n], vals[i%n], 16)
 				if inserted {
 					cache.Get(keys[i%n])
 					inserts++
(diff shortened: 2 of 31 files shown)


pkg/clockcache/cache.go
```diff
@@ -30,9 +32,10 @@ type element struct {
 
 	// CLOCK marker if this is recently used
 	used uint32
+	size uint64
 
 	// pad Elements out to be cache aligned
-	_ [24]byte
+	_ [14]byte
```
JLockerman, line 38: If I did my may right, you added either 8 or 12 bytes (8 for the `size`, and possibly 4 for padding), but only removed 10 from the padding 🤔. Incidentally, I'd recommend putting `size` before `used` to make this calculation easier.

pkg/clockcache/cache.go
```diff
@@ -15,6 +15,8 @@ type Cache struct {
 	// stores indexes into storage
 	elements map[interface{}]*element
 	storage  []element
+	//size of everything stored by storage, does not include size of the cache structure itself
+	dataSize uint64
```
JLockerman, line 19: what lock is this guarded by?

JLockerman (commented):
Some nits around locking documentation, and padding size, but it _looks_ reasonable. My one big question is: why do this instead of getting the info from prof?

pkg/clockcache/cache.go
```diff
@@ -16,6 +16,12 @@ type Cache struct {
 	elements map[interface{}]*element
 	storage  []element
 
+	//Information elements protected by elementsLock
```
antekresic, line 19: Go idiom is to group variables that are being protected by a lock together. Makes the code easier to read. 

I suggest removing the blank line above this one and cleaning up the comments.

pkg/clockcache/cache.go
```diff
@@ -45,11 +52,11 @@ func WithMax(max uint64) *Cache {
 // Insert a key/value mapping into the cache if the key is not already present
 // returns the canonical version of the value
 // and if the value is in the map
-func (self *Cache) Insert(key interface{}, value interface{}) (canonicalValue interface{}, in_cache bool) {
+func (self *Cache) Insert(key interface{}, value interface{}, size uint64) (canonicalValue interface{}, in_cache bool) {
```
antekresic, line 55: This and `InsertBatch` need updates to the comments to include the new argument.

~~Also, would it be a good idea to create convenience methods that calculate the sizes themselves? Manually setting the size every time is a bit error-prone, could lead to subtle mistakes.~~

On second thought, I guess it should be fine since we don't really use this directly anyways.

antekresic (approved):
Couple of minor nits regarding comments, other than that looks good to go. :+1:

pkg/clockcache/cache.go
Harkishen-Singh: One stable way i know of is converting anything to a byte slice. For example:

https://golang.org/pkg/encoding/binary/#Write
```
buf := new(bytes.Buffer)
var num int = 1234
err := binary.Write(buf, binary.LittleEndian, num)
if err != nil {
    fmt.Println("binary.Write failed:", err)
}
```

But I am not sure if that's production recommended.

pkg/clockcache/cache.go
```diff
@@ -95,11 +102,15 @@ func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey inte
...
+		self.dataSize += size
+		self.evictions++
+		*insertLocation = element{key: key, value: value, size: size}
 	} else {
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
-		self.storage = append(self.storage, element{key: key, value: value})
+		self.storage = append(self.storage, element{key: key, value: value, size: size})
+		self.dataSize += size
 		insertLocation = &self.storage[len(self.storage)-1]
```
Harkishen-Singh, line 114: [unrelated to PR] Another question, rather a doubt. Why don't we simply store the index of the element in the storage, rather than storing its address? In that way, it's more efficient?

pkg/clockcache/cache_bench_test.go
```diff
@@ -233,3 +235,8 @@ func BenchmarkInsertConcurrent(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkMemoryEmptyCache(b *testing.B) {
+	b.ReportAllocs()
+	WithMax(uint64(b.N))
+}
```
Harkishen-Singh, line 242: Please paste the benchmark results as a comment to the PR.

pkg/clockcache/cache_bench_test.go
```diff
@@ -17,7 +17,7 @@ func BenchmarkIntCache(b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		cache.Insert(keys[i], keys[i])
+		cache.Insert(keys[i], keys[i], 32)
```
Harkishen-Singh, line 20: This is the weird part. How do we know if that's `32`? unsafe tells me `interface` is 16 bits.

pkg/clockcache/cache.go
```diff
@@ -288,6 +299,20 @@ func (self *Cache) Len() int {
 	return len(self.storage)
 }
 
+func (self *Cache) Evictions() uint64 {
```
Harkishen-Singh, line 302: Wouldn't we have the rate calculation functionality for evictions as per plans? (I was expecting some background goroutine to be doing this job somewhere here)

pkg/clockcache/cache.go
```diff
@@ -288,6 +299,20 @@ func (self *Cache) Len() int {
...
+	defer self.elementsLock.RUnlock()
+	return self.evictions
+}
+
+func (self *Cache) SizeBytes() uint64 {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	//derived from BenchmarkMemoryEmptyCache 120 bytes per element
+	cacheSize := cap(self.storage) * 120
+	return uint64(cacheSize) + self.dataSize
```
Harkishen-Singh, line 313: Aren't we overcalculating here? The cap already has the `dataSize` in itself, so the addition of that is not clear to me. Do we have any tests to make sure about this? (as unit tests, not benchmarks, so that its easier to review)

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/timescale/promscale/pkg/limits/mem"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+var (
+	MemoryTargetMetric = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Name:      "memory_target_bytes",
```
Harkishen-Singh, line 22: as per best practices: https://prometheus.io/docs/practices/naming/#metric-names

```suggestion
			Name:      "target_memory_usage_bytes",
```

Harkishen-Singh: The word `target` is misleading. Something like `memory_usage_bytes` seems good to me., as the target system is the current system and current is an obvious word in this context, so it's not needed. Hence, memory_usage_bytes.

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+			Help:      "The number of bytes of memory the system will target using",
```
Harkishen-Singh, line 23: Will target using? Does the kernel reserves itself a constant memory or variable memory?

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/timescale/promscale/pkg/limits/mem"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+var (
+	MemoryTargetMetric = prometheus.NewGauge(
```
Harkishen-Singh, line 19: If we have just a single metric to be declared, I think we should use `promauto`.
https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promauto@v1.9.0

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Name:      "memory_target_bytes",
+			Help:      "The number of bytes of memory the system will target using",
+		})
+)
+
+type PercentageBytes struct {
+	percentage int    //integer [1-100], only set if specified as percentage
```
Harkishen-Singh, line 28: If the range is always 100, shouldn't we use a smaller int, like int8?

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+var (
+	MemoryTargetMetric = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Name:      "memory_target_bytes",
+			Help:      "The number of bytes of memory the system will target using",
+		})
+)
+
+type PercentageBytes struct {
```
Harkishen-Singh, line 27: Comment for this please.

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+func (t *PercentageBytes) SetBytes(bytes uint64) {
+	t.percentage = 0
+	t.bytes = bytes
+}
+
+func (t *PercentageBytes) Get() (int, uint64) {
+	return t.percentage, t.bytes
+}
+
+func (t *PercentageBytes) Set(val string) error {
```
Harkishen-Singh, line 46: This is too generic. A descriptive comment for this should help.

Harkishen-Singh: Also, since the caller will be using either just Percentage or just Bytes, so shouldn't we make this an enum and be more readable?

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+	val = strings.TrimSpace(val)
+	percentage := false
+	if val[len(val)-1] == '%' {
```
Harkishen-Singh, line 49: There are chances of corrupt inputs here. What happens if user inputs a decimal number? We wouldn't do that but shouldn't this be generic? (that's why the enum way)

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+		val = val[:len(val)-1]
+	}
+
+	numeric, err := strconv.ParseInt(val, 10, 64)
+	if err != nil {
+		return fmt.Errorf("Cannot parse target memory: %w", err)
+	}
+
+	if percentage {
+		if numeric < 1 || numeric > 100 {
```
Harkishen-Singh, line 60: How about this?
```suggestion
		if (numeric - 100) < 0 {
```

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+			return fmt.Errorf("Cannot set target memory: percentage must be in the [1,100] range")
+		}
+		t.SetPercent(int(numeric))
```
Harkishen-Singh, line 63: BTW, can you please tell what a '100%' will signify here?

pkg/limits/flags.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// ParseFlags parses the configuration flags for logging.
+func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
+	/* set defaults */
+	sysMem := mem.SystemMemory()
+	if sysMem > 0 {
+		cfg.targetMemoryFlag.SetPercent(80)
+	} else {
+		cfg.targetMemoryFlag.SetBytes(1e9) //1 GB if cannot determine system memory.
```
Harkishen-Singh, line 96: Will this fail, if promscale is running in a container whose max mem limit is say, 500 MB?

pkg/migration-tool/utils/utils.go
```diff
@@ -106,7 +106,8 @@ func HashLabels(lset prompb.Labels) uint64 {
 		b = append(b, seps[0])
 	}
 	sum := xxhash.Sum64(b)
-	labelsCache.Insert(lset.String(), sum)
+	str := lset.String()
+	labelsCache.Insert(str, sum, uint64(len(str)+8+8))
```
Harkishen-Singh, line 110: Is the best way to know the memory of string is by length? (I recall unsafe for a string always gave me 16 bits)

https://play.golang.com/p/_2_g5UTy2pT

IMO, we should do something like,
```suggestion
	labelsCache.Insert(str, sum, uint64(len([]byte(str))+8+8))
```
is more acurate.

Harkishen-Singh (changes requested):
partial review

pkg/pgclient/config.go
```diff
@@ -69,15 +70,16 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	fs.DurationVar(&cfg.DbConnectionTimeout, "db-connection-timeout", defaultConnectionTime, "Timeout for establishing the connection between Promscale and TimescaleDB.")
 	fs.BoolVar(&cfg.AsyncAcks, "async-acks", false, "Acknowledge asynchronous inserts. If this is true, the inserter will not wait after insertion of metric data in the database. This increases throughput at the cost of a small chance of data loss.")
 	fs.IntVar(&cfg.ReportInterval, "tput-report", 0, "Interval in seconds at which throughput should be reported.")
-	fs.Uint64Var(&cfg.LabelsCacheSize, "labels-cache-size", 10000, "Maximum number of labels to cache.")
-	fs.Uint64Var(&cfg.MetricsCacheSize, "metrics-cache-size", cache.DefaultMetricCacheSize, "Maximum number of metric names to cache.")
-	fs.Uint64Var(&cfg.SeriesCacheSize, "series-cache-size", cache.DefaultSeriesCacheSize, "Maximum number of series to cache.")
```
Harkishen-Singh, line 74: Since we are refactoring these, we should refactor the `docs/cli.md` to a separate section of **cache**.

pkg/pgmodel/cache/series_cache.go
```diff
@@ -10,31 +10,92 @@ import (
...
+	Evictions() uint64
 }
 
 type SeriesCacheImpl struct {
-	cache *clockcache.Cache
+	cache        *clockcache.Cache
+	maxSizeBytes uint64
+}
+
+func NewSeriesCache(config Config, closer <-chan struct{}) *SeriesCacheImpl {
```
Harkishen-Singh, line 42: Just as a growing series cache is important, isn't a growing label cache important (which is again a clockcache)? This takes me to next suggestion which is, we should have evictions rate checking and growth in the clockcache, running in a background goroutine.

PS: this looks like Prom's ewma rate.

pkg/pgmodel/cache/series_cache.go
```diff
@@ -10,31 +10,92 @@ import (
...
+
+	if closer != nil {
+		go cache.runSizeCheck(closer)
+	}
+	return cache
+}
+
+func (t *SeriesCacheImpl) runSizeCheck(closer <-chan struct{}) {
+	prev := uint64(0)
+	ticker := time.NewTicker(GrowCheckDuration)
```
Harkishen-Singh, line 56:
```suggestion
	var (
	    prev uint64
	    ticker = time.NewTicker(GrowCheckDuration)
	)
```

pkg/pgmodel/model/series.go
```diff
@@ -98,6 +99,13 @@ func (l *Series) IsSeriesIDSet() bool {
 	return l.isSeriesIDSetNoLock()
 }
 
+//FinalSizeBytes returns the size in bytes /after/ the seriesID is set
+func (l *Series) FinalSizeBytes() uint64 {
+	//size is the base size of the struct + the str and metricName strings
+	//names and values are not counted since they will be nilled out
+	return uint64(unsafe.Sizeof(*l)) + uint64(len(l.str)+len(l.metricName)) // #nosec
```
Harkishen-Singh, line 106: I think we should avoid unsafe package in our builds. I may be wrong but I heard that they have security vulnerabilities.

https://dev.to/jlauinger/exploitation-exercise-with-unsafe-pointer-in-go-information-leak-part-1-1kga

Harkishen-Singh (changes requested):
Another quick review of the remaining.

pkg/pgmodel/model/series.go
Harkishen-Singh: This is more easy to understand my point

https://hackernoon.com/golang-unsafe-type-conversions-and-memory-access-odz3yrl

pkg/clockcache/cache.go
cevian: Added comments and changed variable name to make it clear this is bytes. We are not doing the binary encoding since that is very expensive

pkg/clockcache/cache.go
```diff
@@ -95,11 +102,15 @@ func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey inte
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
 		delete(self.elements, insertLocation.key)
```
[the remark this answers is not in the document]
cevian: We can't do that because the map key is changing

pkg/clockcache/cache.go
```diff
@@ -95,11 +102,15 @@ func (self *Cache) insert(key interface{}, value interface{}) (canonicalKey inte
...
-		*insertLocation = element{key: key, value: value}
+		self.dataSize -= insertLocation.size
+		self.dataSize += size
+		self.evictions++
```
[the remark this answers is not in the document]
cevian: No, we can't do that since we take the elementsLock only after evict(). We need to do that for performance reasons

pkg/clockcache/cache.go
cevian: I think the lookup is more efficient for pointer dereference than an array index but I'm not positive. we can ask Josh.

pkg/clockcache/cache_bench_test.go
cevian: It was 100-120 bytes per element. Which is where I got the 120 for the estimate

pkg/clockcache/cache_bench_test.go
cevian: its for both they key and value

pkg/clockcache/cache.go
cevian: No, the element has only a pointer (interface{} is a pointer) to the key and value 
```
type element struct {
	// The value stored with this element.
	key   interface{}
	value interface{}
...
```
so the data size is in addition to the size of the cache itself

pkg/clockcache/cache.go
cevian: it is but  this loop does not belong in the datastructure itself but in a wrapper

pkg/limits/flags.go
cevian: This gauge is not the current usage it is the upper limit we are targeting

pkg/limits/flags.go
cevian: This is used when deciding whether we can grow internal datastructures (like cache) to stay below the memory target

pkg/limits/flags.go
cevian: I think that's less readable

pkg/pgmodel/cache/series_cache.go
cevian: I don't think this should be part of clockcache, that code is complicated enough and should probably be just the plain datastructure (it's extremely performance critical). We may write a wrapper for clockcache that does this, but let's do in another PR. Also we can do the labels cache in another PR. This is complicated as-is.

cevian: We've never seen the labels cache get too full, so we can wait for another PR

pkg/clockcache/cache_bench_test.go
cevian: this is a test anyway, so shouldn't matter

pkg/limits/flags.go
cevian: I'm not against another name, but I don't think the one you suggest is better

pkg/limits/flags.go
cevian: if we switch to promauto we should do that across the entire codebase and in a separate PR? I think being consistent with the other stuff we have is better for now

pkg/clockcache/cache.go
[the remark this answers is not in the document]
Harkishen-Singh: Oh yes, I missed that. That makes the demand of a reduceable cache more. (Earlier we had issues in high churn when we wanted a reduce in cache size on my proposal, and now keys are never cleared, which makes the case more stronger). We can move forward for now but i think this may be important in future.

pkg/clockcache/cache.go
Harkishen-Singh: If both are nearly same, then if index way is possible, then we should do that as it will make gc work a bit less.

pkg/clockcache/cache.go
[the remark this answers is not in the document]
Harkishen-Singh: Can you please add this in todo? We should really not miss this. @cevian

pkg/clockcache/cache.go
Harkishen-Singh: ToDo in this PR

Harkishen-Singh: Todo for next PR.

pkg/limits/flags.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/timescale/promscale/pkg/limits/mem"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+var (
+	MemoryTargetMetric = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Name:      "max_memory_target_bytes",
+			Help:      "The system will try to keep memory usage below this target",
```
Harkishen-Singh, line 23: This feels less like a description, and more like a conclusion.

Hence how about this?
```suggestion
			Help:      "Maximum memory that can be targeted, in bytes. The system will try to keep memory usage below this target.",
```

pkg/limits/flags.go
```diff
@@ -0,0 +1,138 @@
...
+	case Absolute:
+		cfg.TargetMemoryBytes = cfg.targetMemoryFlag.value
+	default:
+		return fmt.Errorf("Unknown kind of input")
+	}
+	MemoryTargetMetric.Set(float64(cfg.TargetMemoryBytes))
+	return nil
+}
+
+func init() {
```
Harkishen-Singh, line 134: Let's keep this towards the top.

pkg/pgclient/client.go
```diff
@@ -95,14 +95,14 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 
 // NewClientWithPool creates a new PostgreSQL client with an existing connection pool.
 func NewClientWithPool(cfg *Config, numCopiers int, dbConn pgxconn.PgxConn) (*Client, error) {
-	metricsCache := &cache.MetricNameCache{Metrics: clockcache.WithMax(cfg.MetricsCacheSize)}
-	labelsCache := clockcache.WithMax(cfg.LabelsCacheSize)
-	seriesCache := cache.NewSeriesCache(cfg.SeriesCacheSize)
+	closer := make(chan struct{})
```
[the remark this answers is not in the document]
Harkishen-Singh: I think we should implement this for better readability.

pkg/pgmodel/cache/flags.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"flag"
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/timescale/promscale/pkg/limits"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+// This file and its contents are licensed under the Apache License 2.0.
```
Harkishen-Singh, line 12: Header misplaced.

Harkishen-Singh (changes requested):

pkg/pgclient/client.go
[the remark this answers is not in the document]
Harkishen-Singh: @cevian

Harkishen-Singh (approved):
LGTM!

PR #530, opened 2021-03-15
Adds support for handling null chars.

Harkishen-Singh:

### DIFF
--- a/pkg/pgmodel/ingestor/handler.go
+++ b/pkg/pgmodel/ingestor/handler.go
@@ -111,9 +111,6 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 				//was already set
 				continue
 			}
-			// Filter null chars from label names and label values. If this is done in fillLabelIDs(), then labelMap will
-			// throw error since the sanitized key does not have an entry in the map. Hence, the sanitization is done here
-			// to ensure that the labelMap also contains it.
 			for i := range names {
 				key := labels.Label{Name: names[i], Value: values[i]}
 				_, ok = labelMap[key]
@@ -225,11 +222,11 @@ func (h *insertHandler) fillLabelIDs(metricName string, names *pgsafetype.TextAr
 	defer rows.Close()
 
 	count := 0
+	labelName := new(pgsafetype.Text)
+	labelValue := new(pgsafetype.Text)
 	for rows.Next() {
-		labelName := new(pgsafetype.Text)
-		labelValue := new(pgsafetype.Text)
 		res := labelInfo{}
-		err := rows.Scan(&res.Pos, &res.labelID, labelName, labelValue)
+		err := rows.Scan(&res.Pos, &res.labelID, &labelName.String, &labelValue.String)
 		if err != nil {
 			return dbEpoch, 0, fmt.Errorf("Error filling labels in scan: %w", err)
 		}
--- a/pkg/pgmodel/ingestor/handler.go
+++ b/pkg/pgmodel/ingestor/handler.go
@@ -11,7 +11,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
-	"github.com/timescale/promscale/pkg/pgmodel/model/pgsafetype"
+	"github.com/timescale/promscale/pkg/pgmodel/model/pgutf8str"
 	"github.com/timescale/promscale/pkg/pgxconn"
 )
 
@@ -116,7 +116,9 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 				_, ok = labelMap[key]
 				if !ok {
 					labelMap[key] = labelInfo{}
-					labelList.Add(names[i], values[i])
+					if err := labelList.Add(names[i], values[i]); err != nil {
+						return fmt.Errorf("failed to add label to labelList: %w", err)
+					}
 				}
 			}
 		}
@@ -124,21 +126,13 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	if len(labelMap) == 0 {
 		return nil
 	}
-	safeListName := new(pgsafetype.TextArray)
-	if err := safeListName.Set(labelList.Names); err != nil {
-		return err
-	}
-	safeListValue := new(pgsafetype.TextArray)
-	if err := safeListValue.Set(labelList.Values); err != nil {
-		return err
-	}
 
 	//labels have to be created before series are since we need a canonical
 	//ordering for label creation to avoid deadlocks. Otherwise, if we create
 	//the labels for multiple series in same txn as we are creating the series,
 	//the ordering of label creation can only be canonical within a series and
 	//not across series.
-	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, safeListName, safeListValue, labelMap)
+	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, labelList, labelMap)
 	if err != nil {
 		return fmt.Errorf("Error setting series ids: %w", err)
 	}
@@ -188,7 +182,7 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	return nil
 }
 
-func (h *insertHandler) fillLabelIDs(metricName string, names *pgsafetype.TextArray, values *pgsafetype.TextArray, labelMap map[labels.Label]labelInfo) (model.SeriesEpoch, int, error) {
+func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelList, labelMap map[labels.Label]labelInfo) (model.SeriesEpoch, int, error) {
 	//we cannot use the label cache here because that maps label ids => name, value.
 	//what we need here is name, value => id.
 	//we may want a new cache for that, at a later time.
@@ -197,6 +191,7 @@ func (h *insertHandler) fillLabelIDs(metricName string, names *pgsafetype.TextAr
 	var dbEpoch model.SeriesEpoch
 	maxPos := 0
 
+	names, values := labelList.Get()
 	items := len(names.Elements)
 	if items != len(labelMap) {
 		return dbEpoch, 0, fmt.Errorf("Error filling labels: number of items in labelList and labelMap doesn't match")
@@ -221,12 +216,15 @@ func (h *insertHandler) fillLabelIDs(metricName string, names *pgsafetype.TextAr
 	}
 	defer rows.Close()
 
-	count := 0
-	labelName := new(pgsafetype.Text)
-	labelValue := new(pgsafetype.Text)
+	var (
+		count      int
+		labelName  pgutf8str.Text
+		labelValue pgutf8str.Text
+	)
+
 	for rows.Next() {
 		res := labelInfo{}
-		err := rows.Scan(&res.Pos, &res.labelID, &labelName.String, &labelValue.String)
+		err := rows.Scan(&res.Pos, &res.labelID, &labelName, &labelValue)
 		if err != nil {
 			return dbEpoch, 0, fmt.Errorf("Error filling labels in scan: %w", err)
 		}
--- a/pkg/pgmodel/ingestor/handler.go
+++ b/pkg/pgmodel/ingestor/handler.go
@@ -7,11 +7,11 @@ package ingestor
 import (
 	"context"
 	"fmt"
-
 	"github.com/jackc/pgtype"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
+	"github.com/timescale/promscale/pkg/pgmodel/model/pgsafetype"
 	"github.com/timescale/promscale/pkg/pgxconn"
 )
 
@@ -111,6 +111,9 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 				//was already set
 				continue
 			}
+			// Filter null chars from label names and label values. If this is done in fillLabelIDs(), then labelMap will
+			// throw error since the sanitized key does not have an entry in the map. Hence, the sanitization is done here
+			// to ensure that the labelMap also contains it.
 			for i := range names {
 				key := labels.Label{Name: names[i], Value: values[i]}
 				_, ok = labelMap[key]
@@ -124,13 +127,21 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	if len(labelMap) == 0 {
 		return nil
 	}
+	safeListName := new(pgsafetype.TextArray)
+	if err := safeListName.Set(labelList.Names); err != nil {
+		return err
+	}
+	safeListValue := new(pgsafetype.TextArray)
+	if err := safeListValue.Set(labelList.Values); err != nil {
+		return err
+	}
 
 	//labels have to be created before series are since we need a canonical
 	//ordering for label creation to avoid deadlocks. Otherwise, if we create
 	//the labels for multiple series in same txn as we are creating the series,
 	//the ordering of label creation can only be canonical within a series and
 	//not across series.
-	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, labelList, labelMap)
+	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, safeListName, safeListValue, labelMap)
 	if err != nil {
 		return fmt.Errorf("Error setting series ids: %w", err)
 	}
@@ -180,7 +191,7 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	return nil
 }
 
-func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelList, labelMap map[labels.Label]labelInfo) (model.SeriesEpoch, int, error) {
+func (h *insertHandler) fillLabelIDs(metricName string, names *pgsafetype.TextArray, values *pgsafetype.TextArray, labelMap map[labels.Label]labelInfo) (model.SeriesEpoch, int, error) {
 	//we cannot use the label cache here because that maps label ids => name, value.
 	//what we need here is name, value => id.
 	//we may want a new cache for that, at a later time.
@@ -189,14 +200,14 @@ func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelLi
 	var dbEpoch model.SeriesEpoch
 	maxPos := 0
 
-	items := len(labelList.Names)
+	items := len(names.Elements)
 	if items != len(labelMap) {
 		return dbEpoch, 0, fmt.Errorf("Error filling labels: number of items in labelList and labelMap doesn't match")
 	}
 	// The epoch will never decrease, so we can check it once at the beginning,
 	// at worst we'll store too small an epoch, which is always safe
 	batch.Queue(getEpochSQL)
-	batch.Queue("SELECT * FROM "+schema.Catalog+".get_or_create_label_ids($1, $2, $3)", metricName, labelList.Names, labelList.Values)
+	batch.Queue("SELECT * FROM "+schema.Catalog+".get_or_create_label_ids($1, $2, $3)", metricName, names, values)
 	br, err := h.conn.SendBatch(context.Background(), batch)
 	if err != nil {
 		return dbEpoch, 0, fmt.Errorf("Error filling labels: %w", err)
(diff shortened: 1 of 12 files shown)


pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/timescale/promscale/pkg/pgmodel/model"
+)
+
+// Text is a custom pgtype that wraps the pgtype.Text. It is safe from null characters.
+type Text struct {
+	pgtype.Text
+}
+
+func (t *Text) Set(src interface{}) error {
+	text := src.(string)
```
cevian, line 17: this is unsafe we should make sure it's a string and throw an error otherwise

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+// TextArray is a custom pgtype that wraps the pgtype.TextArray. It is safe from null characters.
+type TextArray struct {
+	pgtype.TextArray
+}
+
+func (t *TextArray) Set(src interface{}) error {
+	textArray := src.([]string)
+	for i := range textArray {
+		textArray[i] = sanitizeNullChars(textArray[i])
+	}
```
cevian, line 51: It's probably more efficient to build a pgtype.Text slice here. You should check the work done inside pgtype.TextArray.Set() my hunch would be it does less work when passed a []pgtype.Text

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+func (t *TextArray) Get() interface{} {
+	var (
+		s   []string
+		arr = t.TextArray.Get()
+	)
+	switch v := arr.(type) {
+	case []string:
```
cevian, line 64: how is this ever true? I dont see pgtype.TextArray.Get() ever return a []string

cevian: If it only returns a pgtype.TextArray, perhaps we should too?

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("safe-text: %w", err)
+	}
+	return nil
+}
+
+func (t *Text) Get() interface{} {
+	return revertSanitization(t.Text.Get().(string))
+}
+
+func (t *Text) AssignTo(dst interface{}) error {
```
cevian, line 28: don't we need to revertSanitization after assignment?

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := t.Text.AssignTo(dst); err != nil {
+		return fmt.Errorf("safe-text: %w", err)
+	}
+	return nil
+}
+
+func (t *Text) Scan(src interface{}) error {
```
cevian, line 35: don't we need to sanitizeNullChars here?

cevian (changes requested):
Partial review

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: I will write some go benchmarks and test this along with what is said at https://github.com/timescale/promscale/pull/530#discussion_r595659173

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: I too felt that we will require here, but it seems that the only place that can be used to handle the reverting stuff is the Get(). Seems like Get() is called every time a value is required, which is why you can see from the end to end tests in `null_chars_test.go`.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: Does seem so. The end-to-end tests speak that the current implementation is sufficient to handle this stuff. I will implement some unit tests to make this more easily noticeable/correctable/testable.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (t *TextArray) Scan(src interface{}) error {
+	if err := t.TextArray.Scan(src); err != nil {
+		return fmt.Errorf("safe-text: %w", err)
+	}
+	return nil
+}
+
+func sanitizeNullChars(s string) string {
+	return strings.ReplaceAll(s, model.NullChar, model.NullCharSanitizeString)
```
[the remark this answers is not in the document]
Harkishen-Singh: Benchmark results:

```
// strings.ReplaceALL: BenchmarkSanitizationOnNullChars-8   	  287623	      4669 ns/op
// strings.Map with rune: BenchmarkSanitizationOnNullChars-8   	  308642	      3868 ns/op
```

clearly, strings.Map with rune is the better way.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: Benchmark results

```
// using []pgtype.Text in pgsafetype.TextArray.Set(): BenchmarkTypeArray-8   	   85900	     13161 ns/op
// using []string in pgsafetype.TextArray.Set(): 	  BenchmarkTypeArray-8   	   84028	     16224 ns/op
```
clearly, []pgtype.Text is the better one (as you said) as the results were consistent in repeated benchmarks.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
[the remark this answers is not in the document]
Harkishen-Singh: Updating to strings.Map

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: Updating to pgtype.Text way

pkg/pgmodel/ingestor/handler.go
```diff
@@ -111,6 +111,9 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 				//was already set
 				continue
 			}
+			// Filter null chars from label names and label values. If this is done in fillLabelIDs(), then labelMap will
+			// throw error since the sanitized key does not have an entry in the map. Hence, the sanitization is done here
+			// to ensure that the labelMap also contains it.
```
cevian, line 116: comments are wrong now?

pkg/pgmodel/ingestor/handler.go
```diff
@@ -124,14 +127,25 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	if len(labelMap) == 0 {
 		return nil
 	}
+	safeListName := new(pgsafetype.TextArray)
+	if err := safeListName.Set(labelList.Names); err != nil {
+		fmt.Println("here")
```
cevian, line 132:
```suggestion

```

pkg/pgmodel/ingestor/handler.go
```diff
@@ -124,14 +127,25 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
...
 	//labels have to be created before series are since we need a canonical
 	//ordering for label creation to avoid deadlocks. Otherwise, if we create
 	//the labels for multiple series in same txn as we are creating the series,
 	//the ordering of label creation can only be canonical within a series and
 	//not across series.
-	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, labelList, labelMap)
+	dbEpoch, maxPos, err := h.fillLabelIDs(metricName, safeListName, safeListValue, labelMap)
 	if err != nil {
+		fmt.Println("from here")
+
```
cevian, line 148:
```suggestion
```

pkg/pgmodel/ingestor/handler.go
```diff
@@ -214,13 +228,18 @@ func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelLi
 	defer rows.Close()
 
 	count := 0
+	labelName := new(pgsafetype.Text)
+	labelValue := new(pgsafetype.Text)
 	for rows.Next() {
-		key := labels.Label{}
+		labelName.Status, labelValue.Status = pgtype.Null, pgtype.Null
```
cevian, line 234: this should not be necessary

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,138 @@
...
+type Text struct {
+	pgtype.Text
+}
+
+func (t *Text) Set(src interface{}) error {
+	text, ok := src.(string)
+	if !ok {
+		return fmt.Errorf("pgsafetype.Text.Set(): src is not of 'string' type")
+	}
+	fmt.Println("text is", text)
```
cevian, line 25:
```suggestion

```

pkg/pgmodel/ingestor/handler.go
```diff
@@ -214,13 +228,18 @@ func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelLi
...
 		res := labelInfo{}
-		err := rows.Scan(&res.Pos, &res.labelID, &key.Name, &key.Value)
+		err := rows.Scan(&res.Pos, &res.labelID, &labelName.String, &labelValue.String)
 		if err != nil {
 			return dbEpoch, 0, fmt.Errorf("Error filling labels in scan: %w", err)
 		}
+		labelName.Status, labelValue.Status = pgtype.Present, pgtype.Present
```
cevian, line 240: This should not happen here, but somewhere inside pgsafetype.Text

pkg/pgmodel/ingestor/ingestor_sql_test.go
```diff
@@ -55,8 +56,8 @@ func TestPGXInserterInsertSeries(t *testing.T) {
 					Sql: "SELECT * FROM _prom_catalog.get_or_create_label_ids($1, $2, $3)",
 					Args: []interface{}{
 						"metric_1",
-						[]string{"__name__", "name_1"},
-						[]string{"metric_1", "value_1"},
+						&pgsafetype.TextArray{TextArray: pgtype.TextArray{Elements: []pgtype.Text{pgtype.Text{String: "__name__", Status: 0x2}, pgtype.Text{String: "name_1", Status: 0x2}}, Dimensions: []pgtype.ArrayDimension{pgtype.ArrayDimension{Length: 2, LowerBound: 1}}, Status: 0x2}},
```
cevian, line 59: can't you make this conversion from []string. []string to pgsafetype.TextArray in the test code below? This is much less readable.

cevian (changes requested):
Partial review

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,150 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package pgsafetype
```
cevian, line 5: change package name to pgutf8str

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,150 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
cevian, line 1: change file name to text_types.go?

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("safe text: %w", err)
+	}
+	return nil
+}
+
+func (t *Text) Get() interface{} {
+	// Status is set to present in pgtype.Text when using pgtype.Text.Set(). However, when using a Scan(),
+	// this does not happen. In order to make this work, we need to explicitly tell here that data is present, otherwise
+	// data is not returned, rather a status is returned in pgtype.Text.Get().
+	t.Text.Status = pgtype.Present
```
cevian, line 46: wait, this should absolutely happen in Scan. Is this a mock problem again? Right now this would return wrong result if ever scanning a NULL value.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	arr := val.(pgtype.TextArray)
+	for i := range arr.Elements {
+		tmp, ok := arr.Elements[i].Get().(string)
+		if !ok {
+			panic("'string' type not received from underlying 'pgtype.Text' in 'pgsafetype.TextArray'")
+		}
+		arr.Elements[i].String = revertSanitization(tmp)
+	}
+	t.Elements = arr.Elements
```
cevian, line 106: can't say I love this. if t.Elements is always sanitized that makes the logic easier to follow. Right now it can be sanitizes or unsanitized. This would lead to weird errors if we ever called TextArray.Get() and then sent the textarray as a parameter. Can we have this function return a new []string? This would prevent confusion []string is always real utf-8. pgutf8str.TextArrray is always sanitized

cevian (changes requested):

pkg/pgmodel/ingestor/handler.go
```diff
@@ -124,13 +124,21 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 	if len(labelMap) == 0 {
 		return nil
 	}
+	safeListName := new(pgsafetype.TextArray)
```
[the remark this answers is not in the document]
Harkishen-Singh: We can, but that will be too low level in the pgtype.TextArray. I will update with the implementation and then we can judge it here.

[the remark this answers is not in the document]
Harkishen-Singh: For some reason, playing manually with low-level Elements and appending in it, has a different behaviour than setting with Set() which is happening here.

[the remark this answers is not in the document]
Harkishen-Singh: Ah, we sanitize only in Set(). Making LabelsList have textarray will mean using an append for Elements, which means it might not look/work good.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: Sorry, I cannot understand completely. You will only use a Get(), if the value in the type is already contained in the type, either by Set() or Scan()

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
Harkishen-Singh: This has been removed in recent commits. Hence, resolving.

pkg/pgmodel/ingestor/handler.go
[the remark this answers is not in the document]
Harkishen-Singh: This is done in recent commits and looks much better.

pkg/pgmodel/model/pgsafetype/custom_pgtypes.go
cevian: two changes
1) do not modify internal state during Get()
2. return []string and not TextArray from Get()

Harkishen-Singh: Done, resolving.

pkg/pgmodel/ingestor/handler.go
```diff
@@ -116,11 +116,14 @@ func (h *insertHandler) setSeriesIds(seriesSamples []model.Samples) error {
 				_, ok = labelMap[key]
 				if !ok {
 					labelMap[key] = labelInfo{}
-					labelList.Add(names[i], values[i])
+					if err := labelList.Add(names[i], values[i]); err != nil {
+						panic(err)
```
cevian, line 120: should return err not panic

pkg/pgmodel/ingestor/handler.go
```diff
@@ -205,22 +208,27 @@ func (h *insertHandler) fillLabelIDs(metricName string, labelList *model.LabelLi
 
 	err = br.QueryRow().Scan(&dbEpoch)
 	if err != nil {
-		return dbEpoch, 0, fmt.Errorf("Error filling labels: %w", err)
+		return dbEpoch, 0, fmt.Errorf("Error filling labelsaaaaa: %w", err)
```
cevian, line 211:
```suggestion
		return dbEpoch, 0, fmt.Errorf("Error filling labels: %w", err)
```

pkg/pgmodel/model/label_list.go
```diff
@@ -4,24 +4,72 @@
...
+	}
+	if err := valueT.Set(value); err != nil {
+		return fmt.Errorf("setting pgtype.Text: %w", err)
+	}
+	ls.Names.Elements = append(ls.Names.Elements, nameT.Text)
+	ls.Values.Elements = append(ls.Values.Elements, valueT.Text)
+	return nil
+}
+
+func (ls *LabelList) UpdateArrayDimensions() {
```
cevian, line 60: I don't think this should be a public function. It's too easy to forget to call it. Instead, I would make LabelList.Names and LabelList.Values  private (names and values) and have a Get() function that returns (names pgutf8str.TextArray, values pgutf8str.TextArray). That function would call updateArrayDimensions() before returning. That way the user of LabelList does not have to worry about this.

pkg/pgmodel/model/pgutf8str/text_types.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+const (
+	NullChar         = '\x00'
+	NullCharSanitize = '\ufffe'
+)
+
+// Text is a custom pgtype that wraps the pgtype.Text. It is safe from null characters.
+type Text struct {
+	pgtype.Text
```
cevian, line 21:
```suggestion
	pgtype.Text //contains sanitized string
```

pkg/pgmodel/model/pgutf8str/text_types.go
```diff
@@ -0,0 +1,147 @@
...
+	panic("pgutf8str.Text.AssignTo(): not implemented")
+}
+
+func (t *Text) Scan(_ interface{}) error {
+	panic("pgutf8str.Text.Scan(): not implemented")
+}
+
+// TextArray is a custom pgtype that wraps the pgtype.TextArray. It is safe from null characters.
+type TextArray struct {
+	pgtype.TextArray
```
cevian, line 60:
```suggestion
	pgtype.TextArray //contains sanitized values
```

pkg/pgmodel/model/pgutf8str/text_types.go
```diff
@@ -0,0 +1,147 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package pgutf8str
```
cevian, line 5: optional: can we add a short documentation for the package describing which problem it solves.

cevian (changes requested):
Looking good. Some minor things requested

pkg/pgmodel/ingestor/handler.go
Harkishen-Singh: Ah, I missed in debugging

pkg/pgmodel/ingestor/handler.go
Harkishen-Singh: Debug thing lol

pkg/pgmodel/model/label_list.go
Harkishen-Singh: I think lengths should be set before using the type. Implementing length in Get() will mean, doing it post-execution, which might harm any dependency on total size in TextArray.

pkg/pgmodel/model/label_list.go
```diff
@@ -4,24 +4,78 @@
...
-	ls.Names[i], ls.Names[j] = ls.Names[j], ls.Names[i]
-	ls.Values[i], ls.Values[j] = ls.Values[j], ls.Values[i]
+	ls.names.Elements[i], ls.names.Elements[j] = ls.names.Elements[j], ls.names.Elements[i]
+	ls.values.Elements[i], ls.values.Elements[j] = ls.values.Elements[j], ls.values.Elements[i]
 }
 func (ls LabelList) Less(i, j int) bool {
-	return ls.Names[i] < ls.Names[j] || (ls.Names[i] == ls.Names[j] && ls.Values[i] < ls.Values[j])
+	return ls.names.Elements[i].String < ls.names.Elements[j].String ||
+		(ls.names.Elements[i].String == ls.names.Elements[j].String &&
+			ls.values.Elements[i].String < ls.values.Elements[j].String)
```
atanasovskib, line 80:
```suggestion
        elemI = ls.names.Elements[i]
        elemJ = ls.names.Elements[j[
	return elemI.String < elemJ.String ||
		(elemI.String == elemJ.String && ls.values.Elements[i].String < ls.values.Elements[j].String)
```

cevian (approved):
LGTM

PR #531, opened 2021-03-16
Fix: enable optional features in PromQL engine were not applied

Harkishen-Singh:

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -89,6 +89,7 @@ type Config struct {
 	Auth *Auth
 
 	// PromQL configuration.
+	EnableFeatures       string
 	EnabledFeaturesList  []string
 	MaxQueryTimeout      time.Duration
 	SubQueryStepInterval time.Duration // Default step interval value if the user has not provided.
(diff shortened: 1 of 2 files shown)


pkg/api/common.go
```diff
@@ -108,21 +109,24 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
...
-	}
 	fs.DurationVar(&cfg.MaxQueryTimeout, "promql-query-timeout", 2*time.Minute, "Maximum time a query may take before being aborted. This option sets both the default and maximum value of the 'timeout' parameter in "+
 		"'/api/v1/query.*' endpoints.")
 	fs.DurationVar(&cfg.SubQueryStepInterval, "promql-default-subquery-step-interval", 1*time.Minute, "Default step interval to be used for PromQL subquery evaluation. "+
 		"This value is used if the subquery does not specify the step value explicitly. Example: <metric_name>[30m:]. Note: in Prometheus this setting is set by the evaluation_interval option.")
 	return cfg
 }
 
+// ComputeFlags is required to be executed post parsing of flags in order to fill dependent configuration settings.
+func ComputeFlags(cfg *Config) {
```
cevian, line 122: This can just be part of Validate() below. It does not need to be a separate function. The convention I want to establish is that flag parsing is split into ParseFlags() and Validate() for all packages.

cevian (changes requested):
One minor-ish change requested

pkg/api/common.go
Harkishen-Singh: I agree on that. The compute is also misleading.

cevian (approved):

PR #532, opened 2021-03-16
Make PromQL engine max-samples and look-back delta customizable.

Harkishen-Singh:


cevian (approved):
Let's add the new  cli flags into draft release notes. Otherwise looks good. I am guessing the defaults here are same as in Prometheus?

Harkishen-Singh:
@cevian the default of `max samples` is not the same as Prometheus. The reason is Promscale is LTS and is expected to have higher limits than vanilla prometheus. Is that good to go?

Harkishen-Singh:
Discuss on meet

cc @cevian

PR #533, opened 2021-03-16
Add new docker-compose file to use without build dependency

VineethReddy02:
Remove `build: context: .` to drop dependency to build Promscale image on setting up Promscale. i.e. removed from Promscale service.

```
    build:
      context: .
```

Noticed Error:

```
$ docker-compose up
Building promscale
[+] Building 0.0s (1/2)                                                                                  
 => [internal] load build definition from Dockerfile                                                0.0s
 => => transferring dockerfile: 2B                                                                  0.0s
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount685150454/Dockerfile: no such file or directory
ERROR: Service 'promscale' failed to build
```

**Note**: This error is observed only when `timescale/promscale:latest` image is not available locally (in docker daemon). Then it tries to build the image in which it fails to find `Dockerfile`

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -1,4 +1,5 @@
 version: '3.0'
+
 services:
   db:
     image: timescaledev/timescaledb-ha:pg12-latest
@@ -19,8 +20,6 @@ services:
     image: timescale/promscale:latest
     ports:
       - 9201:9201/tcp
-    build:
-      context: .
     restart: on-failure
     depends_on:
       - db
--- a/docker-compose/high-availability/docker-compose.yaml
+++ b/docker-compose/high-availability/docker-compose.yaml
@@ -28,8 +28,6 @@ services:
     image: timescale/promscale:latest
     ports:
       - 9201:9201/tcp
-    build:
-      context: .
     restart: on-failure
     depends_on:
       - db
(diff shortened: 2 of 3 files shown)


cevian:
@VineethReddy02 I don't understand your point about the error?

VineethReddy02:
In the current docker-compose file. We mention building a Promscale image by providing `build:context: .` this needs Dockerfile to do so. If a user wants to run Promscale this is adding a dependency to clone promscale Github repo or to manually add Promscale Dockerfile. 

Adding the new docker-compose file in this PR removes such dependency. Now users can just use docker-compose file from this PR as the Promscale image build step has been removed.

cevian (changes requested):
I think this new file should be called docker-compose/docker-compose.yaml and the old one be renamed docker-compose/docker-compose-with-build.yaml and we should put comments in both files to update them in sync

VineethReddy02:
@cevian I am removing the `build: context: .` from `docker-compose.yaml` as in `test.sh` we are already explicitly building the Promscale image with cmd `docker build -t timescale/promscale:latest ../.. --file ../../build/Dockerfile` and tagging it to `promscale:latest`.

cevian (approved):

PR #534, opened 2021-03-17
Add error handling to env variable parsing

antekresic:
In the case where you set an environment variable with the wrong type
(for example, set a string where an int flag is defined), parsing errors
were ignored. This commit adds proper error handling for that case.


cevian (approved):
LGTM

PR #537, opened 2021-03-18
Add support for PG13

wilfriedroset:
As promscale support timescaledb extension up to 2.99.0 which is
compatible with PostgreSQL 13, let's make promscale compatible with it.

See: 94d8c0a

Signed-off-by: Wilfried Roset <wilfriedroset@users.noreply.github.com>


wilfriedroset:
I believe that github workflows can be improve like for example with both pg12 and pg13, with and without promscale extension. The image for the test with Promscale extension was not available for example.

Feel free to point me in the right direction 🙏  I will try to include your suggestions as I'm no expert at github workflow.

cevian:
thanks @wilfriedroset will take a look in a bit. Note to self: the promscale extension also needs an update for pg13

cevian:
@wilfriedroset had to add a bunch of changes to make the test work

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -60,16 +62,22 @@ func TestMain(m *testing.M) {
...
+				return "timescaledev/promscale-extension:0.1.1-ts2-pg12", "timescaledev/promscale-extension:latest-ts2-pg12"
+			} else {
+				return "timescaledev/promscale-extension:0.1.1-ts1-pg12", "timescaledev/promscale-extension:latest-ts1-pg12"
+			}
+		}
+		if !extensionState.UsesTimescaleDB() {
+			return "timescale/timescaledb:latest-pg12", "timescale/timescaledb:latest-pg12"
+		}
 	}
+	panic("Unexpected extension state in upgradeTest")
```
antekresic, line 80: Seems like we could panic first here, would make this condition tree a bit more easier to read.

Else is not needed since we use return statements and I would prefer a switch statement honestly. 
```suggestion
	if !extensionState.UsesPG12() {
		panic("Unexpected extension state in upgradeTest")
	}
	switch {
	case extensionState.UsesMultinode():
		return "timescaledev/promscale-extension:0.1.1-ts2-pg12", "timescaledev/promscale-extension:latest-ts2-pg12"
	case !extensionState.UsesTimescaleDB():
		return "timescale/timescaledb:latest-pg12", "timescale/timescaledb:latest-pg12"
 	case extensionState.UsesTimescale2() :
		return "timescaledev/promscale-extension:0.1.1-ts2-pg12", "timescaledev/promscale-extension:latest-ts2-pg12"
	default:
		return "timescaledev/promscale-extension:0.1.1-ts1-pg12", "timescaledev/promscale-extension:latest-ts1-pg12"
	}
```

antekresic (approved):
Looks good with a small refactor suggestion to simplify a conditional tree.

go.mod
```diff
@@ -37,11 +38,9 @@ require (
...
 	github.com/stretchr/testify v1.7.0
-	github.com/testcontainers/testcontainers-go v0.5.1
+	github.com/testcontainers/testcontainers-go v0.10.0
 	github.com/uber/jaeger-client-go v2.25.0+incompatible
 	go.uber.org/atomic v1.7.0
 	go.uber.org/goleak v1.1.10
 	gotest.tools v2.2.0+incompatible // indirect
-)
-
-replace github.com/docker/docker => github.com/docker/engine v0.0.0-20190717161051-705d9623b7c1
```
Harkishen-Singh, line 47: [tests] TODO to self: Are we good without this?

pkg/internal/testhelpers/containers.go
```diff
@@ -71,37 +72,28 @@ func (e *ExtensionState) UseMultinode() {
...
-		return true
-	case MultinodeAndPromscale:
-		return true
-	default:
-		return false
-	}
+	return (e & promscaleBit) != 0
+}
+
+func (e ExtensionState) UsesPG12() bool {
```
Harkishen-Singh, line 95: [nit] can we have this near to `UsePG12()`?

pkg/internal/testhelpers/containers.go
```diff
@@ -646,17 +646,17 @@ func StartConnectorWithImage(ctx context.Context, image string, printLogs bool,
...
-			"-db-host", "172.17.0.1", // IP refering to the docker's host network
+			"-db-host", "172.17.0.1",
 			"-db-port", pgPort.Port(),
 			"-db-user", dbUser,
 			"-db-password", "password",
 			"-db-name", dbname,
 			"-db-ssl-mode", "prefer",
-			"-web-listen-address", "0.0.0.0:" + ConnectorPort.Port(),
 		},
+		//Networks: []string{"promscale-network"},
```
Harkishen-Singh, line 659: debug comment

pkg/internal/testhelpers/containers.go
```diff
@@ -691,10 +691,16 @@ func StartConnectorWithImage(ctx context.Context, image string, printLogs bool,
 
 func StopContainer(ctx context.Context, container testcontainers.Container, printLogs bool) {
 	if printLogs {
-		_ = container.StopLogProducer()
+		err := container.StopLogProducer()
+		if err != nil {
+			fmt.Println("couldn't stop log producer", err)
```
Harkishen-Singh, line 696: [optional] Here and below: should we use `fmt.Fprintln(os.Stderr, ...)`?

pkg/pgmodel/common/extension/extension.go
```diff
@@ -255,6 +255,7 @@ func MigrateExtension(conn *pgx.Conn, extName string, extSchemaName string, vali
 			}
 			log.Error("msg", fmt.Sprintf("Failed to migrate extension %v from %v to %v: %v", extName, currentVersion, newVersion, err))
 		}
+		log.Info("msg", "successfully updated "+extName+" extension to version "+newVersion.String())
```
Harkishen-Singh, line 258: [nit]
```suggestion
		log.Info("msg", fmt.Sprintf("successfully updated %s extension from %s to %s", extName, currentVersion.String(), newVersion.String())
```

Harkishen-Singh (approved):
Mostly nits and few questions. I don't think any review suggestions were blocking, so approving.

Harkishen-Singh:
I think we should clean the commit history before merging.

.github/workflows/go-scheduled.yml
```diff
@@ -100,11 +100,11 @@ jobs:
...
-        - {name: "W/O Promscale Extension", ext: false, tsdb: true,  tsdb2: false, multi: false}
-        - {name: "W/O TimescaleDB Extension", ext: false, tsdb: false, tsdb2: false, multi: false}
-        - {name: "on TimescaleDB 2.0", ext: true,  tsdb: true,  tsdb2: true,  multi: false}
-        - {name: "on TimescaleDB 2.0 W/O Promscale Extension", ext: false, tsdb: true,  tsdb2: true,  multi: false}
-        - {name: "on TimescaleDB 2.0 multinode", ext: true,  tsdb: true,  tsdb2: false, multi: true}
+        - {name: "Multinode",               ext: true,  tsdb: true,  tsdb2: true,  multi: true,  pg: 13}
+        - {name: "W/O Promscale Extension", ext: false, tsdb: true,  tsdb2: true,  multi: false, pg: 13}
+        - {name: "Plain Postgres",          ext: false, tsdb: false, tsdb2: false, multi: false, pg: 13}
+        - {name: "Timescaledb 1.x (pg 12)", ext: true,  tsdb: true,  tsdb2: false, multi: false, pg: 12}
+        - {name: "Plain Postgres (12)",     ext: false, tsdb: false, tsdb2: false, multi: false, pg: 12}
```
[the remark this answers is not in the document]
cevian: Our priority should be to test on latest with a subset of things tested on older versions (but not all). If you want specific older combinations let me know

pkg/internal/testhelpers/containers.go
cevian: That's not the pattern in the file, and redoiing everything would be a larger change than I want right now

PR #544, opened 2021-03-23
Update Promscale helmchart

VineethReddy02:
Disable maintenance cronjob by default as we expect all users to use >= TimescaleDB `2.0` and Also added an option to configure podAffinity & podAntiAffinity rules for Promscale deployment & Maintenance cronjob.


cevian (approved):
Let's put a note in the release notes about users still on 1.x needing to turn maintenance jobs on if upgrading helm.

Harkishen-Singh (approved):

antekresic (approved):

Harkishen-Singh:
@VineethReddy02 just to bring to attention in case we forget, I think commits with merge should be merged. They should be rebased if I am not wrong.

VineethReddy02:
Yup, Rebasing the PR.

PR #545, opened 2021-03-23
Add text format support for ingest pipeline

antekresic:
Common usecase necessary for users is to push metrics directly to
Promscale using Prometheus text format or OpenMetrics text format. This
commit uses Prometheus `textparse` library to add support for these formats
to be able to ingest them directly into Promscale.

### DIFF
--- a/pkg/pgmodel/ingestor/ingestor.go
+++ b/pkg/pgmodel/ingestor/ingestor.go
@@ -6,7 +6,13 @@ package ingestor
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
+	"time"
 
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/textparse"
+	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/timescale/promscale/pkg/clockcache"
 	"github.com/timescale/promscale/pkg/pgmodel/cache"
 	"github.com/timescale/promscale/pkg/pgmodel/common/errors"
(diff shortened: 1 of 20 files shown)


cevian (changes requested):
Looks good overall. some small changes requested

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -46,13 +52,95 @@ func NewPgxIngestor(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+		// Cannot really happen since text parser would throw the error before we ever get here.
+		if metricName == "" {
+			return nil, 0, errors.ErrNoMetricName
+		}
+		sample := model.NewPromSample(seriesLabels, []prompb.Sample{
+			{
+				Timestamp: t,
+				Value:     v,
+			},
+		})
```
Harkishen-Singh, line 129: What happens if the user sends a text exposition format, that contains more than one representation of a series? (I assume duplicate seriesLabel entries in dataSamples map, with different samples slice)

should we keep a map[*model.promSample][]prompb.Sample and after this loop, insert into the map of dataSamples? This will then behave more like what NewPromSamples was designed for.

pkg/pgmodel/ingestor/ingestor_interface.go
```diff
@@ -4,12 +4,22 @@
...
+	"io"
+	"time"
+
+	"github.com/timescale/promscale/pkg/prompb"
+)
 
 // DBInserter is responsible for ingesting the TimeSeries protobuf structs and
 // storing them in the database.
 type DBInserter interface {
 	// Ingest takes an array of TimeSeries and attepts to store it into the database.
```
Harkishen-Singh, line 17: Ingest => IngestProto. Also, I think the comment needs an update.

pkg/pgmodel/ingestor/ingestor_interface.go
```diff
@@ -4,12 +4,22 @@
...
 	// Returns the number of metrics ingested and any error encountered before finishing.
-	Ingest([]prompb.TimeSeries, *prompb.WriteRequest) (uint64, error)
+	IngestProto([]prompb.TimeSeries, *prompb.WriteRequest) (uint64, error)
+	// Ingest takes a Prometheus exporter text format and attepts to parse and store it into the database.
```
Harkishen-Singh, line 20: Ingest => IngestText

Harkishen-Singh (changes requested):
I think, we should also add a doc or update the [writing_to_promscale](https://github.com/timescale/promscale/blob/master/docs/writing_to_promscale.md) in this PR itself, that explains the new text-based ingestion feature.

Issue #472, opened 2021-02-02
Expose prom-migrator version


VineethReddy02:
While using prom-migrator. I do not see any version in prom-migrator. 

Exposing the current prom-migrator version would be great to understand issues faced by the users in a specific version.   

Suggestions:
1. Log version on migration startup. 
2. Also, add a version flag or CMD i.e `./prom-migrator --version` or `./prom-migrator version` which exposes the version of prom-migrator.

niconosenzo:
Hi @VineethReddy02 , I'm wondering if it should match the `promscale` cli pattern of printing out the version when no argument is passed, or it should implement the `--version` flag.

PR #546, opened 2021-03-24
Add version flag to prom-migrator command

niconosenzo:
Closes issue #472 .
Added `-version` flag to print out the promscale current version. Also it will log the version as an INFO message at the very first line.

```bash
$ ./prom-migrator -version
Version:  0.1.4
$ ./prom-migrator
level=info ts=2021-03-24T11:22:52.902Z caller=main.go:60 Prom-migratorVersion=0.1.4
level=error ts=2021-03-24T11:22:52.902Z caller=main.go:63 msg="could not parse flags" error="mint should be provided for the migration to begin"
$ 
```

### DIFF
--- a/cmd/prom-migrator/main.go
+++ b/cmd/prom-migrator/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/timescale/promscale/pkg/migration-tool/reader"
 	"github.com/timescale/promscale/pkg/migration-tool/utils"
 	"github.com/timescale/promscale/pkg/migration-tool/writer"
+	"github.com/timescale/promscale/pkg/runner"
 	"github.com/timescale/promscale/pkg/version"
 )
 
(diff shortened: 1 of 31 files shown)


cmd/prom-migrator/main.go
```diff
@@ -40,17 +41,23 @@ type config struct {
...
 	writerAuth         utils.Auth
 }
 
 func main() {
 	conf := new(config)
 	parseFlags(conf, os.Args[1:])
+	if conf.versionPrint {
+		fmt.Println("Version: ", version.Version)
+		os.Exit(0)
+	}
```
Harkishen-Singh, line 55: Can you please make this like: https://github.com/timescale/promscale/blob/5de3a5ce6c634508bc524e32e9cad3f7470d586d/pkg/runner/args.go#L14
That will enable to easily expand the args in future.

cmd/prom-migrator/main.go
```diff
@@ -40,17 +41,23 @@ type config struct {
...
 	if err := log.Init(log.Config{Format: "logfmt", Level: "debug"}); err != nil {
 		_, _ = fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+	log.Info("Prom-migrator Version", version.Version)
```
Harkishen-Singh, line 60:
```suggestion
	log.Info("Version:", version.Version)
```

cmd/prom-migrator/main.go
```diff
@@ -154,6 +161,7 @@ func parseFlags(conf *config, args []string) {
 	flag.StringVar(&conf.writerAuth.Password, "write-auth-password", "", "Auth password for remote-write storage.")
 	flag.StringVar(&conf.writerAuth.BearerToken, "write-auth-bearer-token", "", "Bearer token for remote-write storage. "+
 		"This should be mutually exclusive with username and password.")
+	flag.BoolVar(&conf.versionPrint, "version", false, "While passing this flag to the cli, the prom-migration version will be printed at first place.")
```
Harkishen-Singh, line 164: The reason why this should be treated as
https://github.com/timescale/promscale/blob/5de3a5ce6c634508bc524e32e9cad3f7470d586d/pkg/runner/args.go#L14
is because flags like `-version`, `-help` do not require any input, other than their actual presence. I agree this will set itself to true if used, but IMHO (not sure what others think), flags that do not have any input needs, should be a simple match. Plus, these flags are self-explanatory, so this saves from an explanation line as well.

Harkishen-Singh (changes requested):
Thank you very much for PR. Just small changes and then, we are good to go!

cmd/prom-migrator/main.go
niconosenzo: Roger that. I just changed it to match promscale CLI.

cmd/prom-migrator/main.go
niconosenzo: Got it, thanks for the reasoning.

Harkishen-Singh:
@niconosenzo thank you for updating that. Can you please pull from master and rebase the PR so that I can merge?

PR #548, opened 2021-03-26
Fix failing PromQL tests, fail on error logs and support for negative-offset in enable-features.

Harkishen-Singh:
This does 3 tasks:
1. Fix PromQL E2E tests that were showing error logs.
2. Make CI fail if PromQL tests actually fail in logs (other than places not intended, eg: real-dataset tests where we want the output to be in line with Prometheus).
3. Support `promql-negative-offset` as the list of enable-features.


cevian (approved):
LGTM

PR #550, opened 2021-03-29
Adds support for multi-tenancy in Promscale

Harkishen-Singh:
- [x] simple multi-tenancy (no token)
- [x] bearer_token multi-tenancy
- [x] Testing with Prometheus (plain and bearer_token)
- [x] Tests
- [x] TODO: update docs/cli.md
- [x] TODO: Final test before merging.


pkg/api/router.go
```diff
@@ -105,10 +105,18 @@ func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
...
+			errLogs := func(token string) {
+				log.Error("msg", "Unauthorized access to endpoint, invalid bearer token: "+token)
+				http.Error(w, "Unauthorized access to endpoint, invalid bearer token.", http.StatusUnauthorized)
+			}
 			splitToken := strings.Split(r.Header.Get("Authorization"), "Bearer ")
+			if len(splitToken) < 2 {
+				// todo: add test for this.
+				errLogs("")
+				return
+			}
```
Harkishen-Singh, line 117: This is no longer part of this PR. Once the conflicts are resolved, this would be done.

pkg/api/common.go
```diff
@@ -127,6 +129,16 @@ func Validate(cfg *Config) error {
 	return cfg.Auth.Validate()
 }
 
+func getTenant(r *http.Request) (tenant string) {
+	if tn := r.Header.Get("TENANT"); tn == "" {
```
cevian, line 133: for documentation there is no X prefix because of https://datatracker.ietf.org/doc/html/rfc6648

pkg/multi-tenancy/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+
+package config
+
+// TenantLabelKey is a label key reserved for multi-tenancy.
+const TenantLabelKey = "__tenant__"
+
+// Config defines the configuration for multi-tenancy.
+type Config struct {
+	ValidTenants []string
+	tenantsCache map[string]struct{}
```
cevian, line 13: not really a cache 
```suggestion
	validTenantsMap map[string]struct{}
```

pkg/multi-tenancy/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// FillTenantCache creates a cache of valid tenants for multi-tenancy.
+func (cfg *Config) FillTenantCache() {
```
cevian, line 17: instead of this function I suggest a NewConfig(tenantList []string) function that always create a valid Config object. That way you can't have an invalid object without the map filled in being made.

pkg/multi-tenancy/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+// IsTenantAllowed returns true if the given tenantName is allowed to be ingested.
+func (cfg *Config) IsTenantAllowed(tenantName string) bool {
+	if tenantName == "" {
+		// Do not allow empty tenant names.
+		return false
+	}
+	if len(cfg.tenantsCache) == 0 {
```
cevian, line 30: I don't get why this returns true

cevian: if we want an allowed all mode, I'd suggest a separate bool in the config object. That way there can be no confusion

cevian: Or better yet make this a different struct altogether that also implements `IsTenantAllowed`

pkg/multi-tenancy/read/interface.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Authorizer tells if a read request is allowed to query via Promscale.
+type Authorizer interface {
+	// ApplySafetyMatcher applies a safety matcher to incoming query matchers. This safety matcher is responsible
+	// from prevent unauthorized query reads from tenants that the incoming query is not supposed to read.
+	ApplySafetyMatcher(ms []*labels.Matcher) []*labels.Matcher
+}
+
+// initSafetyLabelMatcher creates a new safety label matcher, from the given list of valid tenants.
+func initSafetyLabelMatcher(validTenants []string) (*labels.Matcher, error) {
```
cevian, line 32:
```suggestion
func getSafetyLabelMatcher(validTenants []string) (*labels.Matcher, error) {
```

cevian: this function probably doesn't belong in a file called interface.go

pkg/multi-tenancy/read/read.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/timescale/promscale/pkg/multi-tenancy/config"
+)
+
+type readAuthorizer struct {
+	authorizerConfig
```
cevian, line 15: not sure I understand the difference between a readAuthorizer and an authorizeConfig - why not just make this one struct

pkg/multi-tenancy/read/interface.go
cevian, line 32: again really hate this convention of empty validTenants meaning "all"

pkg/runner/flags.go
```diff
@@ -62,6 +66,12 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
 	fs.StringVar(&cfg.TLSCertFile, "tls-cert-file", "", "TLS Certificate file for web server, leave blank to disable TLS.")
 	fs.StringVar(&cfg.TLSKeyFile, "tls-key-file", "", "TLS Key file for web server, leave blank to disable TLS.")
 
+	// Multi-tenancy flags.
+	// TODO(Harkishen): update the docs in docs/cli.md
+	fs.BoolVar(&cfg.EnableMultiTenancy, "multi-tenancy", false, "Use multi-tenancy mode in Promscale.")
```
cevian, line 71: I would have a separate bool for 
fs.BoolVar(&cfg.EnableDataWithoutTenant, "multi-tenancy-no-tenant-name", false, "If multi-tenancy is on, allow data with no tenant")

pkg/runner/flags.go
```diff
@@ -130,3 +140,33 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
...
+	if err := validateMultiTenancyFlags(cfg); err != nil {
+		return fmt.Errorf("error validating mlti-tenancy configuration: %w", err)
+	}
+	return nil
+}
+
+func validateMultiTenancyFlags(cfg *Config) error {
+	if cfg.ValidTenantsListStr != "" {
+		cfg.ValidTenantsList = strings.Split(cfg.ValidTenantsListStr, ",")
+		cfg.ValidTenantsList = removeEmptyTenants(cfg.ValidTenantsList)
```
cevian, line 157: I think cfg.ValidTenantList should already be a type that is explicit between "allow all" and empty list

pkg/multi-tenancy/write/write.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantName string, labels []prompb.Label) ([]prompb.Label, bool) {
+	for _, label := range labels {
+		if label.Name == config.TenantLabelKey {
+			return labels, false
+		}
+	}
+	if tenantName == "" {
+		// Non-multi-tenant write request.
```
cevian, line 33: as I mentioned in the flags section, whether or not to accept data without a tenant name should be a configuration option

pkg/multi-tenancy/write/interface.go
```diff
@@ -0,0 +1,24 @@
...
+
+var ErrInvalidLabels = fmt.Errorf("invalid labels-set: labels-set should not contain __tenant__ label")
+
+// Authorizer tells if a write request is authorized to be written.
+type Authorizer interface {
+	// IsAuthorized verifies if the token and tenant to be inserted are both authorized.
+	// Note: Even though in start of Promscale, the bearer_token in incoming write-request is verified by the authHandler,
+	// we do the token verification again in order to avoid any edge cases and maintain a clean design.
+	IsAuthorized(tenantName string) bool
+	// VerifyAndApplyTenantLabel verifies if the incoming write-request does not contains a __tenant__ label pair.
```
cevian, line 20:
```suggestion
	// VerifyAndApplyTenantLabel verifies that the __tenant__ label in the samples aligns with the TENANT specified by the header in the request.
```

pkg/multi-tenancy/write/write.go
```diff
@@ -0,0 +1,39 @@
...
+	return &writeAuthorizer{config}
+}
+
+func (a *writeAuthorizer) IsAuthorized(tenantName string) bool {
+	return a.IsTenantAllowed(tenantName)
+}
+
+func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantName string, labels []prompb.Label) ([]prompb.Label, bool) {
+	for _, label := range labels {
+		if label.Name == config.TenantLabelKey {
```
cevian, line 28: if the value is the right tenant name this should be fine
```suggestion
		if label.Name == config.TenantLabelKey && label.Value != tenantName {
```

pkg/pgclient/config.go
```diff
@@ -38,6 +38,13 @@ type Config struct {
 	MaxConnections          int
 	UsesHA                  bool
 	DbUri                   string
+	MT                      MTClientConfig
+}
+
+type MTClientConfig struct {
+	ValidTenants []string
+	TenancyType  string
```
cevian, line 46: This is never used

pkg/pgclient/config.go
```diff
@@ -38,6 +38,13 @@ type Config struct {
...
+	Token        string
```
cevian, line 47: Is this ever used?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -25,35 +25,40 @@ type Cfg struct {
...
 	pi, err := newPgxInserter(conn, cache, scache, cfg)
 	if err != nil {
 		return nil, err
 	}
 
-	return &DBIngestor{db: pi, scache: scache}, nil
+	return &DBIngestor{
+		db:           pi,
+		scache:       scache,
+		multiTenancy: mtAuthr,
```
cevian, line 44:
```suggestion
		mtAuth: mtAuth,
```

pkg/pgmodel/querier/querier.go
```diff
@@ -44,11 +44,12 @@ const (
 
 // NewQuerier returns a new pgxQuerier that reads from PostgreSQL using PGX
 // and caches metric table names and label sets using the supplied caches.
-func NewQuerier(conn pgxconn.PgxConn, metricCache cache.MetricCache, labelsReader lreader.LabelsReader) Querier {
+func NewQuerier(conn pgxconn.PgxConn, metricCache cache.MetricCache, labelsReader lreader.LabelsReader, mtAuthr multi_tenancy_read.Authorizer) Querier {
 	return &pgxQuerier{
 		conn:             conn,
 		labelsReader:     labelsReader,
 		metricTableNames: metricCache,
+		multiTenancy:     mtAuthr,
```
cevian, line 52:
```suggestion
		mtAuth:     mtAuth,
```

pkg/runner/client.go
```diff
@@ -85,7 +87,7 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
 			if err != nil {
 				return nil, err
 			}
-			log.Info("msg", "Migration successful, exiting")
+			log.Info("msg", "Migrationvalid successful, exiting")
```
cevian, line 90:
```suggestion
			log.Info("msg", "Migration successful, exiting")
```

pkg/runner/flags.go
```diff
@@ -35,6 +35,10 @@ type Config struct {
 	InstallExtensions           bool
 	UpgradeExtensions           bool
 	UpgradePrereleaseExtensions bool
+	// Multi-tenancy.
+	EnableMultiTenancy  bool
+	ValidTenantsListStr string
```
cevian, line 40: The string should never be used by anything
```suggestion
	validTenantsListStr string
```

cevian (changes requested):
Right direction. Two high level points:
1. We should be explicit on the object/struct level between "all tenants valid" and "List of valid tenants are empty" I wouldn't use the "empty list = all " convention anywhere in the code past the cli flags section. This will make the code more explicit and robust

2. I think whether or not we accept data without a tenant name and whether or not we allow reads on data without a tenant name should be a CLI option.

pkg/api/router.go
```diff
@@ -105,10 +105,18 @@ func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
 
 	if cfg.Auth.BearerToken != "" {
 		return func(w http.ResponseWriter, r *http.Request) {
+			errLogs := func(token string) {
+				log.Error("msg", "Unauthorized access to endpoint, invalid bearer token: "+token)
```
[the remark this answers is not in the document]
Harkishen-Singh: these are fixed after rebasing with master.

pkg/api/write.go
```diff
@@ -53,6 +54,16 @@ func Write(writer ingestor.DBInserter, elector *util.Elector, metrics *Metrics)
 
 		metrics.LeaderGauge.Set(1)
 
+		var tenant, tenantToken string
+		if authr := apiConf.MultiTenancy.WriteAuthorizer(); authr != nil {
+			tenant, tenantToken = getTenantAndToken(r)
+			if !authr.IsAuthorized(tenantToken, tenant) {
+				log.Error("msg", multi_tenancy.ErrUnauthorized.Error()+tenantToken)
+				http.Error(w, multi_tenancy.ErrUnauthorized.Error()+tenantToken, http.StatusUnauthorized)
```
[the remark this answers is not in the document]
Harkishen-Singh: fixed after rebasing with master

pkg/api/common.go
Harkishen-Singh: Wrote as a comment.

pkg/multi-tenancy/config/config.go
Harkishen-Singh: Yes, I was thinking of the same as part of optimization. Plus, the caller does not have to remember to call FillTenantCache() as it gets handled automatically.

pkg/multi-tenancy/config/config.go
Harkishen-Singh: 3rd comment is not clear. But going with bool as its easier to read.

Harkishen-Singh: The better design would be to make an interface that is implemented by a config that allows all tenants and a config that is selective on tenants.

pkg/runner/flags.go
Harkishen-Singh: that will be confusing for user. They will be confused whether to apply true to both `multi-tenancy` and `multi-tenancy-no-tenant-name` or just the later. So I think we should have a bool saying `skip-tenant-validaton` whose purpose is self-explanatory.

Harkishen-Singh: (Note to myself) TODO

pkg/runner/flags.go
Harkishen-Singh: here is the purpose is to catch `tenant-a, tenant-b, ` (note the trailing `,`)

pkg/multi-tenancy/write/write.go
Harkishen-Singh: yup

pkg/multi-tenancy/write/interface.go
Harkishen-Singh: Note: This means we are giving flexibility to the users so they can either use headers to indicate a tenant or external labels.

pkg/multi-tenancy/write/write.go
Harkishen-Singh: This wasn't as per plans. we just wanted headers to be responsible for tenant name indication and not external labels. But seeing the comments, we are changing to accommodate both.

pkg/multi-tenancy/write/write.go
Harkishen-Singh: (note to myself) TODO

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: reason of such nomenclature:

`auth` => `authentication`
`authr` `authorizer`

pkg/pgmodel/querier/querier.go
Harkishen-Singh: made `mtAuthr`

pkg/multi-tenancy/config/config.go
Harkishen-Singh: done

pkg/multi-tenancy/read/interface.go
Harkishen-Singh: thats no more the case now

pkg/runner/flags.go
Harkishen-Singh: I think it's better to follow https://github.com/timescale/promscale/pull/550#discussion_r612686267 for this case.

pkg/runner/runner.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	promLivenessCheck = time.Second
 	schemaLockId      = 0x4D829C732AAFCEDE // chosen randomly.
+	allowAllTenants   = "allow-all"
```
cevian, line 24: this should be defined in. flags.go, since that's where it is used

pkg/multi-tenancy/read/read.go
```diff
@@ -0,0 +1,57 @@
...
+	config.Config
+	// mtSafetyLabelPair is a label-pair that is applied to incoming multi-tenant read requests for security reasons.
+	// This matcher helps prevent a query from querying a tenant for the query has not been authorized.
+	mtSafetyLabelMatcher *labels.Matcher
+}
+
+// NewPlainReadAuthorizer is a authorizer that does not care about the token checks. It just makes sure that the
+// read operations are performed against a valid tenants, which Promscale has been asked for.
+func NewAuthorizer(cfg config.Config) (Authorizer, error) {
+	if _, ok := cfg.(*config.OpenConfig); ok {
```
cevian, line 25: maybe the config interface should have an AllowAllTenants() function instead?

pkg/multi-tenancy/read/read.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("create plain read-authorizer: %w", err)
+	}
+	return &readAuthorizer{
+		Config:               cfg,
+		mtSafetyLabelMatcher: matcher,
+	}, nil
+}
+
+func (a *readAuthorizer) ApplySafetyMatcher(ms []*labels.Matcher) []*labels.Matcher {
```
cevian, line 41: better name?
```suggestion
func (a *readAuthorizer) ApplyTenantMatcher(ms []*labels.Matcher) []*labels.Matcher {
```

pkg/multi-tenancy/read/read.go
```diff
@@ -0,0 +1,57 @@
...
+	if a.mtSafetyLabelMatcher == nil {
+		return ms
+	}
+	ms = append(ms, a.mtSafetyLabelMatcher)
+	return ms
+}
+
+// getSafetyLabelMatcher creates a new safety label matcher, from the given list of valid tenants.
+func getSafetyLabelMatcher(validTenants []string) (*labels.Matcher, error) {
```
cevian, line 50: How does this work with regard to unset tenants? Shouldn't this respect AllowNonMT flag?

pkg/multi-tenancy/write/write.go
cevian: still needs to be done

pkg/multi-tenancy/write/write.go
```diff
@@ -0,0 +1,39 @@
...
+		if label.Name == config.TenantLabelKey && label.Value != tenantName {
+			return labels, false
+		}
+	}
+	if tenantName == "" {
+		// Non-multi-tenant write request.
+		// We accept non-multi-tenant write requests in multi-tenancy mode.
+		return labels, true
+	}
+	labels = append(labels, prompb.Label{Name: config.TenantLabelKey, Value: tenantName})
```
cevian, line 37: This will add a label even if one already is set. That should be fixed, append should happen only if doesn't exist

pkg/multi-tenancy/write/write_test.go
```diff
@@ -0,0 +1,66 @@
...
+func containsAppliedTenantLabel(lbls []prompb.Label) bool {
+	for _, lbl := range lbls {
+		if lbl.Name == config.TenantLabelKey {
+			return true
+		}
+	}
+	return false
+}
+
+func getlbls() [][]prompb.Label {
```
cevian, line 52: should add case with tenant label already set. Also should test case where it's set to the wrong tenant

pkg/pgclient/client.go
```diff
@@ -230,7 +236,7 @@ func (c *Client) HealthCheck() error {
 	return c.healthCheck()
 }
 
-// Queryable returns the Prometheus promql.Queryable interface that's running
+// Queryable returns the Prometheus promql.Queryable samples-parser that's running
```
cevian, line 239: bad change

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -47,7 +47,7 @@ type labelsReader struct {
 	labels cache.LabelsCache
 }
 
-// LabelValues implements the LabelsReader interface. It returns all distinct values
+// LabelValues implements the LabelsReader samples-parser. It returns all distinct values
```
cevian, line 50: bad change

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -72,7 +72,7 @@ func (lr *labelsReader) LabelValues(labelName string) ([]string, error) {
 	return labelValues, nil
 }
 
-// LabelNames implements the LabelReader interface. It returns all distinct
+// LabelNames implements the LabelReader samples-parser. It returns all distinct
```
cevian, line 75: bad change

pkg/pgmodel/ingestor/samples-parser/parser.go
```diff
@@ -2,30 +2,35 @@
 // Please see the included NOTICE for copyright information and
 // LICENSE for a copy of the license.
 
-package ingestor
+package samples_parser
```
cevian, line 5: I don't agree wiil this change. Also note this PR will have to go in after #571 so this has to be re-thought

pkg/pgmodel/model/sql_test_utils.go
```diff
@@ -246,7 +246,7 @@ func (m *MockRows) Next() bool {
 
 // Scan reads the values from the current row into dest values positionally.
 // dest can include pointers to core types, values implementing the Scanner
-// interface, []byte, and nil. []byte will skip the decoding process and directly
+// samples-parser, []byte, and nil. []byte will skip the decoding process and directly
```
cevian, line 249: bad change (many places in file)

pkg/pgmodel/querier/querier.go
```diff
@@ -44,11 +44,12 @@ const (
 
 // NewQuerier returns a new pgxQuerier that reads from PostgreSQL using PGX
 // and caches metric table names and label sets using the supplied caches.
-func NewQuerier(conn pgxconn.PgxConn, metricCache cache.MetricCache, labelsReader lreader.LabelsReader) Querier {
+func NewQuerier(conn pgxconn.PgxConn, metricCache cache.MetricCache, labelsReader lreader.LabelsReader, mtAuthr multi_tenancy_read.Authorizer) Querier {
```
cevian, line 47: mtaAthr -> mtAuth many places

pkg/pgmodel/querier/querier.go
```diff
@@ -62,13 +63,17 @@ type pgxQuerier struct {
 	conn             pgxconn.PgxConn
 	metricTableNames cache.MetricCache
 	labelsReader     lreader.LabelsReader
+	mtAuthr          multi_tenancy_read.Authorizer
 }
 
 var _ Querier = (*pgxQuerier)(nil)
 
-// Select implements the Querier interface. It is the entry point for our
+// Select implements the Querier samples-parser. It is the entry point for our
```
cevian, line 71: bad change

pkg/pgmodel/querier/querier.go
```diff
@@ -78,21 +83,23 @@ func (q *pgxQuerier) Select(mint int64, maxt int64, sortSeries bool, hints *stor
 	return ss, topNode
 }
 
-// Query implements the Querier interface. It is the entry point for
+// Query implements the Querier samples-parser. It is the entry point for
```
cevian, line 86: bad change

pkg/runner/flags.go
```diff
@@ -143,3 +158,40 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
...
+		return fmt.Errorf("'multi-tenancy-valid-tenants' cannot be empty")
+	}
+	cfg.ValidTenantsList = strings.Split(cfg.validTenantsStr, ",")
+	cfg.ValidTenantsList = removeEmptyTenants(cfg.ValidTenantsList)
+	return nil
+}
+
+// removeEmptyTenants protects against corner cases, when the user enters comma separated tenants
+// such that there is a trailing comma towards the end.
+func removeEmptyTenants(t []string) (tenants []string) {
```
cevian, line 189: we should also make sure no tenant is equal to allowAllTenants

pkg/ha/ha_parser.go
```diff
@@ -6,10 +6,13 @@ package ha
 
 import (
 	"fmt"
+
 	promModel "github.com/prometheus/common/model"
 	"github.com/timescale/promscale/pkg/log"
+	multi_tenancy_write "github.com/timescale/promscale/pkg/multi-tenancy/write"
```
cevian, line 12: any time you need to alias a pkg like this it's code smell. It's a sign that the package is badly named and/or organized. I think this should be "github.com/timescale/promscale/pkg/tenancy" and the authorizer should be WriteAuthorizer

cevian (changes requested):
Quite a few changes needed. Some refactorings/renames went bad. Namely "interface"-> "samples-parser" applied to way too many places.  Naming of packages is another very big issue. see - https://golang.org/doc/effective_go#package-names

pkg/api/common.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/util/httputil"
+	multi_tenancy "github.com/timescale/promscale/pkg/multi-tenancy"
```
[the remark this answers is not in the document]
Harkishen-Singh: Because in encoding/base64, they both are different terms. In multi-tenancy, it's a single term.

What happens if tomorrow, we do a package like, soft-tenancy and hard-tenancy? Each of such words corresponds to a single term.

pkg/api/common.go
```diff
@@ -86,7 +87,8 @@ type Config struct {
 	AdminAPIEnabled bool
 	TelemetryPath   string
 
-	Auth *Auth
+	Auth         *Auth
+	MultiTenancy multi_tenancy.MultiTenancy
```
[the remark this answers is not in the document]
Harkishen-Singh: tenancy.<.*> is generic. There can be multiple things of tenancy and I don't think we should put multi-tenancy under the single tenancy name everywhere.

pkg/api/common.go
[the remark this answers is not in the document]
antekresic: `pkg/tenancy`
`pkg/tenancy/multi`
`pkg/tenancy/soft`
`pkg/tenancy/hard`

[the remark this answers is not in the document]
Harkishen-Singh: soft and hard are parts of multi-tenancy. But I got the point. Thank you.

pkg/multi-tenancy/write/write.go
Harkishen-Singh: whether or not to allow multi-tenancy to accept non MT write requests is accepted from `multi-tenancy-allow-non-multi-tenant-writes` flag and is checked below this (in the commit).

pkg/api/write.go
```diff
@@ -53,6 +54,16 @@ func Write(writer ingestor.DBInserter, elector *util.Elector, metrics *Metrics)
 
 		metrics.LeaderGauge.Set(1)
 
+		var tenant, tenantToken string
+		if authr := apiConf.MultiTenancy.WriteAuthorizer(); authr != nil {
+			tenant = getTenant(r)
+			if !authr.IsAuthorized(tenant) {
+				log.Error("msg", multi_tenancy.ErrUnauthorized.Error()+tenantToken)
```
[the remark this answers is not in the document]
Harkishen-Singh: ah, the aim here was to print the tenant and not token. Sorry.

pkg/ha/ha_parser.go
```diff
@@ -82,9 +87,20 @@ func (h *haParser) ParseData(tts []prompb.TimeSeries) (map[string][]model.Sample
...
 		}
 
+		var (
+			ok     bool
+			labels = t.Labels
+		)
+		if h.multiTenancy != nil {
+			labels, ok = h.multiTenancy.VerifyAndApplyTenantLabel(tenant, t.Labels)
+			if !ok {
+				return nil, -1, multi_tenancy_write.ErrInvalidLabels
```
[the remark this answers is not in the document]
Harkishen-Singh: That would be better. Fixed this.

pkg/multi-tenancy/read/read.go
Harkishen-Singh: Yes, that would be better

pkg/multi-tenancy/read/read.go
Harkishen-Singh: Yep

pkg/multi-tenancy/read/read.go
Harkishen-Singh: Right now, if AllowNonMT flag is set, then it will not apply safety matcher. The safety matcher will be applied only if the config is selectiveConfig.

> Shouldn't this respect AllowNonMT flag?

I think it happens implicitly. When you choose `AllowNonMT` flag, openConfig is used. When this config is noted in NewReadAuthorizer, it does not applied the safety matcher.

pkg/multi-tenancy/write/write.go
Harkishen-Singh: With the recent commits, if the __tenant__ exists, then it ill not come down to this code, so I think it solves the comment here.

pkg/multi-tenancy/write/write_test.go
Harkishen-Singh: done in recent commit

pkg/pgmodel/ingestor/samples-parser/parser.go
Harkishen-Singh: Yep, even i didnt like this, but that I guess was done to avoid cyclic import. Will rebase and see if this can be done in a better manner.

pkg/runner/flags.go
Harkishen-Singh: If allowAllTenants is the option, then this function will not be called.

This takes care of it (first case inside `validateMultiTenancyFlags()`)

```golang
	if cfg.validTenantsStr == allowAllTenants {
		cfg.SkipTenantValidation = true
		return nil
	} else if cfg.validTenantsStr == "" {
		return fmt.Errorf("'multi-tenancy-valid-tenants' cannot be empty")
	}
```

pkg/ha/ha_parser.go
```diff
@@ -18,22 +21,24 @@ const ReplicaNameLabel = "__replica__"
 const ClusterNameLabel = "cluster"
 
 type haParser struct {
-	service *Service
-	scache  cache.SeriesCache
+	service      *Service
+	scache       cache.SeriesCache
+	multiTenancy multi_tenancy_write.Authorizer
```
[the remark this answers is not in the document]
Harkishen-Singh: Seems like rebasing resolved this. Please feel free to unresolved in case it's not.

pkg/multi-tenancy/read/read.go
Harkishen-Singh: With recent commits, it should allow writing and reading from non-tenants only if `AllowNonMT` flag is enabled.

pkg/pgmodel/ingestor/samples-parser/parser.go
Harkishen-Singh: Luckily, rebasing made this not a requirement. Hence, got rid of samples-parser package.

pkg/ha/ha_parser.go
Harkishen-Singh: That is very true. However, i feel keeping respective authorizer in their own packages is better for readability.
I have changed `multi_tenancy_write` -> `mtw` and it looks more pleasing. 

Let me know if ore changes are required here.

pkg/tenancy/write/write.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (a *writeAuthorizer) IsAuthorized(tenantName string) error {
+	if a.IsTenantAllowed(tenantName) {
+		return nil
+	}
+	return fmt.Errorf("%s %s", ErrUnauthorizedTenant.Error(), tenantName)
+}
+
+func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantNameFromHeader string, labels []prompb.Label) ([]prompb.Label, error) {
```
cevian, line 33: This seems really complex. Can we simplify by splitting into a few functions:

```
func (a *writeAuthorizer) getTenantLabelMatchingHeader(tenantNameFromHeader string, labels []prompb.Label) ([]prompb.Label, error) {
	for _, label := range labels {
		if label.Name == config.TenantLabelKey {
			if label.Value == tenantNameFromHeader {
				return labels, nil
			}
			if label.Value = "" {
				// Tenant label exists but no tenant value. This is invalid.
				return labels, fmt.Errorf("%s exists with an empty value", config.TenantLabelKey)
			}
			return labels, errTenantMismatch
		}
	}
	labels = append(labels, prompb.Label{Name: config.TenantLabelKey, Value: tenantNameFromHeader})
	return labels, nil
}



func (a *writeAuthorizer) getTenantNameFromLabel(labels []prompb.Label) (string) {
	for _, label := range labels {
		if label.Name == config.TenantLabelKey {
			//any value including empty is valid 
			return labels.Value
		}
	}
	return  ""
}

func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantNameFromHeader string, labels []prompb.Label) ([]prompb.Label, error) {
	if tenantNameFromHeader != "" {
		return getTenantLabelMatchingHeader(tenantName, labels)
	}

	tenantName := getTenantNameFromLabel(labels)	
	if tenantName == "" {
		//can we integrate this inside of IsAuthorized if given the empty string?
		if a.AllowNonTenants() {
			// Non-multi-tenant write request.
			// We accept non-multi-tenant write requests in tenancy mode.
			return labels, nil
		} else {
			return labels, fmt.Errorf("non multi-tenant writes not allowed")
		}
	}
	return labels, a.IsAuthorized(tenantName)
}
```

cevian: The code as-is has high Cyclomatic complexity making it hard to verify

cevian (changes requested):
Looking much better. Suggested a few code simplifications. Also recommend moving tenancy/read tenancy/write and tenancy/config into tenancy/. I think that would vastly simplify the API.

Will review test on the next iteration

pkg/runner/flags.go
Harkishen-Singh: Ah, that is an interesting case. I definitely missed that!

pkg/tenancy/read/read.go
```diff
@@ -0,0 +1,58 @@
...
+	config.Config
+	// mtSafetyLabelPair is a label-pair that is applied to incoming multi-tenant read requests for security reasons.
+	// This matcher helps prevent a query from querying a tenant for the query has not been authorized.
+	mtSafetyLabelMatcher *labels.Matcher
+}
+
+// NewPlainReadAuthorizer is a authorizer that does not care about the token checks. It just makes sure that the
+// read operations are performed against a valid tenants, which Promscale has been asked for.
+func NewAuthorizer(cfg config.Config) (Authorizer, error) {
+	if _, ok := cfg.(*config.AllowAllTenantsConfig); ok || cfg.AllowNonTenants() {
```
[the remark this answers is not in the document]
Harkishen-Singh: That was clearly a bug! Thank you.

pkg/tenancy/write/write.go
Harkishen-Singh: Simplified with another but similar function splitting.

pkg/runner/flags_test.go
```diff
@@ -85,6 +82,41 @@ func TestParseFlags(t *testing.T) {
 				return c
 			},
 		},
+		{
```
cevian, line 85: I think these tests should be in pkg/tenancy/flags_test.go

pkg/pgclient/client.go
```diff
@@ -180,9 +183,13 @@ func (c *Client) Close() {
...
 }
 
+func (c *Client) Ingestor() *ingestor.DBIngestor {
+	return c.ingestor
+}
+
 // Ingest writes the timeseries object into the DB
-func (c *Client) Ingest(tts []prompb.TimeSeries, req *prompb.WriteRequest) (uint64, error) {
-	return c.ingestor.Ingest(tts, req)
+func (c *Client) Ingest(tenant string, tts []prompb.TimeSeries, req *prompb.WriteRequest) (uint64, error) {
```
cevian, line 191: can't say I love this change, but I don't have a good idea of how to avoid. Maybe @antekresic will have an idea here.

cevian (changes requested):
Getting closer. I am gonna add @antekresic as a second reviewer now.

pkg/pgclient/client.go
Harkishen-Singh: We can pass there params as single encapsulated struct.

pkg/tenancy/read.go
```diff
@@ -0,0 +1,78 @@
...
+	TypeConfig
+	// mtSafetyLabelPair is a label-pair that is applied to incoming multi-tenant read requests for security reasons.
+	// This matcher helps prevent a query from querying a tenant for the query has not been authorized.
+	mtSafetyLabelMatcher *labels.Matcher
+}
+
+// NewPlainReadAuthorizer is a authorizer that does not care about the token checks. It just makes sure that the
+// read operations are performed against a valid tenants, which Promscale has been asked for.
+func NewReadAuthorizer(cfg TypeConfig) (ReadAuthorizer, error) {
+	if _, ok := cfg.(*AllowAllTenantsConfig); ok {
```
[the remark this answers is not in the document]
Harkishen-Singh: I do not understand this.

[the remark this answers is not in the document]
Harkishen-Singh: Safety matcher are just required in read ops. Why does it need to be on the config types?

pkg/tenancy/write.go
```diff
@@ -0,0 +1,66 @@
...
+			} else {
+				return labels, fmt.Errorf("non multi-tenant writes not allowed")
+			}
+		}
+		labels = append(labels, prompb.Label{Name: TenantLabelKey, Value: tenantNameFromHeader})
+		return labels, nil
+	}
+	if tenantNameFromLabels != tenantNameFromHeader {
+		return labels, fmt.Errorf("mismatch: tenant name from labels does not match with that from header")
+	}
```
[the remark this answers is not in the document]
Harkishen-Singh: > TenantNamaeFromHeader == "" and tenantNameFromLabel != "".

If `tenantNameFromLabel` is not empty, then the tenant name is already applied to the label set and we do not need to do anything else. We just need to authorize the label value and that's it.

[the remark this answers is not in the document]
Harkishen-Singh: Added more unit tests to ensure tenant from labels. See `TestVerifyAndApplyTenantLabel` for more info.

pkg/tenancy/write.go
```diff
@@ -0,0 +1,66 @@
...
+	return labels, nil
+}
+
+func getTenantNameFromLabels(labels []prompb.Label) (string, error) {
+	for _, label := range labels {
+		if label.Name == TenantLabelKey {
+			if label.Value == "" {
+				return "", fmt.Errorf("%s exists with an empty value", TenantLabelKey)
```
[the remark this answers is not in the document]
Harkishen-Singh: Oh yes, the authority to determine whether to error or not should be in the calling function's hand.

pkg/runner/flags_test.go
Harkishen-Singh: The test approach here is pretty good and keeps all flags related things here. Plus, the way params are expected here vs in tenancy/flags.go makes it more preferable to write those tests here.

pkg/pgclient/client.go
Harkishen-Singh: Changed this to encapsulated `Request` type.

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,393 @@
...
+		cfg := tenancy.NewSelectiveTenancyConfig(tenants[:2], false) // valid tenant-a & tenant-b.
+		mt, err := tenancy.NewMultiTenancy(cfg)
+		require.NoError(t, err)
+
+		// Ingestion.
+		client, err := pgclient.NewClientWithPool(&pgclient.Config{HAEnabled: false}, 1, pgxconn.NewPgxConn(db), mt)
+		require.NoError(t, err)
+
+		authr := mt.WriteAuthorizer()
+		//for _, tenant := range tenants {
```
Harkishen-Singh, line 162: Debug comments, will be removed after final testing.

pkg/runner/flags_test.go
Harkishen-Singh: refer to `fullyParse`

pkg/tenancy/write.go
```diff
@@ -0,0 +1,76 @@
...
+			return fmt.Errorf("non multi-tenant writes not allowed"), true
+		}
+	}
+	if a.IsTenantAllowed(tenantName) {
+		return nil, false
+	}
+	return fmt.Errorf("%s %s", ErrUnauthorizedTenant.Error(), tenantName), false
+}
+
+func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantNameFromHeader string, labels []prompb.Label) ([]prompb.Label, error) {
```
cevian, line 41: We talked on slack about re-writing a bunch of the logic here.

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,383 @@
...
+			Matchers: []*prompb.LabelMatcher{
+				{
+					Type:  prompb.LabelMatcher_EQ,
+					Name:  model.MetricNameLabelName,
+					Value: "secondMetric",
+				},
+				{
+					Type:  prompb.LabelMatcher_RE,
+					Name:  tenancy.TenantLabelKey,
+					Value: "tenant-a|tenant-c",
```
cevian, line 136: add another case with no matchers on tenants at all

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,383 @@
...
+		require.NoError(t, err)
+
+		// Ingest tenant-a.
+		_, err = client.Ingest(ingstr.Request{Tenant: tenants[0], Ts: ts, Req: ingstr.NewWriteRequest()})
+		require.NoError(t, err)
+		// Ingest tenant-b.
+		_, err = client.Ingest(ingstr.Request{Tenant: tenants[1], Ts: ts, Req: ingstr.NewWriteRequest()})
+		require.NoError(t, err)
+		// Ingest tenant-c.
+		_, err = client.Ingest(ingstr.Request{Tenant: tenants[2], Ts: ts, Req: ingstr.NewWriteRequest()}) // Invalid ingestion, so this should not get ingested.
```
cevian, line 168: Important! shouldn't this return an error? Trying to ingest an invalid tenant should result in auth error

cevian (changes requested):
Mostly minor items now.

pkg/tenancy/interface.go
```diff
@@ -0,0 +1,31 @@
...
+type ReadAuthorizer interface {
+	// ApplyTenantMatcher applies a safety matcher to incoming query matchers. This safety matcher is responsible
+	// from prevent unauthorized query reads from tenants that the incoming query is not supposed to read.
+	ApplyTenantMatcher(ms []*labels.Matcher) []*labels.Matcher
+}
+
+// WriteAuthorizer tells if a write request is authorized to be written.
+type WriteAuthorizer interface {
+	// isAuthorized verifies if the tenant to be inserted is authorized. It returns a ErrUnauthorizedTenant if the tenant is not authorized.
+	isAuthorized(tenantName string) (error, bool)
```
antekresic, line 28: Can you clarify the reason about defining a private function in an interface? Doesn't make much sense to me.

pkg/tenancy/write.go
antekresic: Just looking at the function name, it seems like it needs to be split up into two since its clearly breaking SRP.

pkg/pgmodel/ingestor/parser.go
```diff
@@ -50,9 +62,20 @@ func (d *Parser) ParseData(tts []prompb.TimeSeries) (map[string][]model.Samples,
 			continue
 		}
 
+		var (
+			err    error
+			labels = t.Labels
+		)
+		if d.wAuth != nil {
+			labels, err = d.wAuth.VerifyAndApplyTenantLabel(tenant, t.Labels)
```
antekresic, line 70: I don't think we should be verifying the tenant here. Its too far down the ingestion pipeline, it should be verified at the API level and after validation, pushed down to the parser to apply the necessary tenant label.

pkg/tenancy/interface.go
```diff
@@ -0,0 +1,31 @@
...
+
+const regexOR = "|"
+
+var ErrUnauthorizedTenant = fmt.Errorf("authorized or invalid tenant")
+
+// ReadAuthorizer tells if a read request is allowed to query via Promscale.
+type ReadAuthorizer interface {
+	// ApplyTenantMatcher applies a safety matcher to incoming query matchers. This safety matcher is responsible
+	// from prevent unauthorized query reads from tenants that the incoming query is not supposed to read.
+	ApplyTenantMatcher(ms []*labels.Matcher) []*labels.Matcher
```
antekresic, line 22: Nit: while not a deal breaker, its obvious even from the function signature that we are appending the tenant matcher hence the function should be called `AppendTenantMatcher`.

antekresic (changes requested):
We need to split up the writer logic into verifying and applying tenancy. Verifying can be done in the write API level, while the applying can be done in parser level.

In a later PR, we should think about generalizing label and sample transformation in the parser since we seem to be doing a lot of that with HA and multi-tenancy features.

pkg/api/common.go
```diff
@@ -127,6 +126,16 @@ func Validate(cfg *Config) error {
...
 }
 
+func getTenant(r *http.Request) (tenant string) {
+	// We do not look for `X-` since it has been deprecated as mentioned in https://datatracker.ietf.org/doc/html/rfc6648.
+	if tn := r.Header.Get("TENANT"); tn == "" {
+		return
+	} else {
+		tenant = tn
+	}
+	return
```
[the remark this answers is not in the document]
Harkishen-Singh: Wow, I wrote it very silly. Thank you for this.

pkg/tenancy/interface.go
Harkishen-Singh: It's for tests in write_test.go to access the func. The constructors of configs expose in AuthConfig interface and hence is needed. We can get rid of this by exposing with strict types, and using interface only on the receiving var, but I felt using interface at the very constructor level keeps things well abstracted.

pkg/tenancy/write.go
Harkishen-Singh: after the suggestion which mat gave, it should solve all the concerns here.

pkg/tests/end_to_end_tests/multi_tenancy_test.go
Harkishen-Singh: Ah yes, thank you for catching that. The VerifyAndApplyTenant didn't check at start whether the header tenant is authorized. This is fixed in recent commits.

pkg/pgmodel/ingestor/parser.go
Harkishen-Singh: Verification was first at `write.go` but since we want to support tenants via external labels as well, hence things have to get verified here.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -47,22 +49,46 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
 		if len(ts.Samples) == 0 {
 			continue
 		}
+		labels = ts.Labels
+		if ingestor.wAuth != nil {
+			labels, err = ingestor.wAuth.VerifyAndApplyTenantLabel(r.Tenant, labels)
+			if err != nil {
+				return 0, fmt.Errorf("db-ingestor: %w", err)
+			}
+		}
```
cevian, line 88: Important organizational question: with the new refactor, can't this code go into a parser preProcessor? The preprocessor would just modify ts.Labels with the updated labels and return an error an auth error if needed.
That way the ingestor does not need to know anything about auth or multitenancy, nor would the modification to the ingestor's input (The Request change) be necessary.

cevian (changes requested):
Some minor things but one major thing: I think the VerifyAndApplyTenantLabel call should now be in a parser Preprocessor. Left a detailed comment below in ingestor.go

pkg/tenancy/write.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if a.IsTenantAllowed(tenantName) {
+		return nil
+	}
+	return fmt.Errorf("%s %s", ErrUnauthorizedTenant.Error(), tenantName)
+}
+
+func (a *writeAuthorizer) VerifyAndApplyTenantLabel(tenantNameFromHeader string, labels []prompb.Label) ([]prompb.Label, error) {
+	if tenantNameFromHeader != "" {
+		if err := a.isAuthorized(tenantNameFromHeader); err != nil {
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes, I think the flow of function is changed a bit and now it is easier to follow in terms of error checking.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: I agree thats gonna be more clean but I fear running an extra loop on incoming timeseries will be costly, than handling all this in Ingest(). Is it right or am I wrong?

pkg/tenancy/write.go
```diff
@@ -0,0 +1,76 @@
...
+
+var errTenantMismatch = fmt.Errorf("__tenant__ value and tenant-name from headers are different")
+
+// NewWriteAuthorizer returns a new plainWriteAuthorizer.
+func NewWriteAuthorizer(config AuthConfig) WriteAuthorizer {
+	return &writeAuthorizer{config}
+}
+
+func (a *writeAuthorizer) isAuthorized(tenantName string) error {
+	if tenantName == "" {
```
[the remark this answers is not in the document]
Harkishen-Singh: We do not require this. The IsTenantAllowed was missing a case and hence I had to add this, but now it is fixed.

pkg/api/write.go
```diff
@@ -24,6 +24,85 @@ import (
 	"github.com/timescale/promscale/pkg/util"
 )
 
+type writeStage func(http.ResponseWriter, *http.Request, *Metrics) bool
```
cevian, line 27: why was all of this moved? Are there any changes in the moved code?

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,580 @@
...
+		require.NoError(t, err)
+		require.NoError(t, err)
+
+		// Ingest tenant-c.
+		request = ingstr.NewWriteRequestWithTs(copyMetrics(ts))
+		err = wauth.Process(requestWithHeaderTenant(tenants[2]), request)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), "write-authorizer process: authorization error for tenant tenant-c: unauthorized or invalid tenant")
+
+		// Querying.
```
cevian, line 239: hmm, tenant-c is never ingested though, so showing that query results don't contain tenant-c doesn't test the query path properly.

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,580 @@
...
+					Value: "thirdMetric",
+				},
+			},
+			StartTimestampMs: 2,
+			EndTimestampMs:   4,
+		})
+		require.NoError(t, err)
+
+		// Verifying result.
+		verifyResults(t, expectedResult, result) // Fails: expected to be 1, but shows 0 result.
```
cevian, line 398: what does the comment here mean?

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,580 @@
...
+
+		// ----- query across multiple tenants (tenant-a & tenant-c) -----
```
cevian, line 400: again would be a better test if something with tenant-c was previously ingested.

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -0,0 +1,580 @@
...
+		require.NoError(t, err)
+		_, err = client.Ingest(request)
+		require.NoError(t, err)
+		require.NoError(t, err)
+
+		// Ingest tenant-c.
+		request = ingstr.NewWriteRequestWithTs(applyTenantInLabels(tenants[2], copyMetrics(ts)))
+		err = wauth.Process(&http.Request{}, request)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), "write-authorizer process: authorization error for tenant tenant-c: unauthorized or invalid tenant")
```
cevian, line 480: again this means tenant-c is not in the DB, making the query below less useful.

cevian (changes requested):
Have some concerns about the e2e tests but everything else looks good

pkg/api/write.go
Harkishen-Singh: it's just moved, without any changes. The reason being easier navigation within `api/write.go`, as we can click on can stage function to go directly to the function definition.

pkg/runner/client.go
```diff
@@ -155,9 +156,25 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
...
 	}
+
+	multiTenancy := tenancy.NewNoopAuthorizer()
+	if cfg.TenancyCfg.EnableMultiTenancy {
+		var multiTenancyConfig tenancy.AuthConfig
+		if cfg.TenancyCfg.SkipTenantValidation {
+			multiTenancyConfig = tenancy.NewAllowAllTenantsConfig(cfg.TenancyCfg.AllowNonMTWrites)
+		} else {
+			multiTenancyConfig = tenancy.NewSelectiveTenancyConfig(cfg.TenancyCfg.ValidTenantsList, cfg.TenancyCfg.AllowNonMTWrites)
+		}
```
antekresic, line 167:
```suggestion
		multiTenancyConfig := tenancy.NewAllowAllTenantsConfig(cfg.TenancyCfg.AllowNonMTWrites)
		if !cfg.TenancyCfg.SkipTenantValidation {
			multiTenancyConfig = tenancy.NewSelectiveTenancyConfig(cfg.TenancyCfg.ValidTenantsList, cfg.TenancyCfg.AllowNonMTWrites)
		}
```
Might even want to reverse the options, so we don't use negation in the if clause. Should be cleaner to read.

pkg/tenancy/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+			forbid: true,
+		},
+	}
+	for _, tc := range tcs {
+		c := tc.c
+		if tc.forbid {
+			require.False(t, c.IsTenantAllowed(incomingTenant), tc.name)
+		} else {
+			require.True(t, c.IsTenantAllowed(incomingTenant), tc.name)
+		}
```
antekresic, line 50: I'm not big on using require but these if else statements negate the purpose of using it at all.

Why not:
```suggestion
		require.Equal(t, c.IsTenantAllowed(incomingTenant), !tc.forbid)
```
Or even, changing `forbid` to `allowed` so we don't have negation and it reads much better.

pkg/tenancy/config.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// TenantLabelKey is a label key reserved for tenancy.
+const TenantLabelKey = "__tenant__"
+
+// AuthConfig defines configuration type for tenancy.
+type AuthConfig interface {
+	// AllowNonTenants returns true if tenancy is asked to accept write-requests from non-multi-tenants.
+	allowNonTenants() bool
```
antekresic, line 21: Comment and name don't match. Should this be exported?

antekresic (approved):
Looks in OK shape. Once the comments are addressed, should be able to merge.

pkg/pgmodel/ingestor/insert_ctx.go
```diff
@@ -20,6 +20,12 @@ func NewWriteRequest() *prompb.WriteRequest {
 	return wrPool.Get().(*prompb.WriteRequest)
 }
 
+func NewWriteRequestWithTs(ts []prompb.TimeSeries) *prompb.WriteRequest {
+	wr := wrPool.Get().(*prompb.WriteRequest)
```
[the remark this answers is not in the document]
Harkishen-Singh: And even better to do

```golang
// NewWriteRequest returns a new *prompb.WriteRequest from the pool and applies ts to it if ts is not nil.
func NewWriteRequest(ts []prompb.TimeSeries) *prompb.WriteRequest {
	wr := wrPool.Get().(*prompb.WriteRequest)
	if ts != nil {
		wr.Timeseries = ts
	}
	return wr
}
```

pkg/runner/client.go
Harkishen-Singh: I think this is better than the reverse. Selective config is a case that is non-default so.

pkg/tenancy/config.go
Harkishen-Singh: comment updated. I don't think exporting this is required.

pkg/pgmodel/ingestor/insert_ctx.go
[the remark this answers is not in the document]
antekresic: I actually don't like that we change this like so just because of the tests. I'd prefer to add a new function (but have that function use `NewWriteRequest`) so we can use that but leave all the old code as is, using the `NewWriteRequest` without parameters.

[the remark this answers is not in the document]
antekresic: Since `NewWriteRequestWithTs` is only used by tests, it should be declared in a test file IMO. Less code to test, less to break.

cevian (approved):

PR #551, opened 2021-03-29
rework - Add version flag to prom-migrator command

niconosenzo:
Added back changes from PR #546

### DIFF
--- a/cmd/prom-migrator/main.go
+++ b/cmd/prom-migrator/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/timescale/promscale/pkg/migration-tool/reader"
 	"github.com/timescale/promscale/pkg/migration-tool/utils"
 	"github.com/timescale/promscale/pkg/migration-tool/writer"
+	"github.com/timescale/promscale/pkg/version"
 )
 
 const (
@@ -48,11 +49,19 @@ type config struct {
 
 func main() {
 	conf := new(config)
+	args := os.Args[1:]
+	if shouldProceed := parseArgs(args); !shouldProceed {
+		os.Exit(0)
+	}
+
 	parseFlags(conf, os.Args[1:])
+
 	if err := log.Init(log.Config{Format: "logfmt", Level: "debug"}); err != nil {
+		fmt.Println("Version: ", version.PromMigrator)
 		_, _ = fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+	log.Info("Version", version.PromMigrator)
 	if err := validateConf(conf); err != nil {
 		log.Error("msg", "could not parse flags", "error", err)
 		os.Exit(1)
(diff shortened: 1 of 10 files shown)


niconosenzo:
@Harkishen-Singh hopefully it works this time :crossed_fingers:

niconosenzo:
The tests still failing with `panic: proto: duplicate enum registered: prometheus.ReadRequest_ResponseType`, thought it was a local build issue

cmd/prom-migrator/main.go
```diff
@@ -48,6 +49,11 @@ type config struct {
 
 func main() {
 	conf := new(config)
+    args := os.Args[1:]
+    if shouldProceed := runner.ParseArgs(args); !shouldProceed {
+        os.Exit(0)
+    }
```
Harkishen-Singh, line 55: Importing this from the runner package causes the problem. This might be true in itself because the runner package was aimed to be specific to Promscale only.

This can be solved by making a `parseArgs()` function in `cmd/prom-migrator/main.go`, which is kind of a replica of runner.ParseArgs(). Like,

```golang
func parseArgs(args []string) (shouldProceed bool) {
	shouldProceed = true // Some flags like 'version' are just to get information and not proceed the actual execution. We should stop in such cases.
	for _, flag := range args {
		flag = flag[1:]
		switch flag {
		case "version":
			shouldProceed = false
			fmt.Println(version.VersionPromMigrator)
		}
	}
	return
}
```

And then in `pkg/version/version.go`, we require a version for Prom-migrator that should be printed here. You can add a `VersionPromMigrator` and import that in `main.go` as explained above and that should be enough.

Also, I think it will be better to convert `Version` to `VersionPromscale` and make the related name changes across the repo. That should be good to go.

Harkishen-Singh (changes requested):
Ah, using runner package wasnt the way in which the system would be happy.

If I am not wrong, the reason why `panic: proto: duplicate enum registered: prometheus.ReadRequest_ResponseType` occurs is when you call register a Prometheus metric twice. This is what is happening in the test and is the cause of the failure.

Keeping the parseFlags function (given in suggestions) in the `cmd/prom-migrator/main.go` solves this issue and then I think we are good to go.

Note: Please run `go fmt ./...` before the commit to fix the formatting issue.

PS: Please consider pulling from master branch before you make changes. The merge requirement is to have an updated PR with the master before merging. I am sorry for the extra pain here.

cmd/prom-migrator/main.go
Harkishen-Singh: Lets change `VersionPromscale` => `Promscale` and `VersionPromMigrator` => `PromMigrator` so that while importing, it becomes `version.Promscale` and `version.PromMigrator` which is self-explanatory.

niconosenzo: Wow thanks for the detailed steps @Harkishen-Singh, I just did the proposed changes and `git rebase master` before pushing, let me know if there is something else we need for merging this change.

Harkishen-Singh:
Thank you for updating this. 

cc @cevian , note: the way we are handling version now, both for Promscale and Prom-migrator.

cevian (approved):

cevian:
I'll work on rebasing this

Harkishen-Singh (approved):

cevian:
Thanks agaIN @niconosenzo

PR #552, opened 2021-03-29
Prevent crash with client does not send token when expected.

Harkishen-Singh:

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -106,9 +106,9 @@ func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
 	if cfg.Auth.BearerToken != "" {
 		return func(w http.ResponseWriter, r *http.Request) {
 			splitToken := strings.Split(r.Header.Get("Authorization"), "Bearer ")
-			if cfg.Auth.BearerToken != splitToken[1] {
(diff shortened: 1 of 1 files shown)


pkg/api/router.go
```diff
@@ -105,10 +105,19 @@ func authHandler(cfg *Config, handler http.HandlerFunc) http.HandlerFunc {
 
 	if cfg.Auth.BearerToken != "" {
 		return func(w http.ResponseWriter, r *http.Request) {
+			errFunc := func(token string) {
+				log.Error("msg", "Unauthorized access to endpoint, invalid bearer token: " + token)
```
antekresic, line 109: Logging the token would not be advisable because security concerns i.e. valid token becomes logged by mistake and is persisted where it can be accessed by a 3rd party. I suggest we remove this part.

If we do that, this change would be as simple as adding a check for length in line 109 of the original file.
It would look like this:
```
if len(splitToken) < 2 || cfg.Auth.BearerToken != splitToken[1] {
```

antekresic (changes requested):

cevian (changes requested):
I agree with Ante's comment

antekresic (approved):

Issue #553, opened 2021-03-29
Authentication for Remote write is parsed but not processed with prom-migrator


lkashiwa:
When migrating Prometheus to InfluxDB or Cortex with Authentication needed, the parameters are read and parsed but config ignores that.

~/prom-migrator$ ./prom-migrator -mint=1616997627 -progress-enabled=false -read-url=http://localhost:9090/api/v1/read -write-url=https://xxxxx/write -write-auth-username=xxxx -write-auth-password=yyyyyyy
level=info ts=2021-03-29T10:46:20.158Z caller=main.go:60 msg="&{prom-migrator 1616997627000 1616997627 1617014780000 1617014780 524288000 500MB 1 1 http://localhost:9090/api/v1/read https://xxxxx/write prom_migrator_progress  false {  } {xxxx yyyyyyy }}+"
level=info ts=2021-03-29T10:46:20.158Z caller=planner.go:78 msg="Resuming from where we left off is turned off. Starting at the beginning of the provided time-range."
level=info ts=2021-03-29T10:46:20.158Z caller=writer.go:74 msg="writer is up"
level=info ts=2021-03-29T10:46:20.158Z caller=shards.go:100 msg="shard-0 is active"
level=info ts=2021-03-29T10:46:20.158Z caller=reader.go:61 msg="reader is up"
progress: 0.350% | block-1 time-range: 1 mins | mint: 1616997627 | maxt: 1616997687 | time-range: 1 mins 100% |██████████████████████| ( 4/ 4B, 34.630 kB/s)
progress: 1.049% | block-1 time-range: 2 mins | mint: 1616997687 | maxt: 1616997807 | time-range: 2 mins 100% |██████████████████████| ( 4/ 4B, 22.119 kB/s):0s]
progress: 2.099% | block-1 time-range: 3 mins | mint: 1616997807 | maxt: 1616997987 | time-range: 3 mins 100% |██████████████████████| ( 4/ 4B, 47.988 kB/s):0s]
progress: 3.498% | block-1 time-range: 4 mins | mint: 1616997987 | maxt: 1616998227 | time-range: 4 mins 100% |██████████████████████| ( 4/ 4B, 50.016 kB/s):0s]
progress: 5.247% | block-1 time-range: 5 mins | mint: 1616998227 | maxt: 1616998527 | time-range: 5 mins 100% |██████████████████████| ( 4/ 4B, 62.400 kB/s):0s]
[...]

PR #555, opened 2021-03-29
fix: #553 - WriterAuth are ignored

grzesuav:
Signed-off-by: grzesuav <grzesuav@gmail.com>


Harkishen-Singh (approved):
Ah, that was my mistake. Thank you very much for the fix :muscle:

PR #557, opened 2021-03-31
Add DB connection handling end-to-end test

antekresic:
This test verifies that the system can detect when the connection
to the DB cannot be established. It also checks that operation resumes
as normal as soon as the connection can be established again.

### DIFF
--- a/pkg/pgmodel/model/interface.go
+++ b/pkg/pgmodel/model/interface.go
@@ -47,6 +47,9 @@ func MetricTableName(conn pgxconn.PgxConn, metric string) (string, bool, error) 
 	var possiblyNew bool
 	defer res.Close()
 	if !res.Next() {
+		if res.Err() != nil {
+			return "", true, err
+		}
 		return "", true, errors.ErrMissingTableName
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/model/interface.go
```diff
@@ -47,6 +47,9 @@ func MetricTableName(conn pgxconn.PgxConn, metric string) (string, bool, error)
 	var possiblyNew bool
 	defer res.Close()
 	if !res.Next() {
+		if res.Err() != nil {
```
cevian, line 50: why was this hit? is the metric handling goroutine shutting down and then starting up again?

cevian (approved):
Approved but have one question. Let's hold off on merge

pkg/pgmodel/model/interface.go
antekresic: Its hit because the connection is blocked and the way pgx deals is lazily returning the block error only after you try reading the row. 

That's why we need to check the error if there are no rows.

cevian (approved):

PR #558, opened 2021-03-31
Fix misleading failed samples metric calculation.

Harkishen-Singh:


cevian (approved):
Please merge only after #544

PR #560, opened 2021-04-01
Switch to using a function for data inserts

cevian:
Previously, we used INSERT INTO for copying data into data tables.
This, however requires us to prepare a separate statement for each
unique metric. This ended up creating a lot of prepared statements
(# metrics * # copiers). Each prepare has an overhead of at least
one RTT. This is a problem in environments with high latency.

We switch to using a function that takes the metric name as a parameter
This reduces number of prepared statements for the copier to 1.

We also reduce the statement cache capacity back to the PGX default
of 512.


pkg/pgclient/client.go
```diff
@@ -89,7 +89,7 @@ func getPgConfig(cfg *Config) (*pgxpool.Config, int, error) {
 	} else {
 		connectionArgsFmt = "%s&pool_max_conns=%d&pool_min_conns=%d&statement_cache_capacity=%d"
 	}
-	statementCacheCapacity := cfg.CacheConfig.MetricsCacheSize * 2
+	statementCacheCapacity := 512
```
antekresic, line 92: We should add a comment at this number so we know why we chose it.

pkg/pgmodel/ingestor/insert.go
```diff
@@ -385,13 +383,14 @@ func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
...
 	var affectedMetrics uint64
 	for _, numRows := range numRowsPerInsert {
-		ct, err := results.Exec()
+		var insertedRows int
+		err := results.QueryRow().Scan(&insertedRows)
 		if err != nil {
 			return err
 		}
-		if int64(numRows) != ct.RowsAffected() {
+		if int64(numRows) != int64(insertedRows) {
```
antekresic, line 391: Minor nit: maybe we can simplify these var types so we have to juggle less. I haven't looked at the code too deeply to check if its possible/easy.

antekresic (approved):
Looks good, couple of minor things but its ready for merge even like this.

VineethReddy02:
LGTM!

VineethReddy02 (approved):

PR #564, opened 2021-04-05
Improve locking on metric creation.

cevian:
This PR performs two changes

1) It changes the series lock taken during partition attachment from
access exclusive to share update. The lower lock can be taken
because it turns out PG natively takes the lower lock when attaching a partition
(as opposed to creating a table as a partition).
2) Move the enable-compression code into the metric creation code instead of finalize
code. It is much cheaper to do there instead of in finalize because the operation
takes an exclusive lock on the data tables.

Change 2 also prevents finalize from taking locks on both the data and series tables
in the same txn, which may eliminate a deadlock between finalize and query execution.


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -142,20 +142,19 @@ BEGIN
...
-        --lock is taken.
-        LOCK TABLE ONLY SCHEMA_CATALOG.series IN ACCESS EXCLUSIVE mode;
-
-        EXECUTE format($$
-           ALTER TABLE SCHEMA_CATALOG.series ATTACH PARTITION SCHEMA_DATA_SERIES.%1$I FOR VALUES IN (%2$L)
+        --This may not be strictly necessary but good
+        --to enforce lock ordering (parent->child) explcitly. Note:
+        --creating a table as a partition takes a stronger lock (access exclusive)
+        --so, attaching a partition is better
+        LOCK TABLE ONLY SCHEMA_CATALOG.series IN SHARE UPDATE EXCLUSIVE mode;
```
Harkishen-Singh, line 153: Shouldn't this be `SHARE ROW EXCLUSIVE mode`?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -202,6 +201,14 @@ BEGIN
         END IF;
     END IF;
 
+    --Do not move this into the finalize step, because it's cheap to do while the table is empty
+    --but takes a heavyweight blocking lock otherwise.
+    IF  SCHEMA_CATALOG.is_timescaledb_installed()
+            AND SCHEMA_CATALOG.get_default_compression_setting() THEN
+            PERFORM SCHEMA_PROM.set_compression_on_metric_table(NEW.table_name, TRUE);
```
Harkishen-Singh, line 208: Why do we set compression on a lock metric for maintenance? Wouldnt this lead to unintended result if the user chooses to not compress this metric?

pkg/tests/testdata/expected/info_view-timescaledb-2.out
```diff
@@ -46,7 +46,7 @@ SELECT compress_chunk(show_chunks('prom_data.cpu_usage'));
 SELECT compress_chunk(show_chunks('prom_data.cpu_total'));
              compress_chunk             
 ----------------------------------------
- _timescaledb_internal._hyper_2_2_chunk
+ _timescaledb_internal._hyper_3_2_chunk
```
Harkishen-Singh, line 49: I am not sure why is this change.

pkg/migrations/sql/idempotent/base.sql
cevian: Compression is turned on when the metric is created if `SCHEMA_CATALOG.get_default_compression_setting()` is true here.

pkg/tests/testdata/expected/info_view-timescaledb-2.out
cevian: because we are turning compression on earlier.

pkg/migrations/sql/idempotent/base.sql
cevian: `SHARE UPDATE EXCLUSIVE` is the lock taken for attach partition

pkg/migrations/sql/idempotent/base.sql
Harkishen-Singh: What if user chooses to turn off compression for a metric? Then this will again turn on, even if user didn't want that?

Harkishen-Singh (approved):

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -142,20 +142,19 @@ BEGIN
...
 
         --we will need this lock for attaching the partition so take it now
-        --to avoid lock upgrade. Its critical to minimize work done while this
-        --lock is taken.
-        LOCK TABLE ONLY SCHEMA_CATALOG.series IN ACCESS EXCLUSIVE mode;
-
-        EXECUTE format($$
-           ALTER TABLE SCHEMA_CATALOG.series ATTACH PARTITION SCHEMA_DATA_SERIES.%1$I FOR VALUES IN (%2$L)
+        --This may not be strictly necessary but good
+        --to enforce lock ordering (parent->child) explcitly. Note:
```
antekresic, line 150:
```suggestion
        --to enforce lock ordering (parent->child) explicitly. Note:
```

antekresic (changes requested):
Looks good but the golden files end-to-end test needs to be fixed.

Hard to see what the exact issue is.

antekresic (approved):
Looks good, one typo in comment but should be ready to merge.

PR #567, opened 2021-04-06
Switch to session locks for maintenance tasks

cevian:


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -72,26 +72,34 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.lock_metric_for_maintenance(metric_id
...
+        SELECT pg_try_advisory_lock(12378, metric_id) INTO STRICT res;
 
-        RETURN FOUND;
+        RETURN res;
     ELSE
-        PERFORM m.*
-        FROM SCHEMA_CATALOG.metric m
-        WHERE m.id = metric_id
-        FOR NO KEY UPDATE;
+        PERFORM pg_advisory_lock(12378, metric_id);
```
davidkohn88, line 82: Having a note basically everywhere we use this 12378 thing is probably a good idea, it strikes me as odd every time I see it.

davidkohn88 (approved):
LGTM. Just the one note on the random constant we're using.

pkg/migrations/sql/idempotent/base.sql
Harkishen-Singh: I think this should be done like how we did below

https://github.com/timescale/promscale/blob/9c6e713a96cdb93154cf319702c52619913e8bd6/pkg/pgmodel/migrate.go#L372

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -72,26 +72,34 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.lock_metric_for_maintenance(metric_id
...
 $func$
 LANGUAGE PLPGSQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.lock_metric_for_maintenance(int, boolean) TO prom_writer;
+
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.unlock_metric_for_maintenance(metric_id int)
+    RETURNS VOID
+AS $func$
+DECLARE
+BEGIN
+    PERFORM pg_advisory_unlock(12378, metric_id);
```
Harkishen-Singh, line 96: this needs to be replaced in a similar way

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -72,26 +72,34 @@ CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.lock_metric_for_maintenance(metric_id
...
 AS $func$
 DECLARE
+    res BOOLEAN;
 BEGIN
     IF NOT wait THEN
-        PERFORM m.*
-        FROM SCHEMA_CATALOG.metric m
-        WHERE m.id = metric_id
-        FOR NO KEY UPDATE SKIP LOCKED;
+        SELECT pg_try_advisory_lock(12378, metric_id) INTO STRICT res;
```
Harkishen-Singh, line 78: this too

Harkishen-Singh (approved):
Small nit and then we are good to merge.

PR #568, opened 2021-04-07
Refactor ingest path

cevian:
This is mostly a renaming/reorg refactor. There should not be any logic changes.

- handler->metric batcher
- insert -> copier
- inserter -> dispatcher

Constants are moved to the appropriate files.

Some dead code is removed.

### DIFF
--- a/pkg/pgmodel/ingestor/buffer.go
+++ b/pkg/pgmodel/ingestor/buffer.go
@@ -7,21 +7,32 @@ package ingestor
 import (
 	"sync"
 
-	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
 )
 
-const (
-	// maximum number of insertDataRequests that should be buffered before the
-	// insertHandler flushes to the next layer. We don't want too many as this
-	// increases the number of lost writes if the connector dies. This number
-	// was chosen arbitrarily.
-	flushSize                = 2000
-	getCreateMetricsTableSQL = "SELECT table_name FROM " + schema.Catalog + ".get_or_create_metric_table_name($1)"
-	finalizeMetricCreation   = "CALL " + schema.Catalog + ".finalize_metric_creation()"
-	getEpochSQL              = "SELECT current_epoch FROM " + schema.Catalog + ".ids_epoch LIMIT 1"
-	maxCopyRequestsPerTxn    = 100
-)
+// maximum number of insertDataRequests that should be buffered before the
+// insertHandler flushes to the next layer. We don't want too many as this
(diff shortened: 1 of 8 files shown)


cevian:
Yeah I wasn't sure whether dispatcher or router is better. This class, mainly, routes (or dispatches) samples to the correct metric batcher.

pkg/pgmodel/ingestor/copier.go
```diff
@@ -174,7 +82,7 @@ func runInserter(conn pgxconn.PgxConn, in chan copyRequest) {
 	}
 }
 
-func inserterGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, bool) {
+func copierGetBatch(batch []copyRequest, in chan copyRequest) ([]copyRequest, bool) {
```
Harkishen-Singh, line 85: Should this be `getBatch()` and in comment for this func, we write the copier stuff.

Harkishen-Singh:
If we want a true refactor, there are many things that can be improved. I think we should do them in this PR, otherwise the impact wouldn't be significant. I will comment on those stuff in the PR and then we can make the code sound more pleasant.

antekresic (approved):
Looks good, new names look clearer and better describe the responsibilities.

Harkishen-Singh (changes requested):
Two things that required updation in our discussion:
1. Follow go-vet when writing comments for functions. Please make comments in this PR in accordance with that.
2. rename `dispatcher()` => `seriesDispatcher()` and in the comment, lets add that "this directs the incoming series to their respective `metric batchers` where each `metric batcher` corresponds to a single metric".

pkg/pgmodel/ingestor/copier.go
cevian: I think current name is more descriptive

pkg/pgmodel/ingestor/buffer.go
```diff
@@ -16,13 +15,27 @@ const (
...
 
+type insertDataTask struct {
+	finished *sync.WaitGroup
+	errChan  chan error
+}
+
+// Report that this task is completed, along with any error that may have
+// occurred. Since this is a back-edge on the goroutine graph, it
+// _must never block_: blocking here will cause deadlocks.
+func (idt *insertDataTask) reportResult(err error) {
```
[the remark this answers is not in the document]
cevian: I believe reportResult is a better name as it is used in other contexts too

cevian:
@Harkishen-Singh Dispather is a better name than seriesDispatcher since what is being dispatches is samples not series

Harkishen-Singh:
@cevian I still find comments of functions no following the convention.

Harkishen-Singh (approved):

PR #570, opened 2021-04-08
Prom-migrator: Fix gaps in graphs when migrating on time-based storages.

Harkishen-Singh:

Solves gaps created during the migration process, when using concurrent fetch. The issue was that while we fetch a slab concurrently (dividing into sub-slabs in terms of time-ranges and pulling them concurrently), we need to combine them into a single large time-series before pushing to the writer. However, the combination process did not care about the time order i.e., the order in which concurrent sub-slabs were fetched and combined randomly. This is now fixed and we combine only in the order of creating sub-slabs.

### DIFF
--- a/pkg/migration-tool/planner/slab.go
+++ b/pkg/migration-tool/planner/slab.go
@@ -86,11 +86,13 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
 		if totalRequests == 1 {
 			descriptionMsg = fmt.Sprintf("%s | time-range: %d mins", s.pbarDescriptionPrefix, timeRangeMinutesDelta)
 		}
-		go client.ReadChannels(cctx, readRequest, i, descriptionMsg, responseChan)
+		go client.ReadConcurrent(cctx, readRequest, i, descriptionMsg, responseChan)
 	}
 	var (
 		bytesCompressed   int
 		bytesUncompressed int
+		// pendingResponses is used to ensure that concurrent fetch results are appending in sorted order of time.
+		pendingResponses = make([]*utils.PrompbResponse, totalRequests)
 	)
 	for i := 0; i < totalRequests; i++ {
 		resp := <-responseChan
@@ -98,7 +100,20 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
 		case *utils.PrompbResponse:
 			bytesCompressed += response.NumBytesCompressed
 			bytesUncompressed += response.NumBytesUncompressed
-			s.timeseries = append(s.timeseries, response.Result.Timeseries...)
+			// pendingResponses is in shard order. This means it is in time-order as increasing shard deals with
+			// higher time-ranges of sub-slabs, when fetched concurrently.
+			//
+			// Eg: If a slab (say S, from time-range Ta to Tb) is divided into 5 sub-slabs (s1, s2, s3, s4, s5). Then,
+			// the time-range of those sub-slabs are t1 to t2, t2 to t3, t3 to t4, t4 to t5, t5 to t6 such that
+			// t1 < t2 < t3 < t4 < t5 < t6 and t1 = Ta and t6 = Tb.
+			//
+			// In order to ensure order when we combine these sub-slabs to make a slab, this has to be combined
+			// in the same order (strictly), i.e., s1 + s2 + s3 + s4 + s5, only then the fetched time-series reserve the
+			// order of time.
+			//
+			// Note: Here, response.ID takes care of the 1, 2, 3, 4, 5 respectively, with the only difference that
+			// it starts from 0 in place of 1.
+			pendingResponses[response.ID] = response
 			s.SetDescription(fmt.Sprintf("pulling (%d/%d) ...", i+1, totalRequests), 1)
 		case error:
 			for _, cancelFnc := range cancelFuncs {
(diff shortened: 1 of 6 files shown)


pkg/migration-tool/planner/slab.go
```diff
@@ -108,6 +110,44 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
 		}
 	}
 	close(responseChan)
+	// Combine and append samples.
```
cevian, line 113: I think this should be a separate function

cevian (changes requested):
This PR has no description of the problem being solved, nor the solution implemented.  Both the PR and the commit message should include this information. Is there a corresponding issue as well?

pkg/migration-tool/planner/slab.go
```diff
@@ -108,6 +110,44 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
...
+		s.UpdatePBarMax(s.PBarMax() + 2)
+		s.SetDescription(fmt.Sprintf("combining fetched series from %d responses", totalRequests), 1)
+		timeseries := make(map[string]*prompb.TimeSeries)
+		for i := 0; i < totalRequests; i++ {
+			ts := pendingResponses[i].Result.Timeseries
+			for _, series := range ts {
+				if s, ok := timeseries[getLabelsStr(series.Labels)]; len(series.Samples) > 0 && ok {
+					l := len(s.Samples)
+					if len(s.Samples) > 1 && s.Samples[l-1].Timestamp == series.Samples[0].Timestamp {
+						// Edge case: When concurrent pulls makes the remote-storage send duplicate samples that was sent
```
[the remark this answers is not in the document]
Harkishen-Singh: we do use non-overlapping requests but I kept it in case some storages behave weirdly.

pkg/migration-tool/planner/slab.go
```diff
@@ -108,6 +110,44 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
 		}
 	}
 	close(responseChan)
+	// Combine and append samples.
+	if totalRequests > 1 {
+		s.UpdatePBarMax(s.PBarMax() + 2)
+		s.SetDescription(fmt.Sprintf("combining fetched series from %d responses", totalRequests), 1)
```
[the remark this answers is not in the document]
Harkishen-Singh: we do not really need it, but I did notice that in the case of some very large blocks, the combining process takes a sec or so, so it helps the user know what's happening under the hood. Hence I thought to keep this.

Harkishen-Singh:
Updated the commits and there is no related issue. This was a known problem.

pkg/migration-tool/planner/slab.go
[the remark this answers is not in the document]
cevian: Ok then this should throw an error and not simply cover the issue up.

pkg/migration-tool/planner/slab.go
```diff
@@ -118,6 +144,48 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
...
+func (s *Slab) mergeSubSlabsToSlab(subSlabs []*utils.PrompbResponse) []*prompb.TimeSeries {
+	l := len(subSlabs)
+	s.UpdatePBarMax(s.PBarMax() + 2)
+	s.SetDescription(fmt.Sprintf("combining fetched series from %d responses", l), 1)
+	timeseries := make(map[string]*prompb.TimeSeries)
+	for i := 0; i < l; i++ {
+		ts := subSlabs[i].Result.Timeseries
+		for _, series := range ts {
+			if s, ok := timeseries[getLabelsStr(series.Labels)]; len(series.Samples) > 0 && ok {
+				l := len(s.Samples)
```
cevian, line 156: this masks the l on line 148. Really bad practice. Give it another name

pkg/migration-tool/planner/slab.go
```diff
@@ -118,6 +144,48 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
 	return nil
 }
 
+func (s *Slab) mergeSubSlabsToSlab(subSlabs []*utils.PrompbResponse) []*prompb.TimeSeries {
+	l := len(subSlabs)
```
cevian, line 148: do we really need this var?

pkg/migration-tool/planner/slab.go
```diff
@@ -118,6 +144,48 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
...
+	s.UpdatePBarMax(s.PBarMax() + 2)
+	s.SetDescription(fmt.Sprintf("combining fetched series from %d responses", l), 1)
+	timeseries := make(map[string]*prompb.TimeSeries)
+	for i := 0; i < l; i++ {
```
cevian, line 152:
```suggestion
	for i := 0; i < len(subSlabs); i++ {
```

pkg/migration-tool/planner/slab.go
```diff
@@ -118,6 +144,48 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
...
+		ts := subSlabs[i].Result.Timeseries
+		for _, series := range ts {
+			if s, ok := timeseries[getLabelsStr(series.Labels)]; len(series.Samples) > 0 && ok {
```
cevian, line 155: I would have a variable for  getLabelsStr(series.Labels) as that's an expensive function you don't want to call multiple time

pkg/migration-tool/planner/slab.go
```diff
@@ -118,6 +144,48 @@ func (s *Slab) Fetch(ctx context.Context, client *utils.Client, mint, maxt int64
...
+				l := len(s.Samples)
+				if len(s.Samples) > 1 && s.Samples[l-1].Timestamp == series.Samples[0].Timestamp {
+					// Edge case: When concurrent pulls makes the remote-storage send duplicate samples that was sent
```
cevian, line 158: This branch should be an error.

cevian (approved):
Approving if you make the overlapping time case an error. If you disagree, let's discuss

pkg/migration-tool/planner/slab.go
Harkishen-Singh: Why should it be an error? Do you mean to error and continue with that slab, or error and stop migration?

pkg/migration-tool/planner/slab.go
Harkishen-Singh: I am sorry, that was done unknowingly.

pkg/migration-tool/planner/slab.go
Harkishen-Singh: I have changed this to `numSubuSlabs` and now it helps in reading. Do you think it's trivial even now?

pkg/migration-tool/planner/slab.go
Harkishen-Singh: Do you mean as a cache? Or just for readability?

pkg/migration-tool/planner/slab.go
Harkishen-Singh: @cevian lets discuss about this.

pkg/migration-tool/planner/slab.go
Harkishen-Singh: Lets discuss this on meet

cc @cevian

PR #586, opened 2021-04-15
Fix cross platform docker

### LINKED ISSUE #582: ARM64 builds are packaged with an x86_64 executable
I'm trying to set up promscale on an arm64 platform. However it looks like arm64 docker images are shipped with an x86_64 promscale executable. Running the docker image results in exec format errors and the executable indeed seems to have the wrong arch:

```
$ docker run --rm -it \
    --entrypoint "/bin/cat" \
    timescale/promscale:0.3.0-arm64 \
    promscale \
    > promscale_executable

$ file promscale_executable 
promscale_executable: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```

cevian:
Fixes #582

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,13 +5,13 @@ COPY ./pkg promscale/pkg
 COPY ./cmd promscale/cmd
 COPY ./go.mod promscale/go.mod
 COPY ./go.sum promscale/go.sum
-ARG TARGETOS=linux
-ARG TARGETARCH=amd64
+ARG TARGETOS
+ARG TARGETARCH
 RUN apk update && apk add --no-cache git \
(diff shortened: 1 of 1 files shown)


Harkishen-Singh (approved):
I have't played much with builds, but should be good as per my guess.
Can we have a parallel CI that can build on multiple distros and do a basic run on them? (See what prom does [here](https://app.circleci.com/pipelines/github/prometheus/prometheus/11755/workflows/25d6d866-3d4d-4333-b562-32ebd081d535/jobs/52076/parallel-runs/0?filterBy=ALL), plus we will also run on that platform with a basic test). This will help avoid the problem in any future release.

PR #587, opened 2021-04-15
Fix permissions system

cevian:
3 high level roles
- admin role to install and migrate schema and own all objects
- prom_writer
- prom_reader

Prefer SECURITY DEFINER functions to more permissions
Alter Default priviliges is kinda broken, so avoid.

We now assume timescaledb functions installed in public schema.


pkg/tests/end_to_end_tests/golden_files_test.go
```diff
@@ -104,7 +105,12 @@ func TestSQLGoldenFiles(t *testing.T) {
 			}
 
 			if string(expected) != string(actual) {
-				t.Fatalf("Golden file does not match result: diff\nexpected\n%s\nactual\n%s\n", expected, actual)
+				dmp := diffmatchpatch.New()
+				fileAdmp, fileBdmp, dmpStrings := dmp.DiffLinesToChars(string(expected), string(actual))
+				diffs := dmp.DiffMain(fileAdmp, fileBdmp, false)
+				diffs = dmp.DiffCharsToLines(diffs, dmpStrings)
+				diffs = dmp.DiffCleanupSemantic(diffs)
+				t.Errorf("Golden file does not match result: diff %s %s\n\n%v", expectedFile, actualFile, dmp.DiffPrettyText(diffs))
```
antekresic, line 113: 🔥

scripts/end_to_end_tests.sh
```diff
@@ -174,12 +174,13 @@ compare_connector_and_prom "series?match%5B%5D=ts_prom_sent_samples_total"
 
 # Labels endpoint cannot be compared to Prometheus becuase it will always differ due to direct backfilling of the real dataset.
 # We have to compare it to the correct expected output. Note that `namespace` and `node` labels are from JSON import payload.
-EXPECTED_OUTPUT='{"status":"success","data":["__name__","code","handler","instance","job","le","method","mode","namespace","node","path","quantile","status","version"]}'
+EXPECTED_OUTPUT1='{"status":"success","data":["__name__","code","handler","instance","job","le","method","mode","namespace","node","path","quantile","status","version"]}'
+EXPECTED_OUTPUT2='{"status":"success","data":["__name__","code","handler","instance","job","le","method","mode","namespace","node","path","quantile","status"]}'
```
antekresic, line 178: Why two outputs now?

antekresic (approved):
Looks good to me. 

Have one question which might need a comment but should be good to merge.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1952,13 +2011,20 @@ BEGIN
         END IF;
 END
 $func$
-LANGUAGE PLPGSQL STABLE;
+LANGUAGE PLPGSQL STABLE
+SECURITY DEFINER
+--search path must be set for security definer
+--need to include public(the timescaledb schema) for some timescale functions to work.
+SET search_path = public, pg_temp;
```
davidkohn88, line 2018: The timescale functions don't schema qualify so we need public in the search path? this seems like a bug that we should report upstream?

pkg/migrations/sql/preinstall/000-utils.sql
```diff
@@ -13,11 +13,12 @@
 CREATE OR REPLACE PROCEDURE execute_everywhere(command_key text, command TEXT, transactional BOOLEAN = true)
```
davidkohn88, line 13: I'm not quite sure what this function does, but it scares the crap out of me...

davidkohn88: (from a perms/safety perspective) if it is safe, I'd say we need a comment about why it is safe etc. And that table needs to have write permissions revoked from the writer user most definitely, cause otherwise, the writer could execute arbitrary stuff on new data nodes)

pkg/tests/testdata/expected/views.out
```diff
@@ -49,6 +49,8 @@ SELECT count(compress_chunk(i)) from show_chunks('prom_data.cpu_usage') i;
 (1 row)
 
 \set ON_ERROR_STOP 0
+SET role prom_reader;
```
davidkohn88, line 52: So we connect as prom_writer by default and then set our role to prom_reader right?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -126,6 +126,24 @@ GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.unlock_metric_for_maintenance(int) TO p
 --is insignificant in practice.
 --
 --lock-order: metric table, data_table, series parent, series partition
```
davidkohn88, line 128: It looks like this comment was meant for the next function and you inserted a new function in here?

davidkohn88 (commented):
It seems like it wouldn't be crazy hard to make "timescale_schema" a configurable variable that you just replaced everywhere much like SCHEMA_CATALOG or whatever else (incl in search_path setting from the function definition) but I might be wrong. I might do that just to deal with those odd cases, default it to public, but it might actually be safer that way? Might need a grant for the admin user in the setup? 

Anyway, otherwise looks pretty good, a few small-ish things, but I think overall the approach makes sense and the changes here aren't all that dramatic.

pkg/migrations/sql/preinstall/006-tables_ha.sql
```diff
@@ -14,6 +16,8 @@ CREATE TABLE SCHEMA_CATALOG.ha_leases_logs
     lease_until  TIMESTAMPTZ,          -- exclusive
     PRIMARY KEY (cluster_name, leader_name, lease_start)
 );
+GRANT SELECT ON TABLE SCHEMA_CATALOG.ha_leases_logs TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE SCHEMA_CATALOG.ha_leases_logs TO prom_writer;
```
davidkohn88, line 20: Should this all be done with a function cause otherwise someone with writer perms could pretend to be writing when they're not, bringing down the system?

davidkohn88: But I guess if you're lying about that you could lie in the function too...

davidkohn88: (this could be done as the admin in a separate connection?)

pkg/migrations/sql/versions/dev/0.3.1-dev/1-fix_permissions.sql
```diff
@@ -1,4 +1,6 @@
-CREATE OR REPLACE PROCEDURE execute_everywhere(command_key text, command TEXT, transactional BOOLEAN = true)
+DROP PROCEDURE execute_everywhere(text, TEXT, BOOLEAN);
```
davidkohn88, line 1: Need if exists just in case?

davidkohn88: (I'm not sure why we need to drop rather than create or replace here?

davidkohn88 (approved):
Looks good to me, much better on the execute everywhere stuff. A couple small comments again, but nothing major.

The sequences aren't a problem to do with ALL if you want to simplify that part...

pkg/migrations/sql/preinstall/006-tables_ha.sql
cevian: Yeah I don't think our security model is availability safe to nefarious prom_writers :)

davidkohn88: fair enough :)

pkg/migrations/sql/versions/dev/0.3.1-dev/1-fix_permissions.sql
cevian: will add the if exists. We are moving to a different schema which is why we drop the old

pkg/migrations/sql/preinstall/006-tables_ha.sql
davidkohn88: And I suppose the reader role is the one that you'd be handing out DB users in anyway. Fair enough.

davidkohn88: (Some folks do have a deleter role that's switched to for those things, and that would do this sort of stuff, but seems like overkill)

VineethReddy02 (approved):

PR #589, opened 2021-04-16
Add support to reject samples if they fall on compressed chunks.

Harkishen-Singh:

### DIFF
--- a/pkg/pgmodel/ingestor/copier.go
+++ b/pkg/pgmodel/ingestor/copier.go
@@ -29,7 +29,10 @@ type copyRequest struct {
 	table string
 }
 
-var getBatchMutex = &sync.Mutex{}
+var (
+	getBatchMutex       = &sync.Mutex{}
+	handleDecompression = retryAfterDecompression
+)
 
 // Handles actual insertion into the DB.
 // We have one of these per connection reserved for insertion.
@@ -127,23 +130,12 @@ func insertBatchErrorFallback(conn pgxconn.PgxConn, reqs ...copyRequest) {
 		reqs[i].data.batch.ResetPosition()
 		err := doInsert(conn, reqs[i])
 		if err != nil {
-			err = insertErrorFallback(conn, err, reqs[i])
+			err = tryRecovery(conn, err, reqs[i])
 		}
 
 		reqs[i].data.reportResults(err)
 		reqs[i].data.release()
 	}
-}
-
-// certain errors are recoverable, handle those we can
-//   1. if the table is compressed, decompress and retry the insertion
-func insertErrorFallback(conn pgxconn.PgxConn, err error, req copyRequest) error {
-	err = tryRecovery(conn, err, req)
-	if err != nil {
-		log.Warn("msg", fmt.Sprintf("time out while processing error for %s", req.table), "error", err.Error())
-		return err
-	}
-	return doInsert(conn, req)
 }
 
 // we can currently recover from one error:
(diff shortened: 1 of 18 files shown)


antekresic (changes requested):
This change disrupts too many function signatures, we should come up with a way to setup this configuration by using different recovery functions. 

Best way would be to compose it so that we call two different `runInserter` functions.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -41,10 +42,10 @@ func NewPgxIngestor(conn pgxconn.PgxConn, cache cache.MetricCache, scache cache.
 
 // NewPgxIngestorForTests returns a new Ingestor that write to PostgreSQL using PGX
 // with an empty config, a new default size metrics cache and a non-ha-aware data parser
-func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
+func NewPgxIngestorForTests(conn pgxconn.PgxConn, decompressChunks bool) (*DBIngestor, error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: I feel a config struct for just a bool seems odd. Unless this becomes too long, abstracting with a struct may not be needed. WDYT?

pkg/pgmodel/ingestor/copier.go
```diff
@@ -33,7 +33,7 @@ var getBatchMutex = &sync.Mutex{}
 
 // Handles actual insertion into the DB.
 // We have one of these per connection reserved for insertion.
-func runCopier(conn pgxconn.PgxConn, in chan copyRequest) {
+func runCopier(conn pgxconn.PgxConn, decompressChunks bool, in chan copyRequest) {
```
cevian, line 36: hmm. Why don't we pass down ingestor.Cfg instead and add this as a field to that struct?

pkg/pgmodel/ingestor/copier.go
```diff
@@ -161,6 +168,9 @@ func tryRecovery(conn pgxconn.PgxConn, err error, req copyRequest) error {
 
 	// If the error was that the table is already compressed, decompress and try again.
 	if strings.Contains(pgErr.Message, "insert/update/delete not permitted") {
+		if !decompressChunk {
+			return ErrCompressedChunk
```
cevian, line 172: hmm feels a bit weird to return the error just to catch it one step up and interpret decompressChunk in multiple places. Why don't we instead change tryRecovery to return (retryInsert bool, err error) and return false, nil here?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -39,12 +40,18 @@ func NewPgxIngestor(conn pgxconn.PgxConn, cache cache.MetricCache, sCache cache.
 	return &DBIngestor{dispatcher: dispatcher, sCache: sCache, parser: parser}, nil
 }
 
+// IngestorTestConfig is a config for NewPgxIngestorForTests.
+type IngestorTestConfig struct {
+	Conn             pgxconn.PgxConn
+	DecompressChunks bool
```
cevian, line 46: This should be IgnoreCompressedSamples bool instead so that the default is the zero-value(false). That way we can avoid DecompressChunks: true all over the place.

cevian (changes requested):

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: I had thought about that, but felt that decompressChunks is easier to read and suits our case.

pkg/pgmodel/ingestor/copier.go
Harkishen-Singh: We already have decompressChunks in the struct. I assume you mean to pass cfg just on this level.

antekresic:
As discussed with @Harkishen-Singh, we are going with a handler approach with will be called when you get a "chunk compressed" error and setting it to a noop handler if you disable decompression on insert.

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -58,6 +58,11 @@ func newPgxDispatcher(conn pgxconn.PgxConn, cache cache.MetricCache, scache cach
 	toCopiers := make(chan copyRequest, copierCap)
 	setCopierChannelToMonitor(toCopiers)
 
+	if cfg.IgnoreCompressedChunks {
+		// Handle decompression to not decompress anything.
+		handleDecompression = func(_ pgxconn.PgxConn, _ copyRequest) error { return nil }
```
antekresic, line 63: I think we should add a log here, to let the user know we are not ingesting old data. Which means this handler should be declared elsewhere, not inlined.

antekresic (approved):
Looks good, one comment to add some logging. After that should be good to merge.

cevian (commented):
I'll let Ante review this

Harkishen-Singh:
> I'll let Ante review this

@cevian would you mind cancelling the requested changes request as its blocking from a merge.

pkg/tests/end_to_end_tests/insert_compressed_chunks_test.go
```diff
@@ -0,0 +1,88 @@
...
+		ingestor, err := ingstr.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), nil)
+		require.NoError(t, err)
+		defer ingestor.Close()
+		_, err = ingestor.Ingest(newWriteRequestWithTs(copyMetrics(ts)))
+		require.NoError(t, err)
+		err = ingestor.CompleteMetricCreation()
+		if err != nil {
+			t.Fatal(err)
+		}
+		_, err = db.Exec(context.Background(), "SELECT compress_chunk(i) from show_chunks('prom_data.\"firstMetric\"') i;")
```
antekresic, line 45: You should run this only when running with `useTimescaleDB` flag.

pkg/tests/end_to_end_tests/insert_compressed_chunks_test.go
```diff
@@ -0,0 +1,88 @@
...
+		ingestor, err := ingstr.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), &ingstr.Cfg{IgnoreCompressedChunks: true})
+		require.NoError(t, err)
+		defer ingestor.Close()
+		_, err = ingestor.Ingest(newWriteRequestWithTs(copyMetrics(ts)))
+		require.NoError(t, err)
+		err = ingestor.CompleteMetricCreation()
+		if err != nil {
+			t.Fatal(err)
+		}
+		_, err = db.Exec(context.Background(), "SELECT compress_chunk(i) from show_chunks('prom_data.\"firstMetric\"') i;")
```
antekresic, line 74: Only with `useTimescaleDB`.

antekresic (changes requested):
Logging stuff needs a bit more refactoring.

antekresic (approved):

Harkishen-Singh:
Do not merge.

Harkishen-Singh:
Now mergeable

PR #593, opened 2021-04-19
Improve locking on series and labels

cevian:


antekresic (approved):
I don't like that slicing an existing `TextArray` needs to return an error but thats the reality here. 

I suggest returning the error along with the new `TextArray` in `Slice` method and handling it.

Should be good to merge after that.

PR #594, opened 2021-04-20
Add debugging metrics and log max recv timestamp.

Harkishen-Singh:

This PR adds debugging metrics along with printing out max received timestamp in the log along with `samples/sec`.


pkg/pgmodel/ingestor/insert.go
```diff
@@ -381,6 +388,13 @@ func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
 	}
 	defer results.Close()
 
+	ts := maxtBatch.UnixNano()
+	ts /= 1000000 // Convert nano-secs to milli-secs.
```
cevian, line 392: should we use the model.Time conversion functions here and elsewhere in the code?

pkg/pgmodel/ingestor/metrics.go
```diff
@@ -89,6 +84,13 @@ var (
 		},
 		func() float64 { return float64(len(CopierChannelToMonitor)) },
 	)
+	currentIngestions = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Name:      "current_ingestions",
```
cevian, line 90: should this be write_requests_processing to be more understandable? And should this then be measured from write.go? Should we have both?

cevian (changes requested):

pkg/api/write.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/timescale/promscale/pkg/util"
 )
 
+var maxRecvTimestamp = int64(0)
```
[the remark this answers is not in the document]
Harkishen-Singh: If i do this as math.MinInt64 then the code will be

```golang
if prev := atomic.LoadInt64(&maxRecvTimestamp); maxt > prev {
	atomic.StoreInt64(&maxRecvTimestamp, maxt)
	if prev != math.MinInt64 {
		metrics.MaxRecvTimestamp.Add(float64(maxt - prev))
	} else {
		metrics.MaxRecvTimestamp.Add(float64(maxt))
	}
}
```.

But if we make this 0 (as it is), we just need to

```golang
if prev := atomic.LoadInt64(&maxRecvTimestamp); maxt > prev {
	atomic.StoreInt64(&maxRecvTimestamp, maxt)
	metrics.MaxRecvTimestamp.Add(float64(maxt - prev))
}
```

pkg/api/write.go
```diff
@@ -76,7 +78,7 @@ func Write(writer ingestor.DBInserter, elector *util.Elector, metrics *Metrics)
 		metrics.ReceivedSamples.Add(float64(receivedBatchCount))
 		begin := time.Now()
 
-		numSamples, err := writer.Ingest(timeseries, req)
+		numSamples, maxt, err := writer.Ingest(timeseries, req)
```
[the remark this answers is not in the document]
Harkishen-Singh: Why is the maxt not the highest received timestamp? 

```
for _, si := range data {
	numRows += uint64(si.CountSamples())
	// Since by default, samples are expected to arrive in sorted order with time,
	// the last sample should be the maxt of that series. This assumption helps avoid costly inner loops.
	ls := si.LastSample()
	if maxt < ls.Timestamp {
		maxt = ls.Timestamp
	}
}
```

This is the code for maxt.

For the highest timestamp sent to db, we have `promscale_max_sent_timestamp_milliseconds`. That code lies in line 393 in `ingestor/insert.go`

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -52,7 +52,9 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
 //     tts the []Timeseries to insert
 //     req the WriteRequest backing tts. It will be added to our WriteRequest
 //         pool when it is no longer needed.
-func (ingestor *DBIngestor) Ingest(tts []prompb.TimeSeries, req *prompb.WriteRequest) (uint64, error) {
+func (ingestor *DBIngestor) Ingest(tts []prompb.TimeSeries, req *prompb.WriteRequest) (uint64, int64, error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: If we go by the second approach, then we don't need to return the maxt. Updated accordingly.

pkg/pgmodel/ingestor/metrics.go
Harkishen-Singh: > should this then be measured from write.go

active write requests are part of ingestor. We can declare it here and use it in write.go for better accuracy.

> Should we have both?

Do you mean, one metric over here and another metric in api/write.go? I feel that becomes trivial since that's not our aim here. WDYT?

pkg/api/write.go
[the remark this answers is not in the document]
Harkishen-Singh: Ah got it. In that case, we should keep this (with the updates to names) and remove the ones in insert.go

cevian (changes requested):
The big problem I think this PR exposes is that we have unnecessary complexity in throughput tracking between the sync and async cases. While this PR did not create the issue, I think it should solve it. What do you think of moving the sync tracking into dispatcher where the async stuff lives and making both based on the insertedDataPoints variable in there?

antekresic (changes requested):
I don't really like how we have two different methods of reporting throughput. Seems to me we can unify those regardless if we are using async ack or not.

That would probably be a better change for this.

pkg/pgmodel/ingestor/insert.go
```diff
@@ -320,11 +320,13 @@ func debugInsert() {
 
 // Perform the actual insertion into the DB.
 func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
-	batch := conn.NewBatch()
+	var (
```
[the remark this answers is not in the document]
Harkishen-Singh: I think it looks cleaner and the reason why go supports multiple declarations. Though, this is ssubjective.

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -228,15 +228,24 @@ func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, err
...
 	}
 
 	return numRows, err
 }
 
+func updateMaxSentTs(ts int64) {
+	prev := atomic.LoadInt64(&MaxSentTimestamp)
+	if ts < prev {
+		return
+	}
```
antekresic, line 244:
```suggestion
	if ts < atomic.LoadInt64(&MaxSentTimestamp) {
		return
	}
```

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -195,8 +184,11 @@ func (p *pgxDispatcher) Close() {
 // Though we may insert data to multiple tables concurrently, if asyncAcks is
 // unset this function will wait until _all_ the insert attempts have completed.
 func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, error) {
-	var numRows uint64
-	workFinished := &sync.WaitGroup{}
+	var (
+		numRows      uint64
+		maxt         = int64(math.MinInt64)
```
antekresic, line 189: Why can't this be zero?

pkg/util/throughput-cal/throughput-cal.go
```diff
@@ -0,0 +1,58 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package throughput_cal
```
antekresic, line 5: This is not an idiomatic way to name a package. Take a look at this: 
https://golang.org/doc/effective_go#package-names

You are renaming the package import everywhere, why not call it `throughput`?

Also, one file packages can cause a lot of stutter in the code if not careful.
Examples: 
`throughput.InitThroughputWatcher` vs. `throughput.InitWatcher`
`throughput.NewThroughputCal` vs. `throughput.NewCalc`

pkg/util/throughput-cal/throughput-cal.go
```diff
@@ -0,0 +1,58 @@
...
+		// Watching throughput is disabled.
+		return
+	}
+	watcher.Do(func() {
+		throughputWatcher = NewThroughputCal(calculateEvery)
+		go throughputWatcher.run()
+	})
+}
+
+type throughputCal struct {
```
antekresic, line 32: Should use `Calc` abbreviation, `Cal` makes you think of Calendar.

antekresic (changes requested):
Great improvements, love it :+1: 

Once the package naming issue is sorted out and some minor nits addressed, should be good to go.

antekresic (approved):

pkg/internal/testhelpers/containers.go
```diff
@@ -163,6 +164,8 @@ func WithDB(t testing.TB, DBName string, superuser SuperuserStatus, deferNode2Se
 	f(db, t, PgConnectURL(DBName, superuser))
 }
 
+var connMux sync.Mutex // Needs a mux lock at GetReadOnlyConnection otherwise a DATA RACE warning from go.
```
cevian, line 167: This makes no sense. Can you provide the data race output?

pkg/util/throughput/throughput.go
```diff
@@ -0,0 +1,58 @@
...
+		go throughputWatcher.run()
+	})
+}
+
+type throughputCalc struct {
+	every time.Duration
+	cal   *ewma.Rate
+}
+
+func NewCal(every time.Duration) *throughputCalc {
```
cevian, line 37: does not need to be public
```suggestion
func newThroughputCal(every time.Duration) *throughputCalc {
```

pkg/util/throughput/throughput.go
```diff
@@ -0,0 +1,58 @@
...
+		rate := tc.cal.Rate()
+		tc.cal.Tick()
+		if rate == 0 {
+			continue
+		}
+		log.Info("msg", "Samples write throughput", "samples/sec", int(rate))
+	}
+}
+
+func RegisterSize(size uint64) {
```
cevian, line 56:
```suggestion
func ReportSamplesProcessed(num_samples uint64) {
```

pkg/util/util_test.go
```diff
@@ -31,16 +31,16 @@ func init() {
 }
 
 func TestThroughputCalc(t *testing.T) {
```
cevian, line 33: Can't we delete both this test and the util.ThroughputCalc struct? I don't think it's ever used anymore?

pkg/util/throughput/throughput.go
```diff
@@ -0,0 +1,58 @@
...
+		every: every,
+		cal:   ewma.NewEWMARate(1, every),
+	}
+}
+
+func (tc *throughputCalc) run() {
+	t := time.NewTicker(tc.every)
+	for range t.C {
+		rate := tc.cal.Rate()
+		tc.cal.Tick()
```
cevian, line 48: don't you tick before you get the rate?
```suggestion
                	tc.cal.Tick()
		rate := tc.cal.Rate()
```

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -45,6 +47,10 @@ var (
 	baseExtensionState testhelpers.ExtensionState
 )
 
+func init() {
+	tput.InitWatcher(time.Second)
```
cevian, line 51: why not tput.InitWatcher(0) here? Can you add comment?

pkg/api/write.go
```diff
@@ -187,7 +185,7 @@ func decodeSnappy(w http.ResponseWriter, r *http.Request, m *Metrics) bool {
 	return true
 }
 
-func ingest(inserter ingestor.DBInserter, dataParser *parser.DefaultParser) func(http.ResponseWriter, *http.Request, *Metrics) bool {
+func ingest(inserter ingestor.DBInserter, dataParser *parser.DefaultParser, metrics *Metrics) func(http.ResponseWriter, *http.Request, *Metrics) bool {
```
cevian, line 188: do we want to remove the metrics argument from the returned function then? I'm confused why this change was needed but we shouldn't have metrics passed in both ingest and the returned function

```suggestion
func ingest(inserter ingestor.DBInserter, dataParser *parser.DefaultParser, metrics *Metrics) func(http.ResponseWriter, *http.Request) bool {
```

pkg/pgmodel/ingestor/metrics.go
```diff
@@ -89,6 +84,13 @@ var (
 		},
 		func() float64 { return float64(len(CopierChannelToMonitor)) },
 	)
+	ActiveWriteRequests = prometheus.NewGauge(
```
cevian, line 87: This is no longer set anywhere.

cevian (changes requested):
Looksd mostly good but some comments

pkg/tests/upgrade_tests/upgrade_test.go
Harkishen-Singh: Its to show throughput in upgrade tests, but yes we can disable it here.

pkg/api/write.go
Harkishen-Singh: The reason metrics is returned to satisfy the return type of `writeStage`. But, since the `var metrics *Metrics` is a singleton in `api` package, we can get rid of these _passing of metrics address_ and use the singleton itself.

pkg/internal/testhelpers/containers.go
Harkishen-Singh:
```shell
level=debug ts=2021-06-09T10:20:10.287Z caller=query_engine.go:35 msg="Lookback delta is zero, setting to default value" value=5m0s
=== RUN   TestPromQLSeriesEndpoint/test_series_endpoint
level=info ts=2021-06-09T10:20:10.290Z caller=throughput.go:52 msg="Samples write throughput" samples/sec=15003
==================
WARNING: DATA RACE
Write at 0x00c006e0e2b0 by goroutine 350:
  github.com/timescale/promscale/pkg/internal/testhelpers.GetReadOnlyConnection.func1()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/internal/testhelpers/containers.go:180 +0xd9
  github.com/jackc/pgx/v4/pgxpool.ConnectConfig.func1()
      /home/hsingh/go/pkg/mod/github.com/jackc/pgx/v4@v4.8.0/pgxpool/pool.go:176 +0x346
  github.com/jackc/puddle.(*Pool).constructResourceValue()
      /home/hsingh/go/pkg/mod/github.com/jackc/puddle@v1.1.1/pool.go:470 +0x743
  github.com/jackc/puddle.(*Pool).Acquire()
      /home/hsingh/go/pkg/mod/github.com/jackc/puddle@v1.1.1/pool.go:302 +0x704
  github.com/jackc/pgx/v4/pgxpool.(*Pool).Acquire()
      /home/hsingh/go/pkg/mod/github.com/jackc/pgx/v4@v4.8.0/pgxpool/pool.go:355 +0xab
  github.com/jackc/pgx/v4/pgxpool.(*Pool).Query()
      /home/hsingh/go/pkg/mod/github.com/jackc/pgx/v4@v4.8.0/pgxpool/pool.go:404 +0x6f
  github.com/timescale/promscale/pkg/pgxconn.(*connImpl).Query()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/pgxconn/pgx_conn.go:51 +0xc6
  github.com/timescale/promscale/pkg/pgmodel/querier.(*pgxQuerier).queryMultipleMetrics()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/pgmodel/querier/querier.go:216 +0x154
  github.com/timescale/promscale/pkg/pgmodel/querier.(*pgxQuerier).getResultRows()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/pgmodel/querier/querier.go:172 +0x55e
  github.com/timescale/promscale/pkg/pgmodel/querier.(*pgxQuerier).Select()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/pgmodel/querier/querier.go:75 +0xca
  github.com/timescale/promscale/pkg/query.querier.Select()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/query/queryable.go:57 +0x149
  github.com/timescale/promscale/pkg/query.(*querier).Select()
      <autogenerated>:1 +0x3c
  github.com/timescale/promscale/pkg/api.series.func1()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/api/series.go:79 +0x139c
  github.com/timescale/promscale/pkg/api.corsWrapper.func1()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/api/common.go:161 +0xab
  net/http.HandlerFunc.ServeHTTP()
      /home/hsingh/go/go/src/net/http/server.go:2069 +0x51
  github.com/NYTimes/gziphandler.GzipHandlerWithOpts.func1.1()
      /home/hsingh/go/pkg/mod/github.com/!n!y!times/gziphandler@v1.1.1/gzip.go:336 +0x45e
  net/http.HandlerFunc.ServeHTTP()
      /home/hsingh/go/go/src/net/http/server.go:2069 +0x51
  github.com/timescale/promscale/pkg/api.timeHandler.func1()
      /home/hsingh/go/src/github.com/timescale/promscale/pkg/api/router.go:150 +0xe5
  github.com/prometheus/common/route.(*Router).handle.func1()
      /home/hsingh/go/pkg/mod/github.com/prometheus/common@v0.21.0/route/route.go:83 +0x3d7
  github.com/julienschmidt/httprouter.(*Router).ServeHTTP()
      /home/hsingh/go/pkg/mod/github.com/julienschmidt/httprouter@v1.3.0/router.go:387 +0x1201
  github.com/prometheus/common/route.(*Router).ServeHTTP()
      /home/hsingh/go/pkg/mod/github.com/prometheus/common@v0.21.0/route/route.go:121 +0x69
  net/http.serverHandler.ServeHTTP()
      /home/hsingh/go/go/src/net/http/server.go:2887 +0xca
  net/http.(*conn).serve()
      /home/hsingh/go/go/src/net/http/server.go:1952 +0x87d
```

Harkishen-Singh: But, if we have a lock, no race happens.

Harkishen-Singh: In fact, I think this is after recent changes in postgres jack lib upstream. I got the error in `warning slow ingestion` PR as well. We don't see in our CIs as we don't apply `-race` flag in the master, which we should.
https://github.com/timescale/promscale/blob/66d8a94edf2078180d014c6a294c66ead632637d/.github/workflows/go.yml#L152

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -228,15 +226,27 @@ func (p *pgxDispatcher) InsertData(rows map[string][]model.Samples) (uint64, err
...
 			}
+			postIngestTasks(maxt, numRows)
 		}()
 	}
 
 	return numRows, err
 }
 
+// postIngestTasks performs a set of tasks that are due after ingesting series data.
+func postIngestTasks(ts, numRows int64) {
```
cevian, line 238: may be better in ingestor/metrics.go

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -29,6 +29,9 @@ type DBIngestor struct {
 	sCache     cache.SeriesCache
 }
 
+// MaxSentTimestamp is the max timestamp sent to the database.
+var MaxSentTimestamp = int64(0)
```
cevian, line 33: maybe put this into ingestor/metrics.go

cevian (approved):
Great work. two optional nits left

pkg/pgmodel/ingestor/dispatcher.go
Harkishen-Singh: Why will post ingest task go inside a metrics.go file?

PR #595, opened 2021-04-20
Use small datasets for non-extended unit-tests.

Harkishen-Singh:


antekresic (changes requested):
We should use the large dataset instead of small one.

I think we should do this for the `TestDeleteMetricSQLAPI` test in `zlast_test.go` file as well.

antekresic (approved):

PR #598, opened 2021-04-22
Fix regex anchor

cevian:
Previously, regex anchoring happened as ^str$, which broke when using
a pipe as it interpreted a|b as (^a)|(b$).  while what we want is
^(a|b)$. Anoher words the precedenc of operators broke. Now we found
the anchoring Prometheus does, and just used that.


Harkishen-Singh (approved):

Harkishen-Singh:
@cevian should we add 
https://groups.google.com/g/prometheus-users/c/kxRgVunbe20
As another case in unit tests?

PR #606, opened 2021-04-29
Refactor write pipeline to support multiple formats and preprocessors

antekresic:
Current implementation did not support adding new data formats and data
preprocessors needed to support upcoming new features planned. This change
makes the ingest pipeline easier to work with.


pkg/ha/filter_test.go
```diff
@@ -356,13 +380,6 @@ func TestHaParserParseData(t *testing.T) {
 				t.Fatalf("wanted error from FilterData, got nil")
 			}
```
cevian, line 382: we aren't testing c.proceed anymore are we missing any important cases?

pkg/api/common.go
```diff
@@ -101,6 +102,7 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	cfg.Auth = &Auth{}
 
 	fs.BoolVar(&cfg.ReadOnly, "read-only", false, "Read-only mode for the connector. Operations related to writing or updating the database are disallowed. It is used when pointing the connector to a TimescaleDB read replica.")
+	fs.BoolVar(&cfg.HighAvailability, "high-availability", false, "Enable external_labels based HA.")
```
cevian, line 105: let's leave a note about the cli rename in the draft release nnotes

pkg/api/write.go
```diff
@@ -15,188 +14,176 @@ import (
...
-			req = appendJSONPayload(req, i)
-		}
-	default:
-		// Cannot get here because of header validation.
-		return nil, fmt.Errorf("unsupported data format (not protobuf or json)"), "Write header validation error"
-	}
-
-	return req, nil, ""
+type writeHandler struct {
+	actors  []writeActor
```
cevian, line 158: nit: are these really actors as in the actor model? These are more like stages.

cevian (changes requested):
Looks great. Some changes/questions requested

pkg/api/write.go
```diff
@@ -15,188 +14,176 @@ import (
...
+		for _, ts := range req.Timeseries {
+			receivedBatchCount += uint64(len(ts.Samples))
 		}
 
-		metrics.ReceivedSamples.Add(float64(receivedBatchCount))
+		m.ReceivedSamples.Add(float64(receivedBatchCount))
 		begin := time.Now()
 
-		numSamples, err := writer.Ingest(timeseries, req)
+		numSamples, err := inserter.Ingest(req.Timeseries, req)
```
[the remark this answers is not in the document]
antekresic: Can we do this in a later PR? I wanted to do this but it will be a big change with all the tests.

pkg/api/router.go
```diff
@@ -14,20 +14,29 @@ import (
...
+	"github.com/timescale/promscale/pkg/ha"
+	haClient "github.com/timescale/promscale/pkg/ha/client"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgclient"
 	"github.com/timescale/promscale/pkg/query"
 	"github.com/timescale/promscale/pkg/util"
 )
 
 func GenerateRouter(apiConf *Config, metrics *Metrics, client *pgclient.Client, elector *util.Elector) (http.Handler, error) {
+	writePreprocessors := []parser.Preprocessor{}
```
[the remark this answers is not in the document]
antekresic: The reason is that parser is only for write so we create it in `Write`. We could pull it out if you think that makes more sense or reads better.

pkg/ha/filter_test.go
antekresic: proceed was removed because the signature changed. now we just empty the write request if we don't want to proceed. simplifies the interfaces and we were already checking for empty write requests.

pkg/api/write.go
antekresic: renaming to stages

pkg/api/parser/parser.go
```diff
@@ -0,0 +1,76 @@
...
+	if pre == nil {
+		return
+	}
+	p.preprocessors = append(p.preprocessors, pre)
+}
+
+// ParseRequest runs the correct parser on the format of the request and runs the
+// preprocessors on the payload afterwards.
+func (d DefaultParser) ParseRequest(r *http.Request, req *prompb.WriteRequest) error {
+	format := r.Header.Get("Content-Type")
```
Harkishen-Singh, line 48:
```suggestion
	requestType := r.Header.Get("Content-Type")
```
How is this?

pkg/api/parser/parser.go
```diff
@@ -0,0 +1,76 @@
...
+		err := p.Process(r, req)
+
+		if err != nil {
+			return err
+		}
+
+		if len(req.Timeseries) == 0 {
+			return nil
+		}
+	}
```
Harkishen-Singh, line 73: Do you think we should have a separate local func as `runPreprocessors()` for this?

pkg/api/router.go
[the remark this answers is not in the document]
Harkishen-Singh: Also, I think preprocessors are limited to the write path only, so they should be just inside the `write.go`.

pkg/api/write.go
Harkishen-Singh: I do support `stage`.

pkg/api/write.go
```diff
@@ -15,188 +14,187 @@ import (
...
+type writeHandler struct {
+	stages  []writeStage
+	metrics *Metrics
 }
 
-func isWriter(elector *util.Elector) (bool, error) {
-	if elector != nil {
-		shouldWrite, err := elector.IsLeader()
-		return shouldWrite, err
+func (wh *writeHandler) addStages(stages ...writeStage) {
```
Harkishen-Singh, line 168: I think we should make preprocessor as independent package, that has run as a main function in preprocessor interface and stage as individual stage. The run calls all the run() func in the stage.

```golang
type Stage interface {
  Run() (bool, error)
}

type Processor interface {
  AppendStage()
  Run() (bool, error) // Runs all stage.Run().
}
```

Also, writeStage -> stage? Or, better rename stage to step?

pkg/api/write.go
```diff
@@ -15,188 +14,187 @@ import (
...
-	if err != nil {
-		return nil, err, msg
-	}
-
-	return bytes.NewBuffer(reqBuf), nil, ""
+func (wh *writeHandler) handler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		for _, stage := range wh.stages {
+			if !stage(w, r, wh.metrics) {
+				return
```
Harkishen-Singh, line 181: I think we should also check for errors along with bool, and then report the error with the stage information. This will help understand the error better.

Harkishen-Singh (changes requested):
Awesome PR 👏🏻 !

Few minor points:
1. I think we should add validation of write requests (maybe in some other PR) to make sure metric and label names are in compliance with what's valid in PromQL?
2. I think we should use `require` instead of t.Fatal, etc, as they are more easier to follow and clean.

pkg/api/parser/json/json.go
```diff
@@ -0,0 +1,72 @@
...
+			return fmt.Errorf("invalid sample, should contain only timestamp and value: %v", tv)
+		}
+
+		t := int64(tv[0])
+
+		// Verify that timestamp doesn't contain fractions so we don't silently
+		// drop precision.
+		if float64(t) != tv[0] {
+			return fmt.Errorf("timestamp cannot contain decimal parts in sample: %v", tv)
+		}
```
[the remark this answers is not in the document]
antekresic: You cannot do this in validate since you need to unmarshal the data before you are able to check each value.

The docs state the exact specification of this format already,  this nothing new introduced in this PR:
https://github.com/timescale/promscale/blob/master/docs/writing_to_promscale.md#json-streaming-format

pkg/api/parser/json/json.go
```diff
@@ -0,0 +1,72 @@
...
+	for name, value := range ll {
+		result = append(result, prompb.Label{Name: name, Value: value})
+	}
+	return result
+}
+
+// ParseRequest parses an incoming HTTP request as a JSON payload.
+func ParseRequest(r *http.Request, wr *prompb.WriteRequest) error {
+	dec := json.NewDecoder(r.Body)
+	var i jsonPayload
```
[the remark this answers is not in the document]
antekresic: I don't prefer this change since its adding more lines to the code thus making it harder to read.

pkg/api/parser/json/json.go
```diff
@@ -0,0 +1,72 @@
+package json
```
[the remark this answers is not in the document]
antekresic: We don't have any validation like this on any format, its debatable if we should do it. 

In any case, not for this PR.

pkg/api/write.go
antekresic: I don't understand this comment. The stages handle errors themselves, there is no general way to handle them. They are the top level code which needs to log and handle based on the circumstance.

pkg/api/write.go
antekresic: I would leave this refactor for a different PR and only do it if we refactor both write and read pipelines to use the same interfaces.

pkg/api/parser/parser.go
antekresic: In this case, IMO its fine like it is just because the function is not big.

pkg/api/parser/parser.go
antekresic: Honestly, I think `format` is a better name for content type header. Its not really a request type, its more of a format of the request body.

pkg/api/common.go
antekresic: I think its too late for the experimental part now, it should have been a part of the previous release. Its going to go out of experimental by next release and adding it now doesn't make any sense.

pkg/api/parser/json/json_test.go
```diff
@@ -0,0 +1,148 @@
...
+			err := ParseRequest(req, response)
+
+			if c.responseError != "" {
+				errVal := "nil"
+				if err != nil {
+					errVal = err.Error()
+				}
+				if errVal != c.responseError {
+					t.Fatalf("unexpected error occured:\ngot\n%s\nwanted\n%s\n", errVal, c.responseError)
+				}
```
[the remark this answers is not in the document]
antekresic: Using `require` is something that up for debate. Its output is not as clear as the style used here, which is the style which 95% of the codebase uses.

pkg/api/common.go
cevian: yeah no experimental message

pkg/api/parser/json/json.go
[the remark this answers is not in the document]
Harkishen-Singh: Not for this PR, but we should do otherwise there can be issues while querying data.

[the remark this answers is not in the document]
cevian: I agree with Ante not for this PR

pkg/api/parser/json/json.go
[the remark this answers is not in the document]
Harkishen-Singh: > unmarshal the data before you are able to check each value.

I meant the validation of a sample.

pkg/api/parser/json/json_test.go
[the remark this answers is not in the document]
cevian: I think we should move to require in general but not for this PR

pkg/api/write.go
Harkishen-Singh: I am not sure but this can help increase readability. But if you think another PR is fine for this, let's do that way.

pkg/api/router.go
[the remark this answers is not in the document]
cevian: I see you pulled it out 👍

cevian (approved):

Harkishen-Singh (approved):

PR #607, opened 2021-04-30
Add Prometheus and open-metrics text format support to ingest pipeline

antekresic:


pkg/api/parser/parser.go
```diff
@@ -45,9 +49,9 @@ func (p *DefaultParser) AddPreprocessor(pre Preprocessor) {
...
 // preprocessors on the payload afterwards.
 func (d DefaultParser) ParseRequest(r *http.Request, req *prompb.WriteRequest) error {
-	format := r.Header.Get("Content-Type")
-	parser, ok := d.formatParsers[format]
-	if !ok {
+	mediaType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
+	parser, ok := d.formatParsers[mediaType]
+	if err != nil || !ok {
 		return fmt.Errorf("parser error: unsupported format")
 	}
```
cevian, line 56:
```suggestion
	mediaType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
	if err != nil  {
		return fmt.Errorf("parser error: unable to parse format: %w", err)
	}
	parser, ok := d.formatParsers[mediaType]
	if !ok {
		return fmt.Errorf("parser error: unsupported format")
	}
```

pkg/api/parser/text/text_test.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+func TestDBIngestorParseTextData(t *testing.T) {
+	defaultScrapeTime := time.Now()
+
+	timeProvider = func() time.Time {
+		return defaultScrapeTime
+	}
+
+	testCases := []struct {
```
cevian, line 24: please add 1 test cases:
- multiple metrics with at least one metric with multiple series in one input string

cevian (approved):
Looks good. 2 changes requested but no need for re-review if you agree with the changes.

PR #616, opened 2021-05-05
Cleanup of SQL permissions

cevian:
Some permission changes:
- Add prom_modifier role that allows you to modify (update/delete) metric data.
- Adjust prom_writer to not allow any modifications other than what's absolutely
  neceassary to ingest data.
- Add prom_admin role for all functions that modify configuration options.
- Add prom_maintenance role to execute background jobs.


antekresic (approved):
Looks good. 

As part of the cleanup, I suggest we get rid of `testhelpers.GetReadOnlyConnection` in e2e tests, it should probably be using `prom_reader` role now.

cevian:
@antekresic changed GetReadOnlyConnection to use prom_reader but kept the function as it also tests the readonly txn level

docs/sql_permissions.md
```diff
@@ -0,0 +1,82 @@
...
+- **prom_writer** is allowed to read Promscale data as well as write (insert) new data. This
+  role is not allowed to modify or delete data. Includes all permissions of prom_reader.
+- **prom_modifier** is able to read, write, and modify all Promscale data. Includes all permissions of prom_writer.
+- **prom_admin** is allowed to change the configuration options associated with Promscale. This
+  includes data retention policies, chunk intervals, etc. Includes all permissions of prom_writer.
+- **prom_maintenance** is allowed to execute maintenance tasks such as compression and data retention jobs.
+  mostly only used externally unless using a CRON job to execute_maintenance() instead of the jobs framework.
+  Includes all permissions of prom_reader.
+
+To run Promscale without schema migration you need to use a PostgreSQL user that has the `prom_modifier` role
```
davidkohn88, line 25: This is confusingly worded, partially because of lack of symmetry with the following line. Perhaps a table with the users running as starting with owner and what they can and cannot do would be useful? 

ie 
```
user , schema_migration, delete_series_api, write_api, read_api
owner   x.                            x                                   x             x
maint                                   x                                   x.            x
```

etc.?

davidkohn88: well that didn't work because whitespace, but yeah, you get the idea

docs/sql_permissions.md
```diff
@@ -0,0 +1,82 @@
...
+
+Promscale adopts a Role Based Access Control (RBAC) philosophy to managing permissions
+at the SQL layer. Many users will choose to simply use a single PostgreSQL user for all
+Promscale operations (using a single owner user as defined belows).
+But we also allow users more granular control over permissions management.
+
+Promscale defines several roles:
+- **Owner** this is the owner of the Promscale schema and objects. It is defined
+  as the user that originally installed/migrated Promscale. This user needs to be
+  able to install several PostgreSQL extensions and thus either needs to be a
```
davidkohn88, line 11: Maybe just link to the whitelist discussion below rather than state a similar thing here that distracts?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -250,7 +253,8 @@ BEGIN
         ) WITH (autovacuum_vacuum_threshold = 100, autovacuum_analyze_threshold = 100)
     $$, NEW.table_name, label_id, NEW.id);
    EXECUTE format('GRANT SELECT ON TABLE SCHEMA_DATA_SERIES.%I TO prom_reader', NEW.table_name);
-   EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE SCHEMA_DATA_SERIES.%I TO prom_writer', NEW.table_name);
+   EXECUTE format('GRANT SELECT, INSERT ON TABLE SCHEMA_DATA_SERIES.%I TO prom_writer', NEW.table_name);
```
davidkohn88, line 256: The set of perms is repeated all over the place...perhaps you could just make a function like `grant_promscale_default_perms(schema, table)`?

davidkohn88 (commented):
I've only gotten partway through, will look again in a bit, but there are a few things you could potentially work on now.

pkg/migrations/sql/idempotent/base.sql
cevian: only used in two places in same function. Don't think it's worth refactoring

davidkohn88: fair enough.

pkg/migrations/sql/versions/dev/0.3.1-dev/2-fix_more_permissions.sql
```diff
@@ -0,0 +1,67 @@
...
+    $$;
+$ee$);
+
+CALL SCHEMA_CATALOG.execute_everywhere('grant_prom_reader_prom_writer',$ee$
+    DO $$
+    BEGIN
+        GRANT prom_reader TO prom_writer;
+        GRANT prom_reader TO prom_maintenance;
+        GRANT prom_writer TO prom_modifier;
+        GRANT prom_modifier TO prom_admin;
```
davidkohn88, line 40: probably also want to grant maintenance to admin so that if you connect as admin you can set role prom_maintenance and see what's going on if there's a perms issue or something like that.

docs/sql_permissions.md
```diff
@@ -0,0 +1,89 @@
...
+`GRANT <role> to <user>`. A user can be assigned multiple roles. More information
+can be found in the [PostgreSQL docs about permissions](https://www.postgresql.org/docs/current/user-manag.html).
+
+
+Promscale can be run by the owner or a user with the `prom_modifier`, `prom_writer` or `prom_reader` role.
+The following table describes what features of promscale are available with each role:
+
+|user / role| schema migration | delete series api | write api | read api |
+| --- | --- | --- | --- | --- |
+| owner | ✅ | ✅ | ✅ | ✅|
```
davidkohn88, line 35: should prom_admin be in this table? Is it any different than modifier? Or is prom_admin the owner user? If so, why the two names? Probably want to add a note about what the admin user is for if it's not in this hierarchy.

davidkohn88 (approved):
LGTM, a couple small nits, but otherwise seems like good splitting of roles. I didn't audit every function to figure out if I think it's assigned to the right role, will just trust that that happened in a reasonable way. If there were ones you were on the fence about, it might be a good idea to just go back and add a comment about why it was assigned to the user it was.

docs/sql_permissions.md
davidkohn88: Found the note above, sorry. Missed that, that definitely covers that bit. I do think if you shouldn't run it as admin for some reason you should put a note here, but it looks like it would work...

Issue #597, opened 2021-04-21
Remote read with high-availability enabled


aburkevics:
Hello!
I'm trying to configure Prometheus high-availability setup with Promscale using the new setup described here - https://github.com/timescale/promscale/blob/master/docs/high-avaliability/prometheus-HA.md

I'm having an issue that when Promscale is started with `-enable-ha=true` flag, Prometheus doesn't seem to fetch the data from "remote_read" after re-deployment.

This is the related configuration in Prometheus that I use.
external labels:
```
  external_labels:
    __replica__: ha-prometheus-server-0
    cluster: ha-prometheus-poc
```
write:
```
remote_write:
- queue_config:
    batch_send_deadline: 10s
    capacity: 10000
    max_backoff: 10s
    max_samples_per_send: 3000
    max_shards: 200
    min_backoff: 100ms
    min_shards: 4
  remote_timeout: 30s
  url: http://promscale-connector:9201/write
```
read:
```
remote_read:
- read_recent: true
  url: http://promscale-connector:9201/read
```

For reference, I'm running Timescaledb docker image `timescale/timescaledb:2.2.0-pg12` & Promscale in kubernetes with image - `timescale/promscale:0.3.0`.

Currently for simplicity I'm running single nodes of Prometheus, Promscale & TimescaleDB.

I would expect that while Promscale & TimescaleDB are running and data is retained, I can wipe my Prometheus instance, bring it back up with the same config and it will read the historical data via Promscale. That is what I've tested without `-enable-ha=true` in Promscale and it was working.

If I had deployed promscale without HA before and later deploy it with it, then Prometheus will successfully read the metrics from before HA was enabled.

I see that Promscale is writing data to TimescaleDB when HA is enabled:
`level=info ts=2021-04-21T09:22:36.577Z caller=write.go:96 msg="Samples write throughput" samples/sec=12000`

Is there a way we should tell the Promscale read API to fetch metrics for specific `cluster` value?

VineethReddy02:
Hi!

So the samples ingested into Promscale will hold distinct label sets for different Prometheus clusters i.e. `cluster: <clustername>` so querying with cluster label set should let you query data specific to Prometheus cluster from Promscale.

aburkevics:
Hi!
Do you mean adding something like this to remote_read config?
```
     required_matchers:
       cluster: ha-prometheus-poc
```
Where "ha-prometheus-poc" is the cluster label value we write?

VineethReddy02:
[...]

PR #619, opened 2021-05-05
Fix new HA read path by dropping __replica__ labelset

VineethReddy02:
In the new HA setup read path fails when the read query comes from prometheus as promethues appends replica labelset from external labelset for readRead requests. So dropping the __replica__ labelSet in labelmatchers for read requests, when Promscale is in HA mode.


cevian (approved):

VineethReddy02:
fixes #597

PR #621, opened 2021-05-07
Add backfilling support for HA mode

antekresic:
In the case where Prometheus instances are lagging behind the leases, we
can use the past lease logs to backfill the old data thus not rejecting it,
causing more retries and more lag.


pkg/ha/filter.go
```diff
@@ -49,42 +51,125 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
-			if value.Name == ReplicaNameLabel && value.Value == replicaName {
+			if value.Name == ReplicaNameLabel {
 				t.Labels = append(t.Labels[:ind], t.Labels[ind+1:]...)
 				break
 			}
 		}
 	}
-
-	return nil
+	wr.Timeseries = wr.Timeseries[:numAccepted]
```
cevian, line 172: don't we need to clear the memory of the tail of wr.Timeseries for memory purposes. The equivalent of line 146-148. 
like 
```
		for j := numAccepted; j < len(wr.Timeseries); j++ {
			wr.Timeseries[j] = prompb.Timeseries{}
		}
```

pkg/ha/filter.go
```diff
@@ -49,42 +51,125 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
-			t.Samples = filterSamples(t.Samples, acceptedMinTUnix)
+	switch {
+	case !hasBackfill && allowInsert:
+		filterOutSampleRange(wr, minTUnix, toPromModelTime(leaseStart))
+	case hasBackfill && !allowInsert:
+		filterOutSampleRange(wr, toPromModelTime(leaseStart), maxTUnix+1) //maxTUnix+1 because filterSamples treats end as exclusive.
+	case !hasBackfill && !allowInsert:
+		wr.Timeseries = wr.Timeseries[:0]
+		log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert")
+	default:
```
cevian, line 81: nit: I think this would be more clear
```suggestion
	case hasBackfill && allowInsert:
```

pkg/ha/filter.go
```diff
@@ -49,42 +51,125 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
+		if !backfillStart.Before(maxT) {
+			break
+		}
+
+		keepRangeStart, keepRangeEnd, err := h.service.GetBackfillLeaseRange(backfillStart, maxT, cluster, replica)
+		if err != nil {
+			return false, fmt.Errorf("could not check backfill ha lease: %#v", err)
+		}
+		// Nothing else to keep.
+		if keepRangeStart.IsZero() || keepRangeEnd.IsZero() {
```
cevian, line 104: Nit: personally don't love testing with IsZero so would prefer the functions return an ok boolean to indicate whether the values are valid, but won't insist on this.

cevian (approved):
LGTM. Have mostly nits. Clearing the memory is the only serious comment. I'd like one other reviewer on this PR.

pkg/ha/client/client.go
```diff
@@ -86,16 +96,31 @@ func (l *leaseClientDB) UpdateLease(ctx context.Context, cluster, leader string,
...
 	}
 	return dbState, nil
 }
 
+func (l *leaseClientDB) GetPastLeaseInfo(ctx context.Context, cluster, replica string, start, end time.Time) (LeaseDBState, error) {
+	dbState := LeaseDBState{
+		Cluster: cluster,
+		Leader:  replica,
+	}
+	row := l.dbConn.QueryRow(ctx, getPastLeaseInfoSQL, cluster, replica, start, end)
```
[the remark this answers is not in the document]
antekresic: Yes, will update.

pkg/ha/filter.go
```diff
@@ -49,42 +51,125 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
+		//nothing to do
+	}
+	return nil
+}
+
+func toPromModelTime(t time.Time) int64 {
+	return int64(model.TimeFromUnixNano(t.UnixNano()))
+}
+
+func (h *Filter) filterBackfill(wr *prompb.WriteRequest, minT, maxT time.Time, cluster, replica string) (bool, error) {
```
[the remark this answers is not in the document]
antekresic: Yes, will update.

pkg/ha/filter.go
antekresic: Would you be OK with a comment for the default case? Saying that's the case it covers and why nothing needs to be done?

pkg/ha/filter.go
antekresic: I'm torn on this line too. While Go way is to make zero values useful, it does hurt readability in this case.

Will try to improve.

pkg/ha/filter.go
antekresic: This is done in the deferred function `filterLabels` which needs a rename since that's not all what it all does.

That function will called at the end of this one to remove HA replica labels and remove any `TimeSeries` which don't contain any samples.

Will changing the name to `finalFiltering` be good enough? Will add a comment what that actually means.

VineethReddy02 (approved):

pkg/ha/filter.go
```diff
@@ -49,42 +51,133 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
-			if value.Name == ReplicaNameLabel && value.Value == replicaName {
+			if value.Name == ReplicaNameLabel {
 				t.Labels = append(t.Labels[:ind], t.Labels[ind+1:]...)
 				break
 			}
 		}
 	}
-
-	return nil
+	wr.Timeseries = wr.Timeseries[:numAccepted]
```
VineethReddy02, line 180: Can we add debug log here as below to understand writes that are being accepted?  
```
log.Debug("msg", fmt.Sprintf("ingesting data from replica: %s", replicaName))
```

pkg/ha/filter.go
cevian: I am confused isn't this inside finalFiltering? What I mean here is how do all the timeseries after numAccepted get cleared?

pkg/ha/filter.go
antekresic: I'm not sure this is necessary. 

Not against it, just thinking if its too much logging and might add unnecessary noise making the logs hard to use.

pkg/ha/filter.go
antekresic: You are right, I don't know what I was thinking. Added it to the filtering function.

pkg/ha/filter.go
```diff
@@ -49,42 +51,136 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
...
+	for j := numAccepted; j < len(wr.Timeseries); j++ {
```
cevian, line 181: this loop has to come before the previous line `wr.Timeseries = wr.Timeseries[:numAccepted]` because that line sets len(wr.Timeseries)

cevian (changes requested):

pkg/ha/filter.go
cevian: Yes please add a comment

cevian (approved):

Issue #363, opened 2020-12-01
Implement metric metadata ingestion in catalog


Harkishen-Singh:
The recent PR (#6815) at prometheus upstream adds support for sending prometheus metric metadata (metric type, unit, help text) over remote write component (via promb, see integration PR: #361 ). We should implement the same in our catalog or in hypertables.

cc @cevian

franck102:
We would love to see this, we need the metric type to automate aggregation in Timescale.
We also need it to properly deal with counter resets for counters.

Currently we have a dedicated process looking up the metric types from Prometheus and injecting them into timescale.

Harkishen-Singh:
This issue is being fixed by #624

PR #624, opened 2021-05-13
Adds support for metric-metadata in Promscale.

Harkishen-Singh:

Fixes: #363 

This PR adds support for metric metadata sent by Prometheus into Promscale. The metadata belonging to metric are stored in `_prom_catalog.metadata` which has the following schema:

```shell
mt_test=# \d _prom_catalog.metadata
                                     Table "_prom_catalog.metadata"
    Column     |           Type           | Collation | Nullable |               Default                
---------------+--------------------------+-----------+----------+--------------------------------------
 id            | integer                  |           | not null | nextval('metadata_id_seq'::regclass)
 last_seen     | timestamp with time zone |           | not null | 
 metric_family | text                     |           | not null | 
 type          | text                     |           |          | 
 unit          | text                     |           |          | 
 help          | text                     |           |          | 
Indexes:
    "metadata_pkey" PRIMARY KEY, btree (id)
    "metadata_index" UNIQUE, btree (last_seen, metric_family)
    "metadata_metric_family_type_unit_help_key" UNIQUE CONSTRAINT, btree (metric_family, type, unit, help)


```

TODO:
- [ ] Docs related to metric metadata in another PR. Link the PR here.
- [ ] Update in draft notes. https://github.com/timescale/promscale/pull/624#discussion_r647743016


pkg/api/metadata.go
```diff
@@ -0,0 +1,100 @@
...
+				return
+			}
+		} else {
+			matcher, err := labels.NewMatcher(labels.MatchRegexp, "job", ".*")
+			if err != nil {
+				respondError(w, http.StatusInternalServerError, err, "creating all jobs matcher")
+				return
+			}
+			matchers = append(matchers, matcher)
+		}
```
antekresic, line 92: Instead of doing it like this, how about setting the `matchTarget` to something like `job=~".*"` in case its omitted?

Would clean up this part of the code quite a bit.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -52,12 +57,73 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+func (ingestor *DBIngestor) Ingest(r *prompb.WriteRequest) (numSamples uint64, numMetadata uint64, err error) {
+	defer func() {
+		// WriteRequests can contain pointers into the original buffer we deserialized
+		// them out of, and can be quite large in and of themselves. In order to prevent
+		// memory blowup, and to allow faster deserializing, we recycle the WriteRequest
+		// here, allowing it to be either garbage collected or reused for a new request.
+		// In order for this to work correctly, any data we wish to keep using (e.g.
+		// samples) must no longer be reachable from req.
+		FinishWriteRequest(r)
+	}()
```
antekresic, line 69: I'm not sure about this... I understand why you did it like this but we are not freeing the memory as soon as its not used like before. 

I believe there is a better way to support this by checking for samples and/or metadata. If either is missing, you can flush the write request. If both are present, just remove the samples or metadata from the request.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -52,12 +57,73 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+		n, err := ingestor.ingestSamples(r)
+		return n, 0, err
+	case numMeta > 0 && numTs == 0:
+		// Write request contains only metadata.
+		n, err := ingestor.ingestMetadata(r)
+		return 0, n, err
+	default:
+		// Write request contains both samples and metadata, hence we ingest concurrently.
+		type result struct {
+			id      int8
```
antekresic, line 84: This needs to be a type with constant values so its a bit clearer.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -52,12 +57,73 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+			numRows uint64
+			err     error
+		}
+		res := make(chan result, 2)
+		defer close(res)
+		go func() {
+			n, err := ingestor.ingestSamples(r)
+			res <- result{1, n, err}
```
antekresic, line 92: Don't like magic numbers like this and the below goroutine call.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -52,12 +57,73 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+		n, err := ingestor.ingestMetadata(r)
+		return 0, n, err
+	default:
+		// Write request contains both samples and metadata, hence we ingest concurrently.
+		type result struct {
+			id      int8
+			numRows uint64
+			err     error
+		}
+		res := make(chan result, 2)
```
antekresic, line 88: I don't think you need a buffered channel here. Best to avoid them unless there is a clear need for them.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -74,34 +140,46 @@ func (ingestor *DBIngestor) Ingest(r *prompb.WriteRequest) (uint64, error) {
...
 		// so nil the field
 		ts.Samples = nil
 	}
-	// WriteRequests can contain pointers into the original buffer we deserialized
-	// them out of, and can be quite large in and of themselves. In order to prevent
-	// memory blowup, and to allow faster deserializing, we recycle the WriteRequest
-	// here, allowing it to be either garbage collected or reused for a new request.
-	// In order for this to work correctly, any data we wish to keep using (e.g.
-	// samples) must no longer be reachable from req.
-	FinishWriteRequest(r)
```
antekresic, line 90: Memory needs to be cleared here, since the data is no longer used. Check previous comment for a suggestion.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -74,34 +140,46 @@ func (ingestor *DBIngestor) Ingest(r *prompb.WriteRequest) (uint64, error) {
...
+	for i := 0; i < metadataRows; i++ {
+		tmp := r.Metadata[i]
+		metadata[i] = model.Metadata{
+			MetricFamily: tmp.MetricFamilyName,
+			Unit:         tmp.Unit,
+			Type:         tmp.Type.String(),
+			Help:         tmp.Help,
+		}
+	}
+	rowsInserted, errMetadata := ingestor.metadataDispatcher.Insert(metadata)
```
antekresic, line 170: WriteRequest should be flushed before this point.

pkg/pgmodel/ingestor/metadata-dispatcher.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return insertedRows, err
+	}
+	if totalRows != insertedRows {
+		return insertedRows, fmt.Errorf("failed to insert all metadata: inserted %d rows out of %d rows in total", insertedRows, totalRows)
+	}
+	return insertedRows, nil
+}
+
+func insertMetadataWithBackoff(conn pgxconn.PgxConn, data []model.Metadata) (uint64, error) {
```
antekresic, line 50: This does not insert with backoff, it just sleeps for half of second before inserting. Also, does not retry like the comment below says.

Is there a reason for introducing retrying with backoff?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -24,19 +24,24 @@ type Cfg struct {
 
 // DBIngestor ingest the TimeSeries data into Timescale database.
 type DBIngestor struct {
-	dispatcher model.Dispatcher
-	sCache     cache.SeriesCache
+	sCache             cache.SeriesCache
+	samplesDispatcher  model.Dispatcher
+	metadataDispatcher model.Dispatcher
```
antekresic, line 29: I don't think metadata needs to conform to the `model.Dispatcher` interface. There is no clear reason to do it since the data is not comparable. Changing the interface to accept an empty interface removes the type safety without any gain.

Metadata needs a simple inserter (just like the one you made here) without implementing the dispatcher interface. I would prefer it if you changed `model.Dispatcher` by adding another method `InsertMetadata`. 

Then it would be simple to keep everything under `pgxDispatcher` and makes a lot more sense.

antekresic (changes requested):
Need to think over the logical structure of the code a bit more.

pkg/pgmodel/ingestor/metadata-dispatcher.go
Harkishen-Singh: > does not insert with backoff

I wrote it as backoff as what prometheus calls a backoff. 

> Is there a reason for introducing retrying with backoff

Network connections, interrupts, can be some

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -47,20 +49,91 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+	// samples) must no longer be reachable from req.
+	release := func() { FinishWriteRequest(r) }
+	switch numTs, numMeta := len(timeseries), len(metadata); {
+	case numTs == 0 && numMeta == 0:
+		release()
+		return 0, 0, nil
+	case numTs > 0 && numMeta == 0:
+		// Write request contains only time-series.
+		n, err := ingestor.ingestTimeseries(timeseries, release)
+		return n, 0, err
```
antekresic, line 81: Would put this check first since its the most frequent one.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -47,20 +49,91 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+	case numMeta > 0 && numTs == 0:
+		// Write request contains only metadata.
+		n, err := ingestor.ingestMetadata(metadata, release)
+		return 0, n, err
+	default:
+		release = func() {}
```
antekresic, line 87: Might be a good idea to leave the request cleanup to samples ingestion in this case, since metadata data should not be that big most cases and it would clean up the logic .

antekresic (changes requested):
First pass, will do one more once we rebase this on throughput calculator changes.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: Sorry, I do not get the point. Passing the cleanup in release callback I think looks cleaner so I kept it, plus we need to release on metadata. Metadata can be large especially if Prometheus is scraping kube-state-metrics target, as that contains many metrics and usually takes time to scrape by Prometheus server.

Harkishen-Singh: The recent commit push cleans up this part too.

pkg/pgmodel/ingestor/copier.go
```diff
@@ -311,3 +311,26 @@ func doInsert(conn pgxconn.PgxConn, reqs ...copyRequest) (err error) {
 	DbBatchInsertDuration.Observe(time.Since(start).Seconds())
 	return nil
 }
+
+func insertMetadata(conn pgxconn.PgxConn, reqs []pgmodel.Metadata) (insertedRows uint64, err error) {
+	batch := conn.NewBatch()
+	for _, m := range reqs {
+		batch.Queue("SELECT "+schema.Prom+".insert_metric_metadata($1::TIMESTAMPTZ, $2, $3, $4, $5)", time.Now().Local(), m.MetricFamily, m.Type, m.Unit, m.Help)
```
cevian, line 318: I think this should be time.Now() without the local?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -47,20 +49,90 @@ func NewPgxIngestorForTests(conn pgxconn.PgxConn) (*DBIngestor, error) {
...
+		timeseries = r.Timeseries
+		metadata   = r.Metadata
+	)
+	// WriteRequests can contain pointers into the original buffer we deserialized
+	// them out of, and can be quite large in and of themselves. In order to prevent
+	// memory blowup, and to allow faster deserializing, we recycle the WriteRequest
+	// here, allowing it to be either garbage collected or reused for a new request.
+	// In order for this to work correctly, any data we wish to keep using (e.g.
+	// samples) must no longer be reachable from req.
+	release := func() { FinishWriteRequest(r) }
```
cevian, line 73: There is no reason to do this kind of renaming. You should just call `FinishWriteRequest` when needed.

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+
+const (
+	metricMetadata = "SELECT metric_family, type, unit, help from " + schema.Catalog + ".metadata"
+	middle         = " ORDER BY last_seen"
+)
+
+func MetricMetadata(conn pgxconn.PgxConn, metric string, limit int) (map[string][]model.Metadata, error) {
+	query := metricMetadata
+	if metric != "" {
+		query += fmt.Sprintf(" WHERE metric_family='%s'", metric)
```
cevian, line 27: this is unsafe for sql injection. This should be passed down as a parameter

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	query += middle
```
cevian, line 29:
```suggestion
	query +=" ORDER BY last_seen"
```

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const (
+	metricMetadata = "SELECT metric_family, type, unit, help from " + schema.Catalog + ".metadata"
+	middle         = " ORDER BY last_seen"
+)
+
+func MetricMetadata(conn pgxconn.PgxConn, metric string, limit int) (map[string][]model.Metadata, error) {
+	query := metricMetadata
```
cevian, line 25:
```suggestion
	query := "SELECT metric_family, type, unit, help FROM " + schema.Catalog + ".metadata"
```

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,58 @@
...
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_targets(metric_family_name TEXT, series_ids BIGINT[]) RETURNS TABLE (series_id BIGINT, target_keys TEXT[], target_values TEXT[])
+AS
+$$
+DECLARE
+    metric_table_name TEXT;
+    q TEXT;
+BEGIN
+    SELECT table_name INTO metric_table_name FROM SCHEMA_CATALOG.metric WHERE metric_name = metric_family_name;
+    q = FORMAT('select s.id series_id, array_agg(l.key) target_keys, array_agg(l.value) target_values from SCHEMA_CATALOG.label l inner join SCHEMA_DATA_SERIES.%I s on (true) where l.id = any(s.labels) and (key = ''job'' or key = ''instance'') and s.id = any(%L::bigint[]) group by series_id;', metric_table_name, series_ids);
```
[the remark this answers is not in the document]
cevian: is the job and instance keys special somehow? Where is this documented?

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,58 @@
...
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_metric_metadata(metric_family_name TEXT)
+RETURNS TABLE (metric_family TEXT, type TEXT, unit TEXT, help TEXT)
+AS
+$$
+    SELECT metric_family, type, unit, help FROM SCHEMA_CATALOG.metadata WHERE metric_family = metric_family_name ORDER BY last_seen LIMIT 1
+$$ LANGUAGE SQL;
+GRANT EXECUTE ON FUNCTION SCHEMA_PROM.get_metric_metadata(TEXT) TO prom_writer;
+
+-- metric_families should have unique elements, otherwise there will be duplicate rows in the returned table.
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_multiple_metric_metadata(metric_families TEXT[])
```
cevian, line 26: is this ever used?

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,58 @@
...
+        INSERT INTO _prom_catalog.metadata (last_seen, metric_family, type, unit, help)  VALUES (t, metric_family_name, metric_type, metric_unit, metric_help)
+            ON CONFLICT (metric_family, type, unit, help) DO
+                UPDATE SET last_seen = t;
+        GET DIAGNOSTICS num_rows = ROW_COUNT;
+        RETURN num_rows;
+    END;
+$$ LANGUAGE plpgsql;
+GRANT EXECUTE ON FUNCTION SCHEMA_PROM.insert_metric_metadata(TIMESTAMPTZ, TEXT, TEXT, TEXT, TEXT) TO prom_writer;
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_metric_metadata(metric_family_name TEXT)
```
cevian, line 17: is this ever used?

pkg/migrations/sql/versions/dev/0.3.1-dev/3-metric_metadata_objects.sql
```diff
@@ -0,0 +1,15 @@
...
+    id SERIAL NOT NULL PRIMARY KEY,
+    last_seen TIMESTAMPTZ NOT NULL,
+    metric_family TEXT NOT NULL,
+    type TEXT DEFAULT NULL,
+    unit TEXT DEFAULT NULL,
+    help TEXT DEFAULT NULL,
+    UNIQUE (metric_family, type, unit, help)
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS metadata_index ON SCHEMA_CATALOG.metadata
```
cevian, line 12: this should not be unique

pkg/migrations/sql/versions/dev/0.3.1-dev/3-metric_metadata_objects.sql
```diff
@@ -0,0 +1,15 @@
...
+(
+    last_seen, metric_family
```
cevian, line 14:
```suggestion
    metric_family, last_seen 
```

pkg/migrations/sql/versions/dev/0.3.1-dev/3-metric_metadata_objects.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS SCHEMA_CATALOG.metadata
+(
+    id SERIAL NOT NULL PRIMARY KEY,
```
cevian, line 3: do we need an id here?

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+	if metric != "" {
+		query += fmt.Sprintf(" WHERE metric_family='%s'", metric)
+	}
+	query += middle
+	if limit > 0 {
+		query += fmt.Sprintf(" LIMIT %d", limit)
```
cevian, line 31: shouldn't this limit be on the number off metric families, not metadata entries?

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+	for i := range metrics {
+		if metrics[i] == metric {
+			return seriesIds[i]
+		}
+	}
+	// Corresponding series id does not exists.
+	return []model.SeriesID{}
+}
+
+func getTargetInfo(conn pgxconn.PgxConn, metrics []string, seriesIds [][]model.SeriesID) (map[string]map[Target]struct{}, error) {
```
cevian, line 163: I absoluteley do not understand what this does. Can we explain in a comment?

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,215 @@
...
+				}
+				metadataSlice = append(metadataSlice, TargetMetadataType{
+					Target:   j,
+					Metadata: meta,
+				})
+			}
+		}
+	} else {
+		var metricFamily, typ, unit, help string
+		row := conn.QueryRow(context.Background(), "SELECT metric_family, type, unit, help from "+schema.Catalog+".metadata WHERE metric_family=$1 order by last_seen desc limit 1", metric)
```
cevian, line 115: very hard to read

pkg/tests/end_to_end_tests/metadata_test.go
```diff
@@ -0,0 +1,202 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
cevian, line 1: these tests need to use the priviliges of prom_reader or prom_writer etc. You should use testhelpers.PgxPoolWithRole function as in the other end_to_end tests

cevian (changes requested):
Partial review. This still needs quite a lot of work.

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,58 @@
+CREATE OR REPLACE FUNCTION prom_api.insert_metric_metadata(t TIMESTAMPTZ, metric_family_name TEXT, metric_type TEXT, metric_unit TEXT, metric_help TEXT)
```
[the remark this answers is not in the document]
Harkishen-Singh: I think that will lead to removing return num_rows inserted which is required.

pkg/migrations/sql/idempotent/metric-metadata.sql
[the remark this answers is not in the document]
Harkishen-Singh: > is the job and instance keys special somehow?

When we query target, job and instance keys are only needed. This is done for a SQL API that comes in line with the HTTP API (in both Prometheus and Promscale).

>  Where is this documented?

I will add a doc in the docs PR that is focused on usability around metadata table and the SQL APIs we provide in line with Prometheus.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: Reason why `release()` is used than just calling `FinishWriteRequest` is, we need to clear mem:
1. Only insertMetadata()
2. Only insertSamples()
3. Both insertMetadata() & insertSamples()

In 1 & 2, its simple, just call the `FinishWriteRequest`. But if its 3, then wait for 1 and 2 to complete, and only then clear the request since this is done concurrently. If we hard code `FinishWriteRequest` then case 3 is not possible.

pkg/pgmodel/metadata/metadata.go
Harkishen-Singh: Yes, I missed this. But now, we need to fetch all metadata from db, then form the map. While forming the map, we respect the limit which is what Prometheus does.

pkg/migrations/sql/idempotent/metric-metadata.sql
Harkishen-Singh: No, this is for users to write easier querying. I will document these in the docs PR.

pkg/migrations/sql/idempotent/metric-metadata.sql
Harkishen-Singh: Same as https://github.com/timescale/promscale/pull/624#discussion_r642233084

pkg/migrations/sql/versions/dev/0.3.1-dev/3-metric_metadata_objects.sql
Harkishen-Singh: I kept it for easier reference but you are right, we do not need this.

pkg/pgmodel/metadata/metadata.go
Harkishen-Singh: Added comments here and in the line that calls this function.

pkg/pgmodel/metadata/metadata.go
Harkishen-Singh: Should be easier to read now.

pkg/pgmodel/metadata/metadata.go
```diff
@@ -0,0 +1,49 @@
...
+	} else {
+		rows, err = conn.Query(context.Background(), "SELECT metric_family, type, unit, help from "+schema.Catalog+".metadata ORDER BY last_seen")
+	}
+	if err != nil {
+		return nil, fmt.Errorf("query metric metadata: %w", err)
+	}
+	defer rows.Close()
+	metricFamilies := make(map[string][]model.Metadata)
+	for rows.Next() {
+		if limit != 0 && len(metricFamilies) >= limit {
```
cevian, line 34: this is not right since we may not have all the metadata for the chosen metrics in the map yet.

cevian: Maybe we can change the all-metrics query to (note the order by):
```
		rows, err = conn.Query(context.Background(), "SELECT metric_family, type, unit, help from "+schema.Catalog+".metadata ORDER BY metric_name, last_seen")
```

That way we know metrics are all batched together. It's unclear how we should order the metrics to return for the limit

cevian (changes requested):
Full review. Some changes needed. Please keep as separate commits for now.

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,61 @@
...
+        RETURN num_rows;
+    END;
+$$ LANGUAGE plpgsql;
+GRANT EXECUTE ON FUNCTION SCHEMA_PROM.insert_metric_metadata(TIMESTAMPTZ, TEXT, TEXT, TEXT, TEXT) TO prom_writer;
+
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_metric_metadata(metric_family_name TEXT)
+RETURNS TABLE (metric_family TEXT, type TEXT, unit TEXT, help TEXT)
+AS
+$$
+    SELECT metric_family, type, unit, help FROM SCHEMA_CATALOG.metadata WHERE metric_family = metric_family_name ORDER BY last_seen
```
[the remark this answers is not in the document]
Harkishen-Singh: Oh yes, thank you for pointing out.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: Sorry, but I don't understand how that helps in keeping things than just overriding the function.

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,61 @@
...
+$$ LANGUAGE SQL;
+GRANT EXECUTE ON FUNCTION SCHEMA_PROM.get_metric_metadata(TEXT) TO prom_reader;
+
+-- metric_families should have unique elements, otherwise there will be duplicate rows in the returned table.
+CREATE OR REPLACE FUNCTION SCHEMA_PROM.get_multiple_metric_metadata(metric_families TEXT[])
+RETURNS TABLE (metric_family TEXT, type TEXT, unit TEXT, help TEXT)
+AS
+$$
+    SELECT DISTINCT ON
```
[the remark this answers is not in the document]
Harkishen-Singh: TODO with testing

pkg/pgmodel/metadata/metadata.go
Harkishen-Singh: The order is random (w.r.t. metric) as per what Prometheus does. Prometheus cares that metric a metric should be respected completely before moving on to next metric when there's a limit.
See 
https://github.com/prometheus/prometheus/blob/7bc11dcb06640ce22b4e15eb52b2c065f97cf79a/web/api/v1/api.go#L1125

The query you suggested will batch the metric_names and then I think it will be same on our side as well.

pkg/migrations/sql/idempotent/metric-metadata.sql
```diff
@@ -0,0 +1,37 @@
...
+RETURNS BIGINT
+AS
+$$
+    DECLARE
+        num_rows BIGINT;
+    BEGIN
+        INSERT INTO SCHEMA_CATALOG.metadata (last_seen, metric_family, type, unit, help)
+            SELECT * FROM UNNEST($1, $2, $3, $4, $5)
+        ON CONFLICT (metric_family, type, unit, help) DO
+            UPDATE SET last_seen = t[1];
```
cevian, line 11:
```suggestion
            UPDATE SET last_seen = EXCLUDED.last_seen;
```

cevian (changes requested):
Good job! Two small changes needed and let's rebase, combine commits, and get ready for merge.

pkg/pgmodel/ingestor/ingestor.go
antekresic: I'm good with this approach with a small tweak: sending in a nil function when you can't release memory instead of an empty func and have checks for nil funcs in the ingestion functions. 

Its more clear when reading the code and adds a check that is needed anyways.

antekresic (approved):
One minor suggestion, otherwise looks good to merge. :+1:

cevian (approved):

PR #634, opened 2021-05-24
Add support of Thanos Store grpc-service protocol to make able connect Promscale to Thanos Query

2nick:


Harkishen-Singh:
Hey @2nick! Thank you for contributing a PR for this.

Can you please reply to the questions below in order to provide some context:
1. How is this helpful?
2. Aims behind the implementation. Are you trying to achieve something?

2nick:
> 1. How is this helpful?

This PR adds functionality that allows using Promscale as part of [Thanos](https://github.com/thanos-io/thanos/) infrastructure.

Thanos is usually used with massive Prometheus installations and allows to run queries with metrics located on different Prometheus instances.

Connecting Promscale to Thanos Query component allows making queries that use metrics stored in TimescaleDB.

> 2. Aims behind the implementation. Are you trying to achieve something?

At first, I'd tried to use remote-read in Promscale with Thanos Sidecar which allows adding Prometheus instance to Thanos infrastructure.

But there is a problem that Thanos Sidecar trying to read Prometheus config through [/api/v1/status/config](https://prometheus.io/docs/prometheus/latest/querying/api/#config) to get external labels and Promscale has no such endpoint. Also sidecar won't start without external labels.

The main problem is that when Prometheus has external labels it will add them to samples when sends the stream to remote write and when sends stream through the remote read. 

And if you had added some metrics without those external labels directly to TimescaleDB you wouldn't select them because Prometheus will add external labels to select query.

For me it looks like adding the functionality of Thanos Store to Promscale is the best option here. :)

PR #642, opened 2021-05-26
Fix DB URI issues for URIs with missing `?` parameter settings separator

### LINKED ISSUE #641: Using PROMSCALE_DB_URI causes invalid DB name passed to client libraries
In docker-compose.yml:
```
    - "PROMSCALE_DB_URI=postgresql://postgres:postgres@timescaledb:5432/postgres"
```
Causes:
```
level=info ts=2021-05-26T06:48:59.113Z caller=client.go:116 msg=postgresql://postgres:postgres@timescaledb:5432/postgres numCopiers=32 pool_max_conns=40 pool_min_conns=8 statement_cache="512 statements"
level=error ts=2021-05-26T06:48:59.115Z caller=client.go:58 msg="err creating connection pool for new client" err="failed to connect to `host=timescaledb user=postgres database=postgres&pool_max_conns=40&pool_min_conns=8&statement_cache_capacity=512`: server error (FATAL: database \"postgres&pool_max_conns=40&pool_min_conns=8&statement_cache_cap\" does not exist (SQLSTATE 3D000))"
level=error ts=2021-05-26T06:48:59.205Z caller=runner.go:40 msg="aborting startup due to error" err="client creation error: failed to connect to `host=timescaledb user=postgres database=postgres&pool_max_conns=40&pool_min_conns=8&statement_cache_capacity=512`: server error (FATAL: database \"postgres&pool_max_conns=40&pool_min_conns=8&statement_cache_cap\" does not exist (SQLSTATE 3D000))"
```
It appears to simply concatenates strings to append extra options (why does it add options I did not ask for to begin with?), which causes this error.

Verified that adding `?` at the end of URI fixes this issue.

It should not:

1. Assume that there is already any params exist (hence shouldn't assume `?` already exists in URI)
2. Shouldn't add extra arguments silently

antekresic:
Fixes #641

### DIFF
--- a/pkg/pgclient/config_test.go
+++ b/pkg/pgclient/config_test.go
@@ -58,7 +58,7 @@ func TestConfig_GetConnectionStr(t *testing.T) {
 				UsesHA:                  false,
 				DbUri:                   "",
 			},
-			want:    fmt.Sprintf("application_name=%s host=localhost port=5433 user=postgres dbname=timescale1 password='Timescale123' sslmode=require connect_timeout=120", DefaultApp),
+			want:    fmt.Sprintf("postgresql://postgres:Timescale123@localhost:5433/timescale1?application_name=%s&sslmode=require&connect_timeout=120", DefaultApp),
 			wantErr: false,
 			err:     nil,
 		},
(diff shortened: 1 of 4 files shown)


pkg/pgclient/config_test.go
```diff
@@ -58,7 +58,7 @@ func TestConfig_GetConnectionStr(t *testing.T) {
 				UsesHA:                  false,
```
cevian, line 58: Let's add a test case with a dburi supplied that does not have a ? at the end

cevian (approved):

Harkishen-Singh:
Do we need this after #652? I think that's a super set that solves all problems related to masking if I am not wrong.

PR #646, opened 2021-05-27
Make prom-migrator runtime more customizable.

Harkishen-Singh:

Improvements to the prom-migrator includes:
1. Human readable time
2. Ability to configure timeout for both reader and writer
3. Ability to retry in case of timeout or an error
4. Limit max-retries for both reader and writer
5. Custom handling of behaviour (`skip`, `retry`, `abort`) when faced with error (using `-[reader | writer]-on-error`) or timeout (using `-[reader | writer]-on-timeout`)
6. Increase laIncrement to speed up migration
7. Increase max-read-duration. This is helpful while migrating from non-Prometheus storage to remote storage since there is no more than a 2 hours limit on speed.
8. Updated integration tests to test the retrying behaviour on error
9. Improved flags descriptions based on existing experience
10. Perf improvements: reusable pools for slabs and request byte slices
11. Collect garbage on every push to make memory usage more conservative. This can be turned on with a flag.

TODO:

- [x] Implement reusable pools for slabs
- [x] Implement reusable pools for byte slices
- [x] update docs in docs/cli.md


Harkishen-Singh:
Github buggy in terms of review requests!

cevian:
@antekresic can you please take the first review pass here?

antekresic (changes requested):
Looks good, should be in a decent spot once the comments are addressed.

cmd/prom-migrator/main.go
```diff
@@ -243,34 +314,120 @@ func parseArgs(args []string) (shouldProceed bool) {
 	return
 }
 
-func convertSecFlagToMs(conf *config) {
+// formatTimeInput removes the asterisks from the string input.
```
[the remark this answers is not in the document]
Harkishen-Singh: Oh sorry, definitely single quotes.

cmd/prom-migrator/main.go
```diff
@@ -243,34 +314,120 @@ func parseArgs(args []string) (shouldProceed bool) {
...
+			return fmt.Errorf("'start' flag: %w", err)
+		}
+		conf.mintSec = t
+
+		t, err = timestampSec(conf.end)
+		if err != nil {
+			return fmt.Errorf("'end' flag: %w", err)
+		}
+		conf.maxtSec = t
+	} else {
```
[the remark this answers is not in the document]
Harkishen-Singh: I think that will be done by creating an in-line func, so I am not sure if that will lead to further complications than being linear here.

cmd/prom-migrator/main.go
```diff
@@ -243,34 +314,120 @@ func parseArgs(args []string) (shouldProceed bool) {
...
+	case conf.start == defaultStartTime:
 		return fmt.Errorf("mint should be provided for the migration to begin")
-	case conf.mint < 0:
-		return fmt.Errorf("invalid mint: %d", conf.mint)
-	case conf.maxt < 0:
-		return fmt.Errorf("invalid maxt: %d", conf.maxt)
-	case conf.mint > conf.maxt:
-		return fmt.Errorf("invalid input: minimum timestamp value (mint) cannot be greater than the maximum timestamp value (maxt)")
+	case conf.mintSec > conf.maxtSec:
+		return fmt.Errorf("invalid input: minimum timestamp value (start) cannot be greater than the maximum timestamp value (end)")
```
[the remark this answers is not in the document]
Harkishen-Singh: I assume you mean `validateAndConvertStrFlagsToTs()`. That function just does conversion (I think the name is misleading so will change to `convertStrFlagsToTs()`) and all the validations happens here for now.

cmd/prom-migrator/main.go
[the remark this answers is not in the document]
antekresic: Not sure what are you taking about in-line function. My suggestion is to do something like:

```
func convertTimeStrFlagsToTs(conf *config) error {
    	// remote-storages tend to respond to time in milliseconds. So, we convert the received values in seconds to milliseconds.
    	conf.mint = conf.mintSec * 1000
	conf.maxt = conf.maxtSec * 1000
	if conf.humanReadable {
		...
		return nil
	}
        ...
        return nil
}

**Review by antekresic — APPROVED** [review]
Looks good, one previous comment was not addressed, added a clarification. Once that is resolved, should be good to merge.

File: pkg/migration-tool/utils/client.go [hunk 4]
```diff
@@ -119,26 +195,47 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query, desc string) (re
 	}
 	compressed, err = ioutil.ReadAll(bytes.NewReader(reader.Bytes()))
 	if err != nil {
-		return nil, -1, -1, errors.Wrap(err, fmt.Sprintf("error reading response. HTTP status code: %s", httpResp.Status))
+		err = errors.Wrap(err, fmt.Sprintf("error reading response. HTTP status code: %s", httpResp.Status))
+		if shouldRetry(c.cRuntime.OnErr, c.cRuntime, numAttempts, fmt.Sprintf("error:\n%s\n", err.Error())) {
```
cevian, line 199: couldn't the error here be a timeout, shouldn't we check for that?

pkg/migration-tool/utils/client.go
```diff
@@ -152,19 +249,29 @@ type PrompbResponse struct {
 	NumBytesUncompressed int
 }
 
+var prompbResponsePool = sync.Pool{New: func() interface{} { return new(PrompbResponse) }}
+
+func PutPrompbResponse(p *PrompbResponse) {
+	p.ID = 0
+	p.NumBytesUncompressed = 0
+	p.NumBytesCompressed = 0
+	p.Result.Timeseries = p.Result.Timeseries[:0]
```
cevian, line 258: where is this reused I don't see a check for cap(Result.Timeseries) anywhere

pkg/migration-tool/writer/writer.go
```diff
@@ -120,12 +122,48 @@ func (w *Write) Run(errChan chan<- error) {
...
 	}()
 }
 
-// Blocks returns the total number of blocks pushed to the remote-write storage.
+func (w *Write) collectGarbage() {
+	if w.sigGC == nil {
+		// sigGC is nil if w.GarbageCollectOnPush is set to false.
+		return
+	}
+	if !gcRoutineActive.Load().(bool) {
```
cevian, line 137: why not sync.Once here?

pkg/migration-tool/writer/writer.go
```diff
@@ -59,6 +58,9 @@ func New(config Config) (*Write, error) {
 			Labels: utils.LabelSet(config.ProgressMetricName, config.MigrationJobName),
 		}
 	}
+	if config.GarbageCollectOnPush {
+		write.sigGC = make(chan struct{})
```
cevian, line 62: what's the length of this chan? Will this action be sync or async?

pkg/version/version.go
```diff
@@ -38,7 +38,7 @@ var (
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
 	Promscale                           = "0.4.2-dev.0"
-	PromMigrator                        = "0.0.1"
+	PromMigrator                        = "0.0.2-beta.0"
```
cevian, line 41: why beta?

pkg/migration-tool/writer/writer.go
```diff
@@ -120,12 +122,48 @@ func (w *Write) Run(errChan chan<- error) {
...
+	gcRoutineActive atomic.Value
+	isGCInProgress  atomic.Value
+)
+
+func init() {
+	gcRoutineActive.Store(false)
+	isGCInProgress.Store(false)
+}
+
+func gc(collect <-chan struct{}) {
```
cevian, line 154:
```suggestion
func gc(collect <-chan struct{}) {
        isGCInProgress.Store(false)
```

pkg/migration-tool/writer/writer.go
```diff
@@ -120,12 +122,48 @@ func (w *Write) Run(errChan chan<- error) {
...
+}
+
+var (
+	gcRoutineActive atomic.Value
+	isGCInProgress  atomic.Value
+)
+
+func init() {
+	gcRoutineActive.Store(false)
+	isGCInProgress.Store(false)
```
cevian, line 151:
```suggestion

```

cevian (changes requested):

cmd/prom-migrator/main.go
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, I am saying, cannot do that because then `conf.mintSec = t` which is few lines above, will not be assigned.

cmd/prom-migrator/main.go
```diff
@@ -152,23 +169,69 @@ loop:
...
+		"This allows higher throughput of read by pulling data faster from the remote-read storage. "+
+		"Note: Setting 'concurrent-pull' > 1 will show progress of concurrent fetching of data in the progress-bar and disable real-time transfer rate. "+
+		"High 'concurrent-pull' can consume significant memory, so make sure you balance this with your number of migrating series and available memory. "+
+		"Also, setting this value too high may cause TLS handshake error on the read storage side or may lead to starvation of fetch requests, "+
+		"depending on your network bandwidth.")
+
+	flag.DurationVar(&conf.maxReadDuration, "max-read-duration", defaultMaxReadDuration, "Maximum range of slab that can be achieved through consecutive 'la-increment'. "+
+		"This defaults to '2 hours' since assuming the migration to be from Prometheus TSDB. Increase this duration if you are not fetching from Prometheus "+
+		"or if you are fine with slow read on Prometheus side post 2 hours slab time-range.")
+	flag.DurationVar(&conf.laIncrement, "la-increment", defaultLaIncrement, "Amount of time-range to be incremented in successive slab.")
```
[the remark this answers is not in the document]
Harkishen-Singh: `linear-increment`.

pkg/migration-tool/utils/client.go
```diff
@@ -36,19 +37,60 @@ const (
...
+	// Retry makes the client to retry in case of a timeout or error.
+	Retry
+	// Skip makes the client to skip the current slab.
+	Skip
+	// Abort aborts the client from any further retries and exits the migration process.
+	// This can be applied after a timeout or error.
+	Abort
 )
 
+type ClientRuntime struct {
```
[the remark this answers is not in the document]
Harkishen-Singh: Configuration of client in runtime. Changed this to clientconfig.

cmd/prom-migrator/main.go
```diff
@@ -243,34 +315,120 @@ func parseArgs(args []string) (shouldProceed bool) {
...
+	if s[0] == '\'' {
+		s = s[1:]
+	}
+	if s[len(s)-1] == '\'' {
+		s = s[:len(s)-1]
+	}
+	return s
+}
+
+func parseClientInfo(conf *config) error {
```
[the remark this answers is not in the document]
Harkishen-Singh: (Note to self) Resume from here.

pkg/migration-tool/planner/planner.go
```diff
@@ -177,27 +180,34 @@ func (p *Plan) createSlab(mint, maxt int64) (reference *Slab, err error) {
...
-		stores:    make([]store, p.config.NumStores),
-		mint:      mint,
-		maxt:      maxt,
-		pbarMux:   p.pbarMux,
-		plan:      p,
-	}
-	reference.initStores()
+	baseDescription := fmt.Sprintf("progress: %.2f%% | slab-%d time-range: %d mins | begin: %d | end: %d", percent, id, timeRangeInMinutes, mint/second, maxt/second)
+	if p.config.HumanReadableTime {
+		baseDescription = fmt.Sprintf("progress: %.2f%% | slab-%d time-range: %d mins | begin: %s | end: %s", percent, id, timeRangeInMinutes, timestamp.Time(mint), timestamp.Time(maxt))
```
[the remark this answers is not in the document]
Harkishen-Singh: We should keep both (and use depending on the input) since if a user is using unix, then time range in unix is important to give him an idea when to start from, the next time. Vice-versa for human-readable.

pkg/migration-tool/utils/client.go
```diff
@@ -119,26 +195,47 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query, desc string) (re
...
 	if httpResp.StatusCode/100 != 2 {
+		if shouldRetry(c.cRuntime.OnErr, c.cRuntime, numAttempts, fmt.Sprintf("received status code: %d", httpResp.StatusCode)) {
+			goto retry
+		}
 		return nil, -1, -1, errors.Errorf("remote server %s returned HTTP status %s: %s", c.url.String(), httpResp.Status, strings.TrimSpace(string(compressed)))
 	}
 
 	uncompressed, err := snappy.Decode(nil, compressed)
 	if err != nil {
+		// Snappy errors also happen in case of deadline or context cancel since the read was terminated in middle of the process.
```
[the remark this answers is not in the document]
Harkishen-Singh: No, we don't. I too expected the same behaviour, but in testing, it didn't send the error above rather as corrupt snappy since the timeout occurred. The snappy corruption is because prom-migrator accepts response, and while accepting timeout occurs and gets a partial snappy encoded response, which is bad.

pkg/migration-tool/utils/client.go
Harkishen-Singh: I don't think so. The errors here would be related to reading the underlying byte slice.

pkg/migration-tool/utils/client.go
Harkishen-Singh: In planner.go `mergeSubSlabsToSlab()`. Why do we need a check for cap?

pkg/migration-tool/writer/writer.go
Harkishen-Singh: This would be async since GC can be a blocking call. We don't need the length here since `isGCInProgress` boolean rejects the signal in case GC is already running, so indirectly async.

pkg/version/version.go
Harkishen-Singh: I wasn't sure of this. I think we should let users try retrying before we mark it stable.

pkg/migration-tool/writer/writer.go
cevian: This is NOT async right now since the channel length is 0. and the `gc` function will block  at `runtime.GC()` before taking things off the queue at `for range collect {`. You either need to run two `gc` goroutines or add slack to the channel length here

pkg/version/version.go
cevian: do you think there is anything in this PR that can break existing functionality, if so, what?

cevian (changes requested):

cmd/prom-migrator/main.go
```diff
@@ -152,23 +170,66 @@ loop:
...
-	flag.StringVar(&conf.writeURL, "write-url", "", "URL address for the storage where the data migration is to be written.")
+	flag.IntVar(&conf.concurrentPull, "concurrent-pull", 1, "Concurrent pull enables fetching of data concurrently. "+
+		"Each fetch query is divided into 'concurrent-pull' (value) parts and then fetched concurrently. "+
+		"This allows higher throughput of read by pulling data faster from the remote-read storage. "+
+		"Note: Setting 'concurrent-pull' > 1 will show progress of concurrent fetching of data in the progress-bar and disable real-time transfer rate. "+
+		"High 'concurrent-pull' can consume significant memory, so make sure you balance this with your number of migrating series and available memory. "+
+		"Also, setting this value too high may cause TLS handshake error on the read storage side or may lead to starvation of fetch requests, "+
+		"depending on your network bandwidth.")
+
+	flag.DurationVar(&conf.maxReadDuration, "max-read-duration", defaultMaxReadDuration, "Maximum range of slab that can be achieved through consecutive 'la-increment'. "+
```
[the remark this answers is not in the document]
Harkishen-Singh: Already addressed in the recent commit.

pkg/migration-tool/utils/client.go
```diff
@@ -36,19 +37,108 @@ const (
...
+	// Retry makes the client to retry in case of a timeout or error.
+	Retry
+	// Skip makes the client to skip the current slab.
+	Skip
+	// Abort aborts the client from any further retries and exits the migration process.
+	// This can be applied after a timeout or error.
+	Abort
+
+	onTimeout
+	onError
```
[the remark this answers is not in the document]
Harkishen-Singh: Thank you for suggestion, this will make things a lot more easier 😄

pkg/migration-tool/utils/client.go
Harkishen-Singh: Yes, we aren't reusing `p.Result.Timeseries`, since the p.Result is overwritten by the response that we receive from the PromQL query, so reusing is not possible here. Thank you.

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Oh yes, thank you for pointing it out. Making the length as `2` should do the job.

pkg/version/version.go
Harkishen-Singh: There isn't anything breaking I think. The part of retrying when read times-out is not deterministic. More importantly, I think it will be safer to keep the first release of retrying as beta so that we can know about issues (if they exists as corner cases), than calling it **stable**.

pkg/migration-tool/utils/client.go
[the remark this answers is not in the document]
Harkishen-Singh: Yes, when I tried with local Prometheus, `if errors.Is(err, context.DeadlineExceeded) {` is catching the deadline issue and then retrying. So, we can get rid of this exception at line 214 which you commented. Thank you.

pkg/migration-tool/utils/client.go
```diff
@@ -152,19 +296,29 @@ type PrompbResponse struct {
 	NumBytesUncompressed int
 }
 
+var prompbResponsePool = sync.Pool{New: func() interface{} { return new(PrompbResponse) }}
+
+func PutPrompbResponse(p *PrompbResponse) {
+	p.ID = 0
+	p.NumBytesUncompressed = 0
+	p.NumBytesCompressed = 0
+	p.Result = nil
```
cevian, line 305: what's the point of the pool if we don't reuse the timeseries?

pkg/migration-tool/writer/writer.go
```diff
@@ -120,12 +130,42 @@ func (w *Write) Run(errChan chan<- error) {
...
+
+func (w *Write) collectGarbage() {
+	if w.sigGC == nil {
+		// sigGC is nil if w.GarbageCollectOnPush is set to false.
+		return
+	}
+	gcRunner.Do(func() {
+		go gc(w.sigGC)
+	})
+	w.sigGC <- struct{}{}
```
cevian, line 150: Make the send non-blocking. If the channel already has 2 items `gc` can still block by working in `runtime.GC()`. By making this a non-blocking send you avoid this scenario and still guarantee that the gc will run since the queue has items.
```suggestion
    select {
    case w.sigGC <- struct{}{}:
    default:
    }
```

pkg/migration-tool/writer/writer.go
```diff
@@ -59,6 +59,16 @@ func New(config Config) (*Write, error) {
 			Labels: utils.LabelSet(config.ProgressMetricName, config.MigrationJobName),
 		}
 	}
+	if config.GarbageCollectOnPush {
+		// The channel length is 2 so that the operation of GC can be async. Keeping it 2 makes the first call
```
cevian, line 63: the logic described here is incorrect. see my comment below.

pkg/migration-tool/writer/writer.go
```diff
@@ -59,6 +59,16 @@ func New(config Config) (*Write, error) {
...
+		// run the GC. If the second slab is pushed while GC was running, the second buffer length accepts
+		// the signal from the channel and then isGCInProgress.Load().(bool) tells to return as GC is already running.
+		// This happens for all successive GC calls unless the GC process completes, making the GC run async.
+		//
+		// Calls to the GC should be async since behaviour of garbage collection is blocking. This becomes a lot
+		// more true in case of migrator, since we have to deal with very large data blocks.
+		write.sigGC = make(chan struct{}, 2)
```
cevian, line 70: if we switch to non-blocking send this can be length 1

cevian (changes requested):

pkg/migration-tool/writer/writer.go
Harkishen-Singh: Oh yes, this is much better way

cevian (approved):

PR #650, opened 2021-05-31
Add debian packaging to goreleaser

antekresic:


build/.goreleaser.yml
```diff
@@ -33,6 +27,54 @@ scoop:
...
+
+    license: Apache License 2.0
+    maintainer: Timescale <info@timescale.com>
+    contents:
+      - src: build/systemd/promscale.service
+        dst: /etc/systemd/system/promscale.service
+      - src: build/conf/promscale.conf
+        dst: /etc/sysconfig/promscale.conf
+        type: "config|noreplace"
+        packager: rpm
```
cevian, line 61: how does this conf file work on deb then?

build/scripts/postinstall.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+echo "Promscale installed as a systemd service"
+echo "Enable auto-start with: systemctl enable promscale"
+echo "And start the service now with: systemctl start promscale"
+echo "----------------------------------------"
+CONF_FILE=$(find /etc -name promscale.conf 2>/dev/null | head -n 1)
```
cevian, line 6: why don't we know where this is? How does this work on deb?

cevian (changes requested):
This looks great. some questions about deb packaging. Also was this all manually tested on rhel, debian, and ubuntu?

build/scripts/postinstall.sh
antekresic: Config files go into different places in `/etc` folder. I did miss to add the location for deb package, will add.

build/.goreleaser.yml
antekresic: I somehow missed adding this conf file for deb packaged, will add now.

antekresic:
Updated to add missing deb conf file location. Manually tested with `yum` and `dpkg` package installers.

cevian (approved):

PR #655, opened 2021-06-03
Buffer allocation pool for parsing protobuf

cevian:
Reduces allocations significantly.

### DIFF
--- a/pkg/api/parser/protobuf/protobuf.go
+++ b/pkg/api/parser/protobuf/protobuf.go
@@ -1,9 +1,10 @@
 package protobuf
 
 import (
+	"bytes"
 	"fmt"
-	"io/ioutil"
 	"net/http"
+	"sync"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/timescale/promscale/pkg/prompb"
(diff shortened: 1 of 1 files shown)


pkg/api/parser/protobuf/protobuf.go
```diff
@@ -12,14 +13,24 @@ import (
...
 		return fmt.Errorf("request body read error: %w", err)
 	}
 
-	if err = proto.Unmarshal(reqBuf, wr); err != nil {
+	if err = proto.Unmarshal(b.Bytes(), wr); err != nil {
 		return fmt.Errorf("protobuf unmarshal error: %w", err)
 	}
 
-	return nil
+	bufPool.Put(b)
```
Harkishen-Singh, line 28: I think this should be in `defer` (line 17) as if we error out in line `21` or `25`, we do not put in the pool and this gets GCed.

pkg/api/parser/protobuf/protobuf.go
```diff
@@ -12,14 +13,24 @@ import (
 // ParseRequest is responsible for populating the write request from the
 // data in the request in protobuf format.
 func ParseRequest(r *http.Request, wr *prompb.WriteRequest) error {
-	reqBuf, err := ioutil.ReadAll(r.Body)
+	b := bufPool.Get().(*bytes.Buffer)
+	b.Reset()
+
+	_, err := b.ReadFrom(r.Body)
```
Harkishen-Singh, line 19: Can we do the same for read requests? That will help us save extra allocations.

Harkishen-Singh (changes requested):

antekresic (approved):
I agree with @Harkishen-Singh on the defer change.

Other than that, looks good! :+1:

pkg/api/parser/protobuf/protobuf.go
cevian: Yes but in a separate PR. We should get flamegraphs to see we need this before doing it

pkg/api/parser/protobuf/protobuf.go
cevian: good catch

pkg/api/parser/protobuf/protobuf.go
Harkishen-Singh: @cevian TODO.

Harkishen-Singh (approved):
👍🏻

Harkishen-Singh:
Please note the TODO in the comment. Maybe, let's have that on the project board otherwise.

PR #656, opened 2021-06-03
Fix label fetching during querying

cevian:
There were two problems:
- We fetched labels when we found a series with missing labels.
  For queries that fetch many labels, this meant many separate
  queries per query. Now we query for all missing labels once
  per query by using a map.
- We kept trying to fetch a label with id 0,
  this meant we kept on trying to fetch labels for every series
  even if everything was cached.


pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -97,41 +92,39 @@ func (lr *labelsReader) LabelNames() ([]string, error) {
...
-	for i := range ll {
-		lls[i] = prompb.Label{Name: ll[i].Name, Value: ll[i].Value}
+func (lr *labelsReader) LabelsForIdMap(idMap map[int64]*labels.Label) (err error) {
+	numIds := len(idMap)
+	keys := make([]interface{}, numIds)
+	values := make([]interface{}, numIds)
+
+	_, present := idMap[0]
+	if present {
+		return fmt.Errorf("Looking up a label for id 0")
```
Harkishen-Singh, line 102: Why do we error for `0` label id? I remember in some tests that I had 0 as id which indicated they were not created if I recall. But should we be erroring for this?

Harkishen-Singh: If it's required, then a comment will be helpful.

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -97,41 +92,39 @@ func (lr *labelsReader) LabelNames() ([]string, error) {
...
 	numHits := lr.labels.GetValues(keys, values)
 
-	if numHits < len(ids) {
+	if numHits < numIds {
 		var numFetches int
-		numFetches, err = lr.fetchMissingLabels(keys[numHits:], ids[numHits:], values[numHits:])
+		missingIds := make([]int64, numIds-numHits)
+		numFetches, err = lr.fetchMissingLabels(keys[numHits:], missingIds, values[numHits:])
 		if err != nil {
 			return
```
Harkishen-Singh, line 117: BTW, why don't we return err here (or with concatenation with line `120`)?

pkg/pgmodel/querier/query_builder.go
```diff
@@ -264,9 +278,20 @@ func buildTimeSeries(rows []timescaleRow, lr lreader.LabelsReader) ([]*prompb.Ti
...
+		for _, id := range row.labelIds {
+			if id == 0 {
+				continue
+			}
+			label, ok := labelIDMap[id]
+			if !ok {
+				return nil, fmt.Errorf("Missing label for id %v", id)
+			}
+			if label == nil {
+				return nil, fmt.Errorf("Missing label for id %v", id)
```
Harkishen-Singh, line 291:
```suggestion
				return nil, fmt.Errorf("label not found for id %v", id)
```

pkg/pgmodel/querier/query_builder.go
```diff
@@ -264,9 +278,20 @@ func buildTimeSeries(rows []timescaleRow, lr lreader.LabelsReader) ([]*prompb.Ti
 			return nil, errors.ErrQueryMismatchTimestampValue
 		}
 
-		promLabels, err := lr.PrompbLabelsForIds(row.labelIds)
-		if err != nil {
-			return nil, err
+		promLabels := make([]prompb.Label, 0, len(row.labelIds))
+		for _, id := range row.labelIds {
+			if id == 0 {
```
Harkishen-Singh, line 283: Comment please

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -97,41 +92,39 @@ func (lr *labelsReader) LabelNames() ([]string, error) {
...
 		}
-		values = values[:numHits+numFetches]
+		if numFetches+numHits != numIds {
+			return fmt.Errorf("Missing labels")
```
Harkishen-Singh, line 120:
```suggestion
			return fmt.Errorf("Missing %d labels", numIds - (numFetches + numHits))
```

Optional

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -97,41 +92,39 @@ func (lr *labelsReader) LabelNames() ([]string, error) {
...
+		}
 	}
 
-	lls = make([]labels.Label, 0, len(values))
-	for i := range values {
-		lls = append(lls, values[i].(labels.Label))
+	for i := range keys {
+		label := values[i].(labels.Label)
+		id := keys[i].(int64)
+		idMap[id] = &label
```
Harkishen-Singh, line 127: Are these allocations heavy (the `&label`)? Plus, can we not change the map to simply `map[int64]labels.Label` as I don't strong requirement (after changing type) in the calling functions for pointers?

pkg/pgmodel/querier/series_set.go
```diff
@@ -25,21 +24,38 @@ const (
 
 // pgxSeriesSet implements storage.SeriesSet.
 type pgxSeriesSet struct {
-	rowIdx  int
-	rows    []timescaleRow
-	err     error
-	querier labelQuerier
+	rowIdx     int
+	rows       []timescaleRow
+	labelIDMap map[int64]*labels.Label
```
Harkishen-Singh, line 29: Is it necessary for this to be a pointer? If not, we can save some allocation here even if this is not expensive.

Harkishen-Singh (changes requested):

pkg/pgmodel/lreader/labels_reader.go
antekresic: We do return it since its a named return value but I'd rather we don't use named return values since it creates confusing situations like this and hurts readability. 

They should be used for shorter functions where the code really benefits from them.

pkg/pgmodel/querier/series_set.go
```diff
@@ -25,21 +24,38 @@ const (
...
+			if id == 0 {
+				continue
+			}
+			labelIDMap[id] = nil
+		}
+	}
+
+	err := querier.LabelsForIdMap(labelIDMap)
+	if err != nil {
+		return nil, err
```
antekresic, line 50: Why not return errorSeriesSet here and keep the same func signature?

antekresic (changes requested):

pkg/pgmodel/lreader/labels_reader.go
cevian: added a comment

pkg/pgmodel/lreader/labels_reader.go
cevian: changed

pkg/pgmodel/querier/series_set.go
cevian: good point. fixed

antekresic (approved):

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -97,44 +93,49 @@ func (lr *labelsReader) LabelNames() ([]string, error) {
...
+		)
+
+		missingIds := make([]int64, numIds-numHits)
+		numFetches, err = lr.fetchMissingLabels(ids[numHits:], missingIds, lbs[numHits:])
 		if err != nil {
-			return
+			return err
+		}
+		if numFetches+numHits != numIds {
+			return fmt.Errorf("missing labels")
```
Harkishen-Singh, line 128: should we also mention which all ids were missing the labels for better debugging?

pkg/pgmodel/querier/query_builder.go
```diff
@@ -279,6 +279,21 @@ func (c *clauseBuilder) Build(includeMetricName bool) ([]string, []interface{},
 
 func buildTimeSeries(rows []timescaleRow, lr lreader.LabelsReader) ([]*prompb.TimeSeries, error) {
 	results := make([]*prompb.TimeSeries, 0, len(rows))
+	labelIDMap := make(map[int64]labels.Label)
+	for i := range rows {
```
Harkishen-Singh, line 283: nit: should we move this to a func `initializeLabelIDMap()`?

pkg/pgmodel/querier/query_builder.go
```diff
@@ -279,6 +279,21 @@ func (c *clauseBuilder) Build(includeMetricName bool) ([]string, []interface{},
...
+			if id == 0 {
+				continue
+			}
+			labelIDMap[id] = labels.Label{}
+		}
+	}
+
+	err := lr.LabelsForIdMap(labelIDMap)
+	if err != nil {
+		return nil, err
```
Harkishen-Singh, line 295:
```suggestion
		return nil, fmt.Errorf("build timeseries: %w ", err)
```

Lets wrap the error here, otherwise we can confuse with the error at line 300.

pkg/pgmodel/querier/series_set.go
```diff
@@ -25,20 +24,37 @@ const (
...
+	err        error
+	querier    labelQuerier
 }
 
 // pgxSeriesSet must implement storage.SeriesSet
 var _ storage.SeriesSet = (*pgxSeriesSet)(nil)
 
 func buildSeriesSet(rows []timescaleRow, querier labelQuerier) storage.SeriesSet {
+	labelIDMap := make(map[int64]labels.Label)
+	for i := range rows {
```
Harkishen-Singh, line 39: We should really have a `initializeLabelIDMap()` as said above, and that will reduce the redundancy here.

pkg/pgmodel/querier/series_set.go
```diff
@@ -81,10 +97,21 @@ func (p *pgxSeriesSet) At() storage.Series {
 	// this should pretty much always be non-empty due to __name__, but it
 	// costs little to check here
 	if len(row.labelIds) != 0 {
```
Harkishen-Singh, line 99: I think it will be better to have this as `if len(row.labelIds) == 0` and simply return ps, this will decrease the nesting and look cleaner.

pkg/pgmodel/querier/series_set.go
```diff
@@ -81,10 +97,21 @@ func (p *pgxSeriesSet) At() storage.Series {
...
-		lls, err := p.querier.LabelsForIds(row.labelIds)
-		if err != nil {
-			log.Error("err", err)
-			return nil
+		var lls labels.Labels
+		lls = make([]labels.Label, 0, len(row.labelIds))
```
Harkishen-Singh, line 101:
```suggestion
		lls := make([]labels.Label, 0, len(row.labelIds))
```

pkg/pgmodel/querier/series_set.go
```diff
@@ -81,10 +97,21 @@ func (p *pgxSeriesSet) At() storage.Series {
...
+		for _, id := range row.labelIds {
+			if id != 0 {
```
Harkishen-Singh, line 103:
```suggestion
			if id == 0 {
			  continue
			}
```

Harkishen-Singh (approved):
Some nits, rest 👍🏻

Can we have all errors here starting with small case? That is the convention for errors and we should follow the same.

pkg/pgmodel/querier/series_set.go
cevian: can't do this since want labels.Labels not []labels.Label

PR #657, opened 2021-06-07
Update Prompb with upstream upto May 7.

Harkishen-Singh:

This commit updates the Prompb with Upstream including all changes
till May 7. This includes changes done till commit
8fd73b1d281d3227c337b291e8c797a85251d3d5.

The major feature of this commit is the support of Exemplars.


pkg/prompb/types.pb.go
```diff
@@ -1807,6 +2101,40 @@ func (m *TimeSeries) Unmarshal(dAtA []byte) error {
...
+				return ErrInvalidLengthTypes
+			}
+			postIndex := iNdEx + msglen
+			if postIndex < 0 {
+				return ErrInvalidLengthTypes
+			}
+			if postIndex > l {
+				return io.ErrUnexpectedEOF
+			}
+			m.Exemplars = append(m.Exemplars, Exemplar{})
```
cevian, line 2133: We should implement the same optimizations we do for samples (maybe separate commit same pr)
```suggestion
			if len(m.Examplars) < cap(m.Examplars) {
				m.Examplars = m.Examplars[:len(m.Examplars)+1]
				m.Examplars[len(m.Examplars)-1].Reset()
			} else {
				m.Examplars = append(m.Examplars, Examplars{})
			}
```
in custom.ts.go we need
```
func (m *TimeSeries) Reset() {
	*m = TimeSeries{Labels: m.Labels[:0], Samples: m.Samples[:0], Examplars: m.Exemplars[:0]}
}
```

pkg/prompb/types.pb.go
```diff
@@ -1700,6 +1879,121 @@ func (m *Sample) Unmarshal(dAtA []byte) error {
...
+				return ErrInvalidLengthTypes
+			}
+			postIndex := iNdEx + msglen
+			if postIndex < 0 {
+				return ErrInvalidLengthTypes
+			}
+			if postIndex > l {
+				return io.ErrUnexpectedEOF
+			}
+			m.Labels = append(m.Labels, Label{})
```
cevian, line 1940: We also need to implement the same slice-reuse logic here

pkg/prompb/types.pb.go
```diff
@@ -253,23 +255,101 @@ func (m *Sample) GetTimestamp() int64 {
...
+	Value  float64 `protobuf:"fixed64,2,opt,name=value,proto3" json:"value,omitempty"`
+	// timestamp is in ms format, see pkg/timestamp/timestamp.go for
+	// conversion from time.Time to Prometheus timestamp.
+	Timestamp            int64    `protobuf:"varint,3,opt,name=timestamp,proto3" json:"timestamp,omitempty"`
 	XXX_NoUnkeyedLiteral struct{} `json:"-"`
 	XXX_unrecognized     []byte   `json:"-"`
 	XXX_sizecache        int32    `json:"-"`
 }
 
+func (m *Exemplar) Reset()         { *m = Exemplar{} }
```
cevian, line 270: this should be moved to custom.ts.go and we should implement the slice-reuse logic

cevian (changes requested):
We need to implement our changes for exemplars that would make good reuse of our object pool

Harkishen-Singh:
Bors r+

Issue #583, opened 2021-04-14
Timescale fails after restart


YurkinED:
When first start it works properly, after restart I got an error
Seems some problem with recreating objects

```
level=info ts=2021-04-14T09:18:09.090Z caller=runner.go:32 msg="Version:0.2.1; Commit Hash: "

level=info ts=2021-04-14T09:18:09.090Z caller=runner.go:33 config="&{ListenAddr::9201 PgmodelCfg:{AppName:promscale@0.2.1 Host:localhost Port:5432 User:postgres password: ****Database:promeheus SslMode:allow DbConnectRetries:0 DbConnectionTimeout:1m0s AsyncAcks:false ReportInterval:0 LabelsCacheSize:10000 MetricsCacheSize:10000 SeriesCacheSize:0 WriteConnectionsPerProc:4 MaxConnections:-1 UsesHA:false DbUri:} LogCfg:{Level:debug Format:logfmt} APICfg:{AllowedOrigin:^(?:.*)$ ReadOnly:false AdminAPIEnabled:false TelemetryPath:/metrics Auth:0xc0000b1ef0 EnabledFeaturesList:[] MaxQueryTimeout:2m0s SubQueryStepInterval:1m0s} ConfigFile:config.yml TLSCertFile: TLSKeyFile: HaGroupLockID:0 PrometheusTimeout:-1ns ElectionInterval:5s Migrate:true StopAfterMigrate:false UseVersionLease:true InstallExtensions:true UpgradeExtensions:true UpgradePrereleaseExtensions:false}"

level=warn ts=2021-04-14T09:18:09.197Z caller=client.go:132 msg="Maximum background worker setting is too low for the number of databases in your system. Please increase your timescaledb.max_background_workers setting. See https://docs.timescale.com/latest/getting-started/configuring#workers for more information."

level=warn ts=2021-04-14T09:18:09.197Z caller=client.go:146 msg="No adapter leader election. Group lock id is not set. Possible duplicate write load if running multiple connectors"

level=info ts=2021-04-14T09:18:09.210Z caller=client.go:91 msg="application_name=promscale@0.2.1 host=localhost port=5432 user=postgres dbname=promeheus password='' sslmode=allow connect_timeout=60" numCopiers=8 pool_max_conns=50 pool_min_conns=2

[...]

PR #659, opened 2021-06-08
Fixes Promscale on TimescaleDB-OSS.

Harkishen-Singh:

Fixes: #583 

**Note:** I tried to apply conditional checks where I felt would be necessary. There are high chances that I have missed them out at important places, please keep this in mind while reviewing.

TODO:

- [x] E2E for TimescaleDB-OSS version


antekresic (changes requested):
Missing check in `config_maintenance_jobs`, uses user-defined actions which are not in OSS:
https://docs.timescale.com/api/latest/actions/

cevian (changes requested):

cevian (approved):
Looks good. I do however want to add a big WARNING when promscale first starts up if we are using OSS to warn people: 
```
WARNING: Using the Apache2 version of TimescaleDB. This version does not include compression and thus performance and disk usage will be significantly negatively effected.
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -27,13 +27,35 @@ LANGUAGE SQL STABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_default_retention_period() TO prom_reader;
 
 CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.is_timescaledb_installed()
-    RETURNS BOOLEAN
+RETURNS BOOLEAN
 AS $func$
     SELECT count(*) > 0 FROM pg_extension WHERE extname='timescaledb';
-$func$
+    $func$
```
antekresic, line 33: Minor nit: why add whitespace here? It even makes things harder to read.

pkg/runner/client.go
```diff
@@ -177,6 +186,26 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
...
 }
 
+func isTimescaleDBOSS(conn *pgx.Conn) (bool, error) {
+	var (
+		isTimescaleDB bool
+		isLicenseOSS  bool
+	)
+	err := conn.QueryRow(context.Background(), "SELECT "+schema.Catalog+".is_timescaledb_installed()").Scan(&isTimescaleDB)
+	if err != nil {
+		return false, fmt.Errorf("fetching if TimescaleDB is installed: %w", err)
```
antekresic, line 196:
```suggestion
		return false, fmt.Errorf("error fetching if TimescaleDB is installed: %w", err)
```

pkg/runner/client.go
```diff
@@ -177,6 +186,26 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
...
+	}
+	if !isTimescaleDB {
+		// Return false so that we don't warn for OSS TimescaleDB.
+		return false, nil
+	}
+	err = conn.QueryRow(context.Background(), "SELECT "+schema.Catalog+".is_timescaledb_oss()").Scan(&isLicenseOSS)
+	if err != nil {
+		return false, fmt.Errorf("fetching TimescaleDB license: %w", err)
```
antekresic, line 204:
```suggestion
		return false, fmt.Errorf("error fetching TimescaleDB license: %w", err)
```

antekresic (approved):
Some minor nits, should be good to go.

Issue #660, opened 2021-06-08
Possible deadlock in `delete_expired_series` SQL Function


cevian:
delete_expired_series has a lock escalation issue. The line 
```
SELECT last_update_time, current_epoch-4, current_epoch+1
        FROM SCHEMA_CATALOG.ids_epoch LIMIT 1
        INTO last_epoch_time, deletion_epoch, next_epoch;
```
is followed later by 
```
 LOCK TABLE SCHEMA_CATALOG.ids_epoch IN ACCESS EXCLUSIVE MODE;
```
This escalates the lock and could cause deadlocks if run concurrently - e.g. both the first and second job pass the first line
and now neither can acquire the exclusive lock.

We should fix this by getting the current_epoch and  last_update_time in a separate txn.
This can be done by either:
- Changing delete_expired_series to a procedure (not function) and putting a commit after the SELECT 
- or
- Passing in current_epoch and  last_update_time to delete_expired_series and fetching those values in the first transaction of drop_metric_chunks.

PR #662, opened 2021-06-11
Fix possible deadlock in delete_expired_series().

Harkishen-Singh:

Fixes: #660


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1456,6 +1455,8 @@ BEGIN
         FROM SCHEMA_CATALOG.ids_epoch LIMIT 1
         INTO last_epoch_time, deletion_epoch, next_epoch;
 
+    COMMIT;
```
cevian, line 1458: Please add a comment
```suggestion
    -- We commit here to avoid a lock escalation on the ids_epoch table when we later take an access exclusive lock 
    -- on it
    COMMIT;
```

cevian (approved):

Harkishen-Singh:
@cevian I am getting `invalid transaction` in CI.

Is this because of some invalid condition in nested procedures?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1440,24 +1440,19 @@ SET search_path = pg_temp;
...
+    next_epoch BIGINT;
+    deletion_epoch BIGINT;
 BEGIN
-    -- technically we can delete any ID <= current_epoch - 1
-    -- but it's always safe to leave them around for a bit longer
-    SELECT last_update_time, current_epoch-4, current_epoch+1
-        FROM SCHEMA_CATALOG.ids_epoch LIMIT 1
-        INTO last_epoch_time, deletion_epoch, next_epoch;
+    next_epoch := present_epoch + 1;
+    deletion_epoch := present_epoch - 4;
```
cevian, line 1452:
```suggestion
    -- technically we can delete any ID <= current_epoch - 1
    -- but it's always safe to leave them around for a bit longer
    deletion_epoch := present_epoch - 4;
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1609,7 +1602,9 @@ BEGIN
     IF older_than IS NULL THEN
         -- even though there are no new Ids in need of deletion,
         -- we may still have old ones to delete
-        PERFORM SCHEMA_CATALOG.delete_expired_series(metric_table, ran_at);
+        SELECT current_epoch, last_update_time INTO present_epoch, last_updated FROM
```
cevian, line 1605: This has to happen in a different transaction than the call to delete_expired_series. I suggest moving this up to right before the commit on line 1600

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1622,7 +1617,9 @@ BEGIN
     COMMIT;
 
     -- transaction 4
-        PERFORM SCHEMA_CATALOG.delete_expired_series(metric_table, ran_at);
+        SELECT current_epoch, last_update_time INTO present_epoch, last_updated FROM
```
cevian, line 1620: This has to happen before the previous commit

cevian (changes requested):

cevian (approved):

Harkishen-Singh:
@cevian can you please see why are upgrade tests failing?

PR #666, opened 2021-06-18
Vector selector pushdown

cevian:


pkg/internal/testhelpers/containers.go
```diff
@@ -36,6 +36,8 @@ const (
 
 	Superuser   = true
 	NoSuperuser = false
+
+	LatestDBWithPromscaleImageBase = "cevian/promscale-extension"
```
cevian, line 40: This needs to change before merge but allows the CI to test using the new extension

pkg/pgmodel/querier/querier.go
```diff
@@ -30,13 +31,20 @@ type Reader interface {
 	Read(*prompb.ReadRequest) (*prompb.ReadResponse, error)
 }
 
+type QueryHints struct {
```
cevian, line 34: note we don't control *storage.SelectHints so we add our own hints that are necessary for our pushdowns

antekresic (approved):

pkg/pgmodel/querier/query_builder.go
```diff
@@ -379,13 +411,55 @@ type aggregators struct {
...
+			if hints != nil &&
+				hints.Step > 0 &&
+				hints.Range == 0 &&
+				!calledByTimestamp(path) &&
+				n.OriginalOffset == time.Duration(0) &&
+				n.Offset == time.Duration(0) &&
+				!hasSubquery(path) &&
+				vectorSelectorExtensionRange(extension.PromscaleExtensionVersion) {
+				qf := aggregators{
+					timeClause:  "ARRAY(SELECT generate_series($%d::timestamptz, $%d::timestamptz, $%d))",
```
Harkishen-Singh, line 447: It is difficult for me to understand how it helps. Can you please write a brief summary so that it's easy to follow?

Harkishen-Singh (approved):
Can be merged, but some comments can be added in related parts that explain how performance of vector selectors is improved

PR #667, opened 2021-06-22
Log SQL queries & time consumed in debug level. 

VineethReddy02:
Log time consumed by SQL queries in debug level & drop HA debug logs noise.

### DIFF
--- a/pkg/ha/filter.go
+++ b/pkg/ha/filter.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	"github.com/prometheus/common/model"
-	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
@@ -60,7 +59,6 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
 	if !minT.Before(leaseStart) {
 		if !allowInsert {
 			wr.Timeseries = wr.Timeseries[:0]
-			log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert", "replica", replicaName)
 		}
 		return nil
 	}
@@ -80,7 +78,6 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
 	case !hasBackfill && !allowInsert:
 		// No data to insert.
 		wr.Timeseries = wr.Timeseries[:0]
-		log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert", "replica", replicaName)
 	default:
 		// This case covers the instance when we have backfill and data in current lease to ingest.
 		// The data has already been filtered out so there is nothing to do.
--- a/pkg/pgxconn/pgx_conn.go
+++ b/pkg/pgxconn/pgx_conn.go
@@ -7,6 +7,7 @@ package pgxconn
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgconn"
@@ -39,8 +40,8 @@ type PgxRows interface {
 
 // This is only used by querier to log the time consumed
 // by SQL queries executed using Query(), QueryRow()
-func NewQuerierPgxConn(pool *pgxpool.Pool) PgxConn {
-	return &querierConnImpl{
+func NewQueryLoggingPgxConn(pool *pgxpool.Pool) PgxConn {
+	return &loggingConnImpl{
 		connImpl: connImpl{Conn: pool},
 		Conn:     pool,
 	}
@@ -52,7 +53,7 @@ func NewPgxConn(pool *pgxpool.Pool) PgxConn {
 	}
 }
 
-type querierConnImpl struct {
+type loggingConnImpl struct {
 	connImpl
 	Conn *pgxpool.Pool
 }
@@ -69,14 +70,14 @@ type pgxRows struct {
 	logged    bool
 }
 
-func (p *querierConnImpl) Query(ctx context.Context, sql string, args ...interface{}) (PgxRows, error) {
+func (p *loggingConnImpl) Query(ctx context.Context, sql string, args ...interface{}) (PgxRows, error) {
 	startTime := time.Now()
 	rows, err := p.Conn.Query(ctx, sql, args...)
 	return &pgxRows{Rows: rows, sqlQuery: sql, args: args, startTime: startTime}, err
 }
 
-func (p *querierConnImpl) QueryRow(ctx context.Context, sql string, args ...interface{}) pgx.Row {
-	defer LogQueryStats(sql, time.Time{}, args...)()
+func (p *loggingConnImpl) QueryRow(ctx context.Context, sql string, args ...interface{}) pgx.Row {
+	defer logQueryStats(sql, time.Time{}, args...)()
 	return p.Conn.QueryRow(ctx, sql, args...)
 }
 
@@ -92,19 +93,19 @@ func (p *pgxRows) Next() bool {
 	if !p.logged {
 		p.logged = true
 		res := p.Rows.Next()
-		LogQueryStats(p.sqlQuery, p.startTime, p.args)()
+		logQueryStats(p.sqlQuery, p.startTime, p.args)()
 		return res
 	}
 	return p.Rows.Next()
 }
 
 // calc SQL query execution time
-func LogQueryStats(sql string, startTime time.Time, args ...interface{}) func() {
+func logQueryStats(sql string, startTime time.Time, args ...interface{}) func() {
 	if startTime.IsZero() {
 		startTime = time.Now()
 	}
 	return func() {
-		log.Debug("msg", "SQL query timing", "query", sql, "args", fmt.Sprintf("%v", args...), "time", time.Since(startTime))
+		log.Debug("msg", "SQL query timing", "query", filterIndentChars(sql), "args", fmt.Sprintf("%v", args...), "time", time.Since(startTime))
 	}
 }
 
@@ -134,4 +135,16 @@ func (p *connImpl) NewBatch() PgxBatch {
 
 func (p *connImpl) SendBatch(ctx context.Context, b PgxBatch) (pgx.BatchResults, error) {
 	return p.Conn.SendBatch(ctx, b.(*pgx.Batch)), nil
+}
+
+// filters out indentation characters from the
+// SQL query for better query logging
+func filterIndentChars(query string) string {
+	dropChars := []string{"\n", "\t", "\""}
+	query = strings.ReplaceAll(query, "\n\t", " ")
+	for _, c := range dropChars {
+		query = strings.ReplaceAll(query, c, "")
+	}
+
+	return query
 }
--- a/pkg/pgxconn/pgx_conn.go
+++ b/pkg/pgxconn/pgx_conn.go
@@ -6,10 +6,13 @@ package pgxconn
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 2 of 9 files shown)


pkg/pgxconn/pgx_conn.go
```diff
@@ -37,21 +47,63 @@ type connImpl struct {
...
 
-func (p *connImpl) Exec(ctx context.Context, sql string, arguments ...interface{}) (pgconn.CommandTag, error) {
-	return p.Conn.Exec(ctx, sql, arguments...)
+func (p *pgxRows) Next() bool {
+	if !p.loggged {
+		p.loggged = true
+		LogQueryStats(p.sqlQuery, p.startTime, p.args)()
+	}
+	return p.Rows.Next()
+}
```
cevian, line 70: We have to cann p.Rows.Next before logging to make sure the timing is right
```suggestion
func (p *pgxRows) Next() bool { 
        //The query fetch is async and happens on the first call to next. 
        //so to get timing right, we log timing after first Next() call.
	if !p.loggged {
		p.loggged = true
	         res := p.Rows.Next()
		LogQueryStats(p.sqlQuery, p.startTime, p.args)()
		return res
	}
	return p.Rows.Next()
}
```

pkg/pgxconn/pgx_conn.go
```diff
@@ -37,21 +47,63 @@ type connImpl struct {
...
+	return p.Rows.Scan(dest...)
+}
+
+func (p *pgxRows) Err() error {
+	return p.Rows.Err()
+}
+
+func (p *pgxRows) Close() {
+	p.Rows.Close()
+}
```
cevian, line 82: are these necessary? doesn't go struct embedding create these automatically?

pkg/pgxconn/pgx_conn.go
```diff
@@ -37,21 +47,63 @@ type connImpl struct {
...
+	if startTime.IsZero() {
+		startTime = time.Now()
+	}
+	return func() {
+		log.Debug("msg", fmt.Sprintf("time taken by SQL query: %s with args: %v is %v", sql, args, time.Since(startTime)))
+	}
+}
+
+func (p *connImpl) Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error) {
+	defer LogQueryStats(sql, time.Time{}, args...)()
```
cevian, line 95: I do worry creating this on all connImpl will log the ingest queries as well as the promql queries. I think we probably want just the promql queries logged since otherwise the ingest will just fill the logs. So I'd suggest either having two different connImpl's (one that logs one that doesn't) or add a log bool to connImpl to turn logging on and off

pkg/pgxconn/pgx_conn.go
```diff
@@ -37,21 +47,63 @@ type connImpl struct {
...
+	p.Rows.Close()
+}
+
+// calc SQL query execution time
+func LogQueryStats(sql string, startTime time.Time, args ...interface{}) func() {
+	if startTime.IsZero() {
+		startTime = time.Now()
+	}
+	return func() {
+		log.Debug("msg", fmt.Sprintf("time taken by SQL query: %s with args: %v is %v", sql, args, time.Since(startTime)))
```
cevian, line 90:
```suggestion
		log.Debug("msg", "SQL query timing", "query", sql, "args", args, "time",  time.Since(startTime))
```

cevian (changes requested):

pkg/ha/filter.go
```diff
@@ -60,7 +59,6 @@ func (h *Filter) Process(_ *http.Request, wr *prompb.WriteRequest) error {
 	if !minT.Before(leaseStart) {
 		if !allowInsert {
 			wr.Timeseries = wr.Timeseries[:0]
-			log.Debug("msg", "the samples aren't from the leader prom instance. skipping the insert", "replica", replicaName)
```
Harkishen-Singh, line 63: Why are we dropping this and the one below this?

pkg/pgxconn/pgx_conn.go
Harkishen-Singh: Agrs are gonna be long. Lets do

```suggestion
		log.Debug("msg", "SQL query timing", "duration", time.Since(startTime), "query", sql, "args", args)
```

Harkishen-Singh: Also, @VineethReddy02 can you please paste a sample output of how it looks?

pkg/pgxconn/pgx_conn.go
```diff
@@ -37,21 +47,63 @@ type connImpl struct {
...
 	conn := p.Conn
 	p.Conn = nil
 	conn.Close()
 }
 
-func (p *connImpl) Exec(ctx context.Context, sql string, arguments ...interface{}) (pgconn.CommandTag, error) {
-	return p.Conn.Exec(ctx, sql, arguments...)
+func (p *pgxRows) Next() bool {
+	if !p.loggged {
+		p.loggged = true
```
Harkishen-Singh, line 66:
```suggestion
	if !p.logged {
		p.logged = true
```
I think it should be `hasLogged`.

Harkishen-Singh (changes requested):

Harkishen-Singh:
Also, if I am not wrong, we are also supposed to log SQL time duration and overall PromQL time duration in this PR.

pkg/pgxconn/pgx_conn.go
VineethReddy02: I think the time consumed at the end will b easy to analyze the time taken quickly rather than finding it in the middle of the log lines.

pkg/ha/filter.go
VineethReddy02: To drop the noise from HA debug logs... When running Promscale in debug mode I see too many debug logs from HA. So we decided to drop them.

VineethReddy02:
In this PR we will only log SQL query duration. :) 
@cevian Am I right?

VineethReddy02:
The sample log output looks like

`
level=debug ts=2021-06-28T08:14:59.046Z caller=pgx_conn.go:106 msg="SQL query timing" query="SELECT series.labels,  result.time_array, result.value_array\n\tFROM \"prom_data_series\".\"go_gc_duration_seconds\" series\n\tINNER JOIN LATERAL (\n\t\tSELECT array_agg(time) as time_array, array_agg(value) as value_array\n\t\tFROM\n\t\t(\n\t\t\tSELECT time, value\n\t\t\tFROM \"prom_data\".\"go_gc_duration_seconds\" metric\n\t\t\tWHERE metric.series_id = series.id\n\t\t\tAND time >= '2021-06-28T07:09:45Z'\n\t\t\tAND time <= '2021-06-28T08:14:45Z'\n\t\t\tORDER BY time\n\t\t) as time_ordered_rows\n\t) as result ON (result.value_array is not null)\n\tWHERE\n\t     labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $1 and l.value = $2) AND labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $3 and l.value = $4)" args="[instance localhost:9090 job prometheus]" time=20.93833ms
`

VineethReddy02:
After filtering out indent chars the log looks like:

`
level=debug ts=2021-06-30T15:50:07.609Z caller=pgx_conn.go:108 msg="SQL query timing" query="SELECT series.labels,  result.time_array, result.value_array FROM prom_data_series.go_goroutines series INNER JOIN LATERAL ( SELECT array_agg(time) as time_array, array_agg(value) as value_array FROM ( SELECT time, value FROM prom_data.go_goroutines metric WHERE metric.series_id = series.id AND time >= '2021-06-30T03:45:00Z' AND time <= '2021-06-30T15:50:00Z' ORDER BY time ) as time_ordered_rows ) as result ON (result.value_array is not null) WHERE      labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $1 and l.value = $2)" args="[instance localhost:9090]" time=6.417936ms
`

pkg/pgxconn/pgx_conn.go
```diff
@@ -19,35 +23,97 @@ type PgxBatch interface {
...
+
 func NewPgxConn(pool *pgxpool.Pool) PgxConn {
 	return &connImpl{
 		Conn: pool,
 	}
 }
 
+type loggingConnImpl struct {
+	connImpl
+	Conn *pgxpool.Pool
```
cevian, line 58: you should need this because connImpl already has it you should be able to access via p.connImpl.Conn

pkg/pgxconn/pgx_conn.go
```diff
@@ -19,35 +23,97 @@ type PgxBatch interface {
...
+	pgx.Rows
+	sqlQuery  string
+	args      []interface{}
+	startTime time.Time
+	logged    bool
+}
+
+func (p *loggingConnImpl) Query(ctx context.Context, sql string, args ...interface{}) (PgxRows, error) {
+	startTime := time.Now()
+	rows, err := p.Conn.Query(ctx, sql, args...)
```
cevian, line 75: e.g. you could do 
```suggestion
	rows, err := p.connImpl.Conn.Query(ctx, sql, args...)
```

pkg/pgxconn/pgx_conn.go
```diff
@@ -19,35 +23,97 @@ type PgxBatch interface {
...
+type loggingConnImpl struct {
+	connImpl
+	Conn *pgxpool.Pool
+}
+
 type connImpl struct {
 	Conn *pgxpool.Pool
 }
 
+type pgxRows struct {
```
cevian, line 65:
```suggestion
type loggingPgxRows struct {
```

pkg/pgxconn/pgx_conn.go
```diff
@@ -19,35 +23,97 @@ type PgxBatch interface {
...
+
 func (p *connImpl) Close() {
 	conn := p.Conn
 	p.Conn = nil
 	conn.Close()
 }
 
-func (p *connImpl) Exec(ctx context.Context, sql string, arguments ...interface{}) (pgconn.CommandTag, error) {
-	return p.Conn.Exec(ctx, sql, arguments...)
+func (p *pgxRows) Next() bool {
```
cevian, line 90: nit adding functions from the same class in the same section of the file would be better. right now we have connImpl methods interspersed with pgxRows and loggingConnImpl

pkg/pgxconn/pgx_conn.go
```diff
@@ -70,3 +136,15 @@ func (p *connImpl) NewBatch() PgxBatch {
 func (p *connImpl) SendBatch(ctx context.Context, b PgxBatch) (pgx.BatchResults, error) {
 	return p.Conn.SendBatch(ctx, b.(*pgx.Batch)), nil
 }
+
+// filters out indentation characters from the
+// SQL query for better query logging
+func filterIndentChars(query string) string {
+	dropChars := []string{"\n", "\t", "\""}
```
cevian, line 143: it's probably better to use a replacer here https://golang.org/pkg/strings/#Replacer

pkg/pgxconn/pgx_conn.go
```diff
@@ -19,35 +23,97 @@ type PgxBatch interface {
...
+	if !p.logged {
+		p.logged = true
+		res := p.Rows.Next()
+		logQueryStats(p.sqlQuery, p.startTime, p.args)()
+		return res
+	}
+	return p.Rows.Next()
+}
+
+// calc SQL query execution time
```
cevian, line 102: improve the comment

cevian (changes requested):
Almost there. some nits

pkg/pgxconn/pgx_conn.go
VineethReddy02: Ahh, Again missed the power of embedded types. :)

cevian (approved):

Harkishen-Singh (approved):

Issue #668, opened 2021-06-22
Promscale loosing datapoints


mjf:
It looks like that Promscale is (under certain circumstances) loosing datapoints when writing in Postgres because of the following issue (Prometheus that is using Promscale as remote read/write target in this case can return all datapoints, the datapoints that are missing are corresponding to the following errors that are literaly flooding Prometheus, Promscale and Postgres logs)...

Promscale logs:

``` json
level=warn ts=2021-06-22T11:56:26.925Z caller=copier.go:159 msg="unexpected postgres error while inserting to " err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)"
level=warn ts=2021-06-22T11:56:27.026Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=57
level=warn ts=2021-06-22T11:56:27.033Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=47
level=warn ts=2021-06-22T11:56:27.052Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=42
level=warn ts=2021-06-22T11:56:27.055Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=56
level=warn ts=2021-06-22T11:56:27.055Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=49
level=warn ts=2021-06-22T11:56:27.055Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=54
level=warn ts=2021-06-22T11:56:27.055Z caller=write.go:218 msg="Error sending samples to remote storage" err="ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)" num_samples=47
```

Postgres (TimescaleDB; without Promscale extension) logs:

``` json
[...]

PR #670, opened 2021-06-22
Fix bad check for error in MetricTableName

cevian:
Previously, bad error checking logic in MetricTableName could
have returned a blank table name on error which eventually
led to the following postgres error:

"ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)"

As reported in issue #668

This fixes that and adds defensive checks to avoid blank table name.

It's not clear what the underlying error was though.

### DIFF
--- a/pkg/pgmodel/ingestor/metric_batcher.go
+++ b/pkg/pgmodel/ingestor/metric_batcher.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/jackc/pgtype"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel/cache"
 	"github.com/timescale/promscale/pkg/pgmodel/common/errors"
 	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
@@ -18,7 +19,10 @@ import (
 	"github.com/timescale/promscale/pkg/pgxconn"
 )
 
-const seriesInsertSQL = "SELECT (_prom_catalog.get_or_create_series_id_for_label_array($1, l.elem)).series_id, l.nr FROM unnest($2::prom_api.label_array[]) WITH ORDINALITY l(elem, nr) ORDER BY l.elem"
+const (
+	seriesInsertSQL                 = "SELECT (_prom_catalog.get_or_create_series_id_for_label_array($1, l.elem)).series_id, l.nr FROM unnest($2::prom_api.label_array[]) WITH ORDINALITY l(elem, nr) ORDER BY l.elem"
+	getCreateMetricsTableWithNewSQL = "SELECT table_name, possibly_new FROM " + schema.Catalog + ".get_or_create_metric_table_name($1)"
+)
 
 type metricBatcher struct {
 	conn            pgxconn.PgxConn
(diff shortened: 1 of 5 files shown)


pkg/pgmodel/model/interface.go
```diff
@@ -49,7 +49,7 @@ func MetricTableName(conn pgxconn.PgxConn, metric string) (string, bool, error)
 	var possiblyNew bool
 	defer res.Close()
 	if !res.Next() {
-		if res.Err() != nil {
+		if err := res.Err(); err != nil {
 			return "", true, err
```
cevian, line 53: note this is where the old code could have returned a nil error with a blank tableName

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -72,7 +73,13 @@ func runMetricBatcher(conn pgxconn.PgxConn,
...
 		tableName, err = initializeMetricBatcher(conn, metricName, completeMetricCreationSignal, metricTableNames)
 		if err != nil {
-			firstReq.reportResult(fmt.Errorf("initializing the insert routine has failed with %w", err))
+			err := fmt.Errorf("initializing the insert routine for metric %v has failed with %w", metricName, err)
+			log.Error("msg", err)
+			firstReq.reportResult(err)
+		} else if tableName == "" {
+			err := fmt.Errorf("initializing the insert routine has returned an empty table name for metric name %v", metricName)
+			log.Error("msg", err)
+			firstReq.reportResult(err)
```
antekresic, line 82: Instead of doing it like this, why not return an error from `model.MetricTableName` in case table name is empty? That looks like a cleaner approach.

antekresic (changes requested):

pkg/pgmodel/ingestor/metric_batcher.go
Harkishen-Singh: I agree here.

pkg/pgmodel/model/interface.go
```diff
@@ -49,7 +49,7 @@ func MetricTableName(conn pgxconn.PgxConn, metric string) (string, bool, error)
 	var possiblyNew bool
 	defer res.Close()
 	if !res.Next() {
-		if res.Err() != nil {
+		if err := res.Err(); err != nil {
```
Harkishen-Singh, line 52: Can we check this before this block and if no error, then verify if metric_name is empty. if yes, send a metric not exists error by reusing some previously declared error.

pkg/pgmodel/model/interface.go
```diff
@@ -59,6 +59,10 @@ func MetricTableName(conn pgxconn.PgxConn, metric string) (string, bool, error)
 		return "", true, err
 	}
 
+	if err := res.Err(); err != nil {
```
Harkishen-Singh, line 62: If we are going to check `res.Err` twice (when res.Next() is false), why not check after line 50?

pkg/tests/end_to_end_tests/concurrent_sql_test.go
```diff
@@ -23,11 +23,11 @@ import (
 func testConcurrentMetricTable(t testing.TB, db *pgxpool.Pool, metricName string) int64 {
```
Harkishen-Singh, line 23: Can we have a test case where metric name is `""`?

Harkishen-Singh (changes requested):
Can we have a test case where the metric name is `""`?

pkg/pgmodel/model/interface.go
cevian: The second check is if res.Next is true

pkg/pgmodel/model/interface.go
cevian: no res.Error can change with res.Next and res.Scan calls

pkg/pgmodel/model/interface.go
cevian: This is if Scan has an error

pkg/pgmodel/ingestor/metric_batcher.go
cevian: ok changed.

antekresic (approved):

pkg/pgmodel/ingestor/metric_batcher_test.go
```diff
@@ -0,0 +1,78 @@
...
+			},
+		},
+		{
+			name:        "empty table name",
+			tableName:   "res2",
+			errExpected: true,
+			sqlQueries: []model.SqlQuery{
+				{
+					Sql:     "SELECT table_name, possibly_new FROM _prom_catalog.get_or_create_metric_table_name($1)",
+					Args:    []interface{}{"t1"},
```
Harkishen-Singh, line 59: I do not understand sql recorder perfectly, but shouldn't this be emty string?

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -72,7 +112,9 @@ func runMetricBatcher(conn pgxconn.PgxConn,
 		var err error
 		tableName, err = initializeMetricBatcher(conn, metricName, completeMetricCreationSignal, metricTableNames)
 		if err != nil {
-			firstReq.reportResult(fmt.Errorf("initializing the insert routine has failed with %w", err))
+			err := fmt.Errorf("initializing the insert routine for metric %v has failed with %w", metricName, err)
+			log.Error("msg", err)
```
Harkishen-Singh, line 116: Won't this log the same error twice, which was already being logged in the line below?

pkg/pgmodel/ingestor/metric_batcher_test.go
cevian: No this is the argument - the metric name - not the table name. The response table name is on the next line and it's empty

pkg/pgmodel/ingestor/metric_batcher.go
cevian: No, the next line will return an error for the request but not necessarily log it.

Harkishen-Singh (approved):

PR #673, opened 2021-06-25
Pushdown rate and increase optimizations

cevian:
These function were actually supported in the extension before
but weren't being pushed down. This commit fixes that. 

Depends and is rebased on #666

### DIFF
--- a/pkg/pgmodel/querier/query_builder.go
+++ b/pkg/pgmodel/querier/query_builder.go
@@ -459,7 +459,7 @@ func calledByTimestamp(path []parser.Node) bool {
 var vectorSelectorExtensionRange = semver.MustParseRange(">= 0.1.3-beta")
 var rateIncreaseExtensionRange = semver.MustParseRange(">= 0.1.3-beta")
 
-func callAggregator(hints *storage.SelectHints, funcName string) *aggregators {
+func callAggregator(hints *storage.SelectHints, funcName string) (*aggregators, error) {
 	queryStart := hints.Start + hints.Range
 	queryEnd := hints.End
 	stepDuration := time.Second
@@ -469,7 +469,7 @@ func callAggregator(hints *storage.SelectHints, funcName string) *aggregators {
 		stepDuration = time.Duration(hints.Step) * time.Millisecond
 	} else {
 		if queryStart != queryEnd {
-			panic("query start should equal query end")
+			return nil, fmt.Errorf("query start should equal query end")
 		}
 	}
 	qf := aggregators{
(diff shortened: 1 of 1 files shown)


pkg/pgmodel/querier/query_builder.go
```diff
@@ -427,6 +427,30 @@ func calledByTimestamp(path []parser.Node) bool {
...
+		timeParams:  []interface{}{model.Time(queryStart).Time(), model.Time(queryEnd).Time(), stepDuration},
+		valueClause: "prom_" + funcName + "($%d, $%d,$%d, $%d, time, value)",
+		valueParams: []interface{}{model.Time(hints.Start).Time(), model.Time(queryEnd).Time(), int64(stepDuration.Milliseconds()), int64(rangeDuration.Milliseconds())},
+		unOrdered:   false,
+	}
+	return &qf
+}
 
 /* The path is the list of ancestors (direct parent last) returned node is the most-ancestral node processed by the pushdown */
 func getAggregators(hints *storage.SelectHints, qh *QueryHints, path []parser.Node) (*aggregators, parser.Node, error) {
```
antekresic, line 456: This function needs a bit of refactoring. 

I suggest handling the various checks that overlap for all pushdowns first and returning the no optimization path as soon as possible based on them. Things like is the extension installed or if there is a subquery should be checked as soon as possible.

Should make the rest of the code much more easier to read and maintain.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -427,6 +427,30 @@ func calledByTimestamp(path []parser.Node) bool {
...
+	queryStart := hints.Start + hints.Range
+	queryEnd := hints.End
+	stepDuration := time.Second
+	rangeDuration := time.Duration(hints.Range) * time.Millisecond
+
+	if hints.Step > 0 {
+		stepDuration = time.Duration(hints.Step) * time.Millisecond
+	} else {
+		if queryStart != queryEnd {
+			panic("query start should equal query end")
```
antekresic, line 442: Should we be panicking here? This will kill Promscale just because hints details were unexpected.

antekresic (changes requested):

Harkishen-Singh (approved):
Looks good except, I agree on Ante's comments.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -457,81 +457,92 @@ func calledByTimestamp(path []parser.Node) bool {
...
-					}
+
+	//switch on the current node being processed
+	switch n := qh.CurrentNode.(type) {
+	case *parser.VectorSelector:
+		/* Try to optimize the aggregation first since that would return less data than a plain vector selector */
+		if len(path) >= 2 {
+			//switch on the 2nd-to-last last path node
+			node := path[len(path)-2]
+			switch n := node.(type) {
```
antekresic, line 506: Minor nit: seems the code would be a bit cleaner if we used switch case on `n.Func.Name` and an if clause for the `node.(type)`. Hopefully, in the future, we will add support for more pushdown functions so this change makes even more sense.

antekresic (approved):
I have a minor suggestion that isn't crucial. Otherwise, this looks good to me :+1:

PR #674, opened 2021-06-28
Add support for metric views to enable downsampling features

antekresic:
This change gives users the ability to register metric views which
will represent a new form of metric data in the system as a new metric.
Views have to be derived from a metric that is already in the system
and cannot be ingested to. Querying the metric view data can be done
only by specifying the exact metric view name in the label matcher.
This is a first step into adding a robust downsampling solution for
Promscale, later steps will remove some of the restrictions and give
more freedom in how the metric view data is represented.


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -204,27 +212,35 @@ DECLARE
...
-   EXECUTE format('CREATE TABLE SCHEMA_DATA.%I(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL) WITH (autovacuum_vacuum_threshold = 50000, autovacuum_analyze_threshold = 50000)',
-                    NEW.table_name);
-   EXECUTE format('GRANT SELECT ON TABLE SCHEMA_DATA.%I TO prom_reader', NEW.table_name);
-   EXECUTE format('GRANT SELECT, INSERT ON TABLE SCHEMA_DATA.%I TO prom_writer', NEW.table_name);
-   EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE SCHEMA_DATA.%I TO prom_modifier', NEW.table_name);
-   EXECUTE format('CREATE UNIQUE INDEX data_series_id_time_%s ON SCHEMA_DATA.%I (series_id, time) INCLUDE (value)',
-                    NEW.id, NEW.table_name);
+
+   -- Note: if the inserted metric is a view, handle the necessary permissions and exit.
+   IF NEW.is_view THEN
```
Harkishen-Singh, line 217: Isn't it misleading in terms of naming of this function? If our task is to just give permissions to te views, wouldn't it be better to have another function as `assign_permissions_to_metric_views`

pkg/migrations/sql/preinstall/003-tables.sql
```diff
@@ -94,13 +94,16 @@ CREATE TABLE SCHEMA_CATALOG.metric (
     id SERIAL PRIMARY KEY,
     metric_name text NOT NULL,
     table_name name NOT NULL,
+    table_schema name NOT NULL DEFAULT 'SCHEMA_DATA',
+    series_table name DEFAULT NULL, -- NULL signifies the series table is named the same as metric table name
```
Harkishen-Singh, line 98: I couldn't relate, what is a `series_table` here?

pkg/pgmodel/cache/cache.go
```diff
@@ -51,26 +56,68 @@ type MetricNameCache struct {
...
+		return "", "", "", fmt.Errorf("invalid number of values stored")
+	}
+
+	return values[0], values[1], values[2], nil
+}
+
 // Get fetches the table name for specified metric.
-func (m *MetricNameCache) Get(metric string) (string, error) {
-	result, ok := m.Metrics.Get(metric)
+func (m *MetricNameCache) Get(schema, metric string) (string, string, string, error) {
```
Harkishen-Singh, line 92: I think named returns will be better here in terms of readability.

pkg/pgmodel/cache/cache.go
```diff
@@ -51,26 +56,68 @@ type MetricNameCache struct {
...
+	keyLen, key := m.key(schema, metric)
+	//size includes an 8-byte overhead for each string
+	m.Metrics.Insert(key, val, uint64(keyLen+valLen+16))
+	return nil
 }
 
-// Set stores table name for specified metric.
-func (m *MetricNameCache) Set(metric string, tableName string) error {
+// Set stores metric info cache for specified metric with default schema.
+func (m *MetricNameCache) SetDefaultSchema(schema, metric, tableName, seriesTable string) error {
```
Harkishen-Singh, line 115: Can we define what a default-schema is, here? I assume it's something that will always be static which in that case, why do we accept a `schema` as param if it's gonna be the default?

pkg/pgmodel/querier/query_builder.go
```diff
@@ -164,6 +164,10 @@ func BuildSubQueries(matchers []*labels.Matcher) (*clauseBuilder, error) {
 				cb.SetMetricName(m.Value)
 				continue
 			}
+			if m.Name == pgmodel.SchemaNameLabelName {
+				cb.SetMetricName(m.Value)
+				continue
+			}
 			sq := subQueryEQ
```
Harkishen-Singh, line 171: Are we not blocking people with using special `regexes` or `not` conditions on schema label name as discussed?

pkg/pgmodel/querier/series_set.go
```diff
@@ -86,6 +87,17 @@ func (p *pgxSeriesSet) At() storage.Series {
 			log.Error("err", err)
 			return nil
 		}
+
+		if schemaLabelName, schemaLabelValue := getSchemaLabel(row.schema); schemaLabelName != "" {
```
Harkishen-Singh, line 91: A similar code is repeated in `buildTimeSeries()`.  Can/Should we abstract this down to the function that is responsible for this task? IMHO, It will look the code clean, otherwise its more branches at the moment.

Harkishen-Singh (changes requested):
Initial pass. SQL look still todo.

Some nits that will be helpful:
1. I am finding hard to follow what a series_table is. can you please give comments for that?
2. On a high level, SQL code looks like a reuse of existing code. In such condition, should we rename the function names as most functions are given name for `.*table.*`, which is no longer true. Example: `make_metric_table()`.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1931,6 +1950,72 @@ SET search_path = pg_temp;
...
+    WHERE view_schema = register_metric_view.schema_name
+    AND view_table_usage.view_name = register_metric_view.view_name
+    AND table_schema = 'SCHEMA_DATA';
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'view not based on a metric table from SCHEMA_DATA schema';
+    END IF;
+
+    -- check if the view columns match columns from the underlying metric table
+    PERFORM m.column_name
```
cevian, line 1985: hmm maybe it's better to check 3 things:
- there is a column called series_id (type bigint)
- there is a column called time (type timestamptz)
- there is a column called value (type float)

I don't think we really care if there are other columns or anything and I think this check would be more forward compatible.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1931,6 +1950,72 @@ SET search_path = pg_temp;
...
+        RAISE EXCEPTION 'view columns must match names and data types with the underlying metric table';
+    END IF;
+
+    -- insert into metric table
+    INSERT INTO SCHEMA_CATALOG.metric (metric_name, table_name, table_schema, series_table, is_view, creation_completed)
+    VALUES (register_metric_view.view_name, register_metric_view.view_name, register_metric_view.schema_name, metric_table_name, true, true)
+    ON CONFLICT DO NOTHING;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'metric with the same name and schema already exists, could not register';
```
cevian, line 2005: we probably want an optional argument to this function  "if_not_exists => false" and we don't throw this error if it's true. This would allow people to use this function in a idempotent way

pkg/pgmodel/querier/query_builder.go
```diff
@@ -294,6 +323,16 @@ func buildTimeSeries(rows []timescaleRow, lr lreader.LabelsReader) ([]*prompb.Ti
 			return nil, err
 		}
 
+		if schemaLabelName, schemaLabelValue := getSchemaLabel(row.schema); schemaLabelName != "" {
+			// If its a custom schema, we need to update the metric name as well.
+			for i := range promLabels {
+				if promLabels[i].Name == pgmodel.MetricNameLabelName {
+					promLabels[i].Value = row.metric
```
cevian, line 330:
```suggestion
					promLabels[i].Value = row.metric
					break
```

pkg/pgmodel/querier/querier.go
```diff
@@ -274,64 +286,86 @@ func (q *pgxQuerier) queryMultipleMetrics(filter metricTimeRangeFilter, cases []
...
-		return "", errors.ErrMissingTableName
-	}
+	for res.Next() {
+		if err := res.Scan(&tableName, &tableSchema, &seriesTableName); err != nil {
+			return "", "", "", err
+		}
 
-	if err := res.Scan(&tableName); err != nil {
-		return "", err
+		if schema == "" || schema == tableSchema {
```
cevian, line 349: This logic means if we don't have a schema, this will return the first row. Didn't we want the schema to be prom_data if not specified? Also, do we want this logic in Go or in SQL?

cevian (changes requested):
Partial review. Overall looks good.

pkg/migrations/sql/idempotent/base.sql
antekresic: I don't see a reason to change since it mostly ignores doing anything in case of a view (except for some permission handling).

pkg/migrations/sql/preinstall/003-tables.sql
antekresic: Table that holds series data. Will add a comment to make it more clear.

pkg/pgmodel/cache/cache.go
antekresic: Default schema is the schema which is used if there is no schema specified. Should be determined by the DB state (i.e. if there is only one metric in the system with the name, its schema will be the default schema)

pkg/pgmodel/querier/querier.go
```diff
@@ -274,64 +286,86 @@ func (q *pgxQuerier) queryMultipleMetrics(filter metricTimeRangeFilter, cases []
...
+	if schema == "" {
+		err = q.metricTableNames.SetDefaultSchema(tableSchema, metric, tableName, seriesTableName)
+
+		if err != nil {
+			return "", "", "", err
+		}
 	}
 
-	err = q.metricTableNames.Set(metric, tableName)
+	err = q.metricTableNames.Set(tableSchema, metric, tableName, seriesTableName)
```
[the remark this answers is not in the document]
antekresic: We set default schema for queries which don't specify the schema but we need this to store the cache with the actual schema so if somebody queries that specific schema, it will get a hit.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -242,6 +247,22 @@ func (c *clauseBuilder) GetMetricName() string {
 	return c.metricName
 }
 
+func (c *clauseBuilder) SetSchemaName(name string) {
+	if c.schemaName == "" {
+		c.schemaName = name
```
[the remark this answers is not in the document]
antekresic: I don't see a reason why we shouldn't treat that as the schema not being set at all.

pkg/pgmodel/querier/query_builder.go
antekresic: Good catch, we should be throwing an error in those cases.

pkg/pgmodel/querier/series_set.go
antekresic: It would be nice but because we are dealing with different structs (labels.Label vs. prompb.Label) its not really easy to reuse the code. Its possible but it would add complexity to the code.

pkg/migrations/sql/preinstall/003-tables.sql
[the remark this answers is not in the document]
antekresic: Will update to be not null, should be easier to deal with the value.

I'm not sure about the `series_schema` just because I'm not sure why would we want to have series data outside a single schema. Is there a good reason?

pkg/pgmodel/querier/querier.go
antekresic: Right, I might have mixed it up with getting whatever metric was first created. Will change that. Let me see if I can move this logic to SQL so we only get zero or one rows returned.

pkg/migrations/sql/idempotent/base.sql
antekresic: OK, will change that.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1983,6 +2031,71 @@ SET search_path = pg_temp;
...
+    IF column_count <> 3 THEN
+        RAISE EXCEPTION 'view must contain time, series_id, and value columns with appropriate types';
+    END IF;
+
+    -- insert into metric table
+    INSERT INTO SCHEMA_CATALOG.metric (metric_name, table_name, table_schema, series_table, is_view, creation_completed)
+    VALUES (register_metric_view.view_name, register_metric_view.view_name, register_metric_view.schema_name, metric_table_name, true, true)
+    ON CONFLICT DO NOTHING;
+
+    IF NOT FOUND AND NOT register_metric_view.if_not_exists THEN
```
cevian, line 2084: I think if if_not_exists is true we should raise a NOTICE  to let people know it already existed. That's standard practice in PG

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1983,6 +2031,71 @@ SET search_path = pg_temp;
 REVOKE ALL ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.create_metric_view(text) TO prom_writer;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.register_metric_view(schema_name name, view_name name, if_not_exists BOOLEAN = false)
```
cevian, line 2034: we should also add an unregister function

pkg/migrations/sql/versions/dev/0.4.2-dev/4-downsampling.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN; -- need to run in a single transaction to set series_table column defaults to table_name
```
cevian, line 1: you'll have to move this file now that the version is updated

pkg/pgmodel/cache/cache.go
```diff
@@ -51,26 +56,68 @@ type MetricNameCache struct {
...
 }
 
+func (m *MetricNameCache) key(schema, metric string) (int, string) {
+	// deep copy the strings so the original memory doesn't need to stick around
+	keyBuilder := strings.Builder{}
+	keyBuilder.Grow(len(metric) + len(schema) + len(fieldSeparator))
+	keyBuilder.WriteString(schema)
+	keyBuilder.WriteString(fieldSeparator)
+	keyBuilder.WriteString(metric)
+	return keyBuilder.Len(), keyBuilder.String()
```
cevian, line 66: why do we need the len if we are returning the string anyway? cant we just use `len`

pkg/pgmodel/cache/cache.go
```diff
@@ -13,12 +14,16 @@ import (
 
 const (
 	DefaultMetricCacheSize = 10000
+
+	nullSchemaName = "__null__"
+	fieldSeparator = "*"
```
cevian, line 19: is there a reason `*` can't be a valid schema or table name? How do we protect against that?

pkg/pgmodel/cache/cache.go
cevian: I'm confused why is schema than an argument to this function?

cevian: also how is this different than the logic on line 94? with nullSchemaName

pkg/pgmodel/querier/querier.go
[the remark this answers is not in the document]
cevian: If we want this logic to populate the cache twice, I think it should be inside the cache code.

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -1573,6 +1573,121 @@ func TestConfigMaintenanceJobs(t *testing.T) {
 	})
 }
 
+func TestRegisterMetricView(t *testing.T) {
```
cevian, line 1576: do we have any testing for continous aggregate views?

pkg/pgmodel/querier/querier.go
```diff
@@ -237,15 +243,19 @@ func (q *pgxQuerier) queryMultipleMetrics(filter metricTimeRangeFilter, cases []
...
 		if err != nil {
 			// If the metric table is missing, there are no results for this query.
 			if err == errors.ErrMissingTableName {
 				continue
 			}
 			return nil, nil, err
 		}
+
 		filter.metric = tableName
+		filter.schema = tableSchema
```
cevian, line 256: should we verify this is the data schema and error if not?

pkg/pgmodel/querier/querier.go
```diff
@@ -263,8 +273,9 @@ func (q *pgxQuerier) queryMultipleMetrics(filter metricTimeRangeFilter, cases []
 			rows.Close()
 			return nil, nil, err
 		}
-		// Append all rows into results.
-		results, err = appendTsRows(results, rows, nil)
+		// Append all rows into results. Schema is ignored since we only return
+		// default schema for multi-metric queries.
```
cevian, line 277: how is this guaranteed in the code, I don't see that? Also where is this property tested?

pkg/migrations/sql/idempotent/base.sql
cevian: also we should add a test that we register a view and then drop it without registering. Subsequent queries should error. But, that error should be nice and clear.

pkg/pgmodel/cache/cache.go
```diff
@@ -13,12 +14,16 @@ import (
 
 const (
 	DefaultMetricCacheSize = 10000
+
+	nullSchemaName = "__null__"
```
cevian, line 18: why don't we just use an empty string here?

cevian (changes requested):
Good progress. I am confused by the default schema logic in the cache though. Plus some other issues below.

pkg/pgmodel/cache/cache.go
antekresic: I believe we don't have to, since Postgres doesn't allow it.

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). 

https://www.postgresql.org/docs/current/sql-syntax-lexical.html

I guess the only thing that could contain it is the metric name (which is very unlikely) and should not break anything here.

pkg/tests/end_to_end_tests/create_test.go
antekresic: I didn't think it was necessary tbh. since in essence they behave the same way as plain views.

pkg/pgmodel/querier/querier.go
antekresic: The SQL query uses `series` table to fetch the metric IDs which means it will only fetch non-view metric IDs.

pkg/pgmodel/querier/querier.go
antekresic: The SQL query uses series table to fetch the metric IDs which means it will only fetch non-view metric IDs.

While there are no specific tests, SQL query integration tests confirm this behavior.

pkg/pgmodel/cache/cache.go
antekresic: I guess we could, I'll update.

pkg/pgmodel/cache/cache.go
antekresic: Updated to use single Set function.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,9 +1460,41 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
 IS 'ABORT an INSERT transaction due to the ID epoch being out of date';
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.epoch_abort TO prom_writer;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_used_series(
```
cevian, line 1463: since the set if potential_series ids is much smaller than the set of unique series I suggest something like the following instead 
```
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_confirmed_unused_series(
    series_table TEXT, potential_series_ids BIGINT[]
) RETURNS TABLE (confirmed_series_id BIGINT[]) 
AS $func$
DECLARE
    r RECORD;
BEGIN
    FOR r IN
        SELECT *
        FROM SCHEMA_CATALOG.metric m
        WHERE m.series_table = get_used_series.series_table
    LOOP

    --at each iteration of the loop filter potential_series_ids to only
    --have those series ids that don't exist in the metric tables.
    EXECUTE format(
    $query$
        SELECT array_agg(potential_series.series_id)
				FROM unnest($1) as potential_series(series_id)
				LEFT JOIN LATERAL(
					SELECT 1
				  FROM  %1$I.%2$I
				  WHERE data_exists.series_id = potential_series.series_id
				  --use chunk append + more likely to find something starting at earliest time
				  ORDER BY time ASC
				  LIMIT 1
				) as lateral_exists(indicator) ON (true)
				WHERE lateral_exists.indicator IS NULL
    $query$, r.table_schema, r.table_name)
		USING potential_series_ids
    INTO potential_series_ids;

    END LOOP;
    RETURN potential_series_ids;
END
$func$
LANGUAGE PLPGSQL VOLATILE
```

Which could be used like 
```
SELECT SCHEMA_CATALOG.get_confirmed_unused_series('%1$s', array_agg(series_id))
FROM potentially_drop_series
```

By the way I think the table name parameter should be the series_table not the metric_table. Conceptually, the loop should be all the metric tables that use the same series_table. I also don't think we need the check time in this function.

cevian (changes requested):
Only reviewed the deletion logic so far. Have one request for a change

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -614,15 +634,26 @@ GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.delete_series_catalog_row(name, bigint[
...
+   FROM SCHEMA_CATALOG.metric m
+   WHERE m.table_schema = get_metric_table_name_if_exists.schema
+   AND m.metric_name = get_metric_table_name_if_exists.metric_name
+   UNION ALL
+   SELECT id, table_name::name, table_schema::name, series_table::name, is_view
+   FROM SCHEMA_CATALOG.metric m
+   WHERE m.table_schema = 'SCHEMA_DATA'
+   AND m.metric_name = get_metric_table_name_if_exists.metric_name
+   UNION ALL
+   SELECT id, table_name::name, table_schema::name, series_table::name, is_view
```
cevian, line 650: Everythig looks good except shouldn't the last (3rd case) error if there are more than 2 schemas with same name?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,9 +1460,55 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
...
+    --at each iteration of the loop filter potential_series_ids to only
+    --have those series ids that don't exist in the metric tables.
+    EXECUTE format(
+    $query$
+        SELECT array_agg(potential_series.series_id)
+				FROM unnest($1) as potential_series(series_id)
+				LEFT JOIN LATERAL(
+					SELECT 1
+				  FROM  %1$I.%2$I data_exists
+				  WHERE data_exists.series_id = potential_series.series_id AND time >= %3$L
```
cevian, line 1485: I don't think we should have a checktime here? That was an optimization anyway because we checked all time before previously? Semantically if we have a continuous agg before this point with data, we want to not delete the series

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,9 +1460,55 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
...
+        SELECT *
+        FROM SCHEMA_CATALOG.metric m
+        WHERE m.series_table = get_confirmed_unused_series.series_table
+    LOOP
+
+    --at each iteration of the loop filter potential_series_ids to only
+    --have those series ids that don't exist in the metric tables.
+    EXECUTE format(
+    $query$
+        SELECT array_agg(potential_series.series_id)
```
cevian, line 1480: nit: indentation here seems odd.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2070,6 +2141,106 @@ SET search_path = pg_temp;
...
+    WHERE  table_schema = register_metric_view.schema_name
+    AND    table_name   = register_metric_view.view_name;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'cannot register non-existant metric view in specified schema';
+    END IF;
+
+    -- cannot register view in data schema
+    IF schema_name = 'SCHEMA_DATA' THEN
+        RAISE EXCEPTION 'cannot register metric view in data schema';
```
cevian, line 2162:
```suggestion
        RAISE EXCEPTION 'cannot register metric view in SCHEMA_DATA schema';
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2070,6 +2141,106 @@ SET search_path = pg_temp;
...
+    END IF;
+
+    -- insert into metric table
+    INSERT INTO SCHEMA_CATALOG.metric (metric_name, table_name, table_schema, series_table, is_view, creation_completed)
+    VALUES (register_metric_view.view_name, register_metric_view.view_name, register_metric_view.schema_name, metric_table_name, true, true)
+    ON CONFLICT DO NOTHING;
+
+    IF NOT FOUND THEN 
+        IF register_metric_view.if_not_exists THEN
+            RAISE NOTICE 'metric with same name and schema already exists';
```
cevian, line 2196: return false here?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2070,6 +2141,106 @@ SET search_path = pg_temp;
...
+   column_count int;
+BEGIN
+    DELETE FROM SCHEMA_CATALOG.metric 
+    WHERE unregister_metric_view.schema_name = table_schema
+    AND unregister_metric_view.view_name = table_name
+    AND is_view = TRUE;
+
+    IF NOT FOUND THEN 
+        IF unregister_metric_view.if_exists THEN
+            RAISE NOTICE 'metric with specified name and schema does not exist';
```
cevian, line 2227: return false here?

pkg/migrations/sql/preinstall/003-tables.sql
```diff
@@ -99,8 +99,11 @@ CREATE TABLE SCHEMA_CATALOG.metric (
     retention_period INTERVAL DEFAULT NULL, --NULL to use the default retention_period
     default_compression BOOLEAN NOT NULL DEFAULT true,
     delay_compression_until TIMESTAMPTZ DEFAULT NULL,
-    UNIQUE (metric_name) INCLUDE (table_name),
-    UNIQUE(table_name)
+    table_schema name NOT NULL DEFAULT 'SCHEMA_DATA',
```
cevian, line 102: shouldn't we put this as the third column, right before table_name?

pkg/pgmodel/cache/cache.go
cevian: But there is a second kind of identifier 

> There is a second kind of identifier: the delimited identifier or quoted identifier. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). A delimited identifier is always an identifier, never a key word. So "select" could be used to refer to a column or table named “select”, whereas an unquoted select would be taken as a key word and would therefore provoke a parse error when used where a table or column name is expected. The example can be written with quoted identifiers like this:

When they say arbitrary characters they mean it - full unicode - I've seen emoji-named tables


I think the only really safe way to do this is to have length-delimited fields not fields with separators.

pkg/pgmodel/cache/cache.go
```diff
@@ -51,26 +54,63 @@ type MetricNameCache struct {
...
-	tableBuilder := strings.Builder{}
-	tableBuilder.Grow(len(tableName))
-	tableBuilder.WriteString(tableName)
+	val := m.value(tableSchema, tableName, seriesTable)
+	key := m.key(schema, metric)
 	//size includes an 8-byte overhead for each string
-	m.Metrics.Insert(metricBuilder.String(), tableBuilder.String(), uint64(metricBuilder.Len()+tableBuilder.Len()+16))
+	m.Metrics.Insert(key, val, uint64(len(key)+len(val)+16))
+
+	// In case of empty schema, we should set the entry with specified schema.
```
cevian, line 107:
```suggestion
	// if the schema inserted above was empty, also populate the cache with the real schema
```

pkg/pgmodel/querier/querier.go
cevian: yes but then wouldn't that be a good safety check here?

pkg/tests/end_to_end_tests/create_test.go
cevian: yes but since this is the main goal of this thrust of work we should have at least some end_to_end tests using that to make sure the integration with timescale works as expected. I would actually even start a new test go file. `continuous_agg_test.go`

pkg/tests/end_to_end_tests/query_integration_test.go
```diff
@@ -1111,6 +1209,17 @@ func TestPushdownDelta(t *testing.T) {
 				},
 			},
 		},
+		{
```
cevian, line 1212: do we have any end_to_end tests testing the name resolution logic. With and without the schema label?

pkg/migrations/sql/idempotent/base.sql
cevian: I think the logic we discussed was:
if schema is specified use the exact match
else {
   if there is a metric in the data schema return that
   else {
      if there is only one metric across all schemas, return that
       else throw error
    }
}

cevian (changes requested):
full review

pkg/pgmodel/cache/cache.go
Harkishen-Singh: If `fieldSeparator` is just to separate fields, why not use something that is more common in separating values like `;` or even `,`?

Harkishen-Singh (approved):
Looks good in general. Some nits and agreement on Mat's comments.

pkg/migrations/sql/preinstall/003-tables.sql
antekresic: I've done that before but had issues with upgrade tests reporting different column ordering for fresh install vs. upgrading.

Changing the test to account for different column ordering doesn't seem a trivial thing to me (might be wrong).

pkg/migrations/sql/idempotent/base.sql
antekresic: Right... the problem is we do this step (marking the series for deletion) before deleting the metric chunks. I'll try changing it around and see if that will make it work without the time.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -614,15 +634,51 @@ GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.delete_series_catalog_row(name, bigint[
...
+    WHERE m.table_schema = 'SCHEMA_DATA'
+    AND m.metric_name = get_metric_table_name_if_exists.metric_name;
+
+    IF rows_found = 1 THEN
+        RETURN QUERY SELECT m.id, m.table_name::name, m.table_schema::name, m.series_table::name, m.is_view
+        FROM SCHEMA_CATALOG.metric m
+        WHERE m.table_schema = 'SCHEMA_DATA'
+        AND m.metric_name = get_metric_table_name_if_exists.metric_name;
+        RETURN;
+    END IF;
```
cevian, line 663: Slight optimization for the SCHEMA_DATA case we don't need to query for the count, instead we can use the special FOUND variable instead
```suggestion
        RETURN QUERY SELECT m.id, m.table_name::name, m.table_schema::name, m.series_table::name, m.is_view
        FROM SCHEMA_CATALOG.metric m
        WHERE m.table_schema = 'SCHEMA_DATA'
        AND m.metric_name = get_metric_table_name_if_exists.metric_name;
        
        IF FOUND THEN
             RETURN;
         END IF;
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,39 +1486,76 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
...
+		USING potential_series_ids
+    INTO potential_series_ids;
+
+    END LOOP;
+    RETURN potential_series_ids;
+END
+$func$
+LANGUAGE PLPGSQL VOLATILE
+--security definer to add jobs as the logged-in user
+SECURITY DEFINER
```
cevian, line 1527: this does not need to be a security definer function so we can delete this and the search_path below, and the revoke as well

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,39 +1486,76 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
...
+        WHERE lateral_exists.indicator IS NULL
+    $query$, r.table_schema, r.table_name)
+		USING potential_series_ids
+    INTO potential_series_ids;
+
+    END LOOP;
+    RETURN potential_series_ids;
+END
+$func$
+LANGUAGE PLPGSQL VOLATILE
```
cevian, line 1525:
```suggestion
LANGUAGE PLPGSQL STABLE PARALLEL SAFE
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1429,39 +1486,76 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
...
     --series to what is dropped, so first filter by all series that have been dropped
     --but that aren't in that first hour and then make sure they aren't in the dataset
     EXECUTE format(
     $query$
         WITH potentially_drop_series AS (
-            SELECT distinct series_id
-            FROM SCHEMA_DATA.%1$I
-            WHERE time < %2$L
+            SELECT distinct id as series_id
+            FROM SCHEMA_DATA_SERIES.%1$I
```
cevian, line 1546: ugh I don't think this approach will work because the logic here was to take all the series ids that were in the "section to be deleted" and diff them "EXCEPT" the first hour after deletion to get a small set. The change of getting all series in the table and diffing with the first hour will still yield a huge set. I suggest we switch back to the approach of passing the check_time down to the get_confirmed_unused_series but only apply it to the table we are dropping from and not the other tables/views.

cevian (changes requested):
Partial review.

Found a big problem with the change related to check_time and in the change of marking vs deletion order. I think we have to roll those changes back, sorry. I suggested a way of dealing with check_time in a way that changes a lot fewer things. Sorry for steering you wrong.

pkg/pgmodel/model/sql_test_utils.go
```diff
@@ -82,6 +82,7 @@ func (r *SqlRecorder) Query(ctx context.Context, sql string, args ...interface{}
 func (r *SqlRecorder) QueryRow(ctx context.Context, sql string, args ...interface{}) pgx.Row {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+	println(sql)
```
cevian, line 85: needs to be deleted

pkg/pgmodel/querier/query_builder.go
```diff
@@ -40,12 +40,12 @@ const (
 	/* MULTIPLE METRIC PATH (less common case) */
 	/* The following two sql statements are for queries where the metric name is unknown in the query. The first query gets the
 	* metric name and series_id array and the second queries individual metrics while passing down the array */
-	metricNameSeriesIDSQLFormat = `SELECT m.metric_name, array_agg(s.id)
+	metricNameSeriesIDSQLFormat = `SELECT m.metric_name, m.table_schema, array_agg(s.id)
```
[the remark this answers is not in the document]
cevian: This was never changed

cevian (approved):
Only nits remaining. I would like that delete test add though. Also I just realized we will need to create a data retention mechanism for continous aggs because we do need to mark unused series when deleting from them as well. But let's do that in a separate PR.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1455,9 +1507,56 @@ COMMENT ON FUNCTION SCHEMA_CATALOG.epoch_abort(BIGINT)
 IS 'ABORT an INSERT transaction due to the ID epoch being out of date';
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.epoch_abort TO prom_writer;
 
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_confirmed_unused_series(
```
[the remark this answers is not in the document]
antekresic: Added test in continuous agg e2e test. We should add there a test that the series are marked for deletion when we unregister the cagg in a separate PR.

PR #675, opened 2021-06-29
Mem optimizations on query path

cevian:
Look at commit comments

### DIFF
--- a/pkg/pgmodel/querier/querier.go
+++ b/pkg/pgmodel/querier/querier.go
@@ -34,6 +34,8 @@ type QueryHints struct {
 	CurrentNode parser.Node
 	Lookback    time.Duration
 }
+
+// SerieSet adds a Close method to storage.SeriesSet to provide a way to free memory
 type SeriesSet interface {
 	storage.SeriesSet
 	Close()
@@ -335,7 +337,7 @@ func (errorSeriesSet) Next() bool                   { return false }
 func (errorSeriesSet) At() storage.Series           { return nil }
 func (e errorSeriesSet) Err() error                 { return e.err }
 func (e errorSeriesSet) Warnings() storage.Warnings { return nil }
-func (e errorSeriesSet) Close()                     { return }
+func (e errorSeriesSet) Close()                     {}
 
 type labelQuerier interface {
 	LabelsForIdMap(idMap map[int64]labels.Label) (err error)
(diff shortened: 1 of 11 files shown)


pkg/pgmodel/querier/querier.go
```diff
@@ -38,13 +35,19 @@ type QueryHints struct {
 	Lookback    time.Duration
 }
 
+//SerieSet adds a Close method to storage.SeriesSet to provide a way to free memory
```
antekresic, line 38:
```suggestion
// SeriesSet adds a Close method to storage.SeriesSet to provide a way to free memory
```

antekresic (approved):
Looks good!

pkg/pgmodel/querier/row.go
```diff
@@ -0,0 +1,183 @@
...
+	*dst = pgtype.TimestamptzArray{Elements: elements, Dimensions: arrayHeader.Dimensions, Status: pgtype.Present}
+	return nil
+}
+
+//wrapper to to allow DecodeBinary to reuse existing array so that a pool is effective
+type float8ArrayWrapper struct {
+	*pgtype.Float8Array
+}
+
+func (dstwrapper *float8ArrayWrapper) DecodeBinary(ci *pgtype.ConnInfo, src []byte) error {
```
Harkishen-Singh, line 83: If the purpose of these funcs is to just reuse the `pgtype.Float8Array` allocations, why not just do this way:

```golang
func PutFloat8Array(ref *pgtype.Float8Array) {
    ref.ArrayDimensions = pgtype.ArrayDimensions{}
    ref.Set(nil)
    fPool.Put(ref)
}
```

Same applies for TimestamptzArray.

ref: https://github.com/jackc/pgtype/blob/2ca304d4617a72491b1844eb92e9ecd86f7b84e9/float8_array.go#L22

pkg/pgmodel/querier/timestamp_series.go
```diff
@@ -0,0 +1,61 @@
...
+	At(index int) (int64, bool)
+	Len() int
+}
+
+//rowTimestampSeries is a TimestampSeries based on data fetched from a database row
+type rowTimestampSeries struct {
+	times *pgtype.TimestamptzArray
+}
+
+func NewRowTimestampSeries(times *pgtype.TimestamptzArray) *rowTimestampSeries {
```
Harkishen-Singh, line 24: Do we need this exported? If I am not wrong, its being used in the querier package only. Same applies to `NewRegularTimestampSeries()` too.

Harkishen-Singh (changes requested):
Few comments.

Harkishen-Singh (approved):

PR #677, opened 2021-06-30
Allow pushdown of multiple clauses in a single query

cevian:
This is especially important for comparitive queries like:

sum(rate(requests{error="true"})) / sum(rate(requests))

### DIFF
--- a/pkg/promql/engine.go
+++ b/pkg/promql/engine.go
@@ -586,7 +586,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *parser.Eval
 	}
 	defer querier.Close()
 
-	topNode := ng.populateSeries(querier, s)
+	topNodes := ng.populateSeries(querier, s)
 	prepareSpanTimer.Finish()
 
 	// Modify the offset of vector and matrix selectors for the @ modifier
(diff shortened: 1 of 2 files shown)


pkg/promql/engine.go
```diff
@@ -780,14 +780,13 @@ func (ng *Engine) getTimeRangesForSelector(s *parser.EvalStmt, n *parser.VectorS
...
-func (ng *Engine) populateSeries(querier Querier, s *parser.EvalStmt) parser.Node {
+func (ng *Engine) populateSeries(querier Querier, s *parser.EvalStmt) map[parser.Node]bool {
 	var (
 		// Whenever a MatrixSelector is evaluated, evalRange is set to the corresponding range.
 		// The evaluation of the VectorSelector inside then evaluates the given range and unsets
 		// the variable.
-		set       storage.SeriesSet
 		evalRange time.Duration
-		topNode   parser.Node
+		topNodes  map[parser.Node]bool = make(map[parser.Node]bool)
```
antekresic, line 789: In maps like these, where you just check for existence, its idiomatic to use an empty struct as the value since it doesn't use any memory to store and check for existence of the key without caring for the value (it would be set to `true` always).

Its a small thing TBH but posting for knowledge sharing.
https://levelup.gitconnected.com/memory-allocation-and-performance-in-golang-maps-b267b5ad9217

antekresic (approved):

Harkishen-Singh (approved):
I agree with ante on the struct thing. Other than that 👍

PR #678, opened 2021-06-30
Improve delete_expired_series

cevian:
- Add a low-pass filter to label deletion that checks the
  series table the label was deleted from first.
- Turn off jit. Seems like that was causing major slowdowns
  by compiling the check function for every partition.


antekresic (approved):

Harkishen-Singh (commented):
Will leave this upto Ante.

PR #679, opened 2021-07-02
Improve visibility into maintenance tasks

cevian:


antekresic (approved):

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1627,6 +1629,13 @@ BEGIN
     SELECT older_than + INTERVAL '1 hour'
     INTO check_time;
 
+    startT := clock_timestamp();
+
+    PERFORM set_config('application_name', format('promscale maintenance: data retention: metric %s', metric_name), false);
+    IF log_verbose THEN
+        RAISE LOG 'promscale maintenance: data retention: metric %: starting', metric_name;
```
Harkishen-Singh, line 1636: If `log_verbose` is true, we should also log the `startT` here, maybe at the start of the log. This is in order to know the absolute time when the maintenance started.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1659,27 +1668,49 @@ BEGIN
     IF older_than IS NULL THEN
         -- even though there are no new Ids in need of deletion,
         -- we may still have old ones to delete
+        lastT := clock_timestamp();
+        PERFORM set_config('application_name', format('promscale maintenance: data retention: metric %s: delete expired series', metric_name), false);
         PERFORM SCHEMA_CATALOG.delete_expired_series(metric_table, ran_at, present_epoch, last_updated);
+        IF log_verbose THEN
+            RAISE LOG 'promscale maintenance: data retention: metric %: done deleting expired series as only action in %', metric_name, clock_timestamp-lastT;
+            RAISE LOG 'promscale maintenance: data retention: metric %: finished in %', metric_name, clock_timestamp()-startT;
```
Harkishen-Singh, line 1676: Again, when `log_verbose` is true, we should log time towards the start of the line, so that absolute time is noted, along with the time difference.

Harkishen-Singh: Also, we can change the log to warn if the time taken is beyond, say 1 hour, which indicates the user should increase the background workers or more jobs for maintenance. This explaination should be mentioned in the logs as well.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2497,7 +2582,15 @@ BEGIN
             remaining_metrics := remaining_metrics || r;
             CONTINUE;
         END IF;
+        IF log_verbose THEN
+            startT := clock_timestamp();
+            RAISE LOG 'promscale maintenance: compression: metric %: starting, without lock wait', r.metric_name;
+        END IF;
```
Harkishen-Singh, line 2588: I see this a general trend across some functions. Why not make a function that marks the start of time like `markTaskStart(message TEXT) RETURNS TIMESTAMPTZ`, and when the job is done, we send that mark `markTaskComplete(message TEXT, begin TIMESTAMPTZ)` and pass the returned timestamp into this function.

This will reduce redundancy here.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2507,17 +2600,30 @@ BEGIN
         SELECT *
         FROM unnest(remaining_metrics)
     LOOP
+        IF log_verbose THEN
+            lockStartT := clock_timestamp();
+            RAISE LOG 'promscale maintenance: compression: metric %: waiting for lock', r.metric_name;
+        END IF;
```
Harkishen-Singh, line 2606: A double-check: are we fine with too much logging here, since it's a loop and people will have lakhs of metrics. I understand it's verbose only, but a double-check with our motive here.

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -1539,10 +1539,16 @@ func TestConfigMaintenanceJobs(t *testing.T) {
...
-		changeJobs := func(numJobs int, scheduleInterval time.Duration) {
-			_, err = db.Exec(context.Background(), "SELECT config_maintenance_jobs($1, $2)", numJobs, scheduleInterval)
+		changeJobs := func(numJobs int, scheduleInterval time.Duration, config *string, configErr bool) {
+			_, err = db.Exec(context.Background(), "SELECT config_maintenance_jobs($1, $2, $3)", numJobs, scheduleInterval, config)
 			if err != nil {
-				t.Fatal(err)
+				if !configErr {
+					t.Fatal(err)
+				}
+				return
```
Harkishen-Singh, line 1548: If I am not wrong, t.Fatal() stops the test by itself, so return right be trivial.

Harkishen-Singh (changes requested):
We should log timestamp (atleast in the root function) if verbose is set to true. Also, some redundancy can be removed to make the code nicer 😄

pkg/migrations/sql/idempotent/base.sql
cevian: Logs include the timestamp when the logging occurs in Postgres. There is no need to log it in the message.

pkg/migrations/sql/idempotent/base.sql
cevian: yes that's the point here

pkg/migrations/sql/idempotent/base.sql
cevian: Function call is PG are more expensive then I'd like to use for logging and I don't think it will make the code much cleaner

pkg/tests/end_to_end_tests/create_test.go
cevian: this is need if configErr is true

pkg/migrations/sql/idempotent/base.sql
cevian: the warnings are a good idea. Added a commit.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1724,71 +1756,126 @@ BEGIN
...
+        PERFORM set_config('application_name', format('promscale maintenance: compression'), false);
+        CALL SCHEMA_CATALOG.execute_compression_policy(log_verbose=>log_verbose);
+    END IF;
+    IF log_verbose THEN
+        RAISE LOG 'promscale maintenance: finished in %', clock_timestamp()-startT;
+    END IF;
+    IF clock_timestamp()-startT > INTERVAL '12 hours' THEN
+        RAISE WARNING 'promscale maintenance jobs are taking too long (one run took %)', clock_timestamp()-startT
+              USING HINT = 'Please consider increasing the number of maintenance jobs using config_maintenance_jobs()';
```
Harkishen-Singh, line 1818: Should we also add background workers?

```suggestion
              USING HINT = 'Please consider increasing the number of maintenance jobs using config_maintenance_jobs() or timescaledb.background_workers';
```

Harkishen-Singh (approved):
Minor suggestion that is optional

pkg/migrations/sql/idempotent/base.sql
cevian: I don't think so as the default background workers is so much higher and its less likely people hit this

PR #687, opened 2021-07-08
fix the wrong metric table name in metric deletion

### LINKED ISSUE #648: failure deleting metrics with long and winding names
When deleting metrics with the name that merits shortening for the purpose of naming tables/views the operation fails resulting in "relation not found" error or some such. I believe this is due to the fact that the original metric name is passed to the `delete_series_catalog_row` instead of the table name. I *believe* this fixes it (it did for me):

```diff
diff --git a/pkg/migrations/sql/idempotent/base.sql b/pkg/migrations/sql/idempotent/base.sql
index 308e2fe..8cabe3e 100644
--- a/pkg/migrations/sql/idempotent/base.sql
+++ b/pkg/migrations/sql/idempotent/base.sql
@@ -1963,7 +1963,7 @@ BEGIN
         GET DIAGNOSTICS rows_affected = ROW_COUNT;
         num_rows_deleted = num_rows_deleted + rows_affected;
     END IF;
-    PERFORM SCHEMA_CATALOG.delete_series_catalog_row(name, series_ids);
+    PERFORM SCHEMA_CATALOG.delete_series_catalog_row(metric_table, series_ids);
     RETURN num_rows_deleted;
 END;
 $$
```

Sorry, no PR, I'm not acquainted with the code enough, I believe some changes elsewhere are also needed to trigger migration and I won't have time in the nearest future to figure it out (or even look into figuring it out or generally create a PR).

hacker:
since longer metric names are somewhat mangled when used as a relation
name, we should be passing `metric_table`, not the metric name.

fixes #648


hacker:
Had to update the commit. I didn't commit one stray whitespace change to .sql file, but apparently it did affect generated output and tests caught the difference. Hopefully that was the cause and it will pass now.

cevian (approved):

Harkishen-Singh (approved):

Harkishen-Singh:
Thank you @hacker for the fix.

PR #690, opened 2021-07-11
add option to configure namespace in values.yaml

VineethReddy02:
1. Add option to configure namespace in `values.yaml`
2. In helm chart template files add default namespace value as `.Release.Namespace` (better practice used in community helm charts)
3. Rename `connector` to `promscale` in helm chart. 
4. Add helm chart static generation check to CI i.e. PR level tests & scheduled tests.


cevian (approved):
LGTM

PR #694, opened 2021-07-15
Remove promscale extension building from CI process

antekresic:
Promscale extension was a part of the Promscale repo in the past, that
is the reason why we were running extension building in our CI process.
Removing will speed up the Promscale CI process considerably and the
extension building will be moved to extension repo.


VineethReddy02:
LGTM!

VineethReddy02 (approved):

cevian (approved):

PR #696, opened 2021-07-16
Fix bug in execute_data_retention_policy

cevian:
Previously we had a bug in the SQL because the remaining_metrics
array was being concat with a record which resulted in

"ERROR:  operator does not exist: metric[] || record"

this only happened when a locked metric was found in the
first loop and so it wasn't caught by tests. We added a test
specific to this case.


pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +209,110 @@ func TestSQLDropChunk(t *testing.T) {
...
+		go func() {
+			defer wg.Done()
+			//release the lock to allow execute maintenance to complete
+			_, err := tx.Exec(context.Background(), "SELECT _prom_catalog.unlock_metric_for_maintenance($1)", metricId)
+			require.NoError(t, err)
+			err = tx.Commit(context.Background())
+			require.NoError(t, err)
+		}()
+
+		wg.Wait()
```
antekresic, line 304: I don't really like using `time.Sleep`  to synchronize goroutines for a couple of reasons but mostly because it reminds me of Javascript :grin:. 

I suggest removing the wait group and sleep call. Just call the release lock in a separate goroutine and execute maintenance in the main goroutine. That should block until release goroutine manages to catch up.

This way we are guaranteed to run exec maintenance before releasing the lock and we don't need to wait around.

antekresic (changes requested):
Looks good, I'd just like to change the way we handle the goroutines in the test.

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+				},
+				Samples: []prompb.Sample{
+					{Timestamp: int64(model.TimeFromUnixNano(time.Now().UnixNano()) - 1), Value: 0.1},
+				},
+			},
+		}
+		ingestor, err := ingstr.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), nil)
+		if err != nil {
+			t.Fatal(err)
+		}
```
Harkishen-Singh, line 255:
```suggestion
		require.NoError(t, err)
```
lets keep it consistent with above code.

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+		defer ingestor.Close()
+		_, _, err = ingestor.Ingest(newWriteRequestWithTs(copyMetrics(ts)))
+		if err != nil {
+			t.Error(err)
+		}
```
Harkishen-Singh, line 260:
```suggestion
		require.NoError(t, err)
```

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+
+		var tableName string
+		var metricId int
+		err = db.QueryRow(context.Background(), "SELECT id, table_name FROM _prom_catalog.get_metric_table_name_if_exists('test')").Scan(&metricId, &tableName)
+		if err != nil {
+			t.Fatal(err)
+		}
```
Harkishen-Singh, line 267: same here

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+
+		cnt := 0
+		err = db.QueryRow(context.Background(), fmt.Sprintf(`SELECT count(*) FROM show_chunks('prom_data."%s"')`, tableName)).Scan(&cnt)
+		if err != nil {
```
Harkishen-Singh, line 271: same here

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		if cnt != 2 {
+			t.Errorf("Expected there to be a chunk")
+		}
```
Harkishen-Singh, line 276:
```suggestion
		require.Equal(t, 2, int(cnt))
```

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+		defer func() {
+			_ = tx.Rollback(context.Background())
+		}()
+		_, err = tx.Exec(context.Background(), "SELECT _prom_catalog.lock_metric_for_maintenance($1)", metricId)
+		require.NoError(t, err)
+
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+
+		exec_query := "CALL prom_api.execute_maintenance(log_verbose=>true)"
```
Harkishen-Singh, line 290:
```suggestion
		maintenanceQuery := "CALL prom_api.execute_maintenance(log_verbose=>true)"
```

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+			}
+			if waiting {
+				found_wait = true
+				break
+			}
+			time.Sleep(time.Second)
+		}
+		if !found_wait {
+			t.Fatal("failed to find the exec_maintenance query waiting on lock")
+		}
```
Harkishen-Singh, line 318:
```suggestion
		require.False(t, found_wait, "failed to find the exec_maintenance query waiting on lock")
```

pkg/tests/end_to_end_tests/drop_test.go
```diff
@@ -207,6 +210,132 @@ func TestSQLDropChunk(t *testing.T) {
...
+		//make sure the exec_maintenance completes
+		wg.Wait()
+
+		err = db.QueryRow(context.Background(), fmt.Sprintf(`SELECT count(*) FROM show_chunks('prom_data."%s"')`, tableName)).Scan(&cnt)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if cnt != 1 {
+			t.Errorf("Expected the chunk to be dropped")
+		}
```
Harkishen-Singh, line 335:
```suggestion
		require.NoError(t, err)
		require.Equal(t, 1, int(cnt), "Expected the chunk to be dropped")
```
easier to read now.

Harkishen-Singh (approved):
Good to merge after a couple of nits.

PR #697, opened 2021-07-16
Metric view optimization

jgpruitt:
The LEFT JOIN LATERAL subqueries were doing a lot of work on a per hypertable basis. Doing these in a single pass across all hypertables is quite a bit faster. I first tried CTEs, however some of the queries needed to utilize parallelism to be suitably fast, and the optimizer tended to avoid parallelism when the queries were embedded as CTEs. Using `CREATE TEMPORARY TABLE ... AS` allowed for parallelism, but requires the function to be VOLATILE now.

The original version took 1 hour 5 minutes on the test machine. The new version took 15 seconds.


davidkohn88 (commented):
In general, I'm not a huge fan of creating temp tables in functions, it can be a bit of an anti-pattern, but this seems clever and should work. We're using this in a view and we're sort of taking advantage of parallelism inside the function but not outside, which is probably part of why this wasn't horribly efficient before...I'm wondering if all of this could be replaced with a materialized view that we just schedule a refresh on? 

I also think we should let the db team know of the problems we're facing here and why we're having to do some of this, and see if we can work on some of these functions behaving horribly, which I think some of them do. 

I haven't reviewed this in great detail to assess that it does give the same outcome, i'm assuming you have and more wanted feedback on that design decision...

jgpruitt:
> In general, I'm not a huge fan of creating temp tables in functions, it can be a bit of an anti-pattern, but this seems clever and should work. We're using this in a view and we're sort of taking advantage of parallelism inside the function but not outside, which is probably part of why this wasn't horribly efficient before...I'm wondering if all of this could be replaced with a materialized view that we just schedule a refresh on?

Yeah, I wasn't stoked about using temp tables. Mat suggested breaking out each temp table to a separate function which I have now done. It seems to perform almost as well - only one or two seconds slower - and has the benefit of not requiring the functions to be volatile.

> I also think we should let the db team know of the problems we're facing here and why we're having to do some of this, and see if we can work on some of these functions behaving horribly, which I think some of them do.

Agreed.

> I haven't reviewed this in great detail to assess that it does give the same outcome, i'm assuming you have and more wanted feedback on that design decision...

Yes. I have compared outcomes in a single-node database, but have not in a multi-node configuration.

> The other thing I think we need to think about is whether we're changing anything about how this works when you're only fetching a single metric or the like. Do we now have to do the full join no matter what? (It may be that we'd tried to make it lateral for that reason, then changed how we did it so that was happening anyway, I'm not sure, hard for me to remember now).

If I understand this question correctly, I think the answer is "no, we aren't changing anything when you want a single metric." The metric_view() function is plpgsql and thus cannot be "inlined". I think the whole thing has to run regardless of filtering outside the function.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2154,6 +2154,146 @@ SET search_path = pg_temp;
...
+        END, schema_name_in, dht.hypertable_name) x ON true
+    GROUP BY dht.hypertable_name, dht.node_name
+$function$
+LANGUAGE sql
+STRICT STABLE 
+SECURITY DEFINER
+--search path must be set for security definer
+SET search_path = pg_temp;
+REVOKE ALL ON FUNCTION SCHEMA_CATALOG.hypertable_remote_size(name) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.hypertable_remote_size(name) to prom_modifier;
```
cevian, line 2248:
```suggestion
GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.hypertable_remote_size(name) to prom_reader;
```

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2154,6 +2154,146 @@ SET search_path = pg_temp;
...
+        -- local hypertables
+        SELECT
+            ch.hypertable_name,
+            ch.compression_status,
+            ch.uncompressed_total_size AS before_compression_total_bytes,
+            ch.compressed_total_size AS after_compression_total_bytes,
+            NULL::text AS node_name
+        FROM _timescaledb_internal.compressed_chunk_stats ch
+        WHERE ch.hypertable_schema = schema_name_in
+        UNION ALL
```
cevian, line 2270: we use a UNION ALL for local and distributed for compression_stats but not sizes. is there a reason to treat these differently?

cevian (dismissed):
Looks great. Some minorish questions/change requests

pkg/migrations/sql/idempotent/base.sql
jgpruitt: The code in hypertable_local_size takes 13 seconds by itself in the test db when taking advantage of parallelism. Whenever I combined it with the hypertable_remote_size using UNION ALL, the planner stopped using parallel workers, and the runtime shot up to over an hour.
I can try it one more time now that they are in functions. Maybe that makes a difference.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2221,31 +2370,19 @@ BEGIN
...
-                hds.total_bytes::bigint as total_size_bytes,
-                pg_size_pretty(hds.total_bytes) as total_size,
+                coalesce(pcmvd.total_bytes::bigint, pcmvl.total_bytes::bigint) as total_size_bytes,
+                pg_size_pretty(coalesce(pcmvd.total_bytes::bigint, pcmvl.total_bytes::bigint)) as total_size,
                 (1.0 - (hcs.after_compression_total_bytes::NUMERIC / hcs.before_compression_total_bytes::NUMERIC)) * 100 as compression_ratio,
                 hcs.total_chunks::BIGINT,
                 hcs.number_compressed_chunks::BIGINT as compressed_chunks
             FROM SCHEMA_CATALOG.metric m
+            LEFT JOIN SCHEMA_CATALOG.hypertable_local_size('SCHEMA_DATA') pcmvl ON (pcmvl.hypertable_name = m.table_name)
+            LEFT JOIN SCHEMA_CATALOG.hypertable_remote_size('SCHEMA_DATA') pcmvd ON (pcmvd.hypertable_name = m.table_name)
```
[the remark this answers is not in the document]
jgpruitt: I think I understand. Good catch. Will fix.

pkg/migrations/sql/idempotent/base.sql
jgpruitt: Retried a few different options. If I keep the hypertable_local_size logic and hypertable_remote_size logic in separate functions, parallelism is used and it runs quickly. If I put them together in the same function, the planner won't use parallelism.

antekresic (approved):
:+1:

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2350,6 +2358,239 @@ SET search_path = pg_temp;
 REVOKE ALL ON FUNCTION SCHEMA_CATALOG.delete_series_from_metric(text, bigint[])FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.delete_series_from_metric(text, bigint[]) to prom_modifier;
 
+DO $block$
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() AND SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
```
cevian, line 2363: nit: since all these blocks are the same if, do we want to wrap all these functions in one block? instead of a block-per-function

cevian (approved):
Looks great. One nit/suggestion that I'll leave up to your discretion whether to change or not. Feel free to merge using bors when ready.

Harkishen-Singh (approved):

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2350,6 +2358,239 @@ SET search_path = pg_temp;
...
+$block$
+;
+
+DO $block$
+BEGIN
+    IF SCHEMA_CATALOG.is_timescaledb_installed() AND SCHEMA_CATALOG.get_timescale_major_version() >= 2 THEN
+        -- two columns in _timescaledb_internal.compressed_chunk_stats were renamed in version 2.2
+        -- schema_name -> hypertable_schema
+        -- table_name -> hypertable_name
+        -- we explicit aliases to deal with this
```
davidkohn88, line 2527: I"m not sure I understand this line? perhaps you could state a different way? or maybe there's a word missing?

davidkohn88 (commented):
LGTM, one nit on a comment that I don't quite understand, but I think the approach is definitely better than the previous.

pkg/migrations/sql/idempotent/base.sql
jgpruitt: good catch. I did leave out a word. I'll fix and expand on it a bit

pkg/migrations/sql/idempotent/base.sql
jgpruitt: I could go either way on this. I did it this way so each function was in a standalone "chunk" of code, but I don't feel strongly either way. I'll switch.

PR #700, opened 2021-07-21
Add support of metrics selector for prom-migrator

2nick:
> 1. How is this helpful?

This PR adds possibility to pass a vector selector to the prom-migrator tool to filter metrics for migration.

> 2. Aims behind the implementation. Are you trying to achieve something?

If I don't need all metrics from a remote read source I can filter them by using a vector selector syntax.


Harkishen-Singh:
Thank you for the PR! This is something required and I think we should add this before we make the `0.0.2` version stable, since this will be useful for others in the community.

cmd/prom-migrator/main.go
```diff
@@ -106,6 +108,18 @@ func main() {
 		os.Exit(0)
 	}
 
+	ms, err := parser.ParseExpr(conf.readerMetricsMatcher)
+	if err != nil {
+		log.Error("msg", "could not parse reader metrics matcher", "error", err)
+		os.Exit(2)
+	}
```
Harkishen-Singh, line 115: Please shift the validation part in `func validateConf(conf *config) error`. We can keep this as the first check.

cmd/prom-migrator/main.go
```diff
@@ -106,6 +108,18 @@ func main() {
...
+
+	vs, ok := ms.(*parser.VectorSelector)
```
Harkishen-Singh, line 117: lets shift this check to `func validateConf(conf *config) error` same as the above comment, and change the log to return an error as per the `validate()`'s purpose. The error can be like `fmt.Errorf("'-reader-metrics-matcher' should always be a valid vector_selector as per PromQL")`.

pkg/migration-tool/reader/reader.go
```diff
@@ -78,7 +80,10 @@ func (r *Read) Run(errChan chan<- error) {
 				errChan <- fmt.Errorf("remote-run run: %w", err)
 				return
 			}
-			ms := []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")}
+			ms := r.Config.MetricsMatchers
+			if len(ms) == 0 {
+				ms = []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")}
+			}
```
Harkishen-Singh, line 86: If I am not wrong, this will never be the case, as the validation checks with parser & vector_selector will always prevent this. But I think this is good for a sanitary check, so we can just add a warning saying, to open an issue if this block is ever executed, as then it will be some corner case in the code. 

```suggestion
			if len(ms) == 0 {
			        log.Warn("msg", "empty matchers received. Please open an issue regarding this at https://github.com/timescale/promscale/issues")
				ms = []*labels.Matcher{labels.MustNewMatcher(labels.MatchRegexp, labels.MetricName, ".*")}
			}
```

Harkishen-Singh (changes requested):
This will be a useful feature. Thank you for contributing it. Some minor changes and then we are good to merge. Also, we should have this as part of `0.0.2` release of Prom-migrator. cc @cevian

cmd/prom-migrator/main.go
2nick: Though about that but `parseFlags` fills config with simple types and `validateConf` doesn't change config fields.

So should I add raw validation (without setting some field of `config`) to `validateConf` and leave parsing string+type casting somewhere in `main`?

Harkishen-Singh: I think we should not keep the parsing logic in main. We can do something like what `convertTimeStrFlagsToTs()` does in the `validateConf()`, and similarly, do the parsing in an external function that is called from `validateConf()`. WDYT?

2nick: Done

pkg/migration-tool/reader/reader.go
2nick: Done

cmd/prom-migrator/main.go
```diff
@@ -213,6 +218,7 @@ func parseFlags(conf *config, args []string) {
 	flag.StringVar(&conf.readerClientConfig.OnErrStr, "reader-on-error", "abort", "When an error occurs during read process, how should the reader behave. "+
 		"Valid options: ['retry', 'skip', 'abort']. "+
 		"See 'reader-on-timeout' for more information on the above options. ")
+	flag.StringVar(&conf.readerMetricsMatcher, "reader-metrics-matcher", `{__name__=".*"}`, "Metrics vector selector to read data for migration.")
```
Harkishen-Singh, line 221: This should be a regex 😄 

```suggestion
	flag.StringVar(&conf.readerMetricsMatcher, "reader-metrics-matcher", `{__name__=~".*"}`, "Metrics vector selector to read data for migration.")
```

Harkishen-Singh: Also, please add this flag in the `prom-migrator` readme, which is the reference doc for CLI. You can find that in `cmd/prom-migrator/README.md`.

Harkishen-Singh (changes requested):
Thank you. Just a small change, then ready for merge.

cmd/prom-migrator/main.go
Harkishen-Singh: Also, the main_test.go will need to be fixed. Let me know if you want any help there.

Harkishen-Singh:
For fixing the test, you can make a function in `main_test.go` to return the default matcher that is expected.

```go
func getDefaultReaderLabelsMatcher() []*labels.Matcher {
	matcher, err := labels.NewMatcher(labels.MatchRegexp, "__name__", ".+")
	if err != nil {
		panic(fmt.Errorf("getDefaultReaderLabelsMatcher: %w", err))
	}
	return []*labels.Matcher{matcher}
}
```


And then use that in the test cases. An example to fix the first test case with name `pass_normal` would be:

```go
{
			name:  "pass_normal",
			input: []string{"-start=1970-01-01T00:16:40+00:00", "-end=1970-01-01T00:16:41+00:00", "-reader-url=http://localhost:9090/api/v1/read", "-writer-url=http://localhost:9201/write", "-progress-enabled=false"},
			expectedConf: &config{
				name:              "prom-migrator",
				start:             "1970-01-01T00:16:40+00:00",
				end:               "1970-01-01T00:16:41+00:00",
				mint:              1000000,
				mintSec:           1000,
				maxt:              1001000,
				maxtSec:           1001,
				humanReadableTime: true,
				maxSlabSizeBytes:  524288000,
				readerMetricsMatcher: `{__name__=~".+"}`,
				readerLabelsMatcher: getDefaultReaderLabelsMatcher(),
				readerClientConfig: utils.ClientConfig{
					URL:          "http://localhost:9090/api/v1/read",
					Timeout:      defaultTimeout,
					Delay:        defaultRetryDelay,
					OnTimeoutStr: "retry",
					OnErrStr:     "abort",
				},
				writerClientConfig: utils.ClientConfig{
					URL:          "http://localhost:9201/write",
					Timeout:      defaultTimeout,
					Delay:        defaultRetryDelay,
					OnTimeoutStr: "retry",
					OnErrStr:     "abort",
				},
				progressMetricName: "prom_migrator_progress",
				progressMetricURL:  "",
				maxReadDuration:    defaultMaxReadDuration,
				laIncrement:        defaultLaIncrement,
				maxSlabSize:        "500MB",
				concurrentPush:     1,
				concurrentPull:     1,
				progressEnabled:    false,
			},
			failsValidation: false,
		}
```

Let me know if anything else is needed.

Harkishen-Singh (approved):

Harkishen-Singh:
Go tests failing. Will push a commit that fixes the tests.

PR #702, opened 2021-07-27
Add multi-column support for metric views

antekresic:
This commit is an expansion of downsampling support enabling
registration of multi-column views and support for querying individual
columns using the special label `__column__`.


Harkishen-Singh (approved):

pkg/pgmodel/querier/querier.go
```diff
@@ -206,10 +209,15 @@ func (q *pgxQuerier) querySingleMetric(metric string, filter metricTimeRangeFilt
...
-			// existed at some point but the underlying relation was removed from outside of the system.
-			if e.Code == pgerrcode.UndefinedTable {
+			switch e.Code {
+			case pgerrcode.UndefinedTable:
+				// If we are getting undefined table error, it means the metric we are trying to query
+				// existed at some point but the underlying relation was removed from outside of the system.
 				return nil, nil, fmt.Errorf(errors.ErrTmplMissingUnderlyingRelation, mInfo.TableSchema, mInfo.TableName)
+			case pgerrcode.UndefinedColumn:
+				// If we are getting undefined column error, it means the column we are trying to query
+				// does not exist in the metric table so we return empty results.
```
cevian, line 219: what's the rationale behind empty result and not an error? Wouldn't the error be more helpful to the user?

pkg/pgmodel/querier/query_builder.go
```diff
@@ -334,12 +365,15 @@ func initializeLabeIDMap(labelIDMap map[int64]labels.Label, rows []timescaleRow)
 // getSchemaLabel returns the name and value of schema label if its not empty.
 // Empty value signifies that there is no need to add the schema label.
 // Additionally, we only set the schema if its not the default data schema.
-func getSchemaLabel(metricSchema string) (string, string) {
-	if metricSchema == "" || metricSchema == schema.Data {
-		return "", ""
+func getAdditionalLabels(metricSchema, column string) map[string]string {
+	result := make(map[string]string)
```
cevian, line 369:
```suggestion
	result := make(map[string]string, 2)
```

pkg/pgmodel/querier/query_builder.go
```diff
@@ -334,12 +365,15 @@ func initializeLabeIDMap(labelIDMap map[int64]labels.Label, rows []timescaleRow)
 // getSchemaLabel returns the name and value of schema label if its not empty.
 // Empty value signifies that there is no need to add the schema label.
 // Additionally, we only set the schema if its not the default data schema.
-func getSchemaLabel(metricSchema string) (string, string) {
-	if metricSchema == "" || metricSchema == schema.Data {
-		return "", ""
+func getAdditionalLabels(metricSchema, column string) map[string]string {
```
cevian, line 368: optional: In general returning a small map like this can be expensive (memory allocation+gc pressure). I wonder if we can switch this to a visitor pattern where this runs a passed-in-function for any labels that get added.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -376,16 +410,18 @@ func buildTimeSeries(rows []timescaleRow, lr lreader.LabelsReader) ([]*prompb.Ti
 			promLabels = append(promLabels, prompb.Label{Name: label.Name, Value: label.Value})
 
 		}
-
-		if schemaLabelName, schemaLabelValue := getSchemaLabel(row.schema); schemaLabelName != "" {
-			// If its a custom schema, we need to update the metric name as well.
+		if row.metric != "" {
```
cevian, line 413: if row.metric is only for when metrics need to be overriden maybe change name to row.metricOverride?

cevian (approved):
Some suggestions below, but I don't need to review again. My only real question is missing column error or empty.

pkg/pgmodel/querier/query_builder.go
antekresic: No really good way to do it because of different types of labels that is used with. Removed map usage and simplified the code for efficiency sake.

pkg/pgmodel/querier/querier.go
antekresic: Treating missing column as a missing series and returning empty result. Will add a comment.

PR #709, opened 2021-07-28
Change warning log to debug log 

VineethReddy02:
Move warning log to debug log, as it confuses the users. On Promscale start-up users see the warning log by default stating the older HA system is disabled. 

`
level=warn ts=2021-07-28T17:31:24.279Z caller=client.go:154 msg="No adapter leader election. Group lock id is not set. Possible duplicate write load if running multiple connectors"
`

### DIFF
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -150,11 +150,10 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
 		return nil, fmt.Errorf("elector init error: %w", err)
 	}
 
-	if elector == nil && !cfg.APICfg.ReadOnly {
-		log.Warn(
+	if (elector == nil && !cfg.APICfg.HighAvailability) && !cfg.APICfg.ReadOnly {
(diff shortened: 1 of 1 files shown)


pkg/runner/client.go
```diff
@@ -151,7 +151,7 @@ func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, erro
 	}
 
 	if elector == nil && !cfg.APICfg.ReadOnly {
-		log.Warn(
+		log.Debug(
 			"msg",
 			"No adapter leader election. Group lock id is not set. "+
 				"Possible duplicate write load if running multiple connectors",
```
cevian, line 157: Let's change the if to `if (elector == nil && !cfg.APICfg.HighAvailability) && !cfg.APICfg.ReadOnly`
```suggestion
		log.Info(
			"msg",
			"Prometheus HA is not enabled",
```

cevian (changes requested):

cevian (approved):

PR #711, opened 2021-07-28
Move series creation to copier

cevian:
Previously, series creation was done by the metric batcher.
But there was some performance issues with that:
1) There are lots of metric batchers and thus potentially
   lots of db calls to create series. This did cause the series
   creation to be blocked on waiting for a db connection
   from the pool
2) In general, less batching of series creation calls to the DB

To solve this, we move series creation to the copier and factor it out
into a SeriesWriter class. In addition we had to add logic to handle series
creation across metrics.

### DIFF
--- a/pkg/pgmodel/ingestor/copier.go
+++ b/pkg/pgmodel/ingestor/copier.go
@@ -100,7 +100,11 @@ func runCopier(conn pgxconn.PgxConn, in chan copyRequest, sw *seriesWriter) {
 
 		err := sw.WriteSeries(copyBatch(insertBatch))
 		if err != nil {
-			fmt.Println(err)
+			for i := range insertBatch {
+				insertBatch[i].data.reportResults(err)
+				insertBatch[i].data.release()
+			}
+			return
 		}
 
 		doInsertOrFallback(conn, insertBatch...)
--- a/pkg/pgmodel/ingestor/copier.go
+++ b/pkg/pgmodel/ingestor/copier.go
@@ -34,9 +34,32 @@ var (
 	handleDecompression = retryAfterDecompression
 )
 
+type copyBatch []copyRequest
+
+func (c copyBatch) VisitSeries(cb func(s *pgmodel.Series) error) error {
+	for _, req := range c {
+		samples := req.data.batch.GetSeriesSamples()
+		for _, sample := range samples {
+			err := cb(sample.GetSeries())
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (c copyBatch) NumSeries() int {
+	i := 0
+	for _, req := range c {
+		i += len(req.data.batch.GetSeriesSamples())
+	}
+	return i
+}
+
 // Handles actual insertion into the DB.
 // We have one of these per connection reserved for insertion.
-func runCopier(conn pgxconn.PgxConn, in chan copyRequest) {
+func runCopier(conn pgxconn.PgxConn, in chan copyRequest, sw *seriesWriter) {
 	// We grab copyRequests off the channel one at a time. This, and the name is
 	// a legacy from when we used CopyFrom to perform the insertions, and may
 	// change in the future.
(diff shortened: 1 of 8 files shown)


pkg/pgmodel/ingestor/copier.go
```diff
@@ -75,6 +98,11 @@ func runCopier(conn pgxconn.PgxConn, in chan copyRequest) {
 		}
 		insertBatch = insertBatch[:dst+1]
 
+		err := sw.WriteSeries(copyBatch(insertBatch))
+		if err != nil {
+			fmt.Println(err)
```
antekresic, line 103: Log instead?

antekresic (approved):

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -0,0 +1,358 @@
...
+		return err
+	}
+	if len(infos) == 0 {
+		return nil
+	}
+
+	labelMap := make(map[labelKey]labelInfo, seriesCount)
+	//logically should be a separate function but we want
+	//to prevent labelMap from escaping, so keeping inline.
+	{
```
Harkishen-Singh, line 84: I know we do like this in other places, but I really don't like it. I will try to do some refactoring in ingestor.

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -0,0 +1,358 @@
...
+	labelList              *model.LabelList
+	maxPos                 int
+	labelArraySet          *pgtype.ArrayType
+	labelArraySetNumLabels int
+}
+
+// Set all seriesIds for a samples, fetching any missing ones from the DB,
+// and repopulating the cache accordingly.
+func (h *seriesWriter) WriteSeries(sv SeriesVisitor) error {
+	infos := make(map[string]*perMetricInfo)
```
Harkishen-Singh, line 59: We can give better name than `infos`. `pending`?

```suggestion
	pending := make(map[string]*perMetricInfo)
```

Harkishen-Singh (changes requested):
Some questions and nits, other than that, the more PRs I read for ingestor, the more I feel it needs further refactoring. I will try to make a PR this quarter, to make the code more beautiful.

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -208,271 +202,21 @@ func (h *metricBatcher) flush() {
...
-	return dbEpoch, maxPos, nil
-}
+		numSeries := h.pending.batch.CountSeries()
+		select {
+		case h.toCopiers <- copyRequest{h.pending, h.metricTableName}:
+			MetricBatcherFlushSeries.Observe(float64(numSeries))
+			h.pending = NewPendingBuffer()
+			return
+		case req := <-recvChannel:
+			h.pending.addReq(req)
```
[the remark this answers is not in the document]
cevian: Renamed the function. I am setting to nil to avoid receiving data. Setting channel to nil in select is idiomatic Go. https://www.godesignpatterns.com/2014/05/nil-channels-always-block.html

pkg/pgmodel/ingestor/series_writer.go
cevian: This is purely because of the comment above. It should be a function but for performance it's not. This code is not new in the PR.

pkg/pgmodel/ingestor/copier.go
```diff
@@ -34,9 +34,32 @@ var (
...
+type copyBatch []copyRequest
+
+func (c copyBatch) VisitSeries(cb func(s *pgmodel.Series) error) error {
+	for _, req := range c {
+		samples := req.data.batch.GetSeriesSamples()
+		for _, sample := range samples {
+			err := cb(sample.GetSeries())
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
cevian: Discussed

pkg/pgmodel/ingestor/series_writer.go
cevian: rename to metricDataMap

Harkishen-Singh (approved):

PR #712, opened 2021-08-03
Add data retention support for metric views based on continuous aggregates

antekresic:
Continuous aggregates rely on hypertables to store their aggregated data. This
change tries to check for metric views which are based on them to apply the same
data retention system as for raw metric data. Users will be able to set data
retention rules for them same as for other raw metrics.


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1844,7 +1849,50 @@ BEGIN
 END
 $func$
 LANGUAGE PLPGSQL;
-GRANT EXECUTE ON PROCEDURE SCHEMA_CATALOG.drop_metric_chunks(text, timestamptz, timestamptz, boolean) TO prom_maintenance;
+GRANT EXECUTE ON PROCEDURE SCHEMA_CATALOG.drop_metric_chunks(text, text, timestamptz, timestamptz, boolean) TO prom_maintenance;
+
+
+-- Get metric hypertable information for raw and cagg metrics
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_metric_hypertable_info(metric_schema_name text, metric_table_name text, is_view boolean)
```
cevian, line 1856: nit: makes it more clear it's where the data is stored not the raw hypertable
```suggestion
CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_storage_hypertable_info(metric_schema_name text, metric_table_name text, is_view boolean)
```

pkg/tests/end_to_end_tests/continuous_agg_test.go
```diff
@@ -280,9 +298,52 @@ WITH (timescaledb.continuous,  timescaledb.ignore_invalidation_older_than = '1 m
...
 			t.Errorf("unexpected series count: got %v, wanted 0", count)
 		}
+
+		// Count cagg rows before dropping.
+		err = db.QueryRow(context.Background(), `SELECT count(*) FROM cagg_schema.cagg`).Scan(&count)
+		if err != nil {
+			t.Error("error fetching count of raw metric timeseries", err)
+		}
+
+		//time.Sleep(18000 * time.Second)
```
cevian, line 312:
```suggestion
	
```

pkg/tests/end_to_end_tests/continuous_agg_test.go
```diff
@@ -13,9 +13,12 @@ import (
 	"github.com/timescale/promscale/pkg/internal/testhelpers"
```
cevian, line 13: I think we need some tests to make sure that  the top-level `execute_data_retention_policy` does the right thing with views and that `set_metric_retention_period` works on caggs as well.

cevian (changes requested):
Looks good. But, I have asked for some more tests.

pkg/tests/end_to_end_tests/continuous_agg_test.go
```diff
@@ -280,9 +298,52 @@ WITH (timescaledb.continuous,  timescaledb.ignore_invalidation_older_than = '1 m
...
 		}
 
-		// none of the series should be marked for deletion
+		// None of the series should be marked for deletion.
 		if count != 0 {
 			t.Errorf("unexpected series count: got %v, wanted 0", count)
 		}
+
+		// Count cagg rows before dropping.
+		err = db.QueryRow(context.Background(), `SELECT count(*) FROM cagg_schema.cagg`).Scan(&count)
```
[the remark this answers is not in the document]
antekresic: Umm, this is a test? 😀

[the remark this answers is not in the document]
jgpruitt: Ahh... yes. Never mind 🤣

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1516,7 +1516,7 @@ IS 'ABORT an INSERT transaction due to the ID epoch being out of date';
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.epoch_abort TO prom_writer;
 
 CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_confirmed_unused_series(
-    series_table TEXT, potential_series_ids BIGINT[], check_time TIMESTAMPTZ
+    metric_schema TEXT, metric_table TEXT, series_table TEXT, potential_series_ids BIGINT[], check_time TIMESTAMPTZ
 ) RETURNS BIGINT[]
```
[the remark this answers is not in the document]
jgpruitt: Are we worried that `DROP FUNCTION ... CASCADE` could possibly drop users' code if they have built functions, procedures, or views on top of these functions?

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1253,18 +1253,18 @@ COMMENT ON FUNCTION SCHEMA_PROM.reset_metric_chunk_interval(TEXT)
 IS 'resets the chunk interval for a specific metric to using the default';
 GRANT EXECUTE ON FUNCTION SCHEMA_PROM.reset_metric_chunk_interval(TEXT) TO prom_admin;
 
-CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_metric_retention_period(metric_name TEXT)
+CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_metric_retention_period(schema_name TEXT, metric_name TEXT)
```
cevian, line 1256: This isn't a great interface change TBH from the UIUX point of view. Maybe this should be the second parameter but with a default value of 'prom_data'. I know that violates the argument ordering though. I think this needs a larger discussion.

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1844,7 +1849,53 @@ BEGIN
 END
 $func$
 LANGUAGE PLPGSQL;
-GRANT EXECUTE ON PROCEDURE SCHEMA_CATALOG.drop_metric_chunks(text, timestamptz, timestamptz, boolean) TO prom_maintenance;
+GRANT EXECUTE ON PROCEDURE SCHEMA_CATALOG.drop_metric_chunks(text, text, timestamptz, timestamptz, boolean) TO prom_maintenance;
+
+
+-- Get metric hypertable information for raw and cagg metrics
```
cevian, line 1855:
```suggestion
-- Get hypertable information for where data is stored for raw metrics and for 
-- the materialized hypertable for cagg metrics. For non-materialized views return 
-- no rows
```

cevian (commented):
Everything looks good with tests now. the only remaining question is the changes to the metric admin function UIUX

pkg/migrations/sql/idempotent/base.sql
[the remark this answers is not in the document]
antekresic: Removed CASCADE from all DROPs.

jgpruitt (approved):

PR #714, opened 2021-08-07
Document ThroughputInterval is some boolean

stappersg:
Fixes: #701 

The configuration option tput-report is for two things:
0: Disabling the throughput report
other: Changing the interval for the report. When doing so must
an unit provided. Units as seconds, minutes, hours etc.

PR #715, opened 2021-08-09
Document ThroughputInterval is some boolean

stappersg:
The configuration option tput-report is for two things:
0: Disabling the throughput report
other: Changing the interval for the report. When doing so must
an unit provided. Units as seconds, minutes, hours etc.


stappersg:
This merge request replaces #714

Harkishen-Singh (approved):
Thank you for this fix.

PR #730, opened 2021-09-01
Adds support for go-based benchmarking for Querier.

harkishen:

Note: the folder `pkg/tests/testdata/bench_data.zip` is of `20.9 MB`. Should we fetch this from s3 rather keeping in repo?


pkg/tests/benchmarks/integration/configure.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/promql"
+	"github.com/timescale/promscale/pkg/query"
+)
+
+func getQueryEngine(t testing.TB) *promql.Engine {
```
antekresic, line 17:
```suggestion
func newQueryEngine(t testing.TB) *promql.Engine {
```

antekresic (changes requested):
Lets fix the tests, will do another pass after that.

PR #742, opened 2021-09-16
Fix: restart copier when inserter is not instructed to quit.

zizon:
Copier loop can be break because of, like a bad db connection.
And since there is no retry and will eventually cause inserter suck in ingesting.

This patch simply re-run the routine until inserter do instructed to quit(Close was called).


Harkishen-Singh:
Thank you for opening this PR. We will take a look shortly.

cevian:
Hi @zizon, I fixed this in another way in #745, I started looking at why this wasn't caught in our tests and fixed that too. The advantage of 745 is that it requires less goroutines and there is  not really a good reason for the copier to exit in the first place. See if you agree with that solution and thank you!

zizon:
It is better than my monkey patch.
I will closed this.

And it can be better if considering retry of insertion of that failed batch, at least for some
retrable error.

PR #743, opened 2021-09-16
helm chart - typo in cronjob template

wileyj:
change `promscal` -> `promscale` on line 60

```
Error: UPGRADE FAILED: template: promscale/templates/cronjob-maintenance.yaml:60:25: executing "promscale/templates/cronjob-maintenance.yaml" at <include "promscal.fullname" .>: error calling include: template: no template "promscal.fullname" associated with template "gotpl"
```

### DIFF
--- a/helm-chart/templates/cronjob-maintenance.yaml
+++ b/helm-chart/templates/cronjob-maintenance.yaml
@@ -57,7 +57,7 @@ spec:
           volumes:
             - name: script-volume
               configMap:
-                name: {{ include "promscal.fullname" . | trunc 51 }}-maintenance
+                name: {{ include "promscale.fullname" . | trunc 51 }}-maintenance
           restartPolicy: OnFailure
           {{- with .Values.maintenance.nodeSelector }}
           nodeSelector:


VineethReddy02:
Thanks for the fix @wileyj 

But the maintenance job is by default disabled from two releases as TimescaleDB does by using UDJ (user-defined jobs), Any specific reason why are you explicitly enabling the maintenance job?

wileyj:
> Thanks for the fix @wileyj
> 
> But the maintenance job is by default disabled from two releases as TimescaleDB does by using UDJ (user-defined jobs), Any specific reason why are you explicitly enabling the maintenance job?

Yes - using timescale-cloud, there is apparently a bug in the postgres extension `2.4.1` and the suggestion from support was to enable the maintenance jobs in the meantime. 

Changing `false` to `true` (i did notice the comment about it not being needed https://github.com/timescale/promscale/blob/master/helm-chart/values.yaml#L96-L98) produced the error above.

wileyj:
Plus, it never hurts to fix a typo! This one just happened to help me get past some storage usage issues using timescale-cloud.

cevian (approved):

cevian:
@wileyj Thank you!

PR #745, opened 2021-09-20
Fix copier error handling

cevian:
Previously, the copier quit if WriteSeries or orderExemplarLabelValues failed.
This is wrong since errors can be transient and copiers aren't restarted.

Instead, errors should be reported but copiers shouldn't exit.

This PR fixes the logic as well as a test that should have but didn't detect this.


cevian:
Note this is an alternative solution to #742 but one that avoids having extra go-routines to monitor and restart copiers.

antekresic (approved):

PR #756, opened 2021-09-24
Add OTLP GRPC server with trace ingest support

antekresic:
- Add OTLP GRPC server listen address flag
- Add initial implementation of the pdata.Traces ingestor
- Add minimal end-to-end test (to be expanded in coming work)


pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -178,6 +189,250 @@ func (p *pgxDispatcher) Close() {
...
+func getUUID(buf [16]byte) pgtype.UUID {
+	return pgtype.UUID{
+		Bytes:  buf,
+		Status: pgtype.Present,
+	}
+}
+
+func getEventTimeRange(events pdata.SpanEventSlice) (result pgtype.Tstzrange) {
+	if events.Len() == 0 {
+		_ = result.Set(nil)
```
cevian, line 353: to avoid the error ignoring?
```suggestion
		result = pgtype.Tstzrange{ Present: pgtype.Null }
```

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -178,6 +189,250 @@ func (p *pgxDispatcher) Close() {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func convertByteArrayToInt64(buf [8]byte) int64 {
+	ux := binary.BigEndian.Uint64(buf[:])
+
+	x := int64(ux >> 1)
```
cevian, line 336: I don't get this conversion code why can't we simply cast to an int64? Can you add a comment with explanation/link? Also we probably need the reverse operation.

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -178,6 +189,250 @@ func (p *pgxDispatcher) Close() {
...
+		nameID,
+		span.Kind().String(),
+		span.StartTimestamp().AsTime(),
+		span.EndTimestamp().AsTime(),
+		string(jsonTags),
+		span.DroppedAttributesCount(),
+		eventTimeRange,
+		span.DroppedEventsCount(),
+		span.DroppedLinksCount(),
+		span.Status().Code().String(),
```
cevian, line 425: The code needs to follow the logic https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L264 for handling deprecated code logic.

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -178,6 +189,250 @@ func (p *pgxDispatcher) Close() {
...
+
+	x := int64(ux >> 1)
+	if ux&1 != 0 {
+		x = ^x
+	}
+
+	return x
+}
+
+func getUUID(buf [16]byte) pgtype.UUID {
```
cevian, line 344:
```suggestion
func TraceIDtoUUID(buf [16]byte) pgtype.UUID {
```

cevian: probably good to define the reverse function too?

antekresic: Reverse is just:
```
uuid.Bytes
```
This function is just make the code easier to read.

pkg/pgmodel/ingestor/dispatcher.go
antekresic: I'll add a comment and add a reverse function, export them both because one will be unused/dead code right now.

antekresic: Also, if we just cast to int64, it can overflow from an uint64.

antekresic: OK, my bad, after checking the conversion... seems like Go handles it nicely. I'm gonna remove this and just do a cast.

pkg/pgmodel/ingestor/dispatcher.go
antekresic: Slight issue here, we don't have access to the deprecated code value. `pdata` model doesn't expose it and we don't have direct access to the protobuf struct. 

Only solution I see right now is to clamp the value to 3.

cevian (approved):

PR #762, opened 2021-09-28
Fix upgrade scripts for 0.6.0

cevian:
Fix SQL upgrade scripts and add tests to catch
similar problems in the future.


antekresic (approved):

jgpruitt (approved):

Harkishen-Singh (approved):
Let's get this in

PR #763, opened 2021-09-28
Create new ps_tag schema and operators for tag matching

jgpruitt:
- created new ps_tag schema
- created a type for each data type / operator combination
- created a function for each data type / operator combination to create an instance of the type
- created an operator for each function

This may seem like overkill, but it will allow for the most streamlined (performance-wise) code, as there will be no need for CASE statements to switch on type or operator.


jgpruitt:
taking a different direction

PR #767, opened 2021-10-01
Implement jaeger plugin interface with a proxy

cevian:
The proxy will act as a GRPC client that sends GRPC requests
to Promscale. Promscale itself implements the GRPC server that
actually queries the database itself.


pkg/jaegerproxy/proxy.go
```diff
@@ -0,0 +1,161 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type ProxyConfig struct {
+	TLS                bool          `yaml:"promscale-tls,omitempty"`
+	CaFile             string        `yaml:"promscale-cafile,omitempty"`
+	ServerAddr         string        `yaml:"promscale-server,omitempty"`
+	ServerHostOverride string        `yaml:"promscale-server-host-override,omitempty"`
+	Timeout            time.Duration `yaml:"timeout,omitempty"`
```
niksajakovljevic, line 25: It seems that `Timeout` is never used. I guess we want to pass it down when establishing gRPC connection.

cevian: yeah good point, I'll look into it

Harkishen-Singh (changes requested):
Partial review:
1. Should we group `pkg/jaegerproxy` & `pkg/jaegerquery` into `pkg/jaeger-plugin` and the `plugin` contains `proxy` & `query`?
2. `cmd/jaeger-plugin-proxy` seems like proxy and plugin are doing the same purpose. I think it should be `cmd/jaeger-proxy`.

(continue from `jaegerquery`)

pkg/jaegerquery/find_traces.go
```diff
@@ -0,0 +1,71 @@
...
+	return scanTraces(rows)
+}
+
+func scanTraces(rows pgxconn.PgxRows) ([]*model.Trace, error) {
+	traces := pdata.NewTraces()
+	for rows.Next() {
+		if rows.Err() != nil {
+			return nil, fmt.Errorf("trace row iterator: %w", rows.Err())
+		}
+		ScanRow(rows, &traces)
```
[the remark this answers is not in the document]
Harkishen-Singh: 1. I agree with the bot
2. I don't think passing ref of `traces` is required, since internally, `orig` already has a ref. We can remove the ref here.

pkg/jaegerquery/get_operations.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func getOperations(ctx context.Context, conn pgxconn.PgxConn, query spanstore.OperationQueryParameters) ([]spanstore.Operation, error) {
+	var (
+		pgOperationNames, pgSpanKinds pgtype.TextArray
+		operationsResp                []spanstore.Operation
+	)
+
+	// TODO:, till the test dataset is being used. It makes m to "m", hence sql result is empty.
+	//query.ServiceName = query.ServiceName[1 : len(query.ServiceName)-1]
```
Harkishen-Singh, line 42: Can this be removed now?

pkg/jaegerquery/query.go
```diff
@@ -0,0 +1,58 @@
...
+	return findTraces(ctx, p.conn, query)
+}
+
+func (p *Query) FindTraceIDs(ctx context.Context, query *spanstore.TraceQueryParameters) ([]model.TraceID, error) {
+	return findTraceIDs(ctx, p.conn, query)
+}
+
+func (p *Query) GetDependencies(ctx context.Context, endTs time.Time, lookback time.Duration) ([]model.DependencyLink, error) {
+	//query db using pgx here
+	log.Warn("getDependencies")
```
Harkishen-Singh, line 56: This should log, `GetDependencies will be implemented in future releases` if we aren't implementing before the launch, which I think is the case.

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+		query = fmt.Sprintf(subqueryFormat, "TRUE")
+	}
+
+	if q.NumTraces != 0 {
+		query += fmt.Sprintf(" LIMIT %d", q.NumTraces)
+	}
+	return query, params
+}
+
+type spanDBResult struct {
```
Harkishen-Singh, line 189: I think this & related funcs should go into `trace_result.go` file.

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+		&dbRes.droppedEventsCounts,
+		&dbRes.droppedLinkCounts,
+		&dbRes.traceState,
+		&dbRes.schemaUrl,
+		&dbRes.spanName,
+		&dbRes.resourceTags,
+		&dbRes.spanTags,
+
+		// Event table.
+		&dbRes.eventNames, // 14
```
Harkishen-Singh, line 246:
```suggestion
		&dbRes.eventNames,
```

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+		link.SetDroppedAttributesCount(uint32((*dbResult.linksDroppedTagsCount)[i]))
+		link.Attributes().InitFromMap(makeAttributes((*dbResult.linksTags)[i]))
+	}
+	return nil
+}
+
+// makeAttributes makes attribute map using tags.
+func makeAttributes(tags map[string]interface{}) map[string]pdata.AttributeValue {
+	m := make(map[string]pdata.AttributeValue, len(tags))
+	// todo: attribute val as array?
```
Harkishen-Singh, line 395: I think this needs to be discussed.

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+	case "SPAN_KIND_INTERNAL":
+		return pdata.SpanKindInternal
+	case "SPAN_KIND_CONSUMER":
+		return pdata.SpanKindConsumer
+	case "SPAN_KIND_PRODUCER":
+		return pdata.SpanKindProducer
+	case "SPAN_KIND_UNSPECIFIED":
+		return pdata.SpanKindUnspecified
+	default:
+		panic(fmt.Sprintf("unknown span kind: %s", s))
```
Harkishen-Singh, line 448: I don't think this should be panic, as `s` will be read from db, so an error makes more sense than crashing entire promscale.

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+	var uuid pgtype.UUID
+	uuid.Set(b)
+	params := []interface{}{uuid}
+
+	traceIDClause := "s.trace_id = $1"
+	return buildTraceQuery(traceIDClause, false), params, nil
+}
+
+func buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
+	quals := make([]string, 0, 15)
```
Harkishen-Singh, line 126: is `quals` self-explanatory? I find it hard to get it without reading the code. Maybe `clauses` suits better.

```suggestion
	clauses := make([]string, 0, 15)
```

pkg/jaegerquery/trace_query.go
```diff
@@ -0,0 +1,450 @@
...
+	//todo check the inclusive semantics here
+	var defaultTime time.Time
+	if q.StartTimeMin != defaultTime {
+		params = append(params, q.StartTimeMin)
+		quals = append(quals, fmt.Sprintf(`s.start_time >= $%d`, len(params)))
+	}
+	if q.StartTimeMax != defaultTime {
+		params = append(params, q.StartTimeMax)
+		quals = append(quals, fmt.Sprintf(`s.start_time <= $%d`, len(params)))
+	}
```
Harkishen-Singh, line 164: I recall from Matous's tests, using `between` was more performant than using `AND`. Same goes for duration below.

pkg/runner/runner.go
```diff
@@ -92,6 +94,15 @@ func Run(cfg *Config) error {
...
 
+		queryPlugin := shared.StorageGRPCPlugin{
+			Impl: jaegerquery.New(client.Connection),
+		}
+		err := queryPlugin.GRPCServer(nil, grpcServer)
+		if err != nil {
+			log.Error("msg", "Creating jaeger query GRPC server failed", "err", err)
+			return err
+		}
+
```
Harkishen-Singh, line 105: we should log a similar message like we are doing for `line 107`

pkg/runner/runner.go
```diff
@@ -92,6 +94,15 @@ func Run(cfg *Config) error {
 		grpcServer := grpc.NewServer()
 		otlpgrpc.RegisterTracesServer(grpcServer, api.NewTraceServer(client))
 
+		queryPlugin := shared.StorageGRPCPlugin{
+			Impl: jaegerquery.New(client.Connection),
+		}
+		err := queryPlugin.GRPCServer(nil, grpcServer)
+		if err != nil {
+			log.Error("msg", "Creating jaeger query GRPC server failed", "err", err)
+			return err
```
Harkishen-Singh, line 103: Seeing lines around this, to be consistent, maybe not return err? Also makes the logs more clear.

Harkishen-Singh (changes requested):
Looks good except for some places where I marked it as discussion.

pkg/jaegerproxy/proxy.go
```diff
@@ -0,0 +1,173 @@
...
+		opts = append(opts, grpc.WithInsecure())
+	}
+
+	if config.ConnectTimeout == 0 {
+		config.ConnectTimeout = DefaultTimeout
+	}
+
+	opts = append(opts, grpc.WithBlock())
+	ctx, cancel := context.WithTimeout(context.Background(), config.ConnectTimeout)
+	defer cancel()
```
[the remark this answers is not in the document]
niksajakovljevic: Yep, makes sense to have request timeout but guess we want to split connection and request timeout?

cmd/jaeger-plugin-proxy/main.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	conf := new(jaegerproxy.ProxyConfig)
+	if err = yaml.Unmarshal(bSlice, conf); err != nil {
+		logger.Error("Unmarshalling configuration file contents", "err", err.Error())
+		os.Exit(1)
+	}
+
+	//todo: Remove later
+	logger.Warn("configuration", "conf", fmt.Sprintf("%v", conf))
```
niksajakovljevic, line 48: Nit: I guess `Info` would be more appropriate to print configuration. I did notice a comment above but I think logging configuration is useful when running/debugging in prod.

Harkishen-Singh: Yes, Info is useful but Jaeger only allows Warn & Error level logs to be printed out.

niksajakovljevic:
So it seems that the current flow is: Jaeger -> Jaeger proxy (Grpc Server + GrpcClient) -> Grpc Server -> Promscale internal.
I wonder if we can avoid necessary Grpc serde step and have: Jaeger -> Jaeger proxy -> Promscale internal?
I guess you may have had this discussion before but I'm asking for my understanding as well.

pkg/jaegerproxy/proxy.go
```diff
@@ -0,0 +1,173 @@
...
+) (*storage_v1.GetOperationsResponse, error) {
+	return p.spanReaderClient.GetOperations(ctx, r)
+}
+
+func (p *Proxy) FindTraces(r *storage_v1.FindTracesRequest, stream storage_v1.SpanReaderPlugin_FindTracesServer) error {
+	client, err := p.spanReaderClient.FindTraces(stream.Context(), r)
+	if err != nil {
+		return err
+	}
+	for {
```
niksajakovljevic, line 130: Since streaming logic repeats in few places it's a good candidate for a function

pkg/jaegerproxy/proxy.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (p *Proxy) GetOperations(
+	ctx context.Context,
+	r *storage_v1.GetOperationsRequest,
+) (*storage_v1.GetOperationsResponse, error) {
+	return p.spanReaderClient.GetOperations(ctx, r)
+}
+
+func (p *Proxy) FindTraces(r *storage_v1.FindTracesRequest, stream storage_v1.SpanReaderPlugin_FindTracesServer) error {
+	client, err := p.spanReaderClient.FindTraces(stream.Context(), r)
```
niksajakovljevic, line 126: Nit: since this is streaming client consider naming it `streamClient` or just `stream`?

pkg/jaegerquery/get_services.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/jackc/pgtype"
+	"github.com/jackc/pgx/v4"
+	"github.com/pkg/errors"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const getServicesSQL = `
+SELECT
+ 	array_agg(value#>>'{}' ORDER BY value)
```
niksajakovljevic, line 19: Just curious, why we call `array_agg`? I guess we can do the same without it

pkg/jaegerproxy/proxy.go
```diff
@@ -0,0 +1,173 @@
...
+		spanReaderClient:        storage_v1.NewSpanReaderPluginClient(conn),
+		dependencyReaderClient:  storage_v1.NewDependenciesReaderPluginClient(conn),
+		capClient:               storage_v1.NewPluginCapabilitiesClient(conn),
+		archiveSpanReaderClient: storage_v1.NewArchiveSpanReaderPluginClient(conn),
+		archiveSpanWriterClient: storage_v1.NewArchiveSpanWriterPluginClient(conn),
+	}, nil
+}
+
+func (p *Proxy) Close() {
+	p.conn.Close()
```
niksajakovljevic, line 83: I guess we should close all the other clients, like `spanReaderClient`, etc...?

pkg/jaegerquery/find_traces.go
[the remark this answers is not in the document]
cevian: Fixed the error. Keeping as a reference in case the address orig points to needs to change.

pkg/jaegerproxy/proxy.go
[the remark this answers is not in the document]
cevian: Yea and we aren't supporting request timeout in this version (relying on the jaeger timeout instead)

pkg/runner/runner.go
cevian: this is not like 107 since 107 is inside a go func() { block

pkg/runner/runner.go
cevian: Everything is already consistent the top-level func return error, sub-go-routines don't

pkg/jaegerproxy/proxy.go
cevian: I'm using jaeger's naming conventions here

pkg/jaegerquery/get_services.go
cevian: returning and populating arrays is more efficient than returning rows in our performance tests.

pkg/jaegerproxy/proxy.go
cevian: Those don't have a Close function

pkg/jaegerproxy/proxy.go
cevian: Good catch

Harkishen-Singh (changes requested):
Good to make this out of the draft => ready for review.
1. Some files do not have license headers applied. Please add them.

niksajakovljevic (approved):
LGTM 👍

pkg/jaeger/query/find_traces.go
```diff
@@ -0,0 +1,73 @@
...
+	return scanTraces(rows)
+}
+
+func scanTraces(rows pgxconn.PgxRows) ([]*model.Trace, error) {
+	traces := pdata.NewTraces()
+	for rows.Next() {
+		if rows.Err() != nil {
+			return nil, fmt.Errorf("trace row iterator: %w", rows.Err())
+		}
+		if err := ScanRow(rows, &traces); err != nil {
```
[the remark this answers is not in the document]
cevian: that's less safe because I may need to change where the orig pointer points to

pkg/runner/runner.go
cevian: No, the parent should log. That's what we do everywhere else.

pkg/jaeger/query/trace_scan.go
```diff
@@ -0,0 +1,301 @@
...
+		attr, err := makeAttributes((*dbResult.eventTags)[i])
+		if err != nil {
+			return fmt.Errorf("making event tags: %w", err)
+		}
+		event.Attributes().InitFromMap(attr)
+	}
+	return nil
+}
+
+func populateLinks(
```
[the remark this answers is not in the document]
cevian: for next PR

pkg/jaeger/proxy/proxy.go
```diff
@@ -0,0 +1,163 @@
...
+	"google.golang.org/grpc/credentials"
+)
+
+const DefaultTimeout = time.Duration(5 * time.Second)
+
+type ProxyConfig struct {
+	TLS                bool          `yaml:"promscale-tls"`
+	CaFile             string        `yaml:"promscale-cafile,omitempty"`
+	ServerAddr         string        `yaml:"promscale-grpc-server"`
+	ServerHostOverride string        `yaml:"promscale-grpc-server-host-override,omitempty"`
```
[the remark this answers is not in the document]
cevian: Good point

Harkishen-Singh (approved):
Let's merge after header files.

PR #776, opened 2021-10-04
Refine tracing schema

jgpruitt:
1. Add service_name and span_kind to span_name table
2. Remove span_kind from span table
3. Adjust indexes on event and link tables to include trace_id
4. Update ingestor and trace_writer to deal with schema changes


pkg/pgmodel/ingestor/trace_writer.go
```diff
@@ -16,8 +16,31 @@ import (
...
+		WHERE t.key = 'service.name'
+		AND t.value = to_jsonb($3::text)
+		ON CONFLICT (service_name_id, span_name, span_kind) DO NOTHING;`
+	selectSpanNameIdSQL = `
+		SELECT n.id
+		FROM %s.span_name n
+		INNER JOIN %s.tag t on (n.service_name_id = t.id and t.key = 'service.name')
+		WHERE n.span_name = $1
+		AND n.span_kind = $2
+		AND t.value = to_jsonb($3::text)`
```
cevian, line 43: can we encapsulate all this logic in a plpgsql function that just take a service_name, span_name, and type and returns an id?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -97,8 +97,12 @@ func (ingestor *DBIngestor) IngestTraces(ctx context.Context, r pdata.Traces) er
 
 			for k := 0; k < spans.Len(); k++ {
 				span := spans.At(k)
-
-				nameID, err := ingestor.tWriter.InsertSpanName(ctx, span.Name())
+				serviceName := "UNKNOWN"
```
cevian, line 100: how about using https://github.com/open-telemetry/opentelemetry-collector/blob/v0.33.0/translator/trace/protospan_translation.go#L38 instead of "UNKOWN"

Alternatively can we use NULL or 0 for the id?

pkg/migrations/sql/versions/dev/0.6.1-dev/1-tracing.sql
```diff
@@ -136,11 +136,13 @@ END
 $block$
 ;
 
-
 CREATE TABLE IF NOT EXISTS SCHEMA_TRACING.span_name
```
cevian, line 139: did we want to change the table name to `operation`?

cevian (changes requested):
Mostly looks good, just a few comments.

pkg/pgmodel/ingestor/trace_writer.go
jgpruitt: definitely. will do

pkg/pgmodel/ingestor/trace_writer.go
```diff
@@ -26,17 +49,17 @@ const (
...
+		VALUES ($1, $2, $3, $4, $5, $6, $7, %s.get_tag_map($8), $9, $10, $11, $12, $13, $14, $15, %s.get_tag_map($16), $17, $18)`
 )
 
 type traceWriter interface {
 	InsertSpanLinks(ctx context.Context, links pdata.SpanLinkSlice, traceID [16]byte, spanID [8]byte, spanStartTime time.Time) error
 	InsertSpanEvents(ctx context.Context, events pdata.SpanEventSlice, traceID [16]byte, spanID [8]byte) error
 	InsertSpan(ctx context.Context, span pdata.Span, nameID, instLibID, rSchemaURLID pgtype.Int8, resourceTags pdata.AttributeMap) error
 	InsertSchemaURL(ctx context.Context, sURL string) (id pgtype.Int8, err error)
-	InsertSpanName(ctx context.Context, name string) (id pgtype.Int8, err error)
+	InsertSpanName(ctx context.Context, serviceName string, spanName string, spanKind string) (id pgtype.Int8, err error)
```
antekresic, line 62:
```suggestion
	InsertSpanName(ctx context.Context, serviceName, spanName, spanKind string) (id pgtype.Int8, err error)
```

antekresic (approved):
Looks good, should be OK to merge once Mat's comments are addressed.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -97,8 +97,12 @@ func (ingestor *DBIngestor) IngestTraces(ctx context.Context, r pdata.Traces) er
 
 			for k := 0; k < spans.Len(); k++ {
 				span := spans.At(k)
-
-				nameID, err := ingestor.tWriter.InsertSpanName(ctx, span.Name())
+				serviceName := "OTLPResourceNoServiceName"
+				av, found := span.Attributes().Get("service.name")
```
antekresic, line 101: These two strings should be extracted into constants.

PR #780, opened 2021-10-06
Add insert batching when ingesting trace data

antekresic:
Using batching to reduce DB requests and improve ingest performance.


cevian:
Also note there will be some rebase changes due to the refinement PR #776

PR #794, opened 2021-10-08
Handle insert conflicts for schema URL and instrumentation libs

antekresic:


jgpruitt:
We should avoid updating the rows if nothing is actually different. That would increase writes and cause unnecessary bloat. Insert only is the way to go.

If you'd like db functions for these so you don't have to insert and then select, I would be happy to do that for you.

PR #865, opened 2021-11-03
Ensure consistent sequencing of remote commands

jgpruitt:
The execute_everywhere function records the executed code in a
table named remote_commands. These entries have a column denoting
the order in which the commands were executed. The promscale code
runs execute_everywhere in the idempotent, preinstall, and versions
sql scripts. These commands can be executed in different sequences
depending upon whether it was a fresh install or an upgraded
version of the system.

Our tests compare the contents of the remote_commands table for
fresh installs vs upgrades. This inconsistent ordering causes the
tests to fail.

This PR adds a new script to the idempotent scripts which will
update the seq column on the remote_commands table to ensure a
consistent ordering of the entries. Known entries from the
installation/upgrade path are ordered at the front of the line.
Any other entries are listed afterwards in their original order.


cevian (approved):
Great work. should make writing the upgrade scripts much easier.

PR #871, opened 2021-11-08
Add support for PG14

cevian:
This commit mostly just adds test for PostgreSQL 14.
Since the Promscale extension doesn't yet support PG14,
tests with the extension cannot use PG14. This should
be changed as soon as the extension adds support (should
be soon now). TODOs have been added for this.

Also, when all the kinks for testing with PG14 are ironed out,
our default tests should be changed to PG14, both with CLI flags
and with our CI.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -97,6 +97,7 @@ jobs:
         - {name: "Timescaledb 1.x (pg 12)",  shortname: "tsdb1x-pg-12", ext: true,  tsdb: true,  tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "Plain Postgres (12)",      shortname: "plain-pg-12",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "TimescaleDB-OSS",          shortname: "tsdb-oss",     ext: false, tsdb: true,  tsdb2: true,  tsdboss: true,  multi: false, pg: 13}
+        - {name: "pg 14",                    shortname: "pg14",         ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 14}
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
--- a/pkg/internal/testhelpers/postgres_container.go
+++ b/pkg/internal/testhelpers/postgres_container.go
@@ -145,6 +145,51 @@ func (e ExtensionState) UsesTimescaleDBOSS() bool {
 	return (e & timescaleOSSBit) != 0
 }
 
+func (e ExtensionState) GetPGMajor() string {
+	PGMajor := "14"
+	if e.UsesPG13() {
+		PGMajor = "13"
+	} else if e.UsesPG12() {
+		PGMajor = "12"
+	}
+	return PGMajor
+}
+
+func (e ExtensionState) GetDockerImageName() (string, error) {
+	var image string
+	PGMajor := e.GetPGMajor()
+	PGTag := "pg" + PGMajor
+
+	switch e &^ postgres12Bit &^ postgres13Bit {
+	case Timescale1:
+		if PGMajor != "12" {
+			return "", fmt.Errorf("timescaledb 1.x requires pg12")
+		}
+		image = "timescale/timescaledb:1.7.4-pg12"
+	case Timescale1AndPromscale:
+		if PGMajor != "12" {
+			return "", fmt.Errorf("timescaledb 1.x requires pg12")
+		}
+		image = LatestDBWithPromscaleImageBase + ":latest-ts1-pg12"
+	case Timescale2, Multinode:
+		image = "timescale/timescaledb:latest-" + PGTag
+	case Timescale2AndPromscale, MultinodeAndPromscale:
+		image = LatestDBHAPromscaleImageBase + ":" + PGTag + "-latest"
+
+		//TODO: remove split once pg14 ha image published
+		if PGMajor == "14" {
+			image = "timescaledev/promscale-extension:latest-ts2-" + PGTag
+		}
+	case VanillaPostgres:
+		image = "postgres:" + PGMajor
+	case TimescaleOSS:
+		image = "timescale/timescaledb:latest-" + PGTag + "-oss"
+	case TimescaleNightly, TimescaleNightlyMultinode:
+		image = "timescaledev/timescaledb:nightly-" + PGTag
+	}
+	return image, nil
+}
+
 var (
 	pgHost          = "localhost"
 	pgPort nat.Port = "5432/tcp"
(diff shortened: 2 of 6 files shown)


JamesGuthrie:
Locally I've built the extension with PG14 support and ran the tests. All tests passed except for `TestContinuousAgg2StepAgg`, which appears to fail because the `timescale-toolkit` is not packaged in the docker image which the `promscale_extension` builds:

```
    continuous_agg_test.go:472:
        	Error Trace:	continuous_agg_test.go:472
        	            				main_test.go:247
        	            				containers.go:198
        	            				main_test.go:211
        	            				main_test.go:206
        	            				continuous_agg_test.go:465
        	Error:      	Received unexpected error:
        	            	ERROR: could not open extension control file "/usr/local/share/postgresql/extension/timescaledb_toolkit.control": No such file or directory (SQLSTATE 58P01)
        	Test:       	TestContinuousAgg2StepAgg
```

cevian:
@JamesGuthrie yup. we'll have to wait for our HA image to catch up to support that test. I'll fix up the logic once the extension is released.

JamesGuthrie:
@cevian should we also update the version string range for the `promscale` extension in `version.go`?

JamesGuthrie:
@cevian regarding the HA image. I was planning on adding the updated version of the extension to the https://github.com/timescale/timescaledb-docker and https://github.com/timescale/timescaledb-docker-ha repos. Was there anywhere else that the new version needs to be propagated to?

cevian:
@JamesGuthrie oh that range wasn't updated. That probably means the tests didn't actually install the extension. Yes, we have to do that.

cevian:
I think only  https://github.com/timescale/timescaledb-docker-ha. the other repo(https://github.com/timescale/timescaledb-docker ) is the `thin` image and doesn't include promscale.

JamesGuthrie:
@cevian I've extended the version range and re-run the e2e tests locally.

JamesGuthrie (approved):
I had a minor nitpicky thing which I've corrected and added to this PR. (I hope that this is okay, please let me know if you would prefer me not to do this).

Other than that LGTM as far as I can tell.

pkg/tests/end_to_end_tests/main_test.go
```diff
@@ -75,6 +76,8 @@ func setExtensionState() {
 	case 12:
 		extensionState.UsePG12()
 	case 13:
+		extensionState.UsePG13()
+	case 14:
```
niksajakovljevic, line 80: Nit: maybe worth commenting that `14` is the default extension state so no bit manipulation needed.

pkg/internal/testhelpers/postgres_container.go
```diff
@@ -294,13 +303,16 @@ func StartPGContainer(
 	printLogs bool,
 ) (testcontainers.Container, io.Closer, error) {
 	var image string
-	PGMajor := "13"
+	PGMajor := "14"
+	if extensionState.UsesPG13() {
+		PGMajor = "13"
+	}
 	if extensionState.UsesPG12() {
```
niksajakovljevic, line 310: Nit: using `else if` here seems more appropriate.

niksajakovljevic (approved):

pkg/internal/testhelpers/postgres_container.go
antekresic: Why not put this logic into `extensionState` and add a method called `GetPGMajorVersion()`? PG15 will happen and this part will just grow.

pkg/internal/testhelpers/postgres_container.go
```diff
@@ -314,7 +326,12 @@ func StartPGContainer(
 	case Timescale2, Multinode:
 		image = "timescale/timescaledb:latest-" + PGTag
 	case Timescale2AndPromscale, MultinodeAndPromscale:
-		image = LatestDBHAPromscaleImageBase + ":" + PGTag + "-latest"
+		if PGMajor == "14" {
+			//TODO: remove split once pg14 ha image published
+			image = "timescaledev/promscale-extension:latest-ts2-" + PGTag
+		} else {
+			image = LatestDBHAPromscaleImageBase + ":" + PGTag + "-latest"
+		}
```
antekresic, line 334:
```suggestion
		image = LatestDBHAPromscaleImageBase + ":" + PGTag + "-latest"
		
		//TODO: remove split once pg14 ha image published
		if PGMajor == "14" {
			image = "timescaledev/promscale-extension:latest-ts2-" + PGTag
		} 
```

pkg/internal/testhelpers/postgres_container.go
antekresic: Looking at it a bit more... Image used is determined by extension state, this whole logic could be moved in there.

Would nicely encapsulate things for when we update this in the future.

antekresic (approved):
Some comments on how to improve Docker image selection for future updates. Can be done in a separate PR.

.github/workflows/go.yml
```diff
@@ -97,6 +97,7 @@ jobs:
         - {name: "Timescaledb 1.x (pg 12)",  shortname: "tsdb1x-pg-12", ext: true,  tsdb: true,  tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "Plain Postgres (12)",      shortname: "plain-pg-12",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "TimescaleDB-OSS",          shortname: "tsdb-oss",     ext: false, tsdb: true,  tsdb2: true,  tsdboss: true,  multi: false, pg: 13}
+        - {name: "pg 14",                    shortname: "pg14",         ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 14}
```
jgpruitt, line 100: Why are we only running one test config for pg14? Is that the only setup for 14 we will support?

Actually, in general, what is the thinking behind these? We have one config for 12, several for 13, and one for 14. How do we decide which combos to test?

pkg/internal/testhelpers/postgres_container.go
jgpruitt: This would also allow us to write tests to check the logic without spinning up containers.

jgpruitt (approved):
1 question, and agree with Ante's notes on the container logic, but approved.

.github/workflows/go.yml
cevian: So the overall logic is we test most options on latest (OSS, plain postgres, etc) and then add some test for older versions with configs that make sense.  But we aren't quite ready to make pg14 "latest" b/c we are waiting on the timescaledb-single-ha release.

pkg/internal/testhelpers/postgres_container.go
cevian: Yep agreed. Changed

jgpruitt (approved):

antekresic (approved):

PR #872, opened 2021-11-09
Support higher extension versions than auto-upgrade

cevian:
We should support extensions set up at higher-versions
than what we would auto-upgrade to as long as they are
in the valid range.

Previously, we errored if the version was higher than expected.
But this isn't right. Users should be able to use extension versions
at the bleeding edge just fine (e.g. pre-release versions or versions
that are above the default version); we just shouldn't auto-upgrade
tp them. This fixes the logic to not error in such cases.


antekresic (approved):
Looks good, I would like that comparator if else block to be simplified by returning early but its probably not for this PR.

VineethReddy02 (approved):
LGTM!

Issue #727, opened 2021-08-26
-db-password option is inappropriately interpolated


babbottscott:
It appears the value of -db-password (whether passed on command line or by environment variable) is being interpolated when injected into the connection string.

So far we've had trouble with `%%` and other sprintf-like strings, as well as `#`. I haven't looked into the code, so can't give a full scope, but here are some illustrative command line attempts:
```
$ PROMSCALE_DB_HOST=127.0.0.1 PROMSCALE_DB_PORT=5432 PROMSCALE_DB_NAME=prometheus PROMSCALE_DB_USER=postgres PROMSCALE_DB_SSL_MODE=prefer PROMSCALE_WEB_LISTEN_ADDRESS=:9201 promscale -db-password '%s123'
... err="cannot parse `postgresql://postgres:xxxxxx@127.0.0.1:xxxxxx@0.5.1&sslmode=prefer&connect_timeout=60`: failed to parse as URL (parse \"postgresql://postgres:%s123@127.0.0.1:5432/prometheus?application_name=promscale@0.5.1&sslmode=prefer&connect_timeout=60\": invalid URL escape \"%s1\")"

$ ... promscale -db-password '%%s123'
... err="cannot parse `postgresql://postgres:xxxxxx@127.0.0.1:xxxxxx@0.5.1&sslmode=prefer&connect_timeout=60`: failed to parse as URL (parse \"postgresql://postgres:%%s123@127.0.0.1:5432/prometheus?application_name=promscale@0.5.1&sslmode=prefer&connect_timeout=60\": invalid URL escape \"%%s\")"

$ ... promscale -db-password '\%s123'
... err="cannot parse `postgresql://postgres:xxxxxx@127.0.0.1:xxxxxx@0.5.1&sslmode=prefer&connect_timeout=60`: failed to parse as URL (parse \"postgresql://postgres:\\\\%s123@127.0.0.1:5432/prometheus?application_name=promscale@0.5.1&sslmode=prefer&connect_timeout=60\": net/url: invalid userinfo)"

$ ... promscale -db-password '123#456'
... err="failed to connect to `host=postgres user=promscale database=`: hostname resolving error (lookup postgres on <nameserver_ip>:53: no such host)"

$ ... promscale -db-password '123\#456'
[...]

PR #887, opened 2021-11-16
URL encode postgres connection string

jgpruitt:
Postgres connection strings should be valid URLs. User names, passwords, database names,
application names, etc. can all contain characters that must be escaped in a URL. This
change uses the url package to construct the connection string and thus ensures that
it is properly escaped.

Addresses issue #727


cevian (approved):
LGTM

JamesGuthrie (approved):

pkg/pgclient/config.go
```diff
@@ -119,15 +121,18 @@ func (cfg Config) validateConnectionSettings() error {
...
-			cfg.Host,
-			cfg.Port,
-			cfg.Database,
-			cfg.AppName,
-			cfg.SslMode,
-			int(cfg.DbConnectionTimeout.Seconds()))
+		v := url.Values{}
+		v.Set("application_name", cfg.AppName)
+		v.Set("sslmode", cfg.SslMode)
+		v.Set("connect_timeout", strconv.Itoa(int(cfg.DbConnectionTimeout.Seconds())))
```
antekresic, line 127:
```suggestion
		v.Set("connect_timeout", fmt.Sprintf("%.f", cfg.DbConnectionTimeout.Seconds()))
```

antekresic (commented):
Minor change to avoid double conversion.

PR #892, opened 2021-11-17
Fix clashing function names

cevian:
The `hypertable_compression_stats` name sometimes clashed with Timescale's
name of the same name, creating UIUX issues. Rename the function
to prevent issues.

Fixes #888

PR #894, opened 2021-11-17
Move operator creation to idempotent

jgpruitt:
There is no `IF NOT EXISTS` option for `CREATE OPERATOR`
and the scripts were structure with operators defined in the
preinstall section. However, the function backing an operator
has to exist before you can define the operator. So, we had
"empty" functions defined in the preinstall so that the operator
creation would work and the real functions defined in
idempotent.

This is confusing, and it's a lot of extra code.

This PR moves the operator creation into the idempotent scripts
which allows us to delete the "empty" function definitions. We
get around the lack of `IF NOT EXISTS` with anonymous blocks
which catch the duplicate object exceptions that get thrown
when you issue a `CREATE OPERATOR` for an operator that already
exists.

There is also some minor reorganization and renaming of sql
scripts for general clarity.


pkg/migrations/sql/idempotent/tracing-tags.sql
```diff
@@ -1,13 +1,32 @@
 
+-------------------------------------------------------------------------------
+-- get tag id
+-------------------------------------------------------------------------------
 CREATE OR REPLACE FUNCTION SCHEMA_TRACING.get_tag_id(_tag_map SCHEMA_TRACING_PUBLIC.tag_map, _key SCHEMA_TRACING_PUBLIC.tag_k)
-RETURNS jsonb
+RETURNS bigint
```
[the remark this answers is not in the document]
jgpruitt: you're right. :(
should be fixed now

pkg/migrations/sql/versions/dev/0.7.0-beta.1.dev/1-tracing.sql
```diff
@@ -78,4 +78,5 @@ $block$
 ;
 
 DROP FUNCTION IF EXISTS SCHEMA_TRACING_PUBLIC.put_tag_key(SCHEMA_TRACING_PUBLIC.tag_k, SCHEMA_TRACING_PUBLIC.tag_type);
-DROP FUNCTION IF EXISTS SCHEMA_TRACING_PUBLIC.put_tag(SCHEMA_TRACING_PUBLIC.tag_k, SCHEMA_TRACING_PUBLIC.tag_v, SCHEMA_TRACING_PUBLIC.tag_type);
\ No newline at end of file
+DROP FUNCTION IF EXISTS SCHEMA_TRACING_PUBLIC.put_tag(SCHEMA_TRACING_PUBLIC.tag_k, SCHEMA_TRACING_PUBLIC.tag_v, SCHEMA_TRACING_PUBLIC.tag_type);
```
JamesGuthrie, line 81: I have a question regarding conventions of these "versioned" migration files. `0.7.0-beta.1` has already been released. Does that mean that this file contains all changes _since_ `0.7.0-beta.1`?

jgpruitt: Good point.

>Does that mean that this file contains all changes since 0.7.0-beta.1?

I think the answer to that is "no".

JamesGuthrie: Does that mean that we need to create a new `0.7.0-beta.2.dev` dir? Or `0.7.0`?

JamesGuthrie: @cevian I noticed that you have the same thing going on in #892.

jgpruitt: Okay, I'm going to follow the pattern from #892

cevian: The version for this file is actually the correct `0.7.0-beta.1.dev.1` which comes after  `0.7.0-beta.1` according to semver. During development we don't know if we'll release `0.7.0-beta.2` or `0.7.0` So we bump versions as little as possible. See https://github.com/timescale/promscale/blob/master/pkg/version/version.go#L16

cevian: I'd prefer not to modify existing upgrade scripts though, instead only adding new scripts in PRs. So let's move to another script file? 2-blah?

cevian (approved):
Very nice cleanup! One (minor-ish nit)

pkg/migrations/sql/versions/dev/0.7.0-beta.1.dev/1-tracing.sql
jgpruitt: done.

JamesGuthrie (approved):
LGTM. I just wanted to point out that there's going to be a merge conflict with #892.

PR #895, opened 2021-11-18
Adds Promscale instance table + metrics telemetry

Harkishen-Singh:
This is the first PR of the series of PRs implementing Telemetry in Promscale.


pkg/runner/runner.go
```diff
@@ -13,28 +14,39 @@ import (
...
-	elector      *util.Elector
+	elector     *util.Elector
+	PromscaleID uuid.UUID
+
 	startupError = fmt.Errorf("startup error")
 )
 
+func init() {
+	// PromscaleID must always be generated on start, so that it remains constant throughout the lifecycle.
+	PromscaleID = uuid.New()
```
niksajakovljevic, line 47: This seems to imply that on every instance restart you get a new PromscaleID. Shouldn't we keep and re-use the same UUID in the database? So on first/initial run we will populate it in the DB and on restart we look it up and load into memory

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,275 @@
...
+
+	metricMux sync.RWMutex
+	metrics   []telemetry
+}
+
+type Telemetry interface {
+	StartRoutineAsync()
+	// RegisterMetric registers a Prometheus Gauge or Counter metric for telemetry visitor.
+	// It must be called after creating the telemetry engine.
+	RegisterMetric(columnName string, gaugeOrCounterMetric prometheus.Metric) error
```
niksajakovljevic, line 42: I wonder if we really need this generic interface for registering telemetry metrics? It seems we already know which metrics are we going to track in our code. I guess that would eliminate some code like `metricsVisitor` where we match on strings and instead we can directly map `metrics` to `telemetryStats`.

niksajakovljevic (commented):
Looks good overall! I left few comments

pkg/runner/runner.go
Harkishen-Singh: The `promscale_instance_information` table needs a UUID per Promscale instance, to identify them individually, hence we need it per promscale. Reason: when the stats (say num_samples_ingested) are related to a X promscale updates, promscale should know which row it should update and not touch the rows of other promscales that are running.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (t *engine) StartRoutineAsync() {
+	ctx, stopTelemetryRoutine := context.WithCancel(context.Background())
+	go t.telemetrySync(ctx)
+	t.stopTelemetryRoutine = stopTelemetryRoutine
+
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	t.wgTelemetryRoutine = wg
```
jgpruitt, line 126: I would assign to `t.wgTelemetryRoutine` before starting the goroutine since `telemetrySync` accesses it. I know it's not actually possible to call Done on the context before the `wgTelemetryRoutine` gets assigned, but it feels like a matter of good practice

jgpruitt: same for `stopTelemetryRoutine` actually

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,275 @@
...
+// syncInfoTable stats with promscale_instance_information table.
+func syncInfoTable(conn pgxconn.PgxConn, uuid [16]byte, s telemetryStats) error {
+	lastUpdated := time.Now()
+
+	pgUUID := new(pgtype.UUID)
+	if err := pgUUID.Set(uuid); err != nil {
+		return fmt.Errorf("setting pg-uuid: %w", err)
+	}
+
+	query := `INSERT INTO _ps_catalog.promscale_instance_information
```
jgpruitt, line 246: The schema name is hardcoded here, but isn't it technically configurable?

pkg/migrations/sql/preinstall/013-telemetry.sql
```diff
@@ -0,0 +1,22 @@
...
+    telemetry_traces_queries_executed               BIGINT NOT NULL,
+    telemetry_traces_dependency_queries_executed    BIGINT NOT NULL,
+    deletable                                       BOOLEAN DEFAULT TRUE
+);
+GRANT SELECT ON TABLE SCHEMA_PS_CATALOG.promscale_instance_information TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE SCHEMA_PS_CATALOG.promscale_instance_information TO prom_writer;
+
+-- Write a counter reset row, i.e., the first row in the table. Purpose:
+-- The above telemetry_.* rows logically behave as counter. They get deleted by
+-- telemetry-housekeeper promscale when last_updated is too old to be stale. Since
```
jgpruitt, line 17: I don't quite understand this. Where is the code for the telemetry-housekeeper? I want to try to figure it out.

jgpruitt (changes requested):

pkg/telemetry/telemetry.go
Harkishen-Singh: MetricsVisitor will still be there, as we need to travel though the metrics that we are tracking and get the latest stats, but what can be improved here will be instead of `metrics` being a `[]telemetry`, we can make it a map of `name: prometheus_metric`.

Regarding the `RegisterMetric`, we can register two ways:
1. (the current approach). We know the exact metrics to be monitored. Register those metrics in the telemetry instance passed down the constructor function. This avoids the private metric vars from making them public just for telemetry use-case.
2. In the telemetry package itself, make a var `metrics` and simply import the metrics that you want to monitor from respective packages into this var. However, private metric vars need to be made public.

I feel the first approach is minimalist plus, making a function do this gives the option to error out in runtime, in case a bad metric is asked to be monitored. Second approach does not give the erroring out advantage.

pkg/pgmodel/common/schema/schema.go
```diff
@@ -12,6 +12,7 @@ const (
 	Exemplar  = "prom_data_exemplar"
 	Ext       = "_prom_ext"
 	Catalog   = "_prom_catalog"
+	PSCatalog = "_ps_catalog"
```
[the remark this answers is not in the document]
Harkishen-Singh: We plan to make a schema `_ps_catalog` (mentioned in design doc as comment by Mat) to have all promscale related information. `ps` will essentially pronounce as `promscale`. Hence, `_promscale_catalog`

pkg/telemetry/telemetry.go
Harkishen-Singh: No, this is a static schema which is not configurable upon user. This schema will be further expanded in future, by moving contents of `_prom_catalog` here.

pkg/migrations/sql/preinstall/013-telemetry.sql
Harkishen-Singh: Sorry, this wasn't supposed to be part of this PR. Its basically part of the big PR ([here](https://github.com/timescale/promscale/pull/884)), which I broke down into smaller PRs, this being the first in series. Will remove this part. Thank you.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,288 @@
...
+		return fmt.Errorf("metric should only be of type Counter or Gauge")
+	}
+
+	t.metricMux.Lock()
+	defer t.metricMux.Unlock()
+	t.metrics[column(columnName)] = metric
+	return nil
+}
+
+func isCounterOrGauge(metric prometheus.Metric) (bool, error) {
```
antekresic, line 237: Instead of writing to a Metric, why not check underlying type is `prometheus.Gauge` or `prometheus.Counter`?

pkg/telemetry/telemetry.go
antekresic: Still, lets use the `schema.PSCatalog` here instead of hardcoding. While unlikely we will change the name, its safer and easier to navigate code when looking for code that uses that schema.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,288 @@
...
+
+	stopTelemetryRoutine context.CancelFunc
+	wgTelemetryRoutine   *sync.WaitGroup
+
+	metricMux sync.RWMutex
+	metrics   map[column]prometheus.Metric
+}
+
+type Telemetry interface {
+	StartRoutineAsync()
```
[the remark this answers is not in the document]
antekresic: I see that you are not testing this part of the interface at all, probably because of the above reason.

antekresic (changes requested):
Looks good, needs a bit more work to clean the code up.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,288 @@
...
+
+func (t *engine) telemetrySync(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			t.wgTelemetryRoutine.Done()
+			return
+		case <-time.After(time.Minute * 55):
+		}
+		t.metricMux.RLock()
```
niksajakovljevic, line 152: I don't see this lock being released if `len(t.metrics) != 0`

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,288 @@
...
+		if err := syncTimescaleMetadataTable(t.conn, Stats(tobs)); err != nil {
+			return fmt.Errorf("writing metadata for tobs: %w", err)
+		}
+	}
+	return nil
+}
+
+// syncTimescaleMetadataTable syncs the metadata/stats with telemetry metadata table and returns the last error if any.
+func syncTimescaleMetadataTable(conn pgxconn.PgxConn, m Stats) error {
+	batch := conn.NewBatch()
```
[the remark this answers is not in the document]
Harkishen-Singh: Personal preference. It makes the code look cleaner than one big query, also makes it easy to inject data.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,288 @@
...
+			continue
+		}
+		if err := syncInfoTable(t.conn, t.uuid, newStats); err != nil {
+			log.Debug("msg", "syncing new stats", "error", err.Error())
+		}
+	}
+}
+
+// metricsVisitor fills the provided newStats with most recent metric values.
+func (t *engine) metricsVisitor(newStats *telemetryStats) error {
```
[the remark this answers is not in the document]
Harkishen-Singh: the updated implementation uses `.Range()` which i think is more simpler. Please comment on this if you think it can be improved further.

pkg/telemetry/telemetry.go
Harkishen-Singh: The recent commits removes this part. Hence, resolving.

pkg/telemetry/telemetry.go
```diff
@@ -28,32 +29,24 @@ type engine struct {
...
-	wgTelemetryRoutine   *sync.WaitGroup
-
-	metricMux sync.RWMutex
-	metrics   map[column]prometheus.Metric
+	metrics sync.Map
 }
 
 type Telemetry interface {
-	StartRoutineAsync()
+	Run(context.Context, chan<- struct{})
```
antekresic, line 36: Needs comment on the input parameters.

pkg/telemetry/telemetry.go
```diff
@@ -65,7 +58,9 @@ func NewTelemetryEngine(conn pgxconn.PgxConn, uuid [16]byte) (Telemetry, error)
 func (t *engine) writeMetadata() error {
 	promscale, err := promscaleMetadata()
 	if err != nil {
-		return fmt.Errorf("promscale metadata: %w", err)
+		// Err due to Uname failed to fetch. Send the metadata that's left without Uname
+		// and simply debug log with rate-limit.
```
antekresic, line 62: Comment is too specific and refers to underlying implementation. Since the implementation changes over time, it's likely this comment will be wrong in the future. Probably best to remove it since it doesn't add much here.

pkg/telemetry/telemetry.go
```diff
@@ -212,39 +183,21 @@ func extractMetricValue(metric prometheus.Metric) (float64, error) {
...
-	var temp io_prometheus_client.Metric
-	if err := metric.Write(&temp); err != nil {
-		return false, fmt.Errorf("writing metric: %w", err)
-	}
-	if temp.Gauge != nil {
-		return true, nil
-	} else if temp.Counter != nil {
-		return true, nil
+func isCounterOrGauge(metric prometheus.Metric) bool {
+	if _, ok := metric.(prometheus.Counter); ok {
```
antekresic, line 194: Type switch would be cleaner here.

antekresic (approved):

cevian (changes requested):
A few changes needed and maybe there can be some more end-to-end tests.

pkg/telemetry/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+	bSlice := make([]byte, 65)
+	for i := 0; i < 65; i++ {
+		bSlice[i] = byte(prop[i])
+	}
+	return string(bSlice)
+}
+
+const tobsMetadataPrefix = "tobs_telemetry_"
+
+func tobsMetadata() Metadata {
```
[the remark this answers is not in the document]
Harkishen-Singh: added in following commit. Note: permission error even after using promscale extension installed image.

pkg/telemetry/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+	metadata["promscale_os_release"] = toString(uname.Release)
+	metadata["promscale_os_version"] = toString(uname.Version)
+	metadata["promscale_os_machine"] = toString(uname.Machine)
+	metadata["promscale_os_domain_name"] = toString(uname.Domainname)
+	return metadata, nil
+}
+
+func toString(prop [65]int8) string {
+	bSlice := make([]byte, 65)
+	for i := 0; i < 65; i++ {
```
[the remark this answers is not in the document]
Harkishen-Singh: because of uname allows 65 bytes only

```golang
type Utsname struct {
    Sysname    [65]int8
    Nodename   [65]int8
    Release    [65]int8
    Version    [65]int8
    Machine    [65]int8
    Domainname [65]int8
}
```

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,246 @@
...
+	io_prometheus_client "github.com/prometheus/client_model/go"
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
+	"github.com/timescale/promscale/pkg/pgmodel/model/pgutf8str"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+// ExecPlatform To fill this variable in build time, use linker flags.
+// Example: go build -ldflags="-X github.com/timescale/promscale/pkg/telemetry.ExecPlatform=<any_string>" ./cmd/promscale/
```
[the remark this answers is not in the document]
Harkishen-Singh: I gave an attempt and didn't seem straightforward. Will need help there. Plus, there seems like we may need to set github env as well.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,246 @@
...
+				columnName := statName.(string)
+				promMetric := metric.(prometheus.Metric)
+				underlyingValue, err = extractMetricValue(promMetric)
+				if err != nil {
+					err = fmt.Errorf("extracting metric value of stat '%s': %w", columnName, err)
+					return false
+				}
+
+				switch columnName {
+				case "telemetry_metrics_ingested_samples":
```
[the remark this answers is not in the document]
Harkishen-Singh: I agree map is good and that's why the metadata is done as a map. But here, we need to be align with the columns of metadata table, so having a map of arbitrary string key is not possible to load some column of the `promscale_instance_information` table.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,246 @@
...
+	if err := pgUUID.Set(uuid); err != nil {
+		return fmt.Errorf("setting pg-uuid: %w", err)
+	}
+
+	query := fmt.Sprintf(`INSERT INTO %s.promscale_instance_information
+		VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+	ON CONFlICT (uuid) DO
+		UPDATE SET
+			last_updated = $2,
+			telemetry_metrics_ingested_samples = $3,
```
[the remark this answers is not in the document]
Harkishen-Singh: Then how will we address to update independent columns of instance information table. Any suggestions please

pkg/jaeger/query/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/timescale/promscale/pkg/telemetry"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+var (
+	traceRequestsExec = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "trace",
+		Name:      "query_requests_executed_total",
```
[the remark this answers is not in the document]
Harkishen-Singh: It would be bad to name the metric as `telemetry_traces_queries_executed`. This looks mostly like a deisgn discussion that needs to be done at the doc level. The column names were kept as per design doc. But one thing we can do is

key name of metadata table = column name of instance information table = metric name

Updating according to this.

pkg/jaeger/query/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Help:      "Total number of query requests successfully executed by /getTrace and /fetchTraces API.",
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, we only increment these on successful execution.

```
if err == nil {
```
in `jaeger/query.go`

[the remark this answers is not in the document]
Harkishen-Singh: > dependency queries

dependency requests have their own metric

pkg/jaeger/query/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+	traceExecutionTime = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "trace",
+		Name:      "query_execution_duration_seconds",
+		Help:      "Time taken by a trace query for complete execution in /fetchTraces API.",
```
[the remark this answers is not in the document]
Harkishen-Singh: covering all queries will make this metric useless as the stats collected by this metric will be of no reliable use. I agree a better name is the requirement here

Harkishen-Singh:
Can someone please see why upgrade tests fail here?

```
data differs for schema _prom_catalog:
        [{name:default values:[(chunk_interval,08:00:00) (ha_lease_refresh,10s) (ha_lease_timeout,1m) (metric_compression,true) (retention_period,"90 days")]} {name:exemplar values:[]} {name:exemplar_label_key_position values:[]} {name:ha_leases values:[]} {name:ha_leases_logs values:[]} {name:ids_epoch values:[(0,"1970-01-01 00:00:00+00",t)]} {name:label values:[(1,__name__,test) (2,test,test) (3,__name__,test2) (34,testB,testB) (35,__name__,test3) (36,baz,quf) (4,foo,bar)]} {name:label_key values:[(1,__name__,__name__,__name___id) (2,test,test,test_id) (3,foo,foo,foo_id) (34,testB,testB,testB_id) (35,baz,baz,baz_id)]} {name:label_key_position values:[(test,__name__,1) (test,test,2) (test,testB,3) (test2,__name__,1) (test2,foo,2) (test3,__name__,1) (test3,baz,2)]} {name:metadata values:[]} {name:metric values:[(1,test,test,t,t,,t,,prom_data,test,f) (2,test2,test2,t,t,,t,,prom_data,test2,f) (34,test3,test3,t,t,,t,,prom_data,test3,f)]} {name:remote_commands values:[(_prom_catalog.compress_old_chunks,10,t,"
```

antekresic:
> Can someone please see why upgrade tests fail here?
> 
> ```
> data differs for schema _prom_catalog:
>         [{name:default values:[(chunk_interval,08:00:00) (ha_lease_refresh,10s) (ha_lease_timeout,1m) (metric_compression,true) (retention_period,"90 days")]} {name:exemplar values:[]} {name:exemplar_label_key_position values:[]} {name:ha_leases values:[]} {name:ha_leases_logs values:[]} {name:ids_epoch values:[(0,"1970-01-01 00:00:00+00",t)]} {name:label values:[(1,__name__,test) (2,test,test) (3,__name__,test2) (34,testB,testB) (35,__name__,test3) (36,baz,quf) (4,foo,bar)]} {name:label_key values:[(1,__name__,__name__,__name___id) (2,test,test,test_id) (3,foo,foo,foo_id) (34,testB,testB,testB_id) (35,baz,baz,baz_id)]} {name:label_key_position values:[(test,__name__,1) (test,test,2) (test,testB,3) (test2,__name__,1) (test2,foo,2) (test3,__name__,1) (test3,baz,2)]} {name:metadata values:[]} {name:metric values:[(1,test,test,t,t,,t,,prom_data,test,f) (2,test2,test2,t,t,,t,,prom_data,test2,f) (34,test3,test3,t,t,,t,,prom_data,test3,f)]} {name:remote_commands values:[(_prom_catalog.compress_old_chunks,10,t,"
> ```
> 
> But i dont think this PR does any changes to _prom_catalog

Because you changed this comment: https://github.com/timescale/promscale/pull/895/files#diff-897f70c57155ff45834ae883c44f672213f26cf9749c5ad89cc13fc54641b5d5R2

pkg/telemetry/metadata.go
```diff
@@ -0,0 +1,71 @@
...
+	metadata["promscale_os_domain_name"] = toString(uname.Domainname)
+	return metadata, nil
+}
+
+func toString(prop [65]int8) string {
+	bSlice := make([]byte, 0)
+	for i := 0; i < 65; i++ {
+		if prop[i] == 0 {
+			// Prevent adding null chars.
+			continue
```
cevian, line 47: I think the correct way is to set the length of the slice to the first 0 found. There is no guarantee the suffix is all 0 
```suggestion
			break
```

pkg/telemetry/metadata.go
```diff
@@ -0,0 +1,71 @@
...
+	metadata["promscale_os_sys_name"] = toString(uname.Sysname)
+	metadata["promscale_os_node_name"] = toString(uname.Nodename)
+	metadata["promscale_os_release"] = toString(uname.Release)
+	metadata["promscale_os_version"] = toString(uname.Version)
+	metadata["promscale_os_machine"] = toString(uname.Machine)
+	metadata["promscale_os_domain_name"] = toString(uname.Domainname)
+	return metadata, nil
+}
+
+func toString(prop [65]int8) string {
```
cevian, line 42: why dont we use the method in https://gitlab.cncf.ci/prometheus/prometheus/blob/3128875ff4de14527502be787e3f3eb480a11fc9/pkg/runtime/uname_linux.go#L30

`return string(prop[:bytes.IndexByte(prop[:], 0)])`

pkg/telemetry/telemetry.go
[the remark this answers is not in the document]
cevian: I think you can set ldflags in the build section ([docs](https://goreleaser.com/customization/build/)) so for us it would be around [line 16 ](https://github.com/timescale/promscale/blob/master/build/.goreleaser.yml#L16) and it would be 
```
ldflags:
 - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X github.com/timescale/promscale/pkg/telemetry.ExecPlatform=goreleaser
```

Also isn't a better name here BuildPlatform not ExecPlatform? This will indicate goreleaser vs apt vs rpm etc right?

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,266 @@
...
+// syncInfoTable stats with promscale_instance_information table.
+func syncInfoTable(conn pgxconn.PgxConn, uuid [16]byte, s telemetryStats) error {
+	lastUpdated := time.Now()
+
+	pgUUID := new(pgtype.UUID)
+	if err := pgUUID.Set(uuid); err != nil {
+		return fmt.Errorf("setting pg-uuid: %w", err)
+	}
+
+	query := fmt.Sprintf(`INSERT INTO %s.promscale_instance_information
```
cevian, line 240: if we had a map here instead of hard-coded columns we could build this query using the keys in the map 
```
//keys is a []string of map keys
cols := "uuid, "
cols = strings.Join(keys, ",");
indexes := "$1"
lastIndex := 1
updateStatements = make([]string, 0)
for _, col := range keys {
     lastIndex += 1
     indexes += fmt.Stringf(", $%d", lastIndex)
     updateStatements = append(updateStatements, fmt.Sprintf("%s = %d", col, lastIndex) 
}


query := fmt.Sprintf(`INSERT INTO %s.promscale_instance_information(%s)
VALUES(%s)
ON CONFLICT (uuid) DO
    UPDATE SET
         %s`, schema.PSCatalog, cols, indexes, strings.Join(updateStatements, ", "))

```

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,266 @@
...
+			)
+			t.metrics.Range(func(statName, metric interface{}) bool {
+				columnName := statName.(string)
+				promMetric := metric.(prometheus.Metric)
+				underlyingValue, err = extractMetricValue(promMetric)
+				if err != nil {
+					err = fmt.Errorf("extracting metric value of stat '%s': %w", columnName, err)
+					return false
+				}
+
```
cevian, line 169: I still think we can make thus a map. See my comment on `syncInfoTable`

cevian (changes requested):
Looks better. some additional comments

pkg/api/metrics.go
```diff
@@ -102,10 +106,10 @@ func createMetrics() *Metrics {
 				Help:      "Total number of processed metadata which failed on send to remote storage.",
 			},
 		),
-		SentSamples: prometheus.NewCounter(
+		IngestedSamples: prometheus.NewCounter(
 			prometheus.CounterOpts{
 				Namespace: util.PromNamespace,
-				Name:      "sent_samples_total",
+				Name:      "ingested_samples_total",
```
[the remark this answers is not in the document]
Harkishen-Singh: Saved here: https://github.com/timescale/promscale/releases/tag/untagged-a89ef3a8f229be0d9bdb

pkg/telemetry/metadata.go
Harkishen-Singh: That is `unix.Utsname{}`, which has `[65]byte` as type against the `syscall.Utsname{}` which has `[65]int8`. We cannot directly use `bytes.IndexByte()` with a `int8` without first converting into a slice of `byte`, which make the current approach better. Plus I don't think we should be using `unix` as windows systems would not be happy.

pkg/telemetry/telemetry.go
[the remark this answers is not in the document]
Harkishen-Singh: That approach for goreleaser should work. But I am confused here on some points:
1. Does the current master does not implement commit & version information?
2. I thought goreleaser will release for apt & rpm too, not on individual platforms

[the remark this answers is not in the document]
Harkishen-Singh: Lets discuss this on call

pkg/telemetry/telemetry.go
Harkishen-Singh: Ah, this way. An even easier would be this

```golang
columnNames := []string{"uuid", "last_updated"}
	columnValues := []interface{}{pgUUID, lastUpdated}
	indexes := []string{"$1", "$2"}
	updateStatements := []string{"last_updated = $2"}

	index := 2 // Since 1 & 2 are uuid & last_updated.
	for k, v := range stats {
		index++
		indexes = append(indexes, fmt.Sprintf("$%d", index))
		columnNames = append(columnNames, k)
		columnValues = append(columnValues, v)
		updateStatements = append(updateStatements, fmt.Sprintf("%s = $%d", k, index))
	}

	query := fmt.Sprintf(`INSERT INTO %s.promscale_instance_information(%s) VALUES (%s)
	ON CONFLICT (uuid) DO UPDATE SET %s`,
		schema.PromscaleCatalog,
		strings.Join(columnNames, ", "),
		strings.Join(indexes, ", "),
		strings.Join(updateStatements, ", "),
	)
```

But do NOTE: Metrics are registered individually by each module, meaning when the insert starts, some modules might not have registered their metrics. Hence, I will have to remove the `NOT NULL` constrain, even though this won't matter when update happens at `55 mins`.

pkg/telemetry/telemetry.go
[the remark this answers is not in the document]
Harkishen-Singh: `binary, deb, rpm` in ExecPlatform (replaced by BuildPlatform)

pkg/api/metrics.go
```diff
@@ -22,10 +22,12 @@ type Metrics struct {
 	ReceivedMetadata      prometheus.Counter
 	FailedSamples         prometheus.Counter
 	FailedMetadata        prometheus.Counter
-	SentSamples           prometheus.Counter
+	IngestedSamples       prometheus.Counter
 	SentMetadata          prometheus.Counter
 	SentBatchDuration     prometheus.Histogram
```
jgpruitt, line 27: Why was "SentSamples" changed to "IngestedSamples" but "SentMetadata" and "SentBatchDuration" not changed?

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,262 @@
...
+		rows, err := results.Query()
+		if err != nil {
+			// Metadata table not accessible. Try for func in Promscale extension.
+			return fmt.Errorf("error querying results: %w", err)
+		}
+		for rows.Next() {
+			if err != nil {
+				err = rows.Err()
+			}
```
jgpruitt, line 112: I don't this this if statement is correct unless I'm confused. What exactly is it supposed to do? It seems to me err will always be nil here.

pkg/migrations/sql/preinstall/013-telemetry.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS SCHEMA_PS_CATALOG.promscale_instance_information (
+    uuid                                                UUID NOT NULL PRIMARY KEY,
+    last_updated                                        TIMESTAMPTZ NOT NULL,
+    promscale_ingested_samples_total                    BIGINT DEFAULT 0,
+    promscale_metrics_queries_executed_total            BIGINT DEFAULT 0,
+    promscale_metrics_queries_timedout_total            BIGINT DEFAULT 0,
+    promscale_metrics_queries_failed_total              BIGINT DEFAULT 0,
+    promscale_trace_query_requests_executed_total       BIGINT DEFAULT 0,
+    promscale_trace_dependency_requests_executed_total  BIGINT DEFAULT 0,
+    deletable                                           BOOLEAN DEFAULT TRUE
```
jgpruitt, line 10: weren't we going to rename this column to `is_reset_row` or something like that?

jgpruitt (changes requested):

pkg/api/metrics.go
Harkishen-Singh: That is changed in order to make metric names equal to column names in the instance information table

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,262 @@
...
+	// It must be called after creating the telemetry engine.
+	RegisterMetric(columnName string, gaugeOrCounterMetric prometheus.Metric) error
+	Stop()
+}
+
+func NewTelemetryEngine(conn pgxconn.PgxConn, uuid [16]byte) (Telemetry, error) {
+	// Warn the users about telemetry collection.
+	log.Warn("msg", "Promscale collects anonymous usage telemetry data for project improvements while being GDPR compliant. "+
+		"If you wish not to participate, please turn off sending telemetry in TimescaleDB. "+
+		"Details about the process can be found at https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry/#disabling-telemetry")
```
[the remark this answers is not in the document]
Harkishen-Singh: I think we had a discussion on this where we decided to keep GDPR compliance in Promscale. We can discuss this again in the coming check in is call.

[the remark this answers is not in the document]
jgpruitt: I agree that we should be compliant, I just wonder whether or not we should mention GDPR compliance specifically in the message we print.

pkg/telemetry/telemetry.go
Harkishen-Singh: Ah, thanks for catching this. The if here is to catch the first most error and return, but I forgot to return the actual error.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,257 @@
...
+func (t *engine) Start() {
+	t.stop = make(chan struct{})
+	go func() {
+		collect := time.NewTicker(time.Minute * 55) // Collect telemetry info every 55 minutes since housekeeper fills the _timescaledb_catalog.metadata table every 1 hour.
+		defer collect.Stop()
+
+		for {
+			select {
+			case <-t.stop:
+				close(t.stop)
```
cevian, line 138: You always want the writer of a channel to do the close not the reader.
```
func (t *engine) Stop() {
         t.stop <- struct{}{}
	close(t.stop)
}
```

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,257 @@
...
+	switch metric.(type) {
+	case prometheus.Counter, prometheus.Gauge:
+		return true
+	default:
+		return false
+	}
+}
+
+// syncInfoTable stats with promscale_instance_information table.
+func syncInfoTable(conn pgxconn.PgxConn, uuid [16]byte, stats map[string]float64) error {
```
cevian, line 203: This needs to be tested in some tests.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,257 @@
...
+	Start()
+	// RegisterMetric registers a Prometheus Gauge or Counter metric for telemetry visitor.
+	// It must be called after creating the telemetry engine.
+	RegisterMetric(columnName string, gaugeOrCounterMetric prometheus.Metric) error
+	Stop()
+}
+
+func NewTelemetryEngine(conn pgxconn.PgxConn, uuid [16]byte) (Telemetry, error) {
+	// Warn the users about telemetry collection.
+	log.Warn("msg", "Promscale collects anonymous usage telemetry data to help the Promscale team better understand and assist users. "+
```
cevian, line 48: We should check that timescaledb telemetry isn't off before printing this. If people turn telemetry off and still see this they'd get confused/upset.

cevian (changes requested):

pkg/runner/runner.go
```diff
@@ -98,6 +110,20 @@ func Run(cfg *Config) error {
 	log.Info("msg", "Starting up...")
 	log.Info("msg", "Listening", "addr", cfg.ListenAddr)
 
+	telemetryEngine, err := telemetry.NewTelemetryEngine(client.Connection, PromscaleID)
+	if err != nil {
+		log.Error("msg", "aborting startup due to error in setting up telemetry-engine", "err", err.Error())
+		return fmt.Errorf("creating telemetry-engine: %w", err)
+	}
+	defer telemetryEngine.Stop()
+	telemetryEngine.Start()
```
antekresic, line 119: Nit: reverse order of these two lines.

pkg/telemetry/metadata.go
```diff
@@ -0,0 +1,80 @@
...
+		k, v := decode(envVar)
+		if strings.HasPrefix(k, tobsMetadataPrefix) {
+			metadata[strings.ToLower(k)] = v // Convert to lower case as metadata table should have everything in lowercase.
+		}
+	}
+	return metadata
+}
+
+func decode(s string) (key, value string) {
+	arr := strings.Split(s, "=")
```
antekresic, line 78: Let's use `SplitN` so we avoid splitting the value of the env var into multiple parts.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+
+type Telemetry interface {
+	Start()
+	// RegisterMetric registers a Prometheus Gauge or Counter metric for telemetry visitor.
+	// It must be called after creating the telemetry engine.
+	RegisterMetric(columnName string, gaugeOrCounterMetric prometheus.Metric) error
+	Stop()
+	// Sync syncs the registered metric values with the information table. This is used only
+	// in end-to-end tests.
+	Sync() error
```
antekresic, line 42: If this is only for tests, should it really be a part of the public interface?

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+type engine struct {
+	uuid [16]byte
+	conn pgxconn.PgxConn
+
+	stop chan struct{}
+
+	metrics sync.Map
+}
+
+type Telemetry interface {
```
antekresic, line 34: IMO more fitting name here is `Engine`. 

Would reduce stutter: `telemetry.Engine` vs `telemetry.Telemetry` 
and it seems you call it engine in var names everywhere.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+	// RegisterMetric registers a Prometheus Gauge or Counter metric for telemetry visitor.
+	// It must be called after creating the telemetry engine.
+	RegisterMetric(columnName string, gaugeOrCounterMetric prometheus.Metric) error
+	Stop()
+	// Sync syncs the registered metric values with the information table. This is used only
+	// in end-to-end tests.
+	Sync() error
+}
+
+func NewTelemetryEngine(conn pgxconn.PgxConn, uuid [16]byte) (Telemetry, error) {
```
antekresic, line 45: There is a Go proverb that goes:
`Accept interfaces, return structs`

Basically, you return struct (pointers) and the user decides whether he will treat it as an interface or a struct pointer.

I would change `Telemetry` to `*engine` in the return.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+	isTelemetryOff, err := isTelemetryOff(conn)
+	if err != nil {
+		log.Debug("msg", "unable to get TimescaleDB telemetry configuration. Maybe TimescaleDB is not installed", "err", err.Error())
+	}
+	if !isTelemetryOff {
+		// Warn the users about telemetry collection only if telemetry collection is enabled.
+		log.Warn("msg", "Promscale collects anonymous usage telemetry data to help the Promscale team better understand and assist users. "+
+			"This can be disabled via the process described at https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry/#disabling-telemetry")
+	}
+	engine := &engine{
```
antekresic, line 55: Nit: this can really hurt you, shadowing struct names with variable names. I usually take care not to do that.

Just to avoid any confusion when reading the code.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, fmt.Errorf("writing metadata: %w", err)
+	}
+	return engine, nil
+}
+
+func isTelemetryOff(conn pgxconn.PgxConn) (bool, error) {
+	var state string
+	err := conn.QueryRow(context.Background(), "SHOW timescaledb.telemetry_level").Scan(&state)
+	if err != nil {
+		// Maybe TimescaleDB is not installed, in that case, we cannot send telemetry, hence don't log it.
```
antekresic, line 69: This comment confuses me a bit. What don't we log here? I don't see any logs :confused:

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+const (
+	metadataUpdateWithExtension = "SELECT update_tsprom_metadata($1, $2, $3)"
+	metadataUpdateNoExtension   = "INSERT INTO _timescaledb_catalog.metadata(key, value, include_in_telemetry) VALUES ('promscale_' || $1, $2, $3) ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value, include_in_telemetry = EXCLUDED.include_in_telemetry"
+)
+
+// writeToTimescaleMetadataTable syncs the metadata/stats with telemetry metadata table and returns the last error if any.
+func writeToTimescaleMetadataTable(conn pgxconn.PgxConn, m Metadata) error {
+	update := func(queryFormat string) error {
```
antekresic, line 101: What's the reason for having this as an inline function? Don't see much of a reason and it hurts readability by increasing nesting in this function.

pkg/telemetry/telemetry.go
antekresic: Also, why is this not a method on the engine? Seems to be using engine connection and uuid.

antekresic (changes requested):

pkg/telemetry/telemetry.go
Harkishen-Singh: Because there isn't any good reason to write a new function block just to reduce redundancy. Sorry, but this is a very simple function, which just updates the table. Otherwise, we will have to write a func that accepts `conn, m, queryFormat` which I don't think we wanna do.

antekresic (approved):

jgpruitt (approved):

pkg/migrations/sql/preinstall/012-telemetry.sql
```diff
@@ -0,0 +1,15 @@
...
+    last_updated                                        TIMESTAMPTZ NOT NULL,
+    promscale_ingested_samples_total                    BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_executed_total            BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_timedout_total            BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_failed_total              BIGINT DEFAULT 0 NOT NULL,
+    promscale_trace_query_requests_executed_total       BIGINT DEFAULT 0 NOT NULL,
+    promscale_trace_dependency_requests_executed_total  BIGINT DEFAULT 0 NOT NULL,
+    is_counter_reset_row                                BOOLEAN DEFAULT FALSE NOT NULL, -- counter reset row has '00000000-0000-0000-0000-000000000000' uuid
+    CHECK(uuid = '00000000-0000-0000-0000-000000000000' OR NOT is_counter_reset_row),
+    UNIQUE (uuid, is_counter_reset_row)
```
cevian, line 12: we don't actually need this because UUID is a PK
```suggestion
```

pkg/migrations/sql/preinstall/012-telemetry.sql
```diff
@@ -0,0 +1,15 @@
...
+    uuid                                                UUID NOT NULL PRIMARY KEY,
+    last_updated                                        TIMESTAMPTZ NOT NULL,
+    promscale_ingested_samples_total                    BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_executed_total            BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_timedout_total            BIGINT DEFAULT 0 NOT NULL,
+    promscale_metrics_queries_failed_total              BIGINT DEFAULT 0 NOT NULL,
+    promscale_trace_query_requests_executed_total       BIGINT DEFAULT 0 NOT NULL,
+    promscale_trace_dependency_requests_executed_total  BIGINT DEFAULT 0 NOT NULL,
+    is_counter_reset_row                                BOOLEAN DEFAULT FALSE NOT NULL, -- counter reset row has '00000000-0000-0000-0000-000000000000' uuid
+    CHECK(uuid = '00000000-0000-0000-0000-000000000000' OR NOT is_counter_reset_row),
```
cevian, line 11: to prevent uuid = zero and counter_reset_row=false
```suggestion
    CHECK((uuid = '00000000-0000-0000-0000-000000000000' OR NOT is_counter_reset_row) AND(uuid != '00000000-0000-0000-0000-000000000000' OR is_counter_reset_row)),
```

pkg/runner/runner.go
```diff
@@ -98,6 +110,21 @@ func Run(cfg *Config) error {
...
 
+	telemetryEngine, err := telemetry.NewEngine(client.Connection, PromscaleID)
+	if err != nil {
+		log.Error("msg", "aborting startup due to error in setting up telemetry-engine", "err", err.Error())
+		return fmt.Errorf("creating telemetry-engine: %w", err)
+	}
+
+	telemetryEngine.Start()
+	defer telemetryEngine.Stop()
+
```
cevian, line 121: is it safer to start the engine after the `err = api.RegisterMetricsForTelemetry(telemetryEngine)`call?

cevian (approved):
Some changes before merge requested

pkg/runner/runner.go
Harkishen-Singh: Yes, there should not be any problem as we are using concurrency safe `sync.Map`.

PR #899, opened 2021-11-19
Add tests to validate the tag operators

jgpruitt:
There are 15 operators defined in the ps_tag schema. Each
can deal with at least one and in many cases many input types.
Test that the operators return the correct composite types
and that the fields contain the correct values.

These tests do not modify the contents of the database.

### DIFF
--- a/pkg/tests/end_to_end_tests/tag_op_test.go
+++ b/pkg/tests/end_to_end_tests/tag_op_test.go
@@ -1,0 +1,457 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package end_to_end_tests
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+)
+
+type operator struct {
+	operator   string
+	returnType string
+}
+
+var operators = []operator{
+	{
+		operator:   "==",
+		returnType: "tag_op_equals",
+	},
+	{
+		operator:   "!==",
+		returnType: "tag_op_not_equals",
+	},
+	{
+		operator:   "#<",
+		returnType: "tag_op_less_than",
+	},
+	{
+		operator:   "#<=",
+		returnType: "tag_op_less_than_or_equal",
+	},
+	{
+		operator:   "#>",
+		returnType: "tag_op_greater_than",
+	},
+	{
+		operator:   "#>=",
+		returnType: "tag_op_greater_than_or_equal",
+	},
+}
+
+type testValue struct {
+	inputType   string
+	inputValue  string
+	outputType  string
+	outputValue string
+}
+
+var testValues = []testValue{
+	{
+		inputType:   "smallint",
+		inputValue:  `-1::smallint`,
+		outputType:  "number",
+		outputValue: `-1`,
+	},
+	{
+		inputType:   "smallint",
+		inputValue:  ` 0::smallint`,
+		outputType:  "number",
+		outputValue: `0`,
+	},
+	{
+		inputType:   "smallint",
+		inputValue:  ` 1::smallint`,
+		outputType:  "number",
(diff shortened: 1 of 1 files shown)


pkg/tests/end_to_end_tests/tag_op_test.go
```diff
@@ -0,0 +1,460 @@
+package end_to_end_tests
```
Harkishen-Singh, line 1: header

pkg/tests/end_to_end_tests/tag_op_test.go
```diff
@@ -0,0 +1,460 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"testing"
```
Harkishen-Singh, line 7: Nit:

```suggestion
	"context"
	"fmt"
	"testing"
	
	"github.com/jackc/pgx/v4/pgxpool"
```

pkg/tests/end_to_end_tests/tag_op_test.go
```diff
@@ -0,0 +1,460 @@
...
+		}
+		if outputValue != tval.outputValue {
+			t.Errorf("Expected the output values to be equal: %v != %v", outputValue, tval.outputValue)
+		}
+		if outputType != tval.outputType {
+			t.Errorf("Expected the output types to be equal: %v != %v", outputType, tval.outputType)
+		}
+		if returnType != opr.returnType {
+			t.Errorf("Expected the return types to be equal: %v != %v", returnType, opr.returnType)
+		}
```
Harkishen-Singh, line 458: you can use `require.Equal` to simplify this.

pkg/tests/end_to_end_tests/tag_op_test.go
```diff
@@ -0,0 +1,460 @@
...
+	t.Run(name, func(t *testing.T) {
+		row := db.QueryRow(ctx, fmt.Sprintf(qry, opr.operator, tval.inputValue))
+		var (
+			tagKey      string
+			outputValue string
+			outputType  string
+			returnType  string
+		)
+		err := row.Scan(&tagKey, &outputValue, &outputType, &returnType)
+		if err != nil {
```
Harkishen-Singh, line 444:
```suggestion
		require.NoError(t, err)
```

pkg/tests/end_to_end_tests/tag_op_test.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func TestTagOpExpressions(t *testing.T) {
+	var ctx = context.Background()
+	databaseName := fmt.Sprintf("%s_tag_op", *testDatabase)
+	withDB(t, databaseName, func(db *pgxpool.Pool, tb testing.TB) {
+		testTagOperators(t, db, ctx)
+		testTagOpRegexpMatches(t, db, ctx)
+		testTagOpRegexpNotMatches(t, db, ctx)
+		testTagOpJsonPathExists(t, db, ctx)
```
Harkishen-Singh, line 335: instead of passing down `t`, i think it will be better with passing `tb`, since then the error/panic trace will have a more accurate trace.

Harkishen-Singh (approved):
Minor nits, rest looks good.

pkg/tests/end_to_end_tests/tag_op_test.go
jgpruitt: Unfortunately, there is no `Run` method on the `TB` -- only on `T`. I need the `Run` method so I can execute each test as a subtest.

pkg/tests/end_to_end_tests/tag_op_test.go
jgpruitt: Awesome!

pkg/tests/end_to_end_tests/tag_op_test.go
jgpruitt: Awesome!

cevian (approved):
The tests look good (one nit)

PR #907, opened 2021-11-22
Test trace tree database functions

jgpruitt:
Added multiple tests to ensure that the database functions
related to querying the trace tree structure are working
properly.

Functions tested:

* trace_tree
* upstream_spans
* downstream_spans
* span_tree
* sibling_spans


pkg/tests/end_to_end_tests/trace_tree_test.go
```diff
@@ -0,0 +1,392 @@
...
+		require.False(t, rows.Next(), "Found more result rows than expected")
+	}
+
+	t.Run("upstream_spans", func(t *testing.T) {
+		const qry = `
+		select span_id, coalesce(parent_span_id, -1), dist, idx, path::text
+		from upstream_spans('3dadb2bf-0035-433e-b74b-9075cc9260e8', 13)
+		order by span_id`
+		expectedResults := []result{
+			{1, -1, 3, 4, "{13,6,2,1}"},
```
cevian, line 218: hmm I think "{13,6,2,1}" should be "{1,2,6,13}" shouldn't we keep the order in the paths consistently upstream->downstream for all these functions?

pkg/tests/end_to_end_tests/trace_tree_test.go
```diff
@@ -0,0 +1,392 @@
...
+		require.False(t, rows.Next(), "Found more result rows than expected")
+	}
+
+	t.Run("span_tree", func(t *testing.T) {
+		const qry = `
+		select span_id, coalesce(parent_span_id, -1), dist, idx, path::text
+		from span_tree('3dadb2bf-0035-433e-b74b-9075cc9260e8', 6) 
+		order by span_id`
+		expectedResults := []result{
+			{1, -1, -2, 3, "{6,2,1}"},
```
cevian, line 370: same question here about ordering of "{6,2,1}" vs "{1,2,6}"

pkg/tests/end_to_end_tests/trace_tree_test.go
cevian: related question about whether dist should be positive or negative and relation with span_tree

pkg/tests/end_to_end_tests/trace_tree_test.go
```diff
@@ -0,0 +1,392 @@
...
+			{2, 1, -1, 2, "{6,2}"},
```
cevian, line 371: why is dist and not idx negative? was there a reason for this?

cevian (changes requested):
These tests are very good. Have some questions about function semantics

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: I'm not adamant that it stay the way I coded it, but let me explain my reasoning.
All the results are from the perspective of the span id passed as an argument. Paths going upstream (towards the root) have a negative distance, and paths going downstream have a positive distance. This makes it easy to determine the direction, and an `abs(dist)` is easy if you don't care. All the paths start with the span id that was passed with an arg. The index is the index of the span of the result row in the path array.

jgpruitt: Okay. In all of these functions, I rendered the path from the perspective of the span_id that was passed as an argument. It's the path, from the start span to the span identified by the result row.

It doesn't "have" to be that way, but that was my thinking.

jgpruitt: In `span_tree`, I made upstream distances negative and downstream distances positive so that you can easily determine the direction and an `abs(dist)` is easy if you need it.

We can ditch that idea and always use positive distances, or we can make the distances returned from `upstream_spans` negative if you want consistency.

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: Same as above. It's rendered from the "perspective" of the span_id passed as an arg.

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: `idx` is the index into the path array where you will find the span_id identified by the result row.

I'm open to suggestions about changing it.

pkg/tests/end_to_end_tests/trace_tree_test.go
cevian: ahh I think the stuff with idx and dist makes sense now. I also see what you mean by "from the perspective of the span" but I do think having the paths go in a consistent order would make things more straightforward. Also in terms of documentation than a path is a defined thing with a defined order. So I think it would be better to change

jgpruitt: Okay, I don't feel strongly either way. Will make the change and update the tests

jgpruitt: what about `dist`? do you prefer it to be always positive, or what?

cevian (approved):
Oh I see you removed `idx`. I think that's right.

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: removed the `idx` column.

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: changed this. paths are always ordered from upstream to downstream for consistency.

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: `dist` changed to be negative for all upstream paths across the board for consistency

jgpruitt:
ready for you to review @niksajakovljevic 
I need one more approval

pkg/tests/end_to_end_tests/trace_tree_test.go
```diff
@@ -0,0 +1,381 @@
...
+			var actual = result{}
+			err = rows.Scan(&actual.SpanId, &actual.ParentSpanId, &actual.Dist, &actual.Path)
+			require.NoError(t, err, "Failed to scan row. Expected row for span id %v: %v", expected.SpanId, err)
+
+			require.Equal(t, expected, actual)
+		}
+		require.False(t, rows.Next(), "Found more result rows than expected")
+	}
+
+	t.Run("upstream_spans", func(t *testing.T) {
```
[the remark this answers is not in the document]
jgpruitt: There is another function `span_tree` that returns both upstream and downstream paths from the perspective of a given span. In the results of this function, a negative dist is the way to distinguish between the upstream and downstream paths.

So, for consistency, `upstream_spans` also returns negative dists.

pkg/tests/end_to_end_tests/trace_tree_test.go
```diff
@@ -0,0 +1,381 @@
...
+			{6, 2, -1, "{6,13}"},
+			{13, 6, 0, "{13}"},
+		}
+		eval(t, qry, expectedResults)
+	})
+
+	t.Run("upstream_spans_max_dist", func(t *testing.T) {
+		const qry = `
+		select span_id, coalesce(parent_span_id, -1), dist, path::text
+		from upstream_spans('3dadb2bf-0035-433e-b74b-9075cc9260e8', 13, 2)
```
niksajakovljevic, line 224: In the context of previous discussion on absolute numbers: imho it feels a bit awkward to specify max depth argument as positive number but get negatives in the result.

jgpruitt: Good point. What do you suggest? Use positive numbers for dist and a separate boolean flag in the results of `span_tree` to indicate direction?

pkg/tests/end_to_end_tests/trace_tree_test.go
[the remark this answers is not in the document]
niksajakovljevic: Ok so now I understand that negatives are introduces because of the `span_tree` function. Do you have a specific use-case in mind where negatives are helpful? I imagine that if user gets span tree and loads it in memory it is fairly easy to figure out parents or children by traversing the tree.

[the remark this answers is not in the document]
jgpruitt: The negative is an easy and efficient way to determine direction.

niksajakovljevic (approved):
Looks good! My main objection is usage of negative numbers to express distance (wiki: `Both distance and displacement measure the movement of an object. Distance cannot be negative, and never decreases.`). However I see that you and @cevian had been discussing that before so I don't want to make it a blocker...

pkg/tests/end_to_end_tests/trace_tree_test.go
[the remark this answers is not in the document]
jgpruitt: `dist < 0 as is_upstream`
vs
`path[0] = span_id as is_upstream`

cevian:
> Good point. What do you suggest? Use positive numbers for dist and a separate boolean flag in the results of span_tree to indicate direction?

I am fine with this if you think its a better UIUX

pkg/tests/end_to_end_tests/trace_tree_test.go
jgpruitt: will use positive dist across the board and add is_upstream and is_downstream flags

PR #908, opened 2021-11-23
Remove deprecated TS_PROM_ prefixed env-var support

JamesGuthrie:
The TS_PROM_ prefix was deprecated in Promscale release 0.1.4, released
in December 2020.


antekresic (approved):
I'm OK with this, lets have @cevian chime in is it the time to do this.

cevian (approved):
LGTM - let's just add a note in the draft release notes that this is a potential compatibility break.

Issue #910, opened 2021-11-24
Remove cron info from docker install doc 


cevian:
Cron jobs not necessary in timescaledb 2.0

Note #495

PR #911, opened 2021-11-24
Add note in cron jobs step in bare-metal install docs

VineethReddy02:
Today Promscale doesn't need an explicit cronjob to perform maintenance tasks. This is internally performed using User Defined Jobs.


JamesGuthrie:
This fixes #910

JamesGuthrie:
What about in the docker installation notes? https://github.com/timescale/promscale/blob/master/docs/docker.md

VineethReddy02:
Good catch @JamesGuthrie :)

Btw I just found we have two files almost describing the same steps on the installation of bare-metal i.e. 
https://github.com/timescale/promscale/blob/master/docs/bare-metal-promscale-stack.md
https://github.com/timescale/promscale/blob/master/docs/binary.md

It's ti,e to drop one.

JamesGuthrie:
Sorry, one more thing: while we're at it I suppose that we can also remove `docs/scripts/*`.

cevian (approved):

cevian:
@VineethReddy02 the idea for https://github.com/timescale/promscale/blob/master/docs/bare-metal-promscale-stack.md was to do the entire stack (as opposed to just Promscale connector). Do you think that no-longer makes sense?

jgpruitt (approved):

JamesGuthrie (changes requested):
Can we remove the `docs/scripts` directory?

VineethReddy02:
@cevian makes sense for now to maintain two files on bare metal install of Promscale and the entire stack will re-iterate over this approach later. 

@JamesGuthrie I made changes inline to Ramon's changes in #996 to add a note rather than dropping it completely for now considering backward compatibility, for users that are still using TimescaleDB `1.X`.

cevian (approved):

JamesGuthrie (approved):

cevian:
@VineethReddy02 should this be merged?

VineethReddy02:
Yup! Merging this right away.

PR #930, opened 2021-11-29
Tracing put tests

jgpruitt:
### Add tests for tracing put/get db functions

Tests:
- put_schema_url
- put_instrumentation_lib
- put_tag_key
- put_tag
- get_tag_map
- put_operation

Also removed get_operation as it was unused.


jgpruitt:
ready for review!

bitwalker (approved):
LGTM! Granted, I'm coming into this without much context, but I did some poking around enough to feel confident enough in a review.

pkg/version/version.go
```diff
@@ -38,7 +38,7 @@ var (
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
 
-	Promscale                           = "0.7.0-beta.1.dev.3"
+	Promscale                           = "0.7.0-beta.1.dev.4"
```
VineethReddy02, line 41: Any specific reason why are we changing the dev version on every PR? Doesn't this add friction as other PR's will collide with same `dev.X` tags and requires to rebase as other PR's get in with the same tag?

VineethReddy02 (approved):

pkg/version/version.go
jgpruitt: As I understand it, that last number in the `dev.X` controls which SQL scripts get run from the upgrade folder. In other words, if you add a new upgrade script (like 4-drop_get_operation.sql) then you have to increase the version to `dev.4` or your upgrade script won't run.

jgpruitt: I could be mistaken, and/or not answering your question exactly.

PR #934, opened 2021-11-30
Fix memory leak with SamplesQuerier

cevian:
Fixes memory leak by adding Close() calls to SampleQuerier instances
This matters whenever Select is called as in those cases the Close
trickles down to sampleRow.Close() calls which returns memory to pools.

Fixes timescale#933.

### DIFF
--- a/pkg/api/label_values.go
+++ b/pkg/api/label_values.go
@@ -34,6 +34,8 @@ func labelValues(queryable promql.Queryable) http.HandlerFunc {
 			respondError(w, http.StatusInternalServerError, err, "internal")
 			return
 		}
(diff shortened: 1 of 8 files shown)


niksajakovljevic (approved):

bitwalker (approved):
LGTM! The only thing I'm not sure about is how to confirm that this covers all the places where the querier needs to be closed - I'm assuming this is all uses, but wanted to at least note the uncertainty for other reviewers.

Harkishen-Singh (approved):

cevian:
@bitwalker yup this is all uses.

PR #939, opened 2021-12-01
Make prom-migrator unit test independent of current time

antekresic:
Testing that depends on current time is usually flaky. This change wraps
`time.Now().Unix()` calls into an independent function which can be
manipulated to always return the same value thus making test more reliable.


paulfantom (approved):
lgtm

Harkishen-Singh (approved):

PR #943, opened 2021-12-02
Prepare for the 0.7.0 release

JamesGuthrie:


JamesGuthrie:
There were a couple of things which I'm not sure if I got right:

- I'm not sure if I should bump the version number for `prom-migrator` in `pkg/version/version.go`.
- The instructions say to bump the version in `.goreleaser.yml`, but that doesn't seem to be necessary (at least for `promscale`, there are versions for `prom-migrator` and `jaeger-query-proxy`)

antekresic (approved):
Looks good. AFAIK:
- no need to bump other versions if we are not releasing them
- Promscale picks up the version from the git tag, no need to update `.goreleaser.yaml`

pkg/version/version.go
```diff
@@ -38,7 +38,7 @@ var (
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
 
-	Promscale                           = "0.7.0-beta.1.dev.5"
+	Promscale                           = "0.7.0"
 	PrevReleaseVersion                  = "0.7.0-beta.1"
 	PromMigrator                        = "0.0.2"
```
cevian, line 43: We need to bump the patch here as we do have a new commit for prom-migrator https://github.com/timescale/promscale/commit/b2f41d5da7481d8a779568bb778ae0dba7fbf557#diff-2eaa38bcaeeee666a97b66de9efff82d6b4bdf3dda7df693b9e1cf58940cc342

cevian: We also have to bump jaeger-proxy because of https://github.com/timescale/promscale/commit/aaa3f83caf15c9717c4351f0142ca1625cc1ece6#diff-792b22d12d8ee4227140a940fbff83b8d78f52d9f025fe841241d5c266f4b9cc

cevian (changes requested):
Everything is good except the items you highlighted:
- You do have to bump prom-migrator and jaeger proxy. The only times it's safe not to bump is if there had been NO changes at all. That isn't the case here.
- The Promscale version is derived from the tag. No need for any changes there.

cevian (approved):

PR #949, opened 2021-12-03
Fix upgrade scripts

cevian:
We previously had two upgrade branches
0.6.2=>0.7.0
0.7.0-beta.1=>0.7.0

The first upgrade branch was broken because version upgrade scripts
from 0.6.2=>0.7.0-beta.1 were in version 0.6.1-dev.* which is incorrect
Logically, they should be in a version that's > than 0.6.2, so we put
it into 0.6.99-dev.*


jgpruitt (approved):

JamesGuthrie (approved):
LGTM.

cevian:
Upgrades tested locally from the following versions:
- 0.6.2 
- 0.6.1
- 0.7.0
- 0.7.1-beta.1

antekresic (approved):
lgtm

PR #956, opened 2021-12-08
Add tests for the operation_calls database function

jgpruitt:
Also fixes a bug in the operation_calls database function where the parent span's start_time was NOT being compared against the max_start_time argument


pkg/migrations/sql/idempotent/tracing-functions.sql
```diff
@@ -242,18 +242,18 @@ RETURNS TABLE
...
+        parent.operation_id as parent_operation_id,
+        child.operation_id as child_operation_id,
+        count(*) as cnt
+    FROM
+        _ps_trace.span child
+    INNER JOIN
+        _ps_trace.span parent ON (parent.span_id = child.parent_span_id AND parent.trace_id = child.trace_id)
+    WHERE
+        child.start_time > _start_time_min AND child.start_time < _start_time_max AND
+        parent.start_time > _start_time_min AND parent.start_time < _start_time_max
```
Harkishen-Singh, line 255: This isn't a change in the PR, but do we not have an including bound on the lower end of time? I think projects that have time-based responses have this, including our find_traces API.

Basically,
```postgresql
 child.start_time >= _start_time_min AND child.start_time < _start_time_max AND
 parent.start_time >= _start_time_min AND parent.start_time < _start_time_max
```

pkg/tests/end_to_end_tests/trace_operation_calls_test.go
```diff
@@ -0,0 +1,378 @@
+package end_to_end_tests
```
Harkishen-Singh, line 1: file license header please

Harkishen-Singh (approved):
Just a small comment.

pkg/tests/end_to_end_tests/trace_operation_calls_test.go
bitwalker: Is there a reason to add license headers to files? We have a repo-wide LICENSE file, so adding headers to each file is just extra clutter.

pkg/migrations/sql/idempotent/tracing-functions.sql
jgpruitt: @Harkishen-Singh I thought the same thing. I've always used inclusive-exclusive, however @cevian asked me to check how Jaeger does it, since this function is meant to be Jaeger compatible, and he thought it was exclusive-exclusive. I could definitely be wrong, but the Jaeger code appears to be exclusive-exclusive to me.

pkg/tests/end_to_end_tests/trace_operation_calls_test.go
jgpruitt: I have added it, but I also question its utility. If we really need the license in each file, I would argue that we also need the license as a comment embedded in each database function's body.

PR #959, opened 2021-12-08
Add support for Telemetry Housekeeper

Harkishen-Singh:
The second PR in the series of Telemetry enhancements.

Note for reviewers: Please refer to the commit message in case reviewing needs help with additional information. I have tried to keep it descriptive.


bitwalker (commented):
Just one suggestion, but otherwise looks good to me. I haven't yet looked at the work done in the first PR since I wasn't reviewer for that, so that's the only caveat from my perspective.

jgpruitt:
@Harkishen-Singh I have a proposal for you. I think the housekeeping can be both simpler and more resilient. I proposal that ALL promscale connector instances be housekeepers. Each one would:

Harkishen-Singh:
Thank you @jgpruitt  for proposing it. If i understood correctly, in the above proposal, every promscale will update

1. its own row (that matches its own uuid)
2. the counter reset row in case it finds that there is some peomscale instance that has gone stale

I think this can be done and is a simple and better approach provided there wasn't another stuff that's being missed here. The SQL based stats statements.

If all promscales were to be the house keepers, then they will also do updation of SQL stats (from design doc, 'what are the number of unique metrics in database', etc, this will be in the next PR), which according to earlier plan, would be done by only one promscale which was the housekeeper, but here, would be done by all promscales depending on whether the last update on counter reset row, was done an hour ago or not.

I think this can get confusing, since an update to counter reset row should happen only on stale rows.

Does this make sense?

pkg/runner/client.go
```diff
@@ -28,7 +28,7 @@ var (
 	migrationLockError = fmt.Errorf("Could not acquire migration lock. Ensure there are no other connectors running and try again.")
 )
 
-func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, error) {
+func CreateClient(cfg *Config, promMetrics *api.Metrics) (*pgclient.Client, string, error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: Thank you for catching that.

pkg/telemetry/housekeeper.go
```diff
@@ -0,0 +1,148 @@
...
+		// Promscale session.
+		return false, nil
+	}
+	acquired, err := t.lock.GetAdvisoryLock()
+	if err != nil {
+		return false, fmt.Errorf("attemping telemetry pg-advisory-lock: %w", err)
+	}
+	if acquired {
+		t.isHousekeeper = true
+	}
```
antekresic, line 42:
```suggestion
	t.isHousekeeper = acquired
```

pkg/telemetry/housekeeper.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return fmt.Errorf("get instance stats: %w", err)
+	}
+	t.engineCopy.writeToTimescaleMetadataTable(convertStatsToMetadata(informationTableStats))
+	return nil
+}
+
+const housekeepingDuration = time.Hour
+
+func (t *housekeeper) housekeeping() {
```
antekresic, line 79: Might be better to call this `run` instead of `housekeeping`. Seems more idiomatic.

pkg/telemetry/housekeeper.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgxconn"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+const housekeeperLockId = 0x2D829A932AAFCEDE // Random.
+
+type housekeeper struct {
+	engineCopy    *engineImpl
```
antekresic, line 22: Why did you feel the need to call this `engineCopy`? Couldn't it be just `engine`? Especially since its a pointer to the engine and not a internal copy.

pkg/telemetry/housekeeper.go
```diff
@@ -0,0 +1,148 @@
...
+		&samples, &queriesExec, &queriesTimeout, &queriesFailed, &traceQueryReqs, &traceDepReqs,
+	); err != nil {
+		return nil, fmt.Errorf("querying values from information table: %w", err)
+	}
+	stats["promscale_ingested_samples_total"] = samples
+	stats["promscale_metrics_queries_executed_total"] = queriesExec
+	stats["promscale_metrics_queries_timedout_total"] = queriesTimeout
+	stats["promscale_metrics_queries_failed_total"] = queriesFailed
+	stats["promscale_trace_query_requests_executed_total"] = traceQueryReqs
+	stats["promscale_trace_dependency_requests_executed_total"] = traceDepReqs
```
antekresic, line 141: Seems like you are adding the `promscale_` prefix here only to chop it off in the next step. Might make things simpler to produce the `Metadata` here and remove `Stats` type altogether.

pkg/telemetry/housekeeper.go
```diff
@@ -0,0 +1,148 @@
...
+	return stats, nil
+}
+
+func CleanStalePromscalesAfterCounterReset(conn pgxconn.PgxConn) error {
```
antekresic, line 145: Seems like this is exported only to be used in e2e tests. That's always a red flag to me, especially since its not supposed to be used outside of the housekeeper.

pkg/telemetry/telemetry.go
```diff
@@ -32,12 +33,15 @@ type engineImpl struct {
 	uuid [16]byte
 	conn pgxconn.PgxConn
 
+	keeper           housekeeper
```
antekresic, line 36: Something is strange here... Engine has a reference to housekeeper and housekeeper has a reference to engine. Logically, that doesn't make sense. This needs rethinking. I feel like this is causing a bit of confusion since the responsibilities are not clearly defined.

I can help you there if you feel this comment is a bit overwhelming.

pkg/telemetry/housekeeper.go
antekresic: Looking at the test a bit, this function is used for SQL function testing. 

My suggestion is to unexport this function and test the function by manually calling it like you do in this function.  Reduces the public API and causes less confusion that way.

pkg/telemetry/telemetry.go
```diff
@@ -32,12 +33,15 @@ type engineImpl struct {
...
+	stopHousekeeping func() error
+
 	stop chan struct{}
 
 	metrics sync.Map
 }
 
-func NewEngine(conn pgxconn.PgxConn, uuid [16]byte) (*engineImpl, error) {
+func NewEngine(conn pgxconn.PgxConn, uuid [16]byte, connStr string) (*engineImpl, error) {
```
antekresic, line 44: This might need a bit of rethinking as well. Needing a connection and a connection string to create an engine feels wrong.

antekresic (changes requested):
There is some confusion regarding responsibilities of engine and housekeeper. Might be best to merge the structs together since it seems like the housekeeper duties should fall under the engine. It should be a subset of its responsibilities and not a separate entity especially since you can't be a housekeeper without being an engine.

Needs a bit more thought in the logical organization of code.

jgpruitt:
> Thank you @jgpruitt for proposing it. If i understand correctly, in the above proposal, every promscale will update
> 
> 1. its own row (that matches its own uuid)
> 2. the counter reset row in case it finds that there is some promscale instance that has gone stale
> 
> I think this can be done and is a simple and better approach provided there wasn't another stuff that's being missed here. The SQL based stats statements.
> 
> If all promscales were to be the house keepers, then they will also do updation of SQL stats (from design doc, 'what are the number of unique metrics in database', etc, this will be in the next PR), which according to earlier plan, would be done by only one promscale which was the housekeeper, but here, would be done by all promscales depending on whether the last update on counter reset row, was done an hour ago or not.
> 
> I think this can get confusing, since an update to counter reset row should happen only when cleaning stale rows.
> 
> Does this make sense?

Yes, that makes sense. Thanks for the clarification. I do think we could make the simpler approach work even considering your point. In general, I prefer not having a "special" instance of the collector. I believe it's more resilient to have all instances try to do the scheduled work and first-one-wins.

pkg/telemetry/housekeeper.go
Harkishen-Singh: Since this function will now be on `engineImpl` struct, I think `run` there will be confusing.

pkg/telemetry/housekeeper.go
Harkishen-Singh: Thats very right. But now, since these all funcs would be on `engineImpl`, this wont be required.

pkg/telemetry/housekeeper.go
Harkishen-Singh: This function is called by housekeeping() which works on the direction of `time.After()`. Should I make a function on the struct `engineImpl` to call this, which is not part of interface as its used only for tests?

pkg/telemetry/telemetry.go
Harkishen-Singh: I plan to implement all funcs of housekeeper on `engineImpl`. How does that sound?

pkg/telemetry/telemetry.go
Harkishen-Singh: Suggestions will be helpful. We cannot pass in connStr in Housekeeper() because pg_try_advisory_lock will always be accepted for same session.

pkg/telemetry/housekeeper.go
antekresic: Since the function is just running a SQL function, you can make this one private and just test the SQL function directly. That should be good enough.

Harkishen-Singh:
I agree with that @jgpruitt . The recent commit removes this async confusion, and makes all Promscales housekeepers, depending on if the last housekeeping was done beyond an hour or not.

**NOTE FOR REVIEWERS:** Now, all promscales act as housekeepers. If the last housekeeping was done beyond an hour, then do the checking of stale promscales, counter reset increment, and evaluation of SQL-based telemetry stats (in the coming PRs). The state of last housekeeping done, is saved in the last_update attribute of counter reset row, than saving somewhere else, meaning counter reset row's last_updated is no longer the actual meaning.

pkg/telemetry/telemetry.go
```diff
@@ -142,7 +147,7 @@ func (t *engineImpl) Stop() {
 func (t *engineImpl) Start() {
 	t.stop = make(chan struct{})
 	go func() {
-		collect := time.NewTicker(time.Minute * 55) // Collect telemetry info every 55 minutes since housekeeper fills the _timescaledb_catalog.metadata table every 1 hour.
+		collect := time.NewTicker(time.Hour) // Collect telemetry info every hour.
```
bitwalker, line 150: Might be worth adding some jitter to the ticker so that you don't have a bunch of instances that start together racing to perform the same work.

pkg/telemetry/telemetry.go
```diff
@@ -268,6 +275,60 @@ func (t *engineImpl) syncInfoTable(stats map[string]float64) error {
 	return nil
 }
 
+func (t *engineImpl) housekeeping() {
+	var lastRunBeyondHour bool
+	err := t.conn.QueryRow(context.Background(), "SELECT _ps_catalog.promscale_telemetry_housekeeping()").Scan(&lastRunBeyondHour)
+	if err != nil && !strings.Contains(err.Error(), "could not obtain lock on relation") { // Ignore lock errors as we aren't supposed to do anything, since some other Promscale is doing housekeeping at this time.
```
bitwalker, line 281: Is this error text the same across major versions? If not, it might be preferable to catch the exception in the `promscale_telemetry_housekeeping` function and return an additional boolean that reflects whether or not the attempt was successful

bitwalker (approved):
Looks good! Just a couple comments with suggestions, but nothing blocking

pkg/pgmodel/migrate.go
```diff
@@ -52,6 +52,7 @@ var (
 			"tracing-functions.sql",
 			"tracing-views.sql",
 			"remote-commands.sql",
+			"telemetry.sql",
```
jgpruitt, line 55: `remote-commands.sql` needs to be next to last. Would you move it to just above `apply_permissions.sql` and add a comment to that effect?

pkg/telemetry/telemetry.go
jgpruitt: +1 this

pkg/telemetry/telemetry.go
jgpruitt: It might be cleaner to use the error code `55P03` rather than the error message.
`ERROR:  55P03: could not obtain lock on relation "foo"`

https://www.postgresql.org/docs/14/errcodes-appendix.html

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
...
+                    counter_reset.is_counter_reset_row = TRUE AND (stale.is_counter_reset_row = FALSE AND current_timestamp - stale.last_updated > interval '2 HOUR') -- consider only those rows as stale who were last updated beyond 2 hours.
+                GROUP BY
+                    counter_reset.promscale_ingested_samples_total,
+                    counter_reset.promscale_metrics_queries_executed_total,
+                    counter_reset.promscale_metrics_queries_timedout_total,
+                    counter_reset.promscale_metrics_queries_failed_total,
+                    counter_reset.promscale_trace_query_requests_executed_total,
+                    counter_reset.promscale_trace_dependency_requests_executed_total
+            ) AS subquery
+        WHERE is_counter_reset_row = TRUE;
```
jgpruitt, line 86: simplify this update statement

jgpruitt:
```
UPDATE SCHEMA_PS_CATALOG.promscale_instance_information SET
    promscale_ingested_samples_total                    = x.promscale_ingested_samples_total,
    promscale_metrics_queries_executed_total            = x.promscale_metrics_queries_executed_total,
    promscale_metrics_queries_timedout_total            = x.promscale_metrics_queries_timedout_total,
    promscale_metrics_queries_failed_total              = x.promscale_metrics_queries_failed_total,
    promscale_trace_query_requests_executed_total       = x.promscale_trace_query_requests_executed_total,
    promscale_trace_dependency_requests_executed_total  = x.promscale_trace_dependency_requests_executed_total,
    last_updated = current_timestamp
FROM
(
    SELECT
        sum(promscale_ingested_samples_total)  			        as promscale_ingested_samples_total,
        sum(promscale_metrics_queries_executed_total)  		    as promscale_metrics_queries_executed_total,
        sum(promscale_metrics_queries_timedout_total)  		    as promscale_metrics_queries_timedout_total,
        sum(promscale_metrics_queries_failed_total)  		    as promscale_metrics_queries_failed_total,
        sum(promscale_trace_query_requests_executed_total)      as promscale_trace_query_requests_executed_total,
        sum(promscale_trace_dependency_requests_executed_total) as promscale_trace_dependency_requests_executed_total
    FROM
        SCHEMA_PS_CATALOG.promscale_instance_information
    WHERE 
        is_counter_reset_row = TRUE
        OR (is_counter_reset_row = FALSE AND current_timestamp - last_updated > interval '1 HOUR') -- consider only those rows as stale who were last updated beyond an hour.
) x
WHERE is_counter_reset_row = TRUE;
```

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
...
+-- making sure that at one time, only one housekeeping is being done.
+-- 
+-- It returns TRUE if last run was beyond an hour, otherwise FALSE.
+CREATE OR REPLACE FUNCTION SCHEMA_PS_CATALOG.promscale_telemetry_housekeeping()
+RETURNS BOOLEAN AS
+$$
+    DECLARE
+        is_last_run_beyond_hour BOOLEAN := FALSE;
+    BEGIN
+        LOCK TABLE SCHEMA_PS_CATALOG.promscale_instance_information IN ACCESS EXCLUSIVE MODE NOWAIT; -- Do not wait for the lock as some promscale is already cleaning up the stuff.
```
jgpruitt, line 15: Do we need to lock the whole table or just the counter reset row?

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
...
+        -- The last_update of counter_reset row corresponds to the last run of housekeeping, and not its actual meaning, which was
+        -- when the counter_reset row was updated. This is done so that there can be place to store the last executing time,
+        -- allowing all Promscales to be housekeepers, instead of just one, thereby decreasing the async complexity.
+        -- 
+        -- Hence, update the last_update of counter reset row since now is its recent most run.
+        UPDATE SCHEMA_PS_CATALOG.promscale_instance_information SET last_updated = current_timestamp WHERE is_counter_reset_row = TRUE;
+
+        -- Check if any stale row is present.
+        -- Note: Stale interval is twice the scan interval (i.e., 1 Hour). This is to avoid cleaning up any Promscale instance that got delayed
+        -- in registering itself alive, as all Promscales are updating themselves every hour, and also doing a stale check the same time.
```
jgpruitt, line 39: I don't quite understand this. Why does it need to be twice the scan interval? Doesn't this mean that the metrics will be 2 hours old?

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
...
+
+        -- Last run was beyond an hour.
+        is_last_run_beyond_hour := TRUE;
+
+        -- The last_update of counter_reset row corresponds to the last run of housekeeping, and not its actual meaning, which was
+        -- when the counter_reset row was updated. This is done so that there can be place to store the last executing time,
+        -- allowing all Promscales to be housekeepers, instead of just one, thereby decreasing the async complexity.
+        -- 
+        -- Hence, update the last_update of counter reset row since now is its recent most run.
+        UPDATE SCHEMA_PS_CATALOG.promscale_instance_information SET last_updated = current_timestamp WHERE is_counter_reset_row = TRUE;
```
jgpruitt, line 35: If there ARE stale rows present, we end up running 2 update statements on the counter reset row. I suggest rearranging this a bit so that we only ever issue one update against the row in this function.

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
...
+            WHERE
+                is_counter_reset_row = TRUE
+            AND
+                current_timestamp - last_updated < INTERVAL '1 HOUR') THEN
+            -- Some Promscale did the housekeeping work within the expected interval. Hence, nothing to do, so exit.
+            RETURN is_last_run_beyond_hour;
+        END IF;
+
+        -- Last run was beyond an hour.
+        is_last_run_beyond_hour := TRUE;
```
jgpruitt, line 28:
```
IF (SELECT count(*) > 0 INTO STRICT is_last_run_beyond_hour FROM .....
```
This way the boolean variable gets set directly by the count.

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,94 @@
+-- promscale_telemetry_housekeeping() does telemetry housekeeping stuff, which includes
+-- searching the table once an hour, looking for stale promscales, if found,
+-- adding their values into the counter_reset row, and then cleaning up the stale
+-- promscale instances.
+-- It is concurrency safe, since it takes lock on the promscale_instance_information table,
+-- making sure that at one time, only one housekeeping is being done.
+-- 
+-- It returns TRUE if last run was beyond an hour, otherwise FALSE.
+CREATE OR REPLACE FUNCTION SCHEMA_PS_CATALOG.promscale_telemetry_housekeeping()
+RETURNS BOOLEAN AS
```
jgpruitt, line 10: Instead of hardcoding the intervals (1 hour, 2 hours), I would make them parameters to this function. The parameters can have default values of 1 and 2 hours. Having them as parameters will make it easier to test and easier to make it configurable should we choose to.

pkg/telemetry/telemetry.go
jgpruitt: https://github.com/jackc/pgconn/blob/v1.10.1/errors.go#L33

jgpruitt (changes requested):

pkg/migrations/sql/idempotent/telemetry.sql
cevian:
```
WITH deleted_rows AS (
    DELETE FROM SCHEMA_PS_CATALOG.promscale_instance_information 
    WHERE is_counter_reset_row = FALSE AND current_timestamp - last_updated > interval '2 HOUR'
    RETURNING *
)
UPDATE SCHEMA_PS_CATALOG.promscale_instance_information SET
    promscale_ingested_samples_total                    = promscale_ingested_samples_total + x.del_promscale_ingested_samples_total,
    promscale_metrics_queries_executed_total            = promscale_metrics_queries_executed_total + x.del_promscale_metrics_queries_executed_total,
    promscale_metrics_queries_timedout_total            = promscale_metrics_queries_timedout_total + x.del_promscale_metrics_queries_timedout_total,
    promscale_metrics_queries_failed_total              = promscale_metrics_queries_failed_total + x.del_promscale_metrics_queries_failed_total,
    promscale_trace_query_requests_executed_total       = promscale_trace_query_requests_executed_total + x.del_promscale_trace_query_requests_executed_total,
    promscale_trace_dependency_requests_executed_total  = promscale_trace_dependency_requests_executed_total + x.del_promscale_trace_dependency_requests_executed_total,
    last_updated = current_timestamp
FROM
(
    SELECT
        sum(promscale_ingested_samples_total)  			        as del_promscale_ingested_samples_total,
        sum(promscale_metrics_queries_executed_total)  		    as del_promscale_metrics_queries_executed_total,
        sum(promscale_metrics_queries_timedout_total)  		    as del_promscale_metrics_queries_timedout_total,
        sum(promscale_metrics_queries_failed_total)  		    as del_promscale_metrics_queries_failed_total,
        sum(promscale_trace_query_requests_executed_total)      as del_promscale_trace_query_requests_executed_total,
        sum(promscale_trace_dependency_requests_executed_total) as del_promscale_trace_dependency_requests_executed_total
    FROM
        deleted_rows
) x
WHERE is_counter_reset_row = TRUE;
```

cevian: with the above you can get rid of line 89 as well as the check on line 40 and the extra update on line 35

pkg/telemetry/telemetry.go
```diff
@@ -268,6 +275,60 @@ func (t *engineImpl) syncInfoTable(stats map[string]float64) error {
 	return nil
 }
 
+func (t *engineImpl) housekeeping() {
+	var lastRunBeyondHour bool
```
cevian, line 279: From the point of view of the Go code it doesn't care whether the run was beyond an hour. It cares that the housekeeper performed the update. So this should be called 
```suggestion
	var housekeeperPerformedUpdate bool
```

Then the logic is if houskeeperPerformedUpdate=>getInstanceInformationStats + writeThem

This decouples the internal time-based logic in SQL from the go code that just cares if there was an update

pkg/telemetry/telemetry.go
cevian: Can we catch this error in the SQL function + return false?

cevian (changes requested):
This is good work and a nice, simple implementation. Some changes are needed, though

pkg/telemetry/telemetry.go
Harkishen-Singh: Go tickers are not accurate with respect to time. They shift by few milliseconds in some cases, so I think that should be fine for us.

pkg/telemetry/telemetry.go
Harkishen-Singh: I wrote this after the lock statement, but it does not seem to work.

```postgresql
EXCEPTION
            WHEN SQLSTATE '55P03' THEN -- sqlstate when lock on table is not available.
                RETURN FALSE;
```

pkg/migrations/sql/idempotent/telemetry.sql
```diff
@@ -0,0 +1,59 @@
...
+$$
+    BEGIN
+        -- Check if any stale row is present.
+        IF (SELECT count(*) = 0 FROM SCHEMA_PS_CATALOG.promscale_instance_information
+                WHERE
+                      is_counter_reset_row = FALSE
+                  AND
+                      current_timestamp - last_updated > INTERVAL '1 HOUR') THEN
+            RETURN;
+        END IF;
```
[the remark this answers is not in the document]
Harkishen-Singh: yeah, its just a quick check before proceeding

pkg/telemetry/telemetry.go
```diff
@@ -47,10 +51,21 @@ func NewEngine(conn pgxconn.PgxConn, uuid [16]byte) (*engineImpl, error) {
...
+
 	t := &engineImpl{
 		conn: conn,
 		uuid: uuid,
 	}
+	t.keeper = housekeeper{
+		engineCopy: t,
+		lock:       advisoryLock,
+	}
+
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, I think you are not on the latest commit. The recent commit does not have this stuff.

pkg/telemetry/telemetry.go
```diff
@@ -47,10 +51,21 @@ func NewEngine(conn pgxconn.PgxConn, uuid [16]byte) (*engineImpl, error) {
 		log.Warn("msg", "Promscale collects anonymous usage telemetry data to help the Promscale team better understand and assist users. "+
 			"This can be disabled via the process described at https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry/#disabling-telemetry")
 	}
+
```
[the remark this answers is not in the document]
Harkishen-Singh: No, because even if the telemetry collection is disabled, we do the collection and store it in the table as that will help in debugging.

pkg/migrations/sql/idempotent/telemetry.sql
Harkishen-Singh: Why do we need to know the number of deleted rows?

If it's fine, I will love to keep this simple, which is the most recent pushed version. (which John has suggested)

pkg/migrations/sql/idempotent/telemetry.sql
Harkishen-Singh: I think we should lock the whole table, otherwise, while some promscale is doing the counter reset updation stuff, some other promscale will update its values, as updation of respective rows happen before the go function that locks is called, so it seems a race.

pkg/migrations/sql/idempotent/telemetry.sql
Harkishen-Singh: Sorry, the last_update of counter_reset row is not the last_update of that row, rather the last time when the table was touched for housekeeping.

> NOTE FOR REVIEWERS: Now, all promscales act as housekeepers. If the last housekeeping was done beyond an hour, then do the checking of stale promscales, counter reset increment, and evaluation of SQL-based telemetry stats (in the coming PRs). The state of last housekeeping done, is saved in the last_update attribute of counter reset row, than saving somewhere else, meaning counter reset row's last_updated is no longer the actual meaning.

pkg/telemetry/telemetry.go
Harkishen-Singh: I agree, but just to clear the logic, we need to write getInstanceInformationStats, not because housekeeper performed the update, but because no matter if housekeeper performed update or not, its past 1 hour, so we must write the new information stats, which is why I wrote it. This also means that the SQL function should return true, as soon as its beyond the 1 hour duration, and not only when the counter reset row was updated.

antekresic (approved):
Minor nit for the Go side of things. 

Other than that, it should be OK once the rest of the review comments are addressed.

jgpruitt (approved):

pkg/tests/end_to_end_tests/telemetry_test.go
```diff
@@ -116,12 +116,92 @@ func TestOnlyOneHousekeeper(t *testing.T) {
...
+
+		require.NoError(t, engine.Sync())
+
+		err = conn.QueryRow(context.Background(), "select value from _timescaledb_catalog.metadata where key = 'promscale_ingested_samples_total'").Scan(&val)
+		require.NoError(t, err)
+		require.Equal(t, "100", val)
+	})
+}
+
+func TestCleanStalePromscales(t *testing.T) {
```
cevian, line 160: can we add one more new test that runs Sync for the first time when there is nothing to delete.

pkg/tests/end_to_end_tests/telemetry_test.go
```diff
@@ -116,12 +116,92 @@ func TestOnlyOneHousekeeper(t *testing.T) {
...
+		require.NoError(t, engine.Sync())
+
+		err = conn.QueryRow(context.Background(), "SELECT count(*) > 0 FROM _ps_catalog.promscale_instance_information WHERE uuid = '10000000-0000-0000-0000-000000000000'").Scan(&exists)
+		require.NoError(t, err)
+		require.False(t, exists)
+
+		// Check counter reset row's value.
+		err = conn.QueryRow(context.Background(), "SELECT promscale_ingested_samples_total FROM _ps_catalog.promscale_instance_information WHERE is_counter_reset_row = TRUE").Scan(&cnt)
+		require.NoError(t, err)
+		require.Equal(t, 100, int(cnt))
```
cevian, line 205: let's run `require.NoError(t, engine.Sync())` here again to make sure the logic is safe when the timeout hasn't. been reached yet.

pkg/tests/end_to_end_tests/telemetry_test.go
```diff
@@ -116,12 +116,92 @@ func TestOnlyOneHousekeeper(t *testing.T) {
 
 		conn := pgxconn.NewPgxConn(db)
 
+		// Should error due to unique constraint for counter_reset_row, as there already exists a counter reset row.
 		_, err := conn.Exec(context.Background(), `INSERT INTO _ps_catalog.promscale_instance_information (uuid, last_updated, is_counter_reset_row) VALUES ('00000000-0000-0000-0000-000000000000', current_timestamp, TRUE)`)
+		require.Error(t, err)
+		require.True(t, strings.Contains(err.Error(), `violates unique constraint "promscale_instance_information_pkey"`))
+	})
+}
+
```
cevian, line 125: I'd like a test that runs with telemetry = off.

cevian (approved):
Some important changes requested but if you agree with my suggestions there is no need for me to re-review before merge (thus approving).

I'd like some more tests too

pkg/tests/end_to_end_tests/telemetry_test.go
Harkishen-Singh: Sure, but doing Sync() will add insert a row of the existing promscale, since it wasn't there earlier.

PR #960, opened 2021-12-08
Run full test suite on PG14

JamesGuthrie:
We run the full test suite against the most recent postgres version, and
a subset against older versions. This change promotes our full test
suite to test against PG14.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -89,13 +89,15 @@ jobs:
     strategy:
       matrix:
         test-setups:
-        - {name: "Multinode",                shortname: "multinode",    ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: true,  pg: 13}
-        - {name: "W/O Promscale Extension",  shortname: "wo-prom-ext",  ext: false, tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 13}
-        - {name: "Plain Postgres",           shortname: "plain-pg",     ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 13}
         - {name: "Timescaledb 1.x (pg 12)",  shortname: "tsdb1x-pg-12", ext: true,  tsdb: true,  tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "Plain Postgres (12)",      shortname: "plain-pg-12",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 12}
-        - {name: "TimescaleDB-OSS",          shortname: "tsdb-oss",     ext: false, tsdb: true,  tsdb2: true,  tsdboss: true,  multi: false, pg: 13}
-        - {name: "pg 14",                    shortname: "pg14",         ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 14}
+        - {name: "Plain Postgres (13)",      shortname: "plain-pg-13",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 13}
+        - {name: "Singlenode",               shortname: "singlenode",   ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 14}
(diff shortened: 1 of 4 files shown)


jgpruitt (approved):

.github/workflows/go.yml
```diff
@@ -89,13 +89,14 @@ jobs:
...
-        - {name: "W/O Promscale Extension",  shortname: "wo-prom-ext",  ext: false, tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 13}
-        - {name: "Plain Postgres",           shortname: "plain-pg",     ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 13}
         - {name: "Timescaledb 1.x (pg 12)",  shortname: "tsdb1x-pg-12", ext: true,  tsdb: true,  tsdb2: false, tsdboss: false, multi: false, pg: 12}
         - {name: "Plain Postgres (12)",      shortname: "plain-pg-12",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 12}
-        - {name: "TimescaleDB-OSS",          shortname: "tsdb-oss",     ext: false, tsdb: true,  tsdb2: true,  tsdboss: true,  multi: false, pg: 13}
-        - {name: "pg 14",                    shortname: "pg14",         ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: false, pg: 14}
+        - {name: "Plain Postgres (13)",      shortname: "plain-pg-13",  ext: false, tsdb: false, tsdb2: false, tsdboss: false, multi: false, pg: 13}
+        - {name: "Multinode",                shortname: "multinode",    ext: true,  tsdb: true,  tsdb2: true,  tsdboss: false, multi: true,  pg: 14}
```
cevian, line 95: I think we are missing the canonical single node for pg 14 which used to be called "pg 14". and is `ext: true, tsdb: true, tsdb2: true, tsdboss: false, multi: false, pg: 14`

cevian (changes requested):
If we are making pg14 canonical I think we need to rollback a few changes that limited scope of pg14 tests:
- https://github.com/timescale/promscale/pull/871/files#diff-f8ac25eb8f2eca96d9fb4530566d3913d7c1ae6965a14d7e0a8df0daf7f1a1edR180
- https://github.com/timescale/promscale/pull/871/files#diff-48513ed7faf5b2c9cf4e5c4226fe3366eeb5d01ff0b865de28fbf64482340815R464
- https://github.com/timescale/promscale/pull/871/files#diff-f73ce61bb31645bdf6a5483e5242beb9a33e86d3f0c5c2e635ef6fd248102e26R42

.github/workflows/go.yml
JamesGuthrie: I have added this back in.

JamesGuthrie:
> If we are making pg14 canonical I think we need to rollback a few changes that limited scope of pg14 tests:

Thanks for catching this Mat!

pkg/internal/testhelpers/postgres_container.go
```diff
@@ -175,11 +175,6 @@ func (e ExtensionState) GetDockerImageName() (string, error) {
 		image = "timescale/timescaledb:latest-" + PGTag
 	case Timescale2AndPromscale, MultinodeAndPromscale:
 		image = LatestDBHAPromscaleImageBase + ":" + PGTag + "-latest"
-
-		//TODO: remove split once pg14 ha image published
```
Harkishen-Singh, line 179: There is another todo like this to be removed in `end_to_end_tests/continuous_aggs_test.go`. Should that be removed as well?

JamesGuthrie: Do you have a line reference? I could not find it.

JamesGuthrie: https://github.com/timescale/promscale/pull/960/files#diff-48513ed7faf5b2c9cf4e5c4226fe3366eeb5d01ff0b865de28fbf64482340815L464-L467

Harkishen-Singh (approved):

cevian (approved):

PR #965, opened 2021-12-09
Add test exposing broken _prom_catalog.metric_view() function

JamesGuthrie:
This is a test exposing broken functionality which was fixed in #958

### DIFF
--- a/pkg/tests/end_to_end_tests/functions_test.go
+++ b/pkg/tests/end_to_end_tests/functions_test.go
@@ -21,6 +21,33 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+func TestMetricView(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	withDB(t, *testDatabase, func(dbOwner *pgxpool.Pool, t testing.TB) {
+		db := testhelpers.PgxPoolWithRole(t, *testDatabase, "prom_reader")
+		defer db.Close()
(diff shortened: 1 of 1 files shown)


pkg/tests/end_to_end_tests/functions_test.go
```diff
@@ -21,6 +21,33 @@ import (
...
+			t.Fatal(err)
+		}
+
+		var res string
+		err = dbWriter.QueryRow(context.Background(), "SELECT metric_name FROM _prom_catalog.metric_view()").Scan(&res)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if res != "test_metric_name" {
+			t.Fatal("Fail")
```
antekresic, line 46: Could we add a bit more context in the test failure here? If this failed, you would mostly see that it failed without really saying why did it fail? Maybe include that metric view wasn't created with get_or_create_metric_table function call or something like that.

antekresic (approved):
Looks good overall, might be good to include a bit more context for the failure so its easier to understand why the test failed from the output itself.

pkg/tests/end_to_end_tests/functions_test.go
```diff
@@ -21,6 +21,33 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
+func TestMetricView(t *testing.T) {
```
VineethReddy02, line 24: nit: I would add `func` as a suffix in the test func name to indicate it's a test for the SQL func.

VineethReddy02 (approved):

PR #979, opened 2021-12-13
Remove the event_name_check constraint

jgpruitt:
The event_name_check prevents the event name from being an
empty string, however OTEL does not strictly disallow this.
Remove the constraint.


pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
```diff
@@ -0,0 +1,114 @@
...
+
+        IF _timescaledb_major_version >= 2 THEN
+            _is_timescaledb_oss = (current_setting('timescaledb.license') = 'apache');
+        ELSE
+            _is_timescaledb_oss = (SELECT edition = 'apache' FROM timescaledb_information.license);
+        END IF;
+    END IF;
+
+    IF _is_timescaledb_installed AND (NOT _is_timescaledb_oss) AND _is_compression_available THEN
+        -- is compression enabled for the table?
```
cevian, line 74: you need to decompress all chunks here before turning compression off, otherwise disabling compression will fail.

cevian: This probably means we also should think about how to alter our upgrade tests to catch such bugs

cevian (changes requested):

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
jgpruitt: This makes me think that we should NOT have an upgrade script for this particular change. If the event table is gigantic, applying this change could be a very long outage. What do you think?

cevian: The only other way is dropping the table

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
```diff
@@ -0,0 +1,114 @@
...
+        AND c.contype = 'c'
+    )
+    THEN
+        RETURN;
+    END IF;
+
+    /*
+        These functions do not exist until the
+        idempotent scripts are executed, so we have
+        to deal with it "manually"
```
JamesGuthrie, line 38: Is this true? If we are migrating then the idempotent scripts were run on installation right?

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
JamesGuthrie: I wanted to suggest that it might be "acceptable" for us to force this onto the users who are using tracing in beta: to tell them to manually run the necessary steps. Unfortunately, because all users have the schema installed, this also impacts users who are on `0.7.0` and do not have `-enable-feature=tracing` set.

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
jgpruitt: i think you may be right. They may not be available in the preinstall scripts, but as you say I think they should be here for migrations.

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
JamesGuthrie: This has been marked as resolved, but I'm not sure why. @jgpruitt can you add some context?

jgpruitt: I unresolved this because I added the code to decompress any compressed chunks before disabling compression and dropping the constraints. I also recompress those chunks after dropping the constraint.

Do we want to discuss more?

JamesGuthrie: Okay, I think that hits the right middle ground.

Just to be explicit: with this change, we make the constraint dropping fully automatic, with the caveat that any users with an existing promscale tracing installation will have to assess whether uncompressing and recompressing the `event` hypertable will work for them. As you already mentioned, it could cause a long downtime, or even use more disk than is available, and come to a grinding halt.

Do we want to provide users with instructions for how to determine if this will go smoothly? Or do we assume that users who were using the beta feature would know how to do that?

JamesGuthrie (approved):
I'm learning a lot of interesting SQL bits and pieces reviewing this 🎉

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
jgpruitt: It wouldn't hurt to leave something in the release notes. I don't know that there is anything to tell them to DO, just that the upgrade may take a while if the event table is large.

pkg/migrations/sql/versions/dev/0.7.2-dev/3-event_name.sql
```diff
@@ -0,0 +1,140 @@
...
+    _is_timescaledb_oss boolean = true;
+    _timescaledb_version_text text;
+    _timescaledb_major_version int;
+    _timescaledb_minor_version int;
+    _is_compression_available boolean = false;
+    _compression_enabled boolean = false;
+    _chunk_name text;
+BEGIN
+
+    -- determine whether the constraint even exists
```
VineethReddy02, line 18: event*

antekresic (approved):

pkg/migrations/sql/versions/dev/0.7.2-dev/2-event_name.sql
JamesGuthrie: Nope. Current consensus is leaning towards starting a CHANGELOG.md file in the repo, but it's not decided, and nothing has been done.

jgpruitt: > you need to decompress all chunks here before turning compression off, otherwise disabling compression will fail.

@cevian did I satisfy this change request?

cevian (approved):
Looks good although we should discuss how to improve tests so that such errors are caught (need to modify `getUpgradeDbInfo to write and compress traces before upgrade`)

PR #983, opened 2021-12-14
Fix HA e2e test that contained a data race

### LINKED ISSUE #982: Data race in TestConcurrentUpdateLease e2e test
Test is accessing a map from multiple goroutines and fails from time to time. We should update it so it doesn't fail CI builds.

antekresic:
Fixes #982


bitwalker (approved):
LGTM

VineethReddy02 (approved):

PR #992, opened 2021-12-17
Adds support for PromQL based telemetry stats.

Harkishen-Singh:

Telemetry collection also involves calculating quantiles that need the help
of PromQL expression. This commit implements the same. As of now, we
calculate p50, p90, p95, p99 of response durations of /query.* endpoints
and the /fetch_traces API (via GRPC).

### DIFF
--- a/pkg/telemetry/promql.go
+++ b/pkg/telemetry/promql.go
@@ -1,0 +1,86 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package telemetry
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/timescale/promscale/pkg/promql"
+)
+
+type promqlTelemetry struct {
+	name        string
+	query       string
+	parsedQuery promql.Query
+}
+
+func (t *promqlTelemetry) execute(engine *promql.Engine, queryable promql.Queryable) (float64, error) {
+	if t.parsedQuery == nil {
+		qry, err := engine.NewInstantQuery(queryable, t.query, time.Now())
+		if err != nil {
+			return 0, fmt.Errorf("creating instant query: %w", err)
+		}
+		t.parsedQuery = qry
+	}
+
+	result := t.parsedQuery.Exec(context.Background())
+	if result.Err != nil {
+		return 0, fmt.Errorf("error executing promql query '%s': %w", t.query, result.Err)
+	}
+
+	value, err := getValue(result)
+	if err != nil {
+		return 0, fmt.Errorf("error extracting value from promql query '%s' response: %w", t.query, err)
+	}
+	return value, nil
+}
+
+func getValue(result *promql.Result) (float64, error) {
+	vec, err := result.Vector()
+	if err != nil {
+		return 0, fmt.Errorf("error reading vector response: %w", err)
+	}
+	numSamples := len([]promql.Sample(vec))
+	if numSamples > 1 {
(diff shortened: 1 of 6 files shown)


pkg/telemetry/promql.go
```diff
@@ -0,0 +1,41 @@
...
+
+type promqlTelemetry struct {
+	name        string
+	query       string
+	parsedQuery promql.Query
+}
+
+var promqlStats = []promqlTelemetry{
+	{
+		name:  "promql_query_exeuction_time_p50",
```
thiagogcm, line 17: typo: execution

pkg/telemetry/telemetry.go
```diff
@@ -327,6 +347,59 @@ func (t *engineImpl) getInstanceInformationStats() (Metadata, error) {
...
+		if result.Err != nil {
+			log.Debug("msg", "error executing promql query", "query", promqlStats[i].query, "err", result.Err.Error())
+			continue
+		}
+
+		value, err := getValue(result)
+		if err != nil {
+			log.Debug("msg", "error extracting value from response", "query", promqlStats[i].query, "err", err.Error())
+			continue
+		}
```
cevian, line 379: All this logic (along with the getValue function) is better as a method on `promqlTelemetry`. The engine should be able to call promqlStats[I].Execute() and get the value back? Why -- because the engine should know nothing about internal structures like parsedQuery

pkg/telemetry/promql.go
cevian: Good catch. Applies to all of these

pkg/telemetry/telemetry.go
```diff
@@ -66,6 +80,10 @@ func NewEngine(conn pgxconn.PgxConn, uuid [16]byte) (*engineImpl, error) {
 	return t, nil
 }
 
+func durationMilliseconds(d time.Duration) int64 {
```
cevian, line 83: is this not already defined somewhere?

cevian (changes requested):

pkg/telemetry/telemetry.go
antekresic: https://pkg.go.dev/time#Duration.Milliseconds

antekresic (commented):
Lets address Mats comments then we can give it another round of reviews.

pkg/telemetry/promql.go
Harkishen-Singh: Thank you or catching this.

pkg/telemetry/telemetry.go
Harkishen-Singh: Yes.

> https://pkg.go.dev/time#Duration.Milliseconds

This does not return the converted millisecond of `d`, rather the millisecond'th of `d`

antekresic (approved):

pkg/telemetry/telemetry.go
```diff
@@ -327,6 +343,26 @@ func (t *engineImpl) getInstanceInformationStats() (Metadata, error) {
...
+	}
+	stats := make(Metadata)
+	for _, query := range promqlStats {
+		value, err := query.execute(t.promqlEngine, t.promqlQueryable)
+		if err != nil {
+			log.Debug("msg", "error executing promql expression", "err", err.Error())
+			continue
+		}
+		stats[query.name] = fmt.Sprintf("%.4f", value)
+	}
```
cevian, line 362: add timing as `promql_telemetry_evaluation_duration_seconds`

cevian (approved):

PR #995, opened 2021-12-18
Remove cron instructions

ramonguiu:
Setting up cron to run execute maintenance is no longer needed with TimescaleDB 2. Promscale autmatically schedules jobs within the database to run execute_maintenance.

Solves #910


ramonguiu:
Actually, we still have a lot of users using TimescaleDB 1.7. So we shouldn't remove instructions but add a note instead. I'll close this PR and start a new one.

PR #996, opened 2021-12-18
Cron job only needed for TSDB 1.X

ramonguiu:
Add clarification that the cron job to run execute_maintenance is only needed with TimescaleDB 1.X


antekresic (approved):

JamesGuthrie (approved):
LGTM, but this page will disappear soonish, and these instructions don't seem to be present in the changes here: https://github.com/timescale/docs/pull/478

PR #1002, opened 2021-12-21
Prevent row deadlock

niksajakovljevic:
If Prometheus retries sending samples  Promscale can attempt to insert duplicate data.
This can result in a row deadlock. Deadlock is detectable by PG when running on a singlenode.
However in a multinode setup that does not work due to distributed nature.
We make sure to always sort series before inserting to avoid the deadlock.
We don't sort samples in the series because they should already be sorted by Prometheus.

Closes https://github.com/timescale/promscale/issues/975

### DIFF
--- a/pkg/pgmodel/ingestor/metric_batcher.go
+++ b/pkg/pgmodel/ingestor/metric_batcher.go
@@ -132,28 +132,10 @@ func runMetricBatcher(conn pgxconn.PgxConn,
 	labelArrayOID uint32) {
 
 	var (
-		tableName            string
-		firstReq             *insertDataRequest
-		firstReqSet          = false
-		exemplarsInitialized = false
+		tableName   string
+		firstReq    *insertDataRequest
+		firstReqSet = false
 	)
-
-	addReq := func(req *insertDataRequest, buf *pendingBuffer) {
-		if !exemplarsInitialized && containsExemplars(req.data) {
-			if err := initializeExemplars(conn, metricName); err != nil {
-				log.Error("msg", err)
-				req.reportResult(err)
-				return
-			}
-			exemplarsInitialized = true
-		}
-		buf.addReq(req)
-	}
-	//This channel in synchronous (no buffering). This provides backpressure
-	//to the batcher to keep batching until the copier is ready to read.
-	copySender := make(chan copyRequest)
-	defer close(copySender)
-	readRequest := readRequest{copySender: copySender}
 
 	for firstReq = range input {
 		var err error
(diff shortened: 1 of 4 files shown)


pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -208,6 +213,11 @@ func runMetricBatcher(conn pgxconn.PgxConn,
 		}
 
 		numSeries := pending.batch.CountSeries()
+
+		// Since seriesId order is not guaranteed we need to sort it to avoid row deadlock when duplicates are sent (eg. Prometheus retry)
+		// Samples inside series should be sorted by Prometheus
+		sort.Sort(&pending.batch)
```
cevian, line 219: We can't sort here since seriesID is not guaranteed to be set yet (it will be if there is a cache hit but not a cache miss). It's only guaranteed to be set after the WriteSeries call in persistBatch (series.go). I think we should sort there.

pkg/pgmodel/model/batch.go
```diff
@@ -68,3 +68,15 @@ func (t *Batch) Visitor() *batchVisitor {
...
+func (t *Batch) Len() int {
+	return t.CountSeries()
+}
+
+func (t *Batch) Swap(i, j int) {
+	t.data[i], t.data[j] = t.data[j], t.data[i]
+}
+
+func (t *Batch) Less(i, j int) bool {
+	return t.data[i].Series().seriesID < t.data[j].Series().seriesID
```
cevian, line 81: We should make sure the seriesID is set. GetSeriesID() is safer
```suggestion
	s1, _, err := t.data[i].Series().GetSeriesID() 
	if err != nil {
	  panic()
	}
	s2 _, err :=  t.data[j].Series().GetSeriesID()
	if err != nil {
	  panic()
	}
	return s1 < s2
```

cevian (changes requested):
Good catch as to the problem. Unfortunately, I think the solution has some easy to fix issues

antekresic (approved):

pkg/pgmodel/model/batch.go
```diff
@@ -68,3 +70,23 @@ func (t *Batch) Visitor() *batchVisitor {
...
+}
+
+func (t *Batch) Swap(i, j int) {
+	t.data[i], t.data[j] = t.data[j], t.data[i]
+}
+
+func (t *Batch) Less(i, j int) bool {
+	s1, _, err := t.data[i].Series().GetSeriesID()
+	if err != nil {
+		log.Warn("seriesID", "not set but being sorted on")
```
cevian, line 85: These are not recoverable errors in the core logic. These should be panic not logging.

cevian (approved):

pkg/pgmodel/model/batch.go
niksajakovljevic: I was on the fence for this one. Strictly speaking the application can continue to operate even if some of seriesIDs is not being set (meaning it is zero). So I am not entirely sure if we want to crash the app. I don't feel strong either way so let me know.

niksajakovljevic: Another take is that if you would try to sort batch elsewhere you might not want sorting to crash the app.

PR #1008, opened 2021-12-24
Fix passing of async flag

niksajakovljevic:
There were two problems with async flag. First it was present
in two Config structs and parsed in the struct that is not being
used. Second was that async flag was not passed when creating
sub-config structure. This resulted in async flag not being
applied.

Also removing some unused config.

Hopefully in future we will increase test coverage for flags and configs so we catch this easily. 
It seems we would need a decent amount of refactoring to support these tests so I'll leave it for another PR.


antekresic (approved):

VineethReddy02 (approved):
LGTM!

PR #1011, opened 2022-01-03
Add `db-writer-connection-concurrency` flag back

niksajakovljevic:
This is a deprecated flag but we still need it for compatibility reasons.

### DIFF
--- a/pkg/runner/flags.go
+++ b/pkg/runner/flags.go
@@ -61,6 +61,7 @@ var flagAliases = map[string][]string{
 	"db.statements-cache":               {"db-statements-cache"},
 	"db.uri":                            {"db-uri"},
 	"db.user":                           {"db-user"},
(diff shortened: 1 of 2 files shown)


antekresic (approved):

VineethReddy02 (changes requested):
Shouldn't we add this into docs [here](https://github.com/timescale/promscale/blob/master/docs/cli.md)? As we are adding it back, So the users are aware that it's still available in Promscale latest. :)

JamesGuthrie (approved):

niksajakovljevic:
@VineethReddy02 That docs contains valid flags, not deprecated ones. Note that this PR brings back deprecated flag that was probably missed (to make us backwards compatible)

VineethReddy02 (approved):

PR #1012, opened 2022-01-03
Add the ability to delete trace data

jgpruitt:
Currently there is no easy way to delete trace data. A user may have done a test and may want to drop their trace data to remove test data or because they have decided to stop using tracing altogether and they don't want the trace data to consume space on disk.

We want to offer users the ability to delete trace data.
Adding a new SQL function which will delete all trace data (spans, links, events, tags, etc.).


niksajakovljevic (approved):
Looks good. Have you consider adding filters so we can use the same function to delete eg only for specific service or time range?

pkg/tests/end_to_end_tests/trace_delete_test.go
```diff
@@ -0,0 +1,118 @@
...
+		// call the function to delete all the data
+		_, err = db.Exec(ctx, fmt.Sprintf("select %s.delete_all_traces()", schema.TracePublic))
+		require.NoError(t, err, "delete_all_traces() failed.")
+
+		type result struct {
+			Table    string
+			Expected int64
+		}
+
+		results := []result{
```
cevian, line 98: I might get this list from the database as a list of tables in the trace schema. That would guarantee this test find a fault if we ever add a table and forget to update this function.

cevian (approved):
One optional suggestion, otherwise looks good

PR #1013, opened 2022-01-05
Run `go generate` in Makefile and Dockerfile

niksajakovljevic:
Since we statically implement content of SQL files into Go code we need to run `go generate` to make
sure we use the latest SQL code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,17 @@
 .PHONY: build
-build:
+build: generate
 	go build -o dist/promscale ./cmd/promscale
 
-test:
+.PHONY: test
+test: generate
 	go test -v -race ./... -timeout 40m
 
 # traces-dataset.sz is used by ./pkg/tests/end_to_end_tests/ingest_trace_test.go
 pkg/tests/testdata/traces-dataset.sz:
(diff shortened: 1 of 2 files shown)


jgpruitt:
I like this. Forgetting to run `go generate ./...` has bitten me more times than I care to admit.

Is there any chance that running `go generate ./...` in CI could produce different output than what's checked in? Also, if we are running `go generate ./...` in CI, is there any benefit to having the generated code tracked by git? Was there a reason we didn't do this all along?

Makefile
```diff
@@ -1,15 +1,15 @@
 .PHONY: build
-build:
+build: generate
 	go build -o dist/promscale ./cmd/promscale
 
-test:
+test: generate
```
bitwalker, line 5: We should add `.PHONY: test` here (we're actually missing `.PHONY` on all of the phony targets, such as `all`, `generate`, `e2e-test`, `go-fmt`, etc.

bitwalker (approved):
Just one suggestion, but otherwise looks good to me

niksajakovljevic:
@jgpruitt From what I can tell GoReleaser is running `go generate` so that part is covered meaning that released docker images and binaries always have newest SQL embedded. What wasn't covered was when running with Makefile or building images locally. If SQL is unchanged running `go generate` locally doesn't produce any changes so works as expected.

jgpruitt (approved):

PR #1014, opened 2022-01-05
Add CHANGELOG.md

JamesGuthrie:


bitwalker (approved):
LGTM!

jgpruitt (approved):

cevian:
In our release notes we often have a "Breaking Changes" section. Do we want one in the changelog as well or is that assembled from the other sections?

JamesGuthrie:
> In our release notes we often have a "Breaking Changes" section. Do we want one in the changelog as well or is that assembled from the other sections?

Good question. I think that we would assemble it from other sections. To me it's clear that anything `Removed` is a breaking change. Otherwise, I think that the only other source of breaking changes would be `Changed`. But this begs the question: what are the semantics of `Changed`. I assume that not every `Changed` is a breaking change, so how should we surface breaking changes?

bitwalker:
In my own CHANGELOGs that follow this format, I usually just prefix specific entries like so:

* **BREAKING**: Foo now requires X configuration

As @JamesGuthrie mentioned, removals are implicitly breaking changes, so I mostly find the above convention useful for changes to existing APIs or behaviors.

paulfantom (approved):
lgtm

As for breaking changes, they are just like any other changes and IMHO should be included as part of `Changed` section. However with such approach it is important to explicitly mention what is breaking in release notes.

cevian (approved):
Ok cool. I suggest adopting @bitwalker convention of BREAKING as a prefix to breaking changes and the collating them in release notes.

VineethReddy02 (approved):
LGTM!

antekresic (approved):
Love it, thanks for heading this up. It really makes a difference IMO.

PR #1015, opened 2022-01-05
Add data retention policy for trace data

jgpruitt:
Currently there are not retention policies enforced on tracing data. As a result, users could fill their disk space with tracing data without knowing.

We want to provide users the ability to configure retention for tracing data and ensure that retention is applied when the maintenance job is executed.

We will add a new SQL function set_span_retention_period(new_retention_period interval).
The default retention period for spans will be set to 30 days.

As a result of this, some of the oldest traces may be incomplete in the database (some spans of a trace have been deleted). This is an acceptable tradeoff to ensure applying the retention policy is fast.


pkg/migrations/sql/idempotent/maintenance.sql
```diff
@@ -0,0 +1,495 @@
...
+    _start timestamptz;
+BEGIN
+    _start := clock_timestamp();
+
+    PERFORM SCHEMA_CATALOG.set_app_name('promscale maintenance: data retention: tracing');
+    IF log_verbose THEN
+        RAISE LOG 'promscale maintenance: data retention: tracing: starting';
+    END IF;
+
+    -- transaction 1
```
cevian, line 205: I'd add an extra layer of safety such as if there is an error in any txn its caught and logged, but the rest of the actions continue. We've been caught by this before where an error in one part of the maintenance jobs makes the rest of the job not work.

cevian (changes requested):
Overall looks good but we need some tests for setting the retention policy and for running the procedure.

pkg/migrations/sql/idempotent/maintenance.sql
```diff
@@ -0,0 +1,495 @@
...
+    EXCEPTION WHEN SQLSTATE '22007' THEN --invalid input syntax for type interval
+        RAISE WARNING 'Invalid value specified for span_retention_period in SCHEMA_CATALOG.default.';
+    END;
+
+    IF _span_retention_period is null THEN
+        _span_retention_period = interval '30 days';
+        IF log_verbose THEN
+            RAISE LOG 'Using default value of 30 days for span_retention_period.';
+        END IF;
+    END IF;
```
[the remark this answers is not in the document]
jgpruitt: Very good point. Will fix.

pkg/migrations/sql/idempotent/maintenance.sql
jgpruitt: Good idea. Adding...

jgpruitt:
test is passing, but code is failing due to the new exception handling. Still need to fix the permissions issue with `prom_maintainer`

pkg/migrations/sql/versions/dev/0.7.2-dev/4-data_retention.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO SCHEMA_CATALOG.default(key, value)
+VALUES ('span_retention_period', (30 * INTERVAL '1 days')::text);
```
JamesGuthrie, line 2: This is effectively a breaking change, as it will result in data deletion for setups running longer than 30 days. Interestingly this is one of those cases where we would prefer the migration script to do something other than the preinstall.

jgpruitt: Agreed. Yeah, since this is beta, I'm thinking it's okay.

jgpruitt:
> Overall looks good but we need some tests for setting the retention policy and for running the procedure.

Tests added. Ready for review

pkg/migrations/sql/versions/dev/0.7.2-dev/4-data_retention.sql
ramonguiu: Yes, we should mention in the release notes and indicate to users how they can extend their retention.

JamesGuthrie (changes requested):
Overall LGTM, one comment and one minor nit.

docs/tracing.md
```diff
@@ -166,6 +166,31 @@ If your service is instrumented with Zipkin, configure the Zipkin transport you
...
+
+### Compression
+
+The `_ps_trace.span`, `_ps_trace.event`, and `_ps_trace.link` tables are [hypertables](https://docs.timescale.com/timescaledb/latest/how-to-guides/hypertables/#hypertables).
+These hypertables have [compression](https://docs.timescale.com/timescaledb/latest/how-to-guides/compression/about-compression/) enabled by default if available.
+
+### Retention
+
+The `_ps_trace.span`, `_ps_trace.event`, and `_ps_trace.link` hypertables have an automated data retention policy which will
+drop chunks beyond a certain age. This policy is driven by the `span_retention_period` which is defaulted to 30 days in
```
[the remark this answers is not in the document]
jgpruitt: I'll add `SCHEMA_CATALOG.get_default_span_retention_period()` this morning real quick

pkg/migrations/sql/versions/dev/0.7.2-dev/4-data_retention.sql
jgpruitt: Do I mention this [here](https://github.com/timescale/promscale/releases/tag/untagged-a5f5b8393fcc39093e88) for now? When we have the CHANGELOG stuff added, I would just put it in the changelog, correct?

cevian: you can put it in https://github.com/timescale/promscale/releases/tag/untagged-f8c325bead62cae08400

cevian (approved):
LGTM

JamesGuthrie (approved):

PR #1016, opened 2022-01-06
Ease locking to speedup ingestion on startup

niksajakovljevic:
Due to a lot of locking while inserting metadata on series we tend to
have a `slow start`. It can take up to 15-20 minutes to get to full speed
when ingesting heavily (eg. 500K/s).
Trying to ease some locking.

SQL function `get_or_create_series_id_for_label_array` was taking unnecessary
ACCESS SHARE lock on metric data table. The label keys are actually
created in another function so no need to take this lock.

SQL function `get_new_pos_for_key` was taking SHARE UPDATE EXCLUSIVE lock
on metric data table. Using advisory lock here should enable writes to metric
data table while we safely get new key position.

NOTE: This is work in progress as I believe there are few more places where we can improve locking.


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1163,12 +1156,6 @@ BEGIN
    SELECT mtn.id, mtn.table_name FROM SCHEMA_CATALOG.get_or_create_metric_table_name(metric_name) mtn
```
cevian, line 1156: we can also delete this line right?

cevian (changes requested):
I'd rather get all the changes in `SCHEMA_CATALOG.get_or_create_series_id_for_label_array` in first as a separate PR because the changes there are obvious and then move on to the new_pos_for_key changes that seem less obvious to me.

cevian:
@niksajakovljevic can this PR be closed?

niksajakovljevic:
Since SQL logic was moved to extension repository this PR can be closed.

PR #1021, opened 2022-01-07
Use pg14-based timescaledb images in docker-compose

JamesGuthrie:


paulfantom (approved):
lgtm

docker-compose/high-availability/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.0'
 
 services:
   db:
-    image: timescaledev/promscale-extension:latest-pg12
+    image: timescaledev/promscale-extension:latest-ts2-pg14
```
cevian, line 5: lets actually change this to `timescale/timescaledb-ha:pg14-latest`.

cevian (approved):
Lets put this as a not in the changelog as a breaking change. People can't just switch to this docker compose from the old docker compose without manual intervention.

docker-compose/high-availability/docker-compose.yaml
bitwalker: I think we need to produce an HA version of the extension image and use that.

docker-compose/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.0'
 
 services:
   db:
-    image: timescaledev/timescaledb-ha:pg12-latest
+    image: timescale/timescaledb-ha:pg14-latest
```
bitwalker, line 5: Shouldn't this also use the promscale-extension image? Especially now that we're planning to require the extension, might be good to already have these files depend on that image.

JamesGuthrie: timescaledb-ha also has the extension packaged.

JamesGuthrie:
@cevian 

> Lets put this as a not in the changelog as a breaking change. People can't just switch to this docker compose from the old docker compose without manual intervention.

Technically they can. The docker-compose doesn't define volumes or volume mounts which would be reused over invocations of docker-compose, so changing the docker image shouldn't actually affect anything or anyone.

docker-compose/high-availability/docker-compose.yaml
JamesGuthrie: @cevian Do you know why we were not using timescaledb-ha previously?

cevian: @bitwalker the extension should already be in the HA image

cevian: @JamesGuthrie I am guessing legacy oversight

docker-compose/docker-compose.yaml
cevian: Yep its already there

cevian (commented):
I'll approve this change once we switch over to the ha image

docker-compose/high-availability/docker-compose.yaml
JamesGuthrie: done

PR #1024, opened 2022-01-07
Add a simple tracing setup to the docker-compose

JamesGuthrie:
## Description

This change adds a tracing configuration to our docker-compose, leveraging Promscale's trace output.

## Merge requirements

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -20,6 +20,7 @@ services:
     image: timescale/promscale:latest
     ports:
       - 9201:9201/tcp
+      - 9202:9202/tcp
     restart: on-failure
     depends_on:
       - db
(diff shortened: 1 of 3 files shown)


docker-compose/otel-collector-config.yml
```diff
@@ -0,0 +1,28 @@
...
+
+exporters:
+  logging:
+  otlp:
+    endpoint: promscale:9202
+    tls:
+      insecure: true
+
+processors:
+  batch:
```
VineethReddy02, line 14: Any specific reason on this batch config?

docker-compose/docker-compose.yaml
```diff
@@ -28,6 +29,27 @@ services:
...
+  otel-collector:
+    platform: linux/amd64
+    image: "otel/opentelemetry-collector:0.41.0"
+    command: [ "--config=/etc/otel-collector-config.yml" ]
+    volumes:
+      - ./otel-collector-config.yml:/etc/otel-collector-config.yml
+    ports:
+      - "4317:4317" # OTEL GRPC
+
+  tracegen:
```
VineethReddy02, line 44: Sharing the same from the slack thread:

I’m personally against tracegen, As we already used it in past for testing trace ingestion, which didn’t actually help with real-world data or at least something similar. Tracegen was all about random string in the trace data model. This will not be useful for users anyhow to at least get the feel and visualize the traces.

I think the easiest way to generate some traces is to ingest traces from Promscale. Currently, we are already in the implementation phase of exposing traces from Promscale that should be the source of traces rather than going with some random trace generator into docker-compose. On the other side, node-exporter is a valid metrics generator from node so it adds value into docker-compose.

bitwalker (approved):

JamesGuthrie:
I'm gonna wait for some tracing support in Promscale and then use that as a source of traces instead of `tracegen`.

JamesGuthrie:
FYI I have added a config on top of ante's open PR here: https://github.com/timescale/promscale/pull/1052. Once that is merged I will rebase and un-draft this PR.

antekresic (approved):

PR #1032, opened 2022-01-11
Add telemetry CLI flag section

antekresic:
## Description

The introduction of this new section will reduce confusion between the telemetry
configuration for the telemetry that Promscale produces and various other types
of telemetry Promscale is used to store and query.

## Merge requirements

- [x] CHANGELOG entry for user-facing changes (_not needed, changelog already includes CLI renaming changes_)


Harkishen-Singh (approved):

JamesGuthrie:
I'm not sure that I understand the motivation behind this change. I'm also not sure what it clears up exactly. The term `telemetry` seems to have three different meanings at the moment:

1. Tracking/sending promscale usage data (usage telemetry)
2. Promscale's telemetry (logs, metrics and traces that it emits about itself)
3. End-user telemetry (logs, metrics and traces that Promscale stores, the reason it's used).

jgpruitt:
I tend to agree with James. Are these flags related to the logs that the connector spits out about its own activity?

docs/cli.md
```diff
@@ -48,13 +48,13 @@ You can also find information on flags with `promscale_<version> -help`.
 | db.uri | string | | TimescaleDB/Vanilla PostgresSQL URI. Example:`postgres://postgres:password@localhost:5432/timescale?sslmode=require` |
 | db.user | string | postgres | TimescaleDB/Vanilla Postgres user. |
 
-## Logging flags
+## Telemetry flags
```
cevian, line 51:
```suggestion
## Telemetry flags (for telemetry generated by the Promscale connector itself)
```

cevian (approved):

antekresic:
This change is about the Promscale telemetry it emits about itself. I understand there are multiple tracks of telemetry when talking about Promscale, maybe we can figure out a better name for this group.

I'm open to suggestions for naming, the main thing that we separate out these options from the other telemetry which will be ingested into Promscale.

jgpruitt (approved):

PR #1033, opened 2022-01-11
Fix os_version and add os_id in telemetry metadata.

Harkishen-Singh:

## Description

This is a fix for wrong values in `os_version`, a bug in telemetry collection.

### DIFF
--- a/pkg/api/query.go
+++ b/pkg/api/query.go
@@ -78,6 +78,7 @@ func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable, metric
 			metrics.FailedQueries.Add(1)
 			return
 		}
+		metrics.ExecutedQueries.Add(1)
 
 		respondQuery(w, res, res.Warnings)
 	}
--- a/pkg/api/query_range.go
+++ b/pkg/api/query_range.go
@@ -125,6 +125,7 @@ func queryRange(conf *Config, queryEngine *promql.Engine, queryable promql.Query
 			metrics.FailedQueries.Add(1)
 			return
 		}
+		metrics.ExecutedQueries.Add(1)
 
 		respondQuery(w, res, res.Warnings)
 	}
--- a/pkg/api/query_range_test.go
+++ b/pkg/api/query_range_test.go
@@ -171,6 +171,7 @@ func TestRangedQuery(t *testing.T) {
 				InvalidQueryReqs: invalidQueryReqs,
 				QueryDuration:    queryDuration,
 				TimedOutQueries:  &mockMetric{},
+				ExecutedQueries:  &mockMetric{},
 			}
 			handler := queryRange(&Config{MaxPointsPerTs: 11000}, engine, query.NewQueryable(tc.querier, nil), metrics)
 			queryUrl := constructRangedQuery(tc.metric, tc.start, tc.end, tc.step, tc.timeout)
--- a/pkg/api/read.go
+++ b/pkg/api/read.go
@@ -72,6 +72,7 @@ func Read(config *Config, reader querier.Reader, metrics *Metrics) http.Handler 
 			metrics.FailedQueries.Add(queryCount)
 			return
 		}
+		metrics.ExecutedQueries.Add(queryCount)
 
 		duration := time.Since(begin).Seconds()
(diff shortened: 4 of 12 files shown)


cevian (approved):

pkg/telemetry/metadata.go
```diff
@@ -40,12 +42,21 @@ func promscaleMetadata() Metadata {
 	// leading to type mismatch. Hence, we create a slice to handle both cases.
 	metadata["os_sys_name"] = toString(uname.Sysname[:])
 	metadata["os_node_name"] = toString(uname.Nodename[:])
-	metadata["os_release"] = toString(uname.Release[:])
-	metadata["os_version"] = toString(uname.Version[:])
 	metadata["os_machine"] = toString(uname.Machine[:])
+
+	turnOffPackageLogging()
+	distro := linux.DiscoverDistro()
```
bitwalker, line 48: You don't need an external dependency for this. You can simply read `/etc/os-release` for distribution name, version, description, and other metadata on all of the major Linux distros (and probably all the minor ones too). This includes Debian/Ubuntu, CentOS/Fedora/RHEL, and Alpine for sure, possibly others.

The file contains newline-separated environment variable definitions, e.g.:

```text
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.14.3
```

On all of the OSs I've tested (all of those I mentioned above) at the very least NAME, ID, and VERSION_ID are set, and in some cases there are additional vars such as `VERSION_CODENAME` (which contains distro names like Debian/Ubuntu use, e.g. `bullseye`).

Since we only care about the distro name and version, `/etc/os-release` covers this well.

Harkishen-Singh: agreed, but this was a suggestion from Ante, and I do like it since it keeps the code clean and minimal. Plus, the library does the same under the hood, so I don't see anything wrong in using it.

pkg/api/query_range_test.go
```diff
@@ -171,6 +171,7 @@ func TestRangedQuery(t *testing.T) {
 				InvalidQueryReqs: invalidQueryReqs,
 				QueryDuration:    queryDuration,
 				TimedOutQueries:  &mockMetric{},
+				ExecutedQueries: &mockMetric{},
```
antekresic, line 174: Fix formatting, its failing the CI.

pkg/api/query.go
```diff
@@ -77,6 +77,8 @@ func queryHandler(queryEngine *promql.Engine, queryable promql.Queryable, metric
 			respondError(w, http.StatusUnprocessableEntity, res.Err, "execution")
 			metrics.FailedQueries.Add(1)
 			return
+		} else {
+			metrics.ExecutedQueries.Add(1)
```
antekresic, line 81: Unnecessary else.

pkg/api/query_range.go
```diff
@@ -124,6 +124,8 @@ func queryRange(conf *Config, queryEngine *promql.Engine, queryable promql.Query
 			respondError(w, http.StatusUnprocessableEntity, res.Err, "execution")
 			metrics.FailedQueries.Add(1)
 			return
+		} else {
+			metrics.ExecutedQueries.Add(1)
```
antekresic, line 128: Unnecessary else.

pkg/api/read.go
```diff
@@ -71,6 +71,8 @@ func Read(config *Config, reader querier.Reader, metrics *Metrics) http.Handler
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			metrics.FailedQueries.Add(queryCount)
 			return
+		} else {
+			metrics.ExecutedQueries.Add(queryCount)
```
antekresic, line 75: Unnecessary else.

antekresic (changes requested):
Lets fix the CI and remove unnecessary nesting, should be good to go after that.

pkg/telemetry/metadata.go
cevian: @bitwalker It seems the library rules are more exact/specific https://github.com/dekobon/distro-detect/blob/master/linux/distrochecks.go but would be curious if you disagree.

antekresic (approved):

PR #1034, opened 2022-01-11
Clean up changelog

JamesGuthrie:
## Description

This change updates the changelog in preparation for the 0.8.0 release.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Added
-- Add ability to configure chunk interval on startup [#991]
+- Add ability to configure the default chunk interval on startup [#991]
 - Add `ps_trace.delete_all_traces()` function to delete all trace data [#1012]
 - Add `ps_trace.set_span_retention_period(INTERVAL)` function to set span retention period [#1015]
 - Add `ps_trace.get_span_retention_period()` database function to get current span retention period [#1015]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,24 +15,24 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Added
-- Add ability to configure chunk interval on startup
-- Add function ps_trace.delete_all_traces() to delete all trace data
+- Add ability to configure chunk interval on startup [#991]
+- Add function ps_trace.delete_all_traces() to delete all trace data [#1012]
 
 ### Changed
-- Rename CLI flags to improve user interface
+- Rename CLI flags to improve user interface [#964]
 
 ### Deprecated
-- Deprecate `migrate` flag
+- Deprecate `migrate` flag [#964]
 
 ### Removed
-- Remove deprecated `-promql-enable-feature` flag
-- Remove deprecated leader election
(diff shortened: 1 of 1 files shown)


jgpruitt:
@JamesGuthrie I added entries for #1015. Please let me know if I did it wrong.

jgpruitt (approved):

CHANGELOG.md
```diff
@@ -15,23 +15,27 @@ We use the following categories for changes:
...
-- Add ability to configure chunk interval on startup
+- Add ability to configure chunk interval on startup [#991]
+- Add `ps_trace.delete_all_traces()` function to delete all trace data [#1012]
+- Add `ps_trace.set_span_retention_period(INTERVAL)` function to set span retention period. [#1015]
+- Add `ps_trace.get_span_retention_period()` database function to get current span retention period. [#1015]
+- BREAKING: Enable and configure 30 day default retention period for span data. [#1015]
 
 ### Changed
-- Rename CLI flags to improve user interface
+- Rename CLI flags to improve user interface [#964]
```
bitwalker, line 25: This is technically a breaking change, no?

JamesGuthrie: Depends on how you define breaking change. The old CLI flags still work, so there's no breakage when you update to promscale 0.8.0.

CHANGELOG.md
```diff
@@ -15,23 +15,27 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Added
-- Add ability to configure chunk interval on startup
+- Add ability to configure chunk interval on startup [#991]
+- Add `ps_trace.delete_all_traces()` function to delete all trace data [#1012]
+- Add `ps_trace.set_span_retention_period(INTERVAL)` function to set span retention period. [#1015]
+- Add `ps_trace.get_span_retention_period()` database function to get current span retention period. [#1015]
+- BREAKING: Enable and configure 30 day default retention period for span data. [#1015]
```
bitwalker, line 22: This probably belongs under `CHANGED` instead, since it's not adding new behavior, but changing the existing behavior (which is why it's a breaking change). As a rule, I'd assume that anything that could break things for a user must be a change, and not an addition.

CHANGELOG.md
bitwalker: Ah, it read to me like the old names no longer exist and they were given new names - if both sets still exist, then yeah, no worries.

CHANGELOG.md
JamesGuthrie: done

CHANGELOG.md
JamesGuthrie: 👍

bitwalker (approved):

CHANGELOG.md
```diff
@@ -15,23 +15,27 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Added
-- Add ability to configure chunk interval on startup
+- Add ability to configure chunk interval on startup [#991]
```
cevian, line 18:
```suggestion
- Add ability to configure the default chunk interval on startup [#991]
```

cevian (approved):
One minor thing

Harkishen-Singh (approved):

PR #1043, opened 2022-01-12
Remove deprecated `migrate` from CLI.md

JamesGuthrie:
## Description

The `migrate` option is deprecated.

## Merge requirements


cevian (changes requested):
I think this is actually related to `startup.skip-migrate` already defined below. I suggest just removing this line

JamesGuthrie:
> I think this is actually related to `startup.skip-migrate` already defined below. I suggest just removing this line

You're correct - thanks for the catch! 👍

cevian (approved):

PR #1046, opened 2022-01-13
Remove unnecessary locking in `get_or_create_series_id_for_label_array`

antekresic:
## Description

SQL function get_or_create_series_id_for_label_array was taking unnecessary
ACCESS SHARE lock on metric data table. The label keys are actually
created in another function so no need to take this lock.

This is work based on draft PR #1016  by @niksajakovljevic 

## Merge requirements

- ~~[ ] CHANGELOG entry for user-facing changes~~
- ~~[ ] Updated the relevant documentation~~


pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1163,12 +1163,6 @@ BEGIN
    SELECT mtn.id, mtn.table_name FROM SCHEMA_CATALOG.get_or_create_metric_table_name(metric_name) mtn
```
cevian, line 1163: @antekresic for future optimization: we should be able to get rid of this call too if we cache the metric_id and table_name in the connector code. That would optimize this path further.

cevian (approved):

jgpruitt (approved):

PR #1047, opened 2022-01-13
Add database host and port to instructions

ramonguiu:
## Description

The instructions don't tell people how to specify a host and port if TSDB is not running locally on the default port. I've just added the two parameters to the promscale command.


JamesGuthrie (approved):
Looks good to me, also needs to be updated in the docs repo: https://github.com/timescale/docs/blob/latest/promscale/installation/source.md

ramonguiu:
bors+

PR #1052, opened 2022-01-13
Initial metric ingest tracing instrumentation

### LINKED ISSUE #1004: Improve instrumentation of ingest path

antekresic:
## Description

This change is the first step into tracing instrumentation on Promscale
using OpenTelemetry tracing libraries. Right now, its limited to metric
ingest path and only supports exporting traces to Jaeger. Rest of the
codebase (i.e. query path), as well as adding more exporting support,
will be covered in future PRs.

Closes #1004 

## Merge requirements


go.sum
```diff
@@ -1987,6 +1988,7 @@ go.opentelemetry.io/collector v0.42.0/go.mod h1:HiryUIokIPVCspJIAXlGdpfPFCepUAFL
 go.opentelemetry.io/collector/model v0.39.0/go.mod h1:gS8A27wi+8gM3hrXL+dEjTbrbLxktjHjAwwqI31ELgQ=
 go.opentelemetry.io/collector/model v0.42.0 h1:jQb9oi9NwhTJu6H8cOlK/3yeg+cyWxOrQD8A5TlcqQw=
 go.opentelemetry.io/collector/model v0.42.0/go.mod h1:uUgx84gI+G/tE87Oo84305q0MD8tUV9uWxg+ckAE7Ew=
+<<<<<<< HEAD
```
JamesGuthrie, line 1991: Looks like this snuck in uninvited 😂

pkg/pgmodel/ingestor/copier.go
```diff
@@ -124,22 +147,28 @@ func persistBatch(conn pgxconn.PgxConn, sw *seriesWriter, elf *ExemplarLabelForm
...
 	//This mutex is not for safety, but rather for better batching.
 	//It guarantees that only one copier is reading from the channel at one time
 	//This ensures bigger batches as well as less spread of a
 	//single http request to multiple copiers, decreasing latency via less sync
 	//overhead especially on low-pressure systems.
+	span.AddEvent("Acquiring lock")
 	getBatchMutex.Lock()
+	span.AddEvent("Got the lock")
 	defer getBatchMutex.Unlock()
+	defer span.AddEvent("Unlocking")
```
jgpruitt, line 166: Why not combine these two defers into one?

Also, Go question, are defers executed in reverse order? i.e. is the call to `span.End()` guaranteed to happen AFTER the call to `span.AddEvent(...)`?

jgpruitt (approved):

pkg/pgmodel/ingestor/copier.go
antekresic: Right, I guess I can combine them, need to be careful with closures there. Defers are called last-to-first so End will always come after AddEvent in this case.

pkg/tracer/trace.go
```diff
@@ -0,0 +1,52 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	defaultTracerName = "github.com/timescale/promscale"
```
Harkishen-Singh, line 15: Is this a convention?

pkg/tracer/trace.go
```diff
@@ -0,0 +1,52 @@
...
+func InitProvider(cfg *Config) (*tracesdk.TracerProvider, error) {
+	// Create the Jaeger exporter
+	exp, err := jaeger.New(jaeger.WithCollectorEndpoint(jaeger.WithEndpoint(cfg.JaegerCollectorEndpoint)))
+	if err != nil {
+		return nil, err
+	}
+	tp := tracesdk.NewTracerProvider(
+		// Always be sure to batch in production.
+		tracesdk.WithBatcher(exp),
+		tracesdk.WithSampler(tracesdk.TraceIDRatioBased(cfg.SamplingRatio)),
```
Harkishen-Singh, line 44: What happens if the sampling ratio > 1. I don't see any validity checks for this. Does the inner library handle automatically?

pkg/tracer/trace.go
```diff
@@ -0,0 +1,52 @@
...
+		// Record information about this application in an Resource.
+		tracesdk.WithResource(resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String("promscale-service"),
```
Harkishen-Singh, line 48: promscale-service or promscale?

Harkishen-Singh (approved):
Looks good, just few comments.

pkg/tracer/trace.go
antekresic: Needed a name for the instrumentation library, seen a couple of places using repo URL so did it like that. Honestly, makes little difference.

pkg/tracer/trace.go
antekresic: Yep, its clamped to 1 by the lib so I didn't handle it.

pkg/tracer/trace.go
antekresic: promscale-service because other services in the dev cloud were named like that. Makes little difference to me but I sticked to the convention there.

pkg/runner/runner.go
```diff
@@ -74,6 +77,31 @@ func Run(cfg *Config) error {
 		tput.InitWatcher(cfg.ThroughputInterval)
 	}
 
+	if cfg.TracerCfg.JaegerCollectorEndpoint != "" {
```
cevian, line 80: We should make it easy to configure with either Jaeger OR open telemetry endpoint

pkg/tracer/trace.go
cevian, line 48:
```suggestion
			semconv.ServiceNameKey.String("promscale-connector-service"),
```

cevian (changes requested):
Looks great. Mostly minor tweaks and suggestions

pkg/pgmodel/ingestor/copier.go
```diff
@@ -242,7 +277,7 @@ func retryAfterDecompression(conn pgxconn.PgxConn, req copyRequest, minTimeInt i
 
 	metrics.DecompressCalls.Inc()
 	metrics.DecompressEarliest.WithLabelValues(table).Set(float64(minTime.UnixNano()) / 1e9)
-	err, _ := insertSeries(conn, req) // Attempt an insert again.
+	err, _ := insertSeries(context.Background(), conn, req) // Attempt an insert again.
```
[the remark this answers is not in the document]
antekresic: We should but this particular piece was not of interest since it never happened in benchmark runs so I left it for a future addition. Top level function is traced without going into the retry.

pkg/runner/runner.go
antekresic: Adding new types of endpoints is left for next iteration, Jaeger was needed because that what we use in our current setup.

pkg/pgmodel/ingestor/copier.go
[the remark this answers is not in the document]
antekresic: If you think this is important to add now, let me know and I'll do it.

pkg/runner/runner.go
cevian: If we put this off, let's add an issue to keep track of this.

cevian (approved):

pkg/runner/runner.go
antekresic: https://github.com/timescale/promscale/issues/1076

PR #1053, opened 2022-01-13
Rename span_retention_period

jgpruitt:
## Description

Rename `span_retention_period` to `trace_retention_period`. This better reflects that this config setting impacts ALL tracing-related data. Rename associated functions and update docs.

## Merge requirements


cevian (approved):

JamesGuthrie (approved):
In general this looks good to me.

I just wanted to comment an observation: we've been having a discussion about how we want to write migration files, and whether we want to support users who are using code from the master branch.

This change doesn't cleanly remove functions or values which were previously introduced, meaning that `{get,set}_span_retention_period` functions and the `'span_retention_period'` in the `default` table _would_ remain for any users who were on the master branch before this commit.

Obviously things will still work for such users, so there's no breakage now. But there is some pollution which could cause breakage at some point in the future.

I'm not suggesting that we have to take this into account now and rewrite the migration code, but I do think it highlights how easy it is to miss this stuff.

JamesGuthrie:
Regarding my observation: I just saw that this was discussed in slack, and that there was an explicit decision not to do a migration.

jgpruitt:
@JamesGuthrie it **IS** a good point though, and I'd like to explore how we can address it to make development easier. Maybe we should have a branch dedicated to integrating changes to stage for the next release with the expectation that no one should be installing from this branch, and therefore there is no expectation for doing migrations for changes like this which have not been released.

PR #1062, opened 2022-01-17
Cache metric ID and forward the metric info to series writer

antekresic:
## Description

This is an optimization to remove the need to fetch metric ID and
table name when storing new series data since we already fetch
this info in a prior step of the process.

## Merge requirements


JamesGuthrie (changes requested):
There's no CHANGELOG entry, this change might be worth mentioning if it has an impact.

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -32,55 +32,62 @@ type readRequest struct {
...
-		return "", true, fmt.Errorf("failed to get the table name for metric %s: %w", metric, err)
+		return 0, "", true, fmt.Errorf("failed to get the table name for metric %s: %w", metric, err)
 	}
 
-	var tableName string
-	var possiblyNew bool
+	var (
+		metricID    int64
+		tableName   string
+		possiblyNew bool
```
Harkishen-Singh, line 49: Can we use named returns so it helps in making this func be cleaner? It will avoid making `0, ` in the entire func. Just a simple `return`

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -32,55 +32,62 @@ type readRequest struct {
...
+		return 0, "", true, fmt.Errorf("failed to get the table name for metric %s: %w", metric, err)
 	}
 
 	if tableName == "" {
-		return "", true, fmt.Errorf("failed to get the table name for metric %s: empty table name returned", metric)
+		return 0, "", true, fmt.Errorf("failed to get the table name for metric %s: empty table name returned", metric)
+	}
+
+	if metricID == 0 {
+		return 0, "", true, fmt.Errorf("failed to get the table name for metric %s: zero metric ID returned", metric)
```
Harkishen-Singh, line 72:
```suggestion
		return 0, "", true, fmt.Errorf("failed to get the metric id for metric %s: zero metric ID returned", metric)
```

Harkishen-Singh (approved):

pkg/pgmodel/ingestor/metric_batcher.go
antekresic: No, naked returns are a source of confusion and make the code harder to read especially when its not a trivial function. They should be avoided in most cases.

Can switch this to named returns so we don't need to declare new vars.

jgpruitt (approved):

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -1153,7 +1153,6 @@ LANGUAGE PLPGSQL VOLATILE;
 GRANT EXECUTE ON FUNCTION SCHEMA_CATALOG.get_or_create_series_id_for_kv_array(TEXT, text[], text[]) TO prom_writer;
 
 
-
 CREATE OR REPLACE FUNCTION SCHEMA_CATALOG.get_or_create_series_id_for_label_array(metric_name TEXT, larray SCHEMA_PROM.label_array, OUT table_name NAME, OUT series_id BIGINT)
```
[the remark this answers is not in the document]
jgpruitt: +1

antekresic:
@JamesGuthrie I've added an entry to CHANGELOG and update the benchmarks. I've left the old function in and added a new benchmark so we can compare the improvement.

This is what I get on my machine:
```
$ go test -bench=BenchmarkGetSeriesIDForKeyValueArrayNewSeriesNewLabelsBatch -run=XXX -log-level=error -benchtime=10000x
...
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesNewLabelsBatch-12
Labels DB time 722.503306ms per item: 72.214µs N-labels 10005
DB time 4.892545455s per item: 489.254µs N 10000
   10000            563624 ns/op
...
BenchmarkGetSeriesIDForKeyValueArrayNewSeriesNewLabelsBatchOptimized-12
Labels DB time 729.144623ms per item: 72.878µs N-labels 10005
DB time 2.778612362s per item: 277.861µs N 10000
   10000            352828 ns/op
PASS
```

I can always remove the old if we want to do it, I'll let Mat take a look and advise there.

JamesGuthrie (approved):

cevian (changes requested):
This is a very nice improvement. I think it's good to run the benchmark before and after the change and record diff in PR (as you have above). But I don't think we should keep the old functions around just for that purpose. Also I think that in a bunch of places (maybe all places) where you now pass around the metricID and tableName we should pass around a model.MetricInfo instead. That will make the code more future proof.

cevian (approved):

cevian (approved):

PR #1068, opened 2022-01-19
Make helm publish use relative URLs

cevian:
## Description

Add the --relative flag. This makes the url set in the index.yaml
use relative urls instead of s3://chart.timescale.com/* which doesn't
work since the charts are accessed using https.

## Merge requirements


VineethReddy02 (approved):
LGTM

JamesGuthrie (approved):

PR #1075, opened 2022-01-24
Remove generic schemas

JamesGuthrie:
## Description

There was a plan to allow the schema names which promscale uses to be
user-defined. For this reason the SQL migrations use `SCHEMA_PROM` and
the like.

In the process of moving the SQL to the extension, we discussed this and
decided that we don't really need to make the schemas user-defined.
Replacing the generic schemas with the actual schema makes both the SQL
and go code simpler to follow.

This change removes all generic schema names and their support from both
SQL migration scripts, as well as go SQL statements.

## Merge requirements


JamesGuthrie:
@antekresic in the first pass, can I ask you to take a look at the changes that I made to the queries in the go sources? I'm not sure if that's the way that you would have wanted it done. I can also break this up into two changes: one with mostly SQL changes, and one with only go changes, if you would prefer to review it that way.

pkg/pgmodel/common/schema/schema.go
```diff
@@ -5,22 +5,10 @@
...
-	Prom             = "prom_api"
-	Data             = "prom_data"
-	Info             = "prom_info"
-	Exemplar         = "prom_data_exemplar"
-	Ext              = "_prom_ext"
-	Catalog          = "_prom_catalog"
-	PromscaleCatalog = "_ps_catalog"
-	Timescale        = "public"
+	Data     = "prom_data"
+	Exemplar = "prom_data_exemplar"
```
jgpruitt, line 9: We can get rid of these now, right?

pkg/pgmodel/common/schema/schema.go
```diff
@@ -5,22 +5,10 @@
...
 
 	LockID = 0x4D829C732AAFCEDE // Chosen randomly.
 
-	SeriesView = "prom_series"
-	MetricView = "prom_metric"
 	DataSeries = "prom_data_series"
```
jgpruitt, line 13: This one can go too?

jgpruitt (changes requested):

pkg/pgmodel/common/schema/schema.go
JamesGuthrie: I would like to, but I'm honestly not sure exactly how. The pattern that was used with these was slightly different to the other usages which I've already replaced. In particular, these usages all/mostly use `pgx.Identifier`, in combination with some other item, like a table name or similar. I'm not entirely sure a) if we want to replace all of those, or b) how it should best be done. Please give me your thoughts.

jgpruitt: ok, I'll look...

cevian: I think we can just use the strings in pgx.Identifier{} directly. If we want to keep constants, we should rename them to make sure they are only used as appropriate

jgpruitt: Ah, the `pgx.Identifier` pattern is just there to make sure that the table name is properly escaped as a PostgreSQL identifier. I'll fix...

cevian (approved):
Looks good to me except one fairly minor comment

jgpruitt (approved):

JamesGuthrie:
Thanks for the additional fixes @jgpruitt. I've rebased and squashed the changes down into one commit.

pkg/pgmodel/delete/delete.go
```diff
@@ -79,7 +78,7 @@ func getMetricNameSeriesIDFromMatchers(conn pgxconn.PgxConn, matchers []*labels.
 func filterMetricNameSeriesIds(metrics, schemas []string, seriesIds [][]model.SeriesID) (filteredMetrics []string, filteredSeriesIds [][]model.SeriesID) {
 	for i := range metrics {
 		metricSchema := schemas[i]
-		if metricSchema != schema.Data {
+		if metricSchema != "prom_data" {
```
antekresic, line 81: I think we should leave the schema package alone and this reverts as it was. Don't want to litter the code with hardcoded schema names like this.

pkg/pgmodel/migrate.go
```diff
@@ -375,7 +373,7 @@ func orderFilesNaturally(entries []os.FileInfo) []string {
 	return pp.getNames()
 }
 
-func replaceSchemaNames(r io.ReadCloser) (string, error) {
+func readSchemas(r io.ReadCloser) (string, error) {
```
antekresic, line 376: `readSchemas` might not be the best name here.
```suggestion
func readMigrationFile(r io.ReadCloser) (string, error) {
```

antekresic (commented):
I would like to leave the schema constants so we can use them instead of needing to hardcode schema names in a lot of places.

pkg/pgmodel/migrate.go
JamesGuthrie: Done

pkg/pgmodel/delete/delete.go
JamesGuthrie: I think I got all of these.

JamesGuthrie: I also renamed the constant names to be in line with their contents.

JamesGuthrie:
Note: I just rebased against latest master, and included the commit from https://github.com/timescale/promscale/pull/1079 which should be merged before this anyway. Once that's merged I'll rebase again and then merge.

JamesGuthrie:
This is now up to date with master and ready for review.

pkg/tests/end_to_end_tests/continuous_agg_test.go
```diff
@@ -270,7 +269,7 @@ WITH (timescaledb.continuous,  timescaledb.ignore_invalidation_older_than = '1 m
 
 		// Drop some raw metric data and check that the series data is not marked for deletion.
 		// NOTE: we cannot drop all the raw data becuase we are getting `too far behind` issues with 1.x caggs
-		_, err = db.Exec(context.Background(), "CALL _prom_catalog.drop_metric_chunks($1, $2, $3)", schema.Data, "metric_2", time.Unix(endTime/1000-20000, 0))
+		_, err = db.Exec(context.Background(), "CALL _prom_catalog.drop_metric_chunks($1, $2, $3)", "prom_data", "metric_2", time.Unix(endTime/1000-20000, 0))
```
antekresic, line 272:
```suggestion
		_, err = db.Exec(context.Background(), "CALL _prom_catalog.drop_metric_chunks($1, $2, $3)", schema.PromData, "metric_2", time.Unix(endTime/1000-20000, 0))
```

antekresic (approved):

pkg/tests/end_to_end_tests/continuous_agg_test.go
JamesGuthrie: I've updated this

PR #1077, opened 2022-01-25
Fix broken `promscale_packager` telemetry field for docker envs.

Harkishen-Singh:

This commits fixes the `promscale_packager` telemetry field that was
earlier showing `unknown` for all docker deployments. Now, it will show
`promscale_packager: docker` in the `_timescaledb_catalog.metadata`
table.

Metadata table output

```shell
 promscale_packager                                          | docker                               | t
 promscale_os_node_name                                      | 2bce24e46fcc                         | t
```

## Description

## Merge requirements

### DIFF
--- a/pkg/telemetry/metadata.go
+++ b/pkg/telemetry/metadata.go
@@ -61,12 +61,10 @@ func turnOffPackageLogging() {
 
 func getPkgEnv() string {
 	pkg := os.Getenv("PROMSCALE_PKG")
-	switch pkg {
-	case "deb", "rpm", "apk":
-		return pkg
-	default:
-		return "unknown"
+	if pkg == "" {
+		pkg = "unknown"
 	}
+	return pkg
 }
 
 func toString(prop []byte) string {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ We use the following categories for changes:
 ### Added
 - Add support to instrument Promscale's Otel GRPC server with Prometheus metrics [#1061]
 
+### Fixed
+- Fix broken `promscale_packager` telemetry field for docker envs [#1077]
+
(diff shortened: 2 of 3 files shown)


ramonguiu (approved):

pkg/telemetry/metadata.go
```diff
@@ -61,12 +61,10 @@ func turnOffPackageLogging() {
 
 func getPkgEnv() string {
 	pkg := os.Getenv("PROMSCALE_PKG")
-	switch pkg {
-	case "deb", "rpm", "apk":
-		return pkg
-	default:
-		return "unknown"
+	if pkg == "" {
+		pkg = "unknown"
```
antekresic, line 65: Why not just include `docker` here like it was previously set? The reason I say this is because users can change the env variable and add random values which could potentially leak into our telemetry and cause issues for us.

I know it sounds like a long shot but users can, for instance, put an email address in and accuse us of breaking GDPR which would cause legal issues.

antekresic (changes requested):
We should check the value from env variable before sending it to telemetry because telemetry will be stored long-term by us.

pkg/telemetry/metadata.go
ramonguiu: Users can easily add email addresses in the telemetry they send anyway, so I don't know how this would be different. I believe we are already taking values from environment variables with tobs.

I don't think that's the right way to claim GDPR compliance. Our code is available, so anybody can check the code and see what it does and it will become apparent the data is sent by the user.

The problem with limiting this is that we need to change the code every time we want to add a new package which I don't think we need.

antekresic: If you think that it won't be an issue, then I'm OK with the change. I'll resolve this and approve the PR.

antekresic (approved):

PR #1078, opened 2022-01-25
Use PERFORM instead of SELECT when results discarded

JamesGuthrie:
## Description

When moved to the extension, this errors out with 'query has no
destination for result data'.

## Merge requirements


cevian (approved):

JamesGuthrie:
Hmmm... this approach is not going to work like this. I think this will require substitution in our `move_sql.py` script. Closing this PR.

PR #1079, opened 2022-01-25
Inline ADVISORY_LOCK_PREFIX_* values

JamesGuthrie:
## Description

These placeholders were replaced during the migration process. When we
move these SQL scripts to the extension we would either need to
replicate this, or just inline the values here. Inlining seems to be the
simpler solution.

## Merge requirements


cevian (approved):

pkg/migrations/sql/idempotent/base.sql
```diff
@@ -2754,7 +2762,7 @@ BEGIN
         WHERE h.schema_name = 'SCHEMA_DATA' and h.table_name = ht_table;
 
         --alter job schedule is not currently concurrency-safe (timescaledb issue #2165)
-        PERFORM pg_advisory_xact_lock(ADVISORY_LOCK_PREFIX_JOB, bgw_job_id);
+        PERFORM pg_advisory_xact_lock(12377, bgw_job_id);
```
jgpruitt, line 2765: Why did we not create a function for this one? Is it because this is the only instance of the number?

jgpruitt (changes requested):

pkg/migrations/sql/idempotent/base.sql
JamesGuthrie: That was the idea. I can also pull this into a function if you would prefer.

JamesGuthrie: I've pulled this into a function

jgpruitt (approved):
lgtm!

PR #1081, opened 2022-01-26
Fix skipping compression of first to last chunk

antekresic:
## Description

Due to a bug in logic, we were skipping compression of the first
to last chunk which ended over an hour ago from current time. This was
causing to have one more chunk per metric uncompressed, causing
us to use more storage than necessary.

## Merge requirements


JamesGuthrie (changes requested):
It seems as though we have good coverage for the compression tests. Could you update the existing tests to prevent a regression?

jgpruitt (approved):

cevian (approved):
Good catch!

JamesGuthrie (approved):
LGTM!

PR #1090, opened 2022-01-27
Optimize insert path in the DB by avoiding ON CONFLICT

cevian:
Turns out there is a horrible CPU perf penalty on the DB for ON CONFLICT DO NOTHING.
yet in our data, conflicts are rare. So we first try inserting without ON CONFLICT
and fall back if there is a unique constraint violation.

Unlike retriying with COPY this does not require sending the data again
and thus saves latency and bandwidth.

## Description

## Merge requirements


jgpruitt (approved):

bitwalker (approved):
LGTM

antekresic (approved):
Looks good. 

Maybe we should add a Go bench for ingesting metrics (like we have for traces) so we can see the improvement and record here for historical reasons.

We can do that in a separate PR.

cevian:
@antekresic we have some work in https://github.com/cevian/postbench but yeah adding to go benchmarks will also be good

PR #1095, opened 2022-01-31
Fix metric info cache loading when using query path

antekresic:
## Description

With the recent optimization, we have added metric IDs to metric cache
but when running queries on metrics that exist before the ingestion path
generates the cache, cache will be created with invalid metric ID values.
This change fixes that and adds end-to-end test for verifying that
cache is properly generated when this situation occurs.

## Merge requirements

### DIFF
--- a/pkg/pgmodel/querier/common.go
+++ b/pkg/pgmodel/querier/common.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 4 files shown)


bitwalker (approved):
LGTM!

cevian (approved):

Harkishen-Singh (approved):

Issue #1092, opened 2022-01-28
Trying to insert spans with end_time < start_time


jgpruitt:
I'm seeing this error in the promscale logs in the opentelemetry-demo. I'm not sure if the opentelemetry collector is sending spans this way, or if we're doing something to accidentally put the start_time after the end_time. We should probably do something with these violations in the Go code rather than rely on the database to catch them.

```
opentelemetry-demo-promscale-1  | level=error ts=2022-01-28T16:50:35.941Z caller=runner.go:50 msg="error in GRPC call" err="error sending trace batches: ERROR: new row for relation \"_hyper_1_5_chunk\" violates check constraint \"span_check\" (SQLSTATE 23514)"
```

jgpruitt:
Postgres timestamptz has microsecond resolution while the Span struct in Go is using nanosecond resolution

PR #1096, opened 2022-01-31
Ensure that span start <= end

jgpruitt:
## Description

I have seen instances in which promscale attempts to insert spans that have an end time < start time. There is a database check constraint that catches this error, but it results in dropped spans. This patch catches the issue in Go and swaps the start and end times so that the spans will be inserted successfully.

[Issue](https://github.com/timescale/promscale/issues/1092)

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ We use the following categories for changes:
 ### Added
 - Add Prometheus metrics support for Tracing [#1102]
 
+### Fixed
+- Fix spans with end < start. Start and end are swapped in this case. [#1096]
+
 ## [0.9.0] - 2022-02-02
 
 ### Added
--- a/pkg/pgmodel/ingestor/trace/writer.go
+++ b/pkg/pgmodel/ingestor/trace/writer.go
@@ -276,8 +276,16 @@ func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces)
 
 				eventTimeRange := getEventTimeRange(span.Events())
 
-				if maxEndTimestamp < uint64(span.EndTimestamp()) {
-					maxEndTimestamp = uint64(span.EndTimestamp())
+				// postgresql timestamptz only has microsecond precision while time.Time has nanosecond precision
+				start := span.StartTimestamp().AsTime().Truncate(time.Microsecond)
+				end := span.EndTimestamp().AsTime().Truncate(time.Microsecond)
+				// make sure start <= end
+				if end.Before(start) {
+					start, end = end, start
+				}
+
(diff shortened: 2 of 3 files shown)


antekresic (approved):

JamesGuthrie:
@jgpruitt do we know why we're receiving spans with end < start? I wonder if it's safe to just swap start and end, it could end up hiding something weird/buggy in the user's tracing implementation.

jgpruitt:
@JamesGuthrie I don't know why yet. I want to try to track it down later today.

I think these are our possible solutions:
1. Let the database catch `end < start` which will rollback the transaction and drop a whole batch of spans on the floor. This is what we are doing now. I don't think we want to do this.
2. Remove the constraint and allow the bad data to be inserted. Then, all your queries/dashboards may need logic to deal with this bad data.
3. Discard the span with `end < start`. This would break the tree structure of the trace though.
4. Attempt to "fix" the data. Either swap start and end, or set end to start, or set start to end.

IMO, 4 is the best approach, but they're all icky. WDYT?

jgpruitt:
@JamesGuthrie I've spent a couple of hours trying to track down the source of the `end < start` spans. I haven't found the source of the errors yet. I bypassed the opentelemetry-collector by sending my traces directly to promscale, and the errors kept showing up. So, I don't think it is the opentelemetry-collector. I removed the constraint so the spans would get inserted, but nothing about the spans lent any clues.

I *think I'm using the Python tracing libraries correctly. I referenced multiple tutorials.

Harkishen-Singh (changes requested):
I think this can be an issue upstream. Based on the previous comment, it seems that this happened due to python lib, so I think we should open an issue there and report this.

I am not confident in making our own fixes as this in other words is hiding the cause, which is basically data corruption as the user will never expect this. Hence, I don't think we should have this swapping fix, rather drop the entire span batch. We should be strict in ingesting data to avoid problems down the road.

niksajakovljevic:
I agree with @Harkishen-Singh. I think we should try to get to the bottom of this. Does it only happen with Python libs (seems like race condition? ). `start > end` seems like a major data corruption. Imho swapping it makes it harder to reason about semantics of `start` and `end` and we might open the door for similar bugs to go undetected. I think it's great that you caught this @jgpruitt and that we have DB constraint in place!

jgpruitt:
I created another issue to track my efforts at identifying a source. #1127

IMO, having this "fix" in place is of benefit in case of future bugs

cevian (approved):
I like this solution as it follows the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).

PR #1105, opened 2022-02-03
Refactor out RemoteReadQuerier interface

JamesGuthrie:
## Description

The Querier interface was a weird mixture of "method to query" and
"methods to get things with which to query". This change moves "method
to query" into its own interface.


antekresic (approved):
I'm not sure how worthwhile this change is but its definitely an improvement.

If we want to refactor, that Querier interface might benefit from interface composition instead of doing it like we are right now but that requires a bit more involvement to get right.

PR #1110, opened 2022-02-04
New migration logic to use extension

jgpruitt:
## Description

We want to use the existing migration logic up until version X. The first time we migrate past X, we want to switch to new migration logic in which the extension is required and it handles the database changes.

## Merge requirements


JamesGuthrie (commented):
Looks good from a high-level point of view. I need to take another look when I have a bit more mental capacity.

pkg/pgmodel/migrate.go
```diff
@@ -5,336 +5,175 @@
...
-	}
-	defer func() {
-		_ = tx.Rollback(context.Background())
-	}()
-
-	// No version in DB.
-	if dbVersion.Compare(semver.Version{}) == 0 {
-		if err = t.execMigrationDir(tx, preinstallScripts); err != nil {
+		// new way of doing migrations via extension upgrades only
+		extOptions.Install = true // TODO decide whether it is appropriate to force this to true
```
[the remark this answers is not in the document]
jgpruitt: I like this idea.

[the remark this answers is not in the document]
jgpruitt: wait. why would we introduce AND deprecate `-startup.{install,upgrade}-promscale-extension` in the same release? or am I confused?

pkg/pgmodel/common/extension/extension.go
```diff
@@ -141,8 +141,8 @@ func checkTimescaleDBVersion(conn *pgx.Conn) error {
 	return nil
 }
 
-func checkPromscaleExtensionVersion(conn *pgx.Conn, migrationFailedDueToLockError bool, extOptions ExtensionMigrateOptions) error {
-	currentVersion, newVersion, err := extensionVersions(conn, "promscale", version.ExtVersionRange, version.ExtVersionRangeString, extOptions)
+func checkPromscaleExtensionVersion(conn *pgx.Conn, migrationFailedDueToLockError bool, allowPreRelease bool) error {
+	currentVersion, newVersion, err := extensionVersions(conn, "promscale", version.ExtVersionRange, version.ExtVersionRangeString, allowPreRelease)
 	if err != nil {
 		ExtensionIsInstalled = false
 		if err == errors.ErrExtUnavailable {
```
cevian, line 148: This should be deleted right? The extension is now mandatory.

pkg/pgmodel/common/extension/extension.go
```diff
@@ -59,19 +59,19 @@ func InstallUpgradePromscaleExtensions(db *pgx.Conn, extOptions ExtensionMigrate
...
+	if err = CheckExtensionsVersion(db, false, extOptions.UpgradePreRelease); err != nil {
 		return false, fmt.Errorf("error encountered while migrating extension: %w", err)
 	}
 
 	return ExtensionIsInstalled, nil
 }
 
 // CheckVersions is responsible for verifying the version compatibility of installed Postgresql database and extensions.
-func CheckVersions(conn *pgx.Conn, migrationFailedDueToLockError bool, extOptions ExtensionMigrateOptions) error {
+func CheckVersions(conn *pgx.Conn, migrationFailedDueToLockError bool, allowPreRelease bool) error {
```
cevian, line 70: Why this change in API all over the place?

pkg/pgmodel/common/extension/extension.go
```diff
@@ -141,8 +141,8 @@ func checkTimescaleDBVersion(conn *pgx.Conn) error {
 	return nil
 }
 
-func checkPromscaleExtensionVersion(conn *pgx.Conn, migrationFailedDueToLockError bool, extOptions ExtensionMigrateOptions) error {
-	currentVersion, newVersion, err := extensionVersions(conn, "promscale", version.ExtVersionRange, version.ExtVersionRangeString, extOptions)
+func checkPromscaleExtensionVersion(conn *pgx.Conn, migrationFailedDueToLockError bool, allowPreRelease bool) error {
+	currentVersion, newVersion, err := extensionVersions(conn, "promscale", version.ExtVersionRange, version.ExtVersionRangeString, allowPreRelease)
 	if err != nil {
 		ExtensionIsInstalled = false
```
cevian, line 147: we should remove this global if possible. The extension is now mandatory...

pkg/pgmodel/migrate.go
```diff
@@ -5,336 +5,186 @@
...
 
-func NewMigrator(db *pgx.Conn, sqlFiles http.FileSystem, toc map[string][]string) *Migrator {
-	return &Migrator{db: db, sqlFiles: sqlFiles, toc: toc}
-}
-
-func (t *Migrator) Migrate(appVersion semver.Version) error {
-	if err := ensureVersionTable(t.db); err != nil {
-		return fmt.Errorf("error ensuring version table: %w", err)
-	}
+	if appSemver.LT(firstAppVersionWithNewMigration) {
```
cevian, line 88: do we really need this branch? Will we merge this PR before we are ready to release the new migration system? If not, then why keep this branch around?

pkg/pgmodel/common/extension/extension.go
cevian: Also need to change line 165 below too

pkg/pgmodel/migrate.go
```diff
@@ -5,336 +5,186 @@
...
-	return nil
-}
 
-type Migrator struct {
-	db       *pgx.Conn
-	sqlFiles http.FileSystem
-	toc      map[string][]string
-}
+	firstAppVersionWithNewMigration := FirstAppVersionWithNewMigration()
+	appSemver := semver.MustParse(appVersion.Version)
```
cevian, line 86: I think we should be a bit conservative here and double check the extension we need is first available before doing any of this migration logic. Something like
```
_, _, err := extensionVersions(conn, "promscale", version.ExtVersionRange, version.ExtVersionRangeString, extOptions)
// check for err. If there is an err that means that the right extension version is unavailable
```

pkg/pgmodel/old_migrator.go
```diff
@@ -0,0 +1,438 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
cevian, line 1: is this the old system verbatim or were there any logic changes. Hard to tell from this diff

pkg/runner/flags.go
```diff
@@ -127,8 +136,12 @@ func ParseFlags(cfg *Config, args []string) (*Config, error) {
 	fs.BoolVar(&skipMigrate, "startup.skip-migrate", false, "Skip migrating Promscale SQL schema to latest version on startup.")
 
 	fs.BoolVar(&cfg.UseVersionLease, "startup.use-schema-version-lease", true, "Use schema version lease to prevent race conditions during migration.")
-	fs.BoolVar(&cfg.InstallExtensions, "startup.install-extensions", true, "Install TimescaleDB, Promscale extension.")
-	fs.BoolVar(&cfg.UpgradeExtensions, "startup.upgrade-extensions", true, "Upgrades TimescaleDB, Promscale extensions.")
+	fs.BoolVar(&cfg.InstallExtensions, "startup.install-extensions", true, fmt.Sprintf("(DEPRECATED) Will be removed in version %s. Install TimescaleDB, Promscale extensions.", firstAppVersionWithNewMigration))
+	fs.BoolVar(&cfg.UpgradeExtensions, "startup.upgrade-extensions", true, fmt.Sprintf("(DEPRECATED) Will be removed in version %s. Upgrades TimescaleDB, Promscale extensions.", firstAppVersionWithNewMigration))
+	fs.BoolVar(&cfg.InstallTimescaleDBExtension, "startup.install-timescaledb-extension", true, "Install TimescaleDB extension.")
```
cevian, line 141: is `startup.install-extension.timescaledb` a better name? WDYT?

cevian (changes requested):
Biggest question is do we need this merged before the new migration change? I think if we don't then we can simplify the logic here quite a bit.

jgpruitt:
moving to new PR and branch https://github.com/timescale/promscale/pull/1132

PR #1129, opened 2022-02-09
Add failing offset pushdown test

JamesGuthrie:
## Description

## Merge requirements


pkg/pgmodel/querier/query_builder.go
```diff
@@ -242,7 +242,29 @@ func canAttemptPushdown(metadata *promqlMetadata) bool {
...
+	}
+
+	// We can't handle subqueries in pushdowns
+	if hasSubquery(path) {
+		return false
+	}
+
+	// We can't handle offsets in pushdowns
+	vs, isVectorSelector := queryHints.CurrentNode.(*parser.VectorSelector)
+	if isVectorSelector && (vs.Offset != 0 || vs.OriginalOffset != 0) {
```
niksajakovljevic, line 263: Base on Prometheus docs `OriginalOffset` is actual offset set from the query so it seems no need to check `Offset` as that is just derived during query execution.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -242,7 +242,29 @@ func canAttemptPushdown(metadata *promqlMetadata) bool {
...
+		return false
+	}
+
+	// We can't handle subqueries in pushdowns
+	if hasSubquery(path) {
+		return false
+	}
+
+	// We can't handle offsets in pushdowns
+	vs, isVectorSelector := queryHints.CurrentNode.(*parser.VectorSelector)
```
niksajakovljevic, line 262: Shouldn't we handle `MatrixSelector` as well?

pkg/tests/end_to_end_tests/promql_query_endpoint_test.go
```diff
@@ -2506,6 +2506,10 @@ func TestPromQLQueryEndpoint(t *testing.T) {
 			name:  "two pushdowns, same metric different matchers",
 			query: `sum(rate(metric_2{foo = "bar"}[5m]))/sum(rate(metric_2[5m]))`,
 		},
+		{
+			name:  "pushdown with offset",
```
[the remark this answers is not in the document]
JamesGuthrie: As these tests are primarily used to compare the result that Promscale delivers compared to Prometheus, IMO the description of the test should be independent of what Promscale does or doesn't do. I've updated the test name to `delta function over range selector with offset`.

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: I don't think so (but I might be missing something). A MatrixSelector doesn't have an offset, its fields are `Range` and `VectorSelector`. The contained `VectorSelector` may have an offset, and will be handled by this logic.

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: done

pkg/pgmodel/querier/query_builder.go
```diff
@@ -242,7 +242,29 @@ func canAttemptPushdown(metadata *promqlMetadata) bool {
...
+
+	// We can't do pushdowns without the extension
+	if !extension.ExtensionIsInstalled {
+		return false
+	}
+
+	// We can't handle subqueries in pushdowns
+	if hasSubquery(path) {
+		return false
+	}
```
antekresic, line 259: The correct Go style for things like this is to use a switch statement, would make the code more compact and easier to read.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -242,7 +242,29 @@ func canAttemptPushdown(metadata *promqlMetadata) bool {
 	path := metadata.path // PromQL AST.
 	queryHints := metadata.queryHints
 	selectHints := metadata.selectHints
-	return extension.ExtensionIsInstalled && queryHints != nil && !hasSubquery(path) && selectHints != nil
+
+	// We can't push down without hints
```
antekresic, line 246: I'm sorry, I have to do this. :grin: 
```suggestion
	// We can't push down without hints.
```

antekresic (approved):
Couple of nits and a question: do we want to add a TODO to add in the offset support if its something that might be added in the future?

Other than that, I have no problems with the PR itself. Great work :+1:

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: TIL, thanks!

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: No problem! Fixed.

JamesGuthrie:
> do we want to add a TODO to add in the offset support if its something that might be added in the future?

https://github.com/timescale/promscale/issues/1130

pkg/pgmodel/querier/query_builder.go
niksajakovljevic: Hmm `MatrixSelector` does not embed `VectorSelector` but `Expr` which may contain `VectorSelector`. So I guess in case of `MatrixSelector` it would not successfully type assert.

niksajakovljevic (changes requested):
I believe we need to handle `MatrixSelector` explicitly. Check my related comment.

pkg/pgmodel/querier/query_builder.go
niksajakovljevic: In other words in case of `MatrixSelector` we would need to type assert on it's `. VectorSelector` field

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: The plot thickens: this change was not good. We do actually need to validate that both `Offset` and `OriginalOffset` are zero.

pkg/pgmodel/querier/query_builder.go
JamesGuthrie: I've restructured the code by folding the `canAttemptPushdown` into `tryPushDown`. This makes the conditions that we're asserting clearer. In particular, it's more visible that we only attempt pushdowns on `VectorSelector`, which handles your concern here.

niksajakovljevic (approved):

PR #1145, opened 2022-02-14
Log deprecated flag usage

antekresic:
## Description

This PR contains a couple of changes that are connected to the renaming process of old flags. Main feature is the addition of logging usage of deprecated flags with recommendations on how to migrate them.

I'm unsure if any of this should be added to the changelog, perhaps only the change to remove the deprecated flags in 0.11.0 version. I would like some input on that before adding it.

## Merge requirements

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -43,6 +43,17 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	return cfg
 }
 
+// InitDefault is used to start the logger with sane defaults before we can configure it.
+// It's useful in instances where we want to log stuff before the configuration
+// has been successfully parsed. Calling Init function later on overrides this.
+func InitDefault() {
+	l := level.NewFilter(
+		log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)),
+		level.AllowInfo(),
+	)
+	logger = log.With(l, "ts", timestampFormat, "caller", log.Caller(4))
+}
+
 // Init starts logging given the configuration. By default, it uses logfmt format
 // and minimum logging level.
 func Init(cfg Config) error {
--- a/pkg/runner/flags.go
+++ b/pkg/runner/flags.go
@@ -46,58 +46,62 @@ type Config struct {
 	StartupOnly                 bool
 }
 
-var flagAliases = map[string][]string{
-	"auth.tls-cert-file":                {"tls-cert-file"},
-	"auth.tls-key-file":                 {"tls-key-file"},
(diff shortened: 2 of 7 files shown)


pkg/log/log.go
```diff
@@ -43,6 +43,17 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	return cfg
 }
 
+// InitDefault is used to start the logger with sane defaults before the we can configure it.
```
JamesGuthrie, line 46:
```suggestion
// InitDefault is used to start the logger with sane defaults before we can configure it.
```

JamesGuthrie (approved):
Looks good, one minor nit.

pkg/log/log.go
antekresic: :+1:

pkg/runner/flags.go
```diff
@@ -225,3 +246,52 @@ func validate(cfg *Config) error {
...
+			}
+
+			if !set {
+				panic(fmt.Sprintf("trying to set an flag alias for a flag that is missing: %s", f.Name))
+			}
+		}
+	})
+}
+
+func flagAliasesUsed(fs *flag.FlagSet) map[string]string {
```
Harkishen-Singh, line 268:
```suggestion
func deprecatedFlagsUsed(fs *flag.FlagSet) map[string]string {
```
Is this better?

pkg/runner/flags.go
```diff
@@ -225,3 +246,52 @@ func validate(cfg *Config) error {
...
+
+			switch {
+			case f.Name == "migrate":
+				// Migrate flag is a special case since it's not just a rename
+				// of an old flag. We handle it separately when logging
+				// use of deprecated flags.
+				return
+			case err != nil:
+				fallthrough
+			case i != 1:
```
Harkishen-Singh, line 287: shouldn't this be `i < 1` as more than one deprecated flags can be passed?

pkg/util/util.go
```diff
@@ -12,8 +12,7 @@ import (
 )
 
 const (
-	PromNamespace     = "promscale"
-	aliasDescTemplate = "Alias for -%s flag. "
+	PromNamespace = "promscale"
 )
```
Harkishen-Singh, line 16: Nit

```
const PromNamespace = "promscale"
```

Harkishen-Singh (approved):
A question and minor nits

pkg/runner/flags.go
antekresic: Actually no, an alias can have only one flag that it is aliasing.

PR #1148, opened 2022-02-14
Add support for OTEL collector endpoint in Promscale tracing telemetry exporter

antekresic:
## Description

In addition to Jaeger, now Promscale can send its telemetry data to OTEL
collector endpoint. Since this is an endpoint that Promscale itself exposes,
we can actually ingest tracing data from Promscale into Promscale itself.

## Merge requirements


JamesGuthrie (approved):

pkg/tracer/trace.go
```diff
@@ -33,20 +50,70 @@ func Default() trace.Tracer {
...
+			return nil, fmt.Errorf("failed to create trace exporter: %w", err)
+		}
+
 		// Always be sure to batch in production.
-		tracesdk.WithBatcher(exp),
+		opts = append(opts, tracesdk.WithBatcher(exp))
+	}
+
+	if len(opts) == 0 {
+		return nil, fmt.Errorf("no traces collector endpoints provided, please use the telemetry trace endpoint flags to set at least one")
```
cevian, line 104:
```suggestion
		return nil, fmt.Errorf("tracing telemetry requires setting either the open telemetry collector endpoint or the jaeger endpoint")
```

pkg/runner/runner.go
```diff
@@ -79,7 +79,7 @@ func Run(cfg *Config) error {
 		tput.InitWatcher(cfg.ThroughputInterval)
 	}
 
-	if cfg.TracerCfg.JaegerCollectorEndpoint != "" {
+	if cfg.TracerCfg.SamplingRatio > 0 {
```
cevian, line 82: hmm, not sure how much I like this. Enabling tracing would now require setting both an endpoint and a sampling ration by the user. I think I'd prefer a sampling ratio of 1.0 by default and enabling tracing if either the jaeger or the Otel collector endpoints are set.

antekresic: Gotcha, valid point. Will update.

docs/cli.md
```diff
@@ -54,8 +54,11 @@ You can also find information on flags with `promscale_<version> -help`.
...
+| telemetry.trace.otel-endpoint | string | "" (empty) | OpenTelemetry tracing collector GRPC URL endpoint to send telemetry to. (i.e. otel-collector:4317) |
+| telemetry.trace.otel-tls-cert-file | string | "" (empty) | TLS Certificate file used for client authentication against the OTEL tracing collector GRPC endpoint, leave blank to disable TLS. |
+| telemetry.trace.otel-tls-key-file | string | "" (empty) | TLS Key file for client authentication against the OTEL tracing collector GRPC endpoint, leave blank to disable TLS. |
 | telemetry.trace.jaeger-endpoint | string | "" (empty) | Jaeger tracing collector thrift HTTP URL endpoint to send telemetry to (e.g. https://jaeger-collector:14268/api/traces). |
-| telemetry.trace.sample-ratio | float | 0.0 | Trace sampling ratio, amount of spans to send to collector. Valid values from 0.0 (none) to 1.0 (all). |
+| telemetry.trace.sample-ratio | float | 1.0 | Trace sampling ratio, amount of spans to send to collector. Valid values from 0.0 (none) to 1.0 (all). |
```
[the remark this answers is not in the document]
antekresic: Silently convert to 1.0. I'm fine with that since the OTEL library does the same thing and setting to 1.1 or 2.0 isn't really a good reason panic out of execution IMO.

Only thing we might do is log that we are clamping to 1.0 if the value is over. Same with <0.0 I guess.

cevian (approved):

PR #1152, opened 2022-02-15
Increase visibility into trace ingest

niksajakovljevic:
Add few more metrics that will give us more insight into ingest.
Mainly we want to see how big our insert batches are and also the time it
takes to insert various trace related objects. This should help
us in finding weak spots.

### DIFF
--- a/pkg/api/write.go
+++ b/pkg/api/write.go
@@ -19,6 +19,7 @@ import (
 	"github.com/timescale/promscale/pkg/api/parser"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel/ingestor"
+	"github.com/timescale/promscale/pkg/prompb"
 	"github.com/timescale/promscale/pkg/tracer"
 	"github.com/timescale/promscale/pkg/util"
 )
@@ -261,8 +262,14 @@ func ingest(
 			return false
 		}
 
+		// set number of received samples and metadata in case below Ingest fails
+		numSamplesTmp := uint64(getTotalSamples(req))
+		numMetadataTmp := uint64(len(req.Metadata))
+
 		numSamples, numMetadata, err = inserter.Ingest(ctx, req)
 		if err != nil {
+			numSamples = numSamplesTmp
+			numMetadata = numMetadataTmp
 			statusCode = "500"
 			log.Warn("msg", "Error sending samples to remote storage", "err", err, "num_samples", numSamples)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
@@ -295,4 +302,12 @@ func detectSnappyStreamFormat(input []byte) bool {
 	}
 
 	return true
+}
+
+func getTotalSamples(wr *prompb.WriteRequest) int {
+	total := 0
+	for _, ts := range wr.Timeseries {
+		total += len(ts.Samples)
+	}
+	return total
 }
--- a/pkg/pgmodel/ingestor/trace/batch.go
+++ b/pkg/pgmodel/ingestor/trace/batch.go
@@ -53,6 +53,7 @@ func (q sortableItems) Swap(i, j int) {
 
 //batcher queues up items to send to the DB but it sorts before sending
 //this avoids deadlocks in the DB. It also avoids sending the same items repeatedly.
+// batcher is not thread safe
 type batcher struct {
 	batch map[batchItem]interface{}
 	cache cache
@@ -65,9 +66,14 @@ func newBatcher(cache cache) batcher {
 	}
 }
 
-// Queue adds item to the batch to be sent.
+// Queue adds item to the batch to be sent or replaces an existing item
 func (b batcher) Queue(i batchItem) {
 	b.batch[i] = nil
+}
+
+// Number of items in the batch.
+func (b batcher) Len() int {
+	return len(b.batch)
 }
 
 // SendBatch sends the batch over the DB connections and gets the ID results.
--- a/pkg/pgmodel/ingestor/trace/writer.go
+++ b/pkg/pgmodel/ingestor/trace/writer.go
@@ -126,14 +126,22 @@ func getServiceName(rSpan pdata.ResourceSpans) string {
 	return serviceName
 }
 
+var (
+	traceLabel      = prometheus.Labels{"type": "trace"}
+	traceSpanLabel  = prometheus.Labels{"type": "trace", "kind": "span"}
+	traceEventLabel = prometheus.Labels{"type": "trace", "kind": "event"}
+	traceLinkLabel  = prometheus.Labels{"type": "trace", "kind": "link"}
+)
+
 func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces) error {
 	startIngest := time.Now() // Time taken for complete ingestion => Processing + DB insert.
 	code := "500"
-	metrics.IngestorActiveWriteRequests.With(prometheus.Labels{"type": "trace", "kind": "span"}).Inc()
+	metrics.IngestorActiveWriteRequests.With(traceSpanLabel).Inc()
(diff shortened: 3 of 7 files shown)


pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -126,12 +126,18 @@ func getServiceName(rSpan pdata.ResourceSpans) string {
 	return serviceName
 }
 
+var traceLabel = prometheus.Labels{"type": "trace"}
```
Harkishen-Singh, line 129: i think we should unify all vars into one to make it more go-ish.

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -126,12 +126,18 @@ func getServiceName(rSpan pdata.ResourceSpans) string {
...
+var traceSpanLabel = prometheus.Labels{"type": "trace", "kind": "spans"}
```
Harkishen-Singh, line 130: Lets keep the kind as per the convention in metric refactoring.

```suggestion
var traceSpanLabel = prometheus.Labels{"type": "trace", "kind": "span"}
```

lets also change `tags` to `tag` somewhere down.

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -207,6 +213,10 @@ func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces)
 			}
 		}
 	}
+
+	metrics.InsertBatchSize.With(prometheus.Labels{"type": "trace", "kind": "instrumentation"}).Observe(float64(instrLibBatch.b.Len()))
```
Harkishen-Singh, line 217: I think this is now `instrumentation_lib` as per metric renaming.

```suggestion
	metrics.InsertBatchSize.With(prometheus.Labels{"type": "trace", "kind": "instrumentation_lib"}).Observe(float64(instrLibBatch.b.Len()))
```

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -207,6 +213,10 @@ func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces)
...
+	metrics.InsertBatchSize.With(prometheus.Labels{"type": "trace", "kind": "operation"}).Observe(float64(operationBatch.b.Len()))
+	metrics.InsertBatchSize.With(prometheus.Labels{"type": "trace", "kind": "tags"}).Observe(float64(tagsBatch.b.Len()))
```
Harkishen-Singh, line 219:
```suggestion
	metrics.InsertBatchSize.With(prometheus.Labels{"type": "trace", "kind": "tag"}).Observe(float64(tagsBatch.b.Len()))
```

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -313,29 +323,43 @@ func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces)
...
+	metrics.InsertBatchSize.With(traceEventLabel).Observe(float64(eventBatch.Len()))
+	metrics.InsertBatchSize.With(traceLinkLabel).Observe(float64(linkBatch.Len()))
+
 	start := time.Now()
-	if err := t.sendBatches(ctx, eventBatch, linkBatch, spanBatch); err != nil {
-		return fmt.Errorf("error sending trace batches: %w", err)
+	if err := t.sendBatch(ctx, eventBatch); err != nil {
+		return fmt.Errorf("error sending trace event batch: %w", err)
+	}
+	metrics.InsertDuration.With(traceEventLabel).Observe(time.Since(start).Seconds())
```
Harkishen-Singh, line 334: when this PR is rebased, there is a label `code`, that denotes success or failure. success has `2xx` and failure has `500`. We should make this and the below `.Observe()` as deferable and based on the condition, duration for a success series of a failure series should be set.

Harkishen-Singh (changes requested):
Some changes are required after this PR is merged with #1113

pkg/pgmodel/ingestor/trace/batch.go
```diff
@@ -65,11 +66,16 @@ func newBatcher(cache cache) batcher {
 	}
 }
 
-// Queue adds item to the batch to be sent.
+// Queue adds item to the batch to be sent or replaces an existing item
 func (b batcher) Queue(i batchItem) {
```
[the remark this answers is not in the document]
niksajakovljevic: I guess you haven't checked the implementation? This code is adding `batchItem` so makes no sense to rename the function to `Clear`

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -126,14 +126,22 @@ func getServiceName(rSpan pdata.ResourceSpans) string {
...
+	traceEventLabel = prometheus.Labels{"type": "trace", "kind": "event"}
+	traceLinkLabel  = prometheus.Labels{"type": "trace", "kind": "link"}
+)
+
 func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces) error {
 	startIngest := time.Now() // Time taken for complete ingestion => Processing + DB insert.
 	code := "500"
-	metrics.IngestorActiveWriteRequests.With(prometheus.Labels{"type": "trace", "kind": "span"}).Inc()
+	metrics.IngestorActiveWriteRequests.With(traceSpanLabel).Inc()
+	metrics.InsertablesReceivedTotal.With(traceSpanLabel).Add(float64(traces.SpanCount()))
```
Harkishen-Singh, line 140: we should not have this, as its already caught by 354 `metrics.IngestorItems.With(traceSpanLabel).Add(float64(traces.SpanCount()))`

pkg/pgmodel/metrics/ingest.go
```diff
@@ -167,6 +167,25 @@ var (
 			Help:      "Total number of requests to ingestor.",
 		}, []string{"type", "code"},
 	)
+	InsertablesReceivedTotal = prometheus.NewCounterVec(
```
Harkishen-Singh, line 170: We dont need this, since `IngestorItems` handles it.

Harkishen-Singh: If we add this, then this has to be done across samples as well, which it does not. It earlier used to, but there wasn't a need and hence it was removed.

Harkishen-Singh (changes requested):
We should remove `InsertablesReceivedTotal`. Then its good to merge.

pkg/pgmodel/ingestor/trace/writer.go
niksajakovljevic: These two things are not semantically the same. What I wanted is to get number of spans received - doesn't mean they are ingested. Line 354 will only be executed in case there was no failure meaning that we will miss the numbers in case of error.
I will move `IngestorItems` in place where `InsertablesReceivedTotal` was used.
I also noticed that we are not consistent with using `IngestorItems`. For metrics ingestion we increase it even if an error occurs, however for tracing we were not doing that.

Harkishen-Singh (approved):

pkg/api/write.go
```diff
@@ -260,6 +261,9 @@ func ingest(
 			ingestor.FinishWriteRequest(req)
 			return false
 		}
+		// set number of received samples and metadata in case below Ingest fails
+		numSamples = uint64(getTotalSamples(req))
+		numMetadata = uint64(len(req.Metadata))
```
antekresic, line 266: Won't this be always be overwritten by the Ingest below? Am I missing something?

niksajakovljevic: Good catch!

niksajakovljevic: Funny thing is that I added a comment stating that we prevent in case of `Ingest` failing but accidentally put it before the error is returned :) Note to myself: don't work on PRs when it's a public holiday :)

antekresic (approved):

PR #1165, opened 2022-02-17
Fix ingested_samples_total telemetry

Harkishen-Singh:

## Description

This PR fixes ingested_samples telemetry by using another metric name for items (aka insertables) received.

TODO:

- [x] Fix api/write_test.go
- [x] Run ingestion, querying for samples and traces to verify metrics are correctly implemented
- [x] Verify telemetry collection

## Merge requirements

### DIFF
--- a/pkg/api/write.go
+++ b/pkg/api/write.go
@@ -54,14 +54,14 @@ func Write(
 	inserter ingestor.DBInserter,
 	dataParser *parser.DefaultParser,
 	elector *util.Elector,
-	callBack func(code string, durationSeconds, numSamples, numMetadata float64),
+	updateMetrics func(code string, duration, receivedSamples, receivedMetadata float64),
 ) http.Handler {
 	wh := writeHandler{}
 	wh.addStages(
 		validateWriteHeaders,
 		checkLegacyHA(elector, metrics),
 		decodeSnappy,
-		ingest(inserter, dataParser, callBack),
+		ingest(inserter, dataParser, updateMetrics),
 	)
 	return wh.handler()
 }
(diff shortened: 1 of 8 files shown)


niksajakovljevic (changes requested):
Hmmm...This doesn't seem to be inline with `IngestorItems` being items that we've received and `IngestorInsertablesIngested` being items that we've inserted into DB. Let's discuss this again so we are all on the same page.

pkg/api/write.go
```diff
@@ -54,8 +54,12 @@ func Write(
 	inserter ingestor.DBInserter,
 	dataParser *parser.DefaultParser,
 	elector *util.Elector,
-	callBack func(code string, durationSeconds, numSamples, numMetadata float64),
+	callBack func(code string, duration float64, received, ingested items),
```
niksajakovljevic, line 57: Consider naming this function to something more specific :)

pkg/pgmodel/ingestor/copier.go
```diff
@@ -446,8 +446,6 @@ func insertSeries(ctx context.Context, conn pgxconn.PgxConn, reqs ...copyRequest
 			registerDuplicates(numRowsExpected - insertedRows)
 		}
 	}
-	metrics.IngestorInsertablesIngested.With(prometheus.Labels{"type": "metric", "kind": "sample"}).Add(float64(totalSamples))
```
niksajakovljevic, line 449: I'd rather update metrics in the same place where ingest happens than via callbacks ( unless there is a special reason). Mostly for keeping things simple

niksajakovljevic (approved):

antekresic (approved):
lgtm

PR #1177, opened 2022-02-21
Register `promscale_ingest_channel_len` metric and make it a Gauge.

Harkishen-Singh:

## Description

## Merge requirements

### DIFF
--- a/pkg/pgmodel/ingestor/copier.go
+++ b/pkg/pgmodel/ingestor/copier.go
@@ -484,15 +484,3 @@ func insertMetadata(conn pgxconn.PgxConn, reqs []pgmodel.Metadata) (insertedRows
 	metrics.IngestorInsertDuration.With(prometheus.Labels{"type": "metric", "subsystem": "", "kind": "exemplar"}).Observe(time.Since(start).Seconds())
 	return insertedRows, nil
 }
-
-var copierChannelMutex sync.Mutex
-
-func setCopierChannelToMonitor(toSamplesCopiers chan readRequest) {
-	copierChannelMutex.Lock()
-	defer copierChannelMutex.Unlock()
-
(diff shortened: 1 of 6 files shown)


pkg/pgmodel/ingestor/copier.go
```diff
@@ -485,14 +486,30 @@ func insertMetadata(conn pgxconn.PgxConn, reqs []pgmodel.Metadata) (insertedRows
 	return insertedRows, nil
 }
 
-var copierChannelMutex sync.Mutex
+var (
+	copierChannelMutex            sync.Mutex
+	samplesCopierChannelToMonitor chan readRequest
+	// Declare and register here since we need readRequest struct which should not be defined in
+	// pgmodel/metric.
+	ingestorChannelLenCopier = prometheus.NewGaugeFunc(
```
niksajakovljevic, line 494: I failed to understand why can't we just have one gauge metric `IngestorChannelLen` (defined in ingest.go) and use it for both batcher and copier with different label values? That approach would make it for less code and would not require to have a global var `samplesCopierChannelToMonitor`

niksajakovljevic: Similar to what's done here https://github.com/timescale/promscale/pull/1177/files#diff-e157d1ebaa83fd2018edfff00e940023c70044b7df48451a3926d0c36af61826L494

Harkishen-Singh: The recent commit updates this. But basically, one metric (copier channel_len) needs a callback to be updated when prometheus pings the /metric (hence we need NewGaugeFunc), and the other done for metric batcher will be used irrespective of prometheus calls (so NewGauge will work).

antekresic (approved):

niksajakovljevic (approved):

PR #1180, opened 2022-02-22
Log warning when remote read request is cancelled midway

antekresic:
## Description
  Previously, we were trying to update the header but that cannot be
  done in this instance since we are already starting to write the
  request body. Only valid way of handling these situations is to
  log the error and finish the request.

## Merge requirements

### DIFF
--- a/pkg/api/read.go
+++ b/pkg/api/read.go
@@ -87,8 +87,11 @@ func Read(config *Config, reader querier.Reader, metrics *Metrics, updateMetrics
 
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
-			statusCode = "500"
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			// Most likely the request was cancelled from client side.
+			// We use a non-standard code so we can distinguish from actual
+			// internal server errors.
+			statusCode = "499"
+			log.Warn("msg", "Error writing HTTP response", "err", err)
 			return
 		}
 		statusCode = "2xx"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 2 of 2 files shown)


pkg/api/read.go
```diff
@@ -88,7 +88,7 @@ func Read(config *Config, reader querier.Reader, metrics *Metrics, updateMetrics
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
 			statusCode = "500"
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			log.Warn("msg", "Error writing HTTP response", "err", err)
```
Harkishen-Singh, line 91: This should be logged as error i think.

Harkishen-Singh: And do we respond back in HTTP in an error case?

sumerman (commented):
A potentially misguided suggestion 😅

pkg/api/read.go
```diff
@@ -88,7 +88,7 @@ func Read(config *Config, reader querier.Reader, metrics *Metrics, updateMetrics
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
 			statusCode = "500"
```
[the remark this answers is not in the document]
antekresic: Might be useful to do this so we can ignore these things in alerts.

pkg/api/read.go
antekresic: In keeping with previously made assumptions, using Warn here as there is a similar situation above:

```log.Warn("msg", "Error executing query", "query", req, "storage", "PostgreSQL", "err", err)```

Not sure about your HTTP comment. Client aborted the request, whatever we do with the HTTP response its not going to reach them.

pkg/api/read.go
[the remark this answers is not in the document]
antekresic: Updated.

Harkishen-Singh (approved):

sumerman (approved):

PR #1188, opened 2022-02-23
Remove CHECK constraints on span table

niksajakovljevic:
It seems that CHECK constrains come with a huge CPU
penalty for the database. We can instead move checks to Promscale code.
Benchmark shows around 35% improvement in CPU usage.

Note: Currently TimescaleDB has a limitation and does not support dropping constraints on compressed hypertables. We need this fixed in TimescaleDB before upgrade scripts could do the work.


cevian:
This should not be merged. We need to wait for the extension stuff to be done. The SQL code is currently frozen

cevian:
@niksajakovljevic this need to be re-implemented in the extension

niksajakovljevic:
In the context of recent refactoring here are two related PRs:
https://github.com/timescale/promscale_extension/pull/188
https://github.com/timescale/promscale/pull/1302

PR #1233, opened 2022-03-11
Verify telemetry engine as `nil` if TimescaleDB is not installed.

Harkishen-Singh:

Fixes: https://github.com/timescale/o11y-team-applications/issues/145

## Description

Same as the heading.

## Merge requirements

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -79,6 +79,9 @@ func NewEngine(conn pgxconn.PgxConn, uuid [16]byte, promqlQueryable promql.Query
 }
 
 func isTelemetryOff(conn pgxconn.PgxConn) (bool, error) {
+	if !util.IsTimescaleDBInstalled(conn) {
+		return true, nil
+	}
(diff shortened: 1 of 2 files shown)


JamesGuthrie (changes requested):
Two things:

1. Tests aren't passing.
2. I also added skips to `TestTelemetrySQLStats` and `TestPromQLBasedTelemetry`. Should those not also be refactored in the same way?

Harkishen-Singh:
> TestTelemetrySQLStats

No, SQLStats is completely on SQL side of TimescaleDB so we always skip if there is no TimescaleDB. I feel the same way for `TestPromQLBasedTelemetry` as if the image is non-timescaleDB, then setting up client, router goes in vain as the next lines involve checking from database which again needs to be TImescaleDB.

antekresic (approved):

JamesGuthrie (approved):
I think it might make more sense to keep the skips there and add and explicit test which validates that the engine is null when timescaledb is not present.

Issue #1156, opened 2022-02-16
Refactor `RegisterMetric` in telemetry


Harkishen-Singh:
Right now, we need to pass around telemetry interface everywhere so that constructors like `trace.NewWriter`, `jaegerquery.New` can register their metrics for the telemetry engine, which is not good.

We should rather do something like ParseFlags, i.e., pass down a pointer that registers the metrics of a module, only when the telemetry engine is on.

PR #1234, opened 2022-03-11
Refactors telemetry metric registration

Harkishen-Singh:

This commit refactors the metric registration process in telemetry. Now,
the telemetry engine no longer requires to be passed down to each module
that wants to send telemetry, rather just expose a func which accepts a
telemetry engine.

The call of this exposed function happens in runner.go where all modules
that want to send telemetry are called and managed.

Fixes: #1156 

## Merge requirements

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -29,7 +29,7 @@ import (
 	"github.com/timescale/promscale/pkg/util"
 )
 
-func GenerateRouter(apiConf *Config, client *pgclient.Client, elector *util.Elector, t telemetry.Engine) (*mux.Router, error) {
+func GenerateRouter(apiConf *Config, client *pgclient.Client, elector *util.Elector) (*mux.Router, error) {
 	var writePreprocessors []parser.Preprocessor
 	if apiConf.HighAvailability {
 		service := ha.NewService(haClient.NewLeaseClient(client.Connection))
@@ -98,9 +98,7 @@ func GenerateRouter(apiConf *Config, client *pgclient.Client, elector *util.Elec
 	router.Path("/healthz").Methods(http.MethodGet).HandlerFunc(Health(healthChecker))
 	router.Path(apiConf.TelemetryPath).Methods(http.MethodGet).HandlerFunc(promhttp.Handler().ServeHTTP)
 
-	if err = jaeger.ExtendQueryAPIs(router, client.Connection, t); err != nil {
-		return nil, fmt.Errorf("error extending router with Jaeger-query APIs: %w", err)
-	}
+	jaeger.ExtendQueryAPIs(router, client.Connection)
 
 	debugProf := router.PathPrefix("/debug/pprof").Subrouter()
 	debugProf.Path("").Methods(http.MethodGet).HandlerFunc(pprof.Index)
@@ -116,14 +114,10 @@ func GenerateRouter(apiConf *Config, client *pgclient.Client, elector *util.Elec
 	debugProf.Path("/mutex").Methods(http.MethodGet).HandlerFunc(pprof.Handler("mutex").ServeHTTP)
 
 	router.Path("/debug/fgprof").Methods(http.MethodGet).HandlerFunc(fgprof.Handler().ServeHTTP)
-
-	if err = registerMetricsForTelemetry(t); err != nil {
-		log.Error("msg", "error registering metrics for telemetry", "err", err.Error())
-	}
 	return router, nil
 }
 
-func registerMetricsForTelemetry(t telemetry.Engine) error {
+func RegisterTelemetryMetrics(t telemetry.Engine) error {
 	var err error
 	if err = t.RegisterMetric(
 		"promscale_ingested_samples_total",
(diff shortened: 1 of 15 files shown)


pkg/runner/runner.go
```diff
@@ -126,12 +135,27 @@ func Run(cfg *Config) error {
...
+	}
+	if engine == nil {
+		telemetryEngine = telemetry.NewNoopEngine()
+	} else {
+		telemetryEngine = engine
+	}
+	telemetryEngine.Start()
+	defer telemetryEngine.Stop()
+
+	registerMetricsForTelemetry(telemetryEngine)
```
antekresic, line 157: With removal of line 144:
```suggestion
	if engine == nil {
		engine = telemetry.NewNoopEngine()
	} 
	engine.Start()
	defer engine.Stop()

	registerMetricsForTelemetry(engine)
```

pkg/api/router.go
```diff
@@ -116,14 +114,10 @@ func GenerateRouter(apiConf *Config, client *pgclient.Client, elector *util.Elec
...
 	router.Path("/debug/fgprof").Methods(http.MethodGet).HandlerFunc(fgprof.Handler().ServeHTTP)
-
-	if err = registerMetricsForTelemetry(t); err != nil {
-		log.Error("msg", "error registering metrics for telemetry", "err", err.Error())
-	}
 	return router, nil
 }
 
-func registerMetricsForTelemetry(t telemetry.Engine) error {
+func RegisterMetricsForTelemetry(t telemetry.Engine) error {
```
antekresic, line 120: Naming nit: `RegisterTelemetryMetrics` is shorter and more direct.

antekresic (approved):
Couple of minor things, great refactortelemetryEngine :+1:

pkg/telemetry/telemetry.go
```diff
@@ -44,13 +44,13 @@ type engineImpl struct {
...
-func NewEngine(conn pgxconn.PgxConn, uuid [16]byte, promqlQueryable promql.Queryable) (*engineImpl, error) {
+func NewEngine(conn pgxconn.PgxConn, uuid [16]byte, promqlQueryable promql.Queryable) (Engine, error) {
 	isTelemetryOff, err := isTelemetryOff(conn)
 	if err != nil {
 		log.Debug("msg", "unable to get TimescaleDB telemetry configuration. Maybe TimescaleDB is not installed", "err", err.Error())
 	}
 	if isTelemetryOff {
-		return nil, nil
+		return NewNoopEngine(), nil
 	} else {
```
antekresic, line 54: No need for this else since we are returning early.

antekresic (approved):

pkg/tests/end_to_end_tests/telemetry_test.go
```diff
@@ -32,6 +33,17 @@ func setTobsEnv(prop string) error {
 	return os.Setenv(fmt.Sprintf("TOBS_TELEMETRY_%s", prop), prop)
 }
 
+type e2eTelemetry interface {
```
niksajakovljevic, line 36: Please add some comments so it's more obvious why is this needed: `we want to manually trigger Sync for tests..`

niksajakovljevic: Also consider better naming: eg `testTelemetryEngine`

pkg/runner/runner.go
```diff
@@ -255,3 +261,23 @@ func Run(cfg *Config) error {
...
 }
+
+func registerTelemetryMetrics(client *pgclient.Client) {
+	t, err := telemetry.NewEngine(client.Connection, PromscaleID, client.Queryable())
+	if err != nil {
+		log.Debug("msg", "err creating telemetry engine", "err", err.Error())
+		return
+	}
+	t.Start()
+	defer t.Stop()
```
niksajakovljevic, line 272: This doesn't seem right. Why are we stopping the engine upon telemetry registration?

pkg/runner/runner.go
```diff
@@ -255,3 +261,23 @@ func Run(cfg *Config) error {
 	}
 	return err
 }
+
+func registerTelemetryMetrics(client *pgclient.Client) {
```
niksajakovljevic, line 265: Consider better naming: eg `startTelemetryEngine` since this code is starting the engine and registering few metrics...

pkg/runner/runner.go
niksajakovljevic: I'd recommend returning the engine instance from this function and invoking `Stop` in our shutdown/cleanup stage...

niksajakovljevic (changes requested):
Mainly the telemetry engine lifecycle seems wrong.

pkg/runner/runner.go
Harkishen-Singh: Oh yes, you are right

pkg/tests/end_to_end_tests/telemetry_test.go
Harkishen-Singh: > Also consider better naming: eg testTelemetryEngine

That name sounds like name for a test function

pkg/runner/runner.go
Harkishen-Singh: Made this `initTelemetryEngine` since starting here only and stopping in caller doesn't look nice. Hence, it initializes and registers metrics and start and deferred stop is done in the caller.

niksajakovljevic (approved):

PR #1238, opened 2022-03-14
Add information about manually forcing retention

JamesGuthrie:
## Description

This question came up in the community slack, and was not otherwise addressed by the documentation.

## Merge requirements

### DIFF
--- a/docs/metric_deletion_and_retention.md
+++ b/docs/metric_deletion_and_retention.md
@@ -154,3 +154,17 @@ SELECT prom_api.reset_metric_retention_period('container_cpu_usage_seconds_total
 
 For TimescaleDB versions < 2.0, the retention policies are executed using a cron job. Please see the Promscale
 installation instructions for your platform to see how to set up the cron job.
+
+
+### Manually forcing retention to be applied
+
+Promscale leverages TimescaleDB's [User-Defined Actions] to schedule a maintenance job which drops data which is
+over its retention period. By default, it will automatically be executed every 30 minutes.
+
+You can also manually run this maintenance job with the following SQL command:
+
+```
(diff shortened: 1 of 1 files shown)


docs/metric_deletion_and_retention.md
```diff
@@ -154,3 +154,17 @@ SELECT prom_api.reset_metric_retention_period('container_cpu_usage_seconds_total
...
+
+### Manually forcing retention to be applied
+
+Promscale leverages TimescaleDB's [User-Defined Actions] to schedule a maintenance job which drops data which is
+over its retention period. By default, it will automatically be executed every 30 minutes.
+
+You can also manually run this maintenance job with the following SQL command:
+
+```
+CALL prom_api.execute_maintenance();
```
niksajakovljevic, line 167: It's always great to see docs getting improved. However here I have few concerns:
- `prom_api.execute_maintenance()` seems to do compression work as well
- running this manually should be more of an edge case since we already automated this work. So I wonder what are the use cases for this? eg When should users go for it?
- are there any potential drawbacks of running it manually? Lock contention? How about if user runs it in parallel (multiple calls)?

antekresic: Yeah, we should definitely at least mention that this will compress stuff as well (which might be something user doesn't want).

antekresic: I think this is good enough to go in.

antekresic (approved):

JamesGuthrie:
Closing this. The concept of needing docs for this lives on in https://github.com/timescale/promscale_extension/issues/312. The underlying doc has been moved to the docs page anyway, so this PR ain't doin' nothin'.

PR #1246, opened 2022-03-16
Fix batch flush size

niksajakovljevic:
There are two things addressed within this PR: 

  1.  Fix batch `IsFull` check
    We should count the number of samples and exemplars and not the number of series.
    The count of series is usually less then the number of samples/exemplars.
    This makes it hard to reason and control how much rows we insert per one statement. 

2. Lower `FlushSize` and `MaxInsertStmtPerTxn` in order to reduce number of rows per transaction

    Previous settings meant that we can get 200K rows inside one transaction.
    Such a huge number of rows within a transaction decreases system ingest performance,
    leads to increased transaction time and can cause more lock contention. i
    New max rows per transaction limit is 5K (`FlushSize` * `MaxInsertStmtPerTxn`).

### DIFF
--- a/pkg/pgmodel/ingestor/buffer.go
+++ b/pkg/pgmodel/ingestor/buffer.go
@@ -50,7 +50,8 @@ func NewPendingBuffer() *pendingBuffer {
 }
 
 func (p *pendingBuffer) IsFull() bool {
-	return p.batch.CountSeries() > metrics.FlushSize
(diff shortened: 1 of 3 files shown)


niksajakovljevic:
I'll soon share benchmark numbers.

niksajakovljevic:
Batch is bug fixed with https://github.com/timescale/promscale/pull/1247

PR #1248, opened 2022-03-17
Run tests against ha docker

JamesGuthrie:
## Description

The main goal of this change (which is probably too large) is to ensure
that we can build the promscale connector against a version of the
promscale extension which contains all of the necessary bits.

This change is motivated by the fact that there is now a hard dependency
on the extension, and we need to be able to test in-progress changes
across repositories.

The basic mechanism is that for each open PR in the promscale_extension
repo, there will be a docker image built and pushed to the registry at:
`ghcr.io/timescale/dev_promscale_extension:<special-tag-here>`.

`<special-tag-here>` is constructed from the branch name from which the
PR originated. When a PR is opened in the promscale repository, we look
to see if there is already a docker image corresponding to the branch
name. If not, we fall back to the most recent docker image which was
built on the `develop` branch in the extension repository.

`<special-tag-here>` has the basic form of:

  `<extension-version>-<timescale-major>-<pg-version>`

Some examples are:
- `jg-ha-dockerfile-ts2-pg14`
- `develop-ts2-pg14`
- `0.5.0-ts2-pg14`

Local development is supported through two mechanisms:

1. When running e.g. `make docker-image-14` in the promscale_extension
   repo, the docker image is tagged as: `local/dev_promscale_extension`
   with the `<extension-version>` component of the tag set to `head`.
2. When running `make test` in the promscale repo, it first looks to see
   if a local image is available matching the `local/` prefix, then
   checks for a docker image derived from the current branch name, and
   then for built from the `develop` branch.

A number of things have changed tangentially as well:

- end_to_end tests now take the docker image which they should run
  against as an argument instead of using some magic to derive it
- `make test` was split into two components: `make unit` and `make e2e`
- multinode tests are being skipped because we broke them in the develop
  branch of the promscale extension
## Merge requirements


Makefile
```diff
@@ -1,23 +1,30 @@
...
 	go build -ldflags "-X 'github.com/timescale/promscale/pkg/version.Branch=${GIT_BRANCH}' -X 'github.com/timescale/promscale/pkg/version.CommitHash=${GIT_COMMIT}'" -o dist/promscale ./cmd/promscale
 
 .PHONY: test
-test: generate
-	go test -v -race ./... -timeout 40m
+test: unit e2e
+
+.PHONY: unit
+unit: generate
+	go test -v -race $(shell go list ./... | grep -v tests/end_to_end_tests) -timeout 40m
```
cevian, line 15: I think you'll need to exclude upgrade_tests eventually as well

pkg/util/election_test.go
```diff
@@ -24,9 +24,9 @@ var (
 	electionInterval = flag.Duration("election-interval", 1*time.Second, "Scheduled election interval")
 )
 
-const extensionState = testhelpers.Timescale
-
 func TestPgLeaderLock(t *testing.T) {
+	// TODO (james): Replace hardcoded value
```
cevian, line 28: should we move these into end to end tests?

pkg/internal/testhelpers/postgres_container.go
```diff
@@ -49,95 +47,80 @@ var (
...
-	return (e & postgres13Bit) != 0
-}
-
-func (e ExtensionState) GetPGMajor() string {
-	PGMajor := "14"
-	if e.UsesPG13() {
-		PGMajor = "13"
-	} else if e.UsesPG12() {
-		PGMajor = "12"
+func (e TestOptions) TryGetDockerImagePgVersion() (string, error) {
```
cevian, line 114: can we exec into the image and get the version from the pg_config binary instead?

JamesGuthrie: Yes, but only after the container has been started. In the upgrade tests (where this is used), there is no `timescaledev/promscale-extension:0.1.2-ts2-pg14` docker image, so trying to start it will fail. I think that we need to rework the upgrade tests a little anyway to make it a bit smoother.

pkg/util/election_test.go
JamesGuthrie: Hmmm... That would probably be a lot cleaner. Will take a look into doing that.

pkg/tests/end_to_end_tests/main_test.go
```diff
@@ -31,11 +31,12 @@ import (
...
-	updateGoldenFiles     = flag.Bool("update", false, "update the golden files of this test")
-	useDocker             = flag.Bool("use-docker", true, "start database using a docker container")
-	useTimescaleDB        = flag.Bool("use-timescaledb", true, "use TimescaleDB")
-	postgresVersion       = flag.Int("postgres-version-major", 14, "Major version of Postgres")
+	testDatabase      = flag.String("database", "tmp_db_timescale_migrate_test", "database to run integration tests on")
+	updateGoldenFiles = flag.Bool("update", false, "update the golden files of this test")
+	useDocker         = flag.Bool("use-docker", true, "start database using a docker container")
+	useTimescaleDB    = flag.Bool("use-timescaledb", true, "use TimescaleDB")
+	// TODO (james): Replace hardcoded value
+	timescaleDockerImage  = flag.String("timescale-docker-image", "ghcr.io/timecale/dev_promscale_extension:jg-ha-dockerfile-ts2-pg14", "TimescaleDB docker image to run tests against")
```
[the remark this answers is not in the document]
JamesGuthrie: Yes (maybe). I would suggest that the version is named after the branch. We don't have a `main` branch in the `promscale_extension` repo. That is easily fixed.

We don't have a `master` build of the extension yet because the docker stuff hasn't been merged to `master` (or even `develop` for that matter). My plan is to gradually "upgrade" the image specified here from `develop` to `master` as we merge things in the `promscale_extension` repo.

Makefile
```diff
@@ -1,23 +1,30 @@
...
 	wget https://github.com/timescale/promscale-test-data/raw/main/traces-dataset.sz -O ./pkg/tests/testdata/traces-dataset.sz
 
-.PHONY: e2e-test
-e2e-test: pkg/tests/testdata/traces-dataset.sz generate
-	go test -v ./pkg/tests/end_to_end_tests/
-	go test -v ./pkg/tests/end_to_end_tests/ -use-timescaledb=false
-	go test -v ./pkg/tests/end_to_end_tests/ -use-multinode
+.PHONY: e2e
+e2e: CURRENT_BRANCH?=$(shell git branch --show-current | sed 's#/#-#')
+e2e: DOCKER_IMAGE?=$(shell ./scripts/fallback-docker.sh $(LOCAL_DOCKER_BASE):head-ts2-pg14 $(GHCR_DOCKER_BASE):{$(CURRENT_BRANCH)-ts2-pg14,develop-ts2-pg14})
```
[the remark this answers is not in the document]
JamesGuthrie: No, I don't think so. The precedence here is:

1. local image
2. remote image with same branch name
3. remote image on `develop`

If we don't have 3. then as soon as we have something in promscale's `master` branch which depends on the new changes in the extension's `develop` branch (which aren't published/merged to `develop`), running the tests will fail.

Makefile
JamesGuthrie: Possibly. As I mentioned below, I think that we need to rework the upgrade tests a little.

pkg/util/election_test.go
JamesGuthrie: It looks like `pkg/util/election.go` is dead code...

Makefile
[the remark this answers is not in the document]
cevian: I guess I just don't understand how `{$(CURRENT_BRANCH)-ts2-pg14,develop-ts2-pg14}` expands. 

I'd expect the line to be

```
e2e: DOCKER_IMAGE?=$(shell ./scripts/fallback-docker.sh $(LOCAL_DOCKER_BASE):head-ts2-pg14 $(GHCR_DOCKER_BASE):$(CURRENT_BRANCH)-ts2-pg14 $(GHCR_DOCKER_BASE):develop-ts2-pg14)
```

But if you are sure the expansion works I am fine with it

.github/workflows/go.yml
```diff
@@ -95,8 +95,8 @@ jobs:
         - {name: "Singlenode (13)",          shortname: "singlenode-13",  tsdb: true,  multi: false, pg: 13}
         - {name: "Singlenode",               shortname: "singlenode-14",  tsdb: true,  multi: false, pg: 14}
         - {name: "Without TimescaleDB",      shortname: "no-timescaledb", tsdb: false, multi: false, pg: 14}
-        - {name: "Multinode",                shortname: "multinode",      tsdb: true,  multi: true,  pg: 14}
-
+        # TODO (james): Skipping multinode because tests are broken for now
```
cevian, line 98: let's make an issue to re-enable

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -755,7 +776,10 @@ func startDB(t *testing.T, ctx context.Context) (*pgx.Conn, testcontainers.Conta
 		t.Fatal(err)
 	}
 
-	dbContainer, closer, err := testhelpers.StartDatabaseImage(ctx, "timescaledev/promscale-extension:testing-extension-upgrade", tmpDir, dataDir, *printLogs, testhelpers.Timescale)
+	// TODO (james): Replace hardcoded value
```
cevian, line 779: let's make an issue tracking this

cevian (commented):
Changes look good. Minor nits now

pkg/tests/end_to_end_tests/main_test.go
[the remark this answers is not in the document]
JamesGuthrie: done

pkg/util/election_test.go
JamesGuthrie: I've removed this in https://github.com/timescale/promscale/pull/1253, waiting for one more reviewer then I'll rebase.

Makefile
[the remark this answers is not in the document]
JamesGuthrie: Ya I'm pretty sure that it expands to how you expect it to look. I've removed the expansion for readability though.

PR #1251, opened 2022-03-17
Remove trace_id domain type

niksajakovljevic:
TimescaleDB query planner fails to produce optimal query plans when domain type
is used for filtering.  Replacing traceId domain type with PG UUID solves the problem.

Here is related issue for TimescaleDB https://github.com/timescale/timescaledb/issues/4179


niksajakovljevic:
Just a draft PR until we finish ongoing SQL migration.

PR #1254, opened 2022-03-18
Disable running Promscale in both HA and read-only mode

JamesGuthrie:
## Description

This was not possible before with legacy HA, but was missed when
updating to non-legacy HA. It doesn't make sense to try to combine these
two operation modes.

## Merge requirements


cevian (approved):

paulfantom (approved):
lgtm

PR #1262, opened 2022-03-21
Fix overflowing pendingBuffers and make them respect flush limit. [WIP]

antekresic:
## Description

Previously, we have only checked if the pendingBuffer is full before sending to the
copier. In certain instances, if the incoming data is already bigger than the
flush size, we would be sending the oversized buffers which could cause
oversized ingest transactions. This change checks full pendingBuffers and
resizes them to the max while pushing the overflown data into a new pendingBuffer
which can be used later on to batch with other incoming data.
## Merge requirements

### DIFF
--- a/pkg/pgmodel/ingestor/buffer.go
+++ b/pkg/pgmodel/ingestor/buffer.go
@@ -49,13 +49,50 @@ func NewPendingBuffer() *pendingBuffer {
 	return pendingBuffers.Get().(*pendingBuffer)
 }
 
-func (p *pendingBuffer) IsFull() bool {
+func NewPendingBufferFromExisting(existing *pendingBuffer) *pendingBuffer {
+	if existing != nil {
+		return existing
+	}
+	return NewPendingBuffer()
+}
+
+// total returns total size of pendingBuffer, including both samples and exemplars.
+func (p *pendingBuffer) total() int {
 	samples, exemplars := p.batch.Count()
-	return samples+exemplars >= metrics.FlushSize
+	return samples + exemplars
+}
+func (p *pendingBuffer) IsFull() bool {
+	return p.total() >= metrics.FlushSize
 }
 
 func (p *pendingBuffer) IsEmpty() bool {
 	return p.batch.CountSeries() == 0
+}
+
+// ResizeToMax resizes the current buffer to maximum size and
+// returns the overflow as a new *pendingBuffer that can be used
+// to continue batching incoming data.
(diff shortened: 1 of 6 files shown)


pkg/pgmodel/ingestor/buffer.go
```diff
@@ -49,15 +49,52 @@ func NewPendingBuffer() *pendingBuffer {
...
+		return nil
+	}
+
+	// If we are oversized, resize current pendingBuffer and create a
+	// new one with the overflow. Since we check is the pendingBuffer full
+	// with each addition of a request, its safe to assume only the last
+	// request has overflown us.
+
+	lastRequest := p.needsResponse[len(p.needsResponse)-1]
+	// Since we are splitting up a single incoming request in two buffers,
```
niksajakovljevic, line 87: I guess one request can result in X number of buffers depending on the size of request. That will break the assumption here.
I'd recommend a different and simpler approach: iterate through the request items and keep assigning them to our buffer(s) until the whole request is exhausted (this will create one or many buffers). This will avoid a need of ever resizing the buffer since upfront we know how much we can add to buffer. I find it confusing to add to a buffer more than needed.

pkg/pgmodel/model/batch.go
```diff
@@ -65,6 +65,28 @@ func (t *Batch) AppendSlice(s []Insertable) {
...
+	for _, d := range overflow {
+		switch d.Type() {
+		case Sample:
+			t.numSamples -= 1
+		case Exemplar:
+			t.numExemplars -= 1
+		}
+	}
+
+	t.data = t.data[:size]
```
niksajakovljevic, line 86: I guess the upper limit (exclusive) should be `size` + 1 because we can fill up to a `size` into buffer. But let's maybe consider alternative approach here.

pkg/pgmodel/model/batch.go
```diff
@@ -65,6 +65,28 @@ func (t *Batch) AppendSlice(s []Insertable) {
...
+	if len(t.data) <= size {
+		return nil
+	}
+
+	overflow := t.data[size:]
+
+	for _, d := range overflow {
+		switch d.Type() {
+		case Sample:
+			t.numSamples -= 1
```
niksajakovljevic, line 80: Should instead use `d.Count()`

antekresic:
Postponing this fix until we do some proposed refactoring of the code as discussed with @niksajakovljevic

PR #1272, opened 2022-03-25
Avoid sending telemetry in CI tests

Harkishen-Singh:

## Description

## Merge requirements

### DIFF
--- a/pkg/internal/testhelpers/postgres_container.go
+++ b/pkg/internal/testhelpers/postgres_container.go
@@ -476,6 +476,7 @@ func startPGInstance(
 		"-c", "port=" + containerPort.Port(),
 		"-c", "max_prepared_transactions=150",
 		"-c", "log_line_prefix=" + nodeType + " %m [%d]",
+		"-c", "timescaledb.telemetry_level=off",
 		"-i",
 	}
 


cevian (changes requested):
I think this whole PR can be a one-line change.
in postgres_container.go line 478 add "-c", "timescaledb.telemetry_level=off"

antekresic (changes requested):
What Mat said.

Harkishen-Singh:
Ah, so this PR creates a problem for telemetry tests. The telemetry tests are designed to write telemetry metadata if the database telemetry is on. If it is set to off, the tests should not write any telemetry information. This is the expected case.

Now, this works well, which is the problem here. Since this PR sets the telemetry level to off, we can no longer write telemetry to database for testing in E2E tests.

One solution is to use the environment var, but wont work for all cases. Any suggestions please.

cevian:
Ugh. Can we override the `If it is set to off, the tests should not write any telemetry information` behavior for the tests?

Harkishen-Singh:
Let's dig into this PR only if this is a problem in the current telemetry state.

@VineethReddy02 can you please update here if there exist garbage values in telemetry data at the moment?

Harkishen-Singh:
Closing this for the time since we are not sure that we get this in CI at the moment. We can come back to this PR in the future.

Issue #1136, opened 2022-02-10
Docs: Add a note in prom-migrator on HA supportability


VineethReddy02:
Add a note in prom-migrator docs that we do not support migration in HA setup as we discussed previously to do backfilling users need to run a separate Promscale instance separately for backfilling and remove it post-migration. So the already running HA setup isn’t disturbed.
 
 Fixes: #1112

PR #1284, opened 2022-04-06
Add note saying `prom-migrator` cannot migrate to HA setups.

Harkishen-Singh:

Fixes: #1136

### DIFF
--- a/cmd/prom-migrator/README.md
+++ b/cmd/prom-migrator/README.md
@@ -72,6 +72,9 @@ as write endpoints are somewhat constrained.
 ./prom-migrator -start=1606408552 -end=1606415752 -reader-url=<read_endpoint_url_for_remote_read_storage> -writer-url=<write_endpoint_url_for_remote_write_storage> -progress-metric-url=<read_endpoint_url_for_remote_write_storage>
 ```
 
+**Note:** Prom-migrator does not support migrating data to a HA environment. If you have an HA setup,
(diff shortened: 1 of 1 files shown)


cmd/prom-migrator/README.md
```diff
@@ -72,6 +72,9 @@ as write endpoints are somewhat constrained.
 ./prom-migrator -start=1606408552 -end=1606415752 -reader-url=<read_endpoint_url_for_remote_read_storage> -writer-url=<write_endpoint_url_for_remote_write_storage> -progress-metric-url=<read_endpoint_url_for_remote_write_storage>
 ```
 
+**Note:** Prom-migrator does not support migrating data to a HA environment. If you have an HA setup,
+try running it in non-HA mode to ingest data from prom-migrator.
```
VineethReddy02, line 76: Instead of recommending user to run Promscale in non HA mode, let's recommend running a new Promscale connector only for the migrator and shut it down post migration. This should be ideal as it doesn't require effort's of disabling and Re-enabling HA mode.

Harkishen-Singh: The problem is Prom-migrator is beyond Promscale. We know Promscale can do only for migration, but other systems might not. Hence the best is to say don't use HA while ingesting data from the migrator.

VineethReddy02 (approved):

paulfantom (approved):

PR #1286, opened 2022-04-08
Adds support for rule-manager in Promscale

Harkishen-Singh:
Fixes: https://github.com/timescale/o11y-team-applications/issues/179

This PR adds a rules manager in Promscale. This is the first PR in the series of implementing Alerting and Recording rules in Promscale.

This PR does not link the rules manager with the runner. The next PR will implement:
1. loading + validating Prometheus config
2. CLI flags

The following PR will link with the runner. See the `PR and commit breakdown` in the design doc for more info.

Note: I haven't updated the changelog since this module is not linked with the runner, hence there will be no difference in Promscale's current features. So even if we make another release before implementing the ruling feature completely into Promscale, this PR will not be a blocker.

TODO:

- [ ] Open PR upstream to change the link of prom-migrator from `cmd/prom-migrator` to `pkg/migration-tool/cmd/prom-migrator`. @VineethReddy02


pkg/rules/custom.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+			if !alert.ResolvedAt.IsZero() {
+				a.EndsAt = alert.ResolvedAt
+			} else {
+				a.EndsAt = alert.ValidUntil
+			}
+			res = append(res, a)
+		}
+
+		if len(alerts) > 0 {
```
antekresic, line 82: I'd rather do this check at top and return early.

pkg/rules/custom_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/prometheus/prometheus/model/labels"
+	prometheus_promql "github.com/prometheus/prometheus/promql"
+	"github.com/stretchr/testify/require"
+
+	promscale_promql "github.com/timescale/promscale/pkg/promql"
+)
+
+func TestYoloVector(t *testing.T) {
```
antekresic, line 13: Would like to see a couple of more tests with edge cases since this is a yolo func. Thinks like empty values etc.

antekresic (changes requested):

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func (app *appenderAdapter) Append(_ storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	hash := l.Hash()
+	if _, found := app.data[hash]; !found {
+		app.data[hash] = initTs(l)
+	}
+	app.data[hash].Samples = append(app.data[hash].Samples, prompb.Sample{Timestamp: t, Value: v})
+	return 0, nil // return 0 seriesRef since we do not do series level caching.
```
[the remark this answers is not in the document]
Harkishen-Singh: No, because then we will need sync.Map. But that's not it, the rollback will be tough and we no longer know which samples belong to the current session, hence its much cleaner to keep a struct per session (appended). This is the same reason prometheus has written the code in that way.

pkg/rules/custom.go
Harkishen-Singh: its still the same, as the top is just a loop, which if zero will not run at all :)

pkg/rules/rules.go
```diff
@@ -0,0 +1,139 @@
...
+	prom_rules "github.com/prometheus/prometheus/rules"
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgclient"
+	"github.com/timescale/promscale/pkg/query"
+	"github.com/timescale/promscale/pkg/rules/adapters"
+)
+
+const (
+	DefaultQueueCapacity   = 10000
```
cevian, line 26: do these need to be configurable?

pkg/rules/rules.go
```diff
@@ -0,0 +1,139 @@
...
+	ResendDelay     time.Duration
+}
+
+type manager struct {
+	rulesManager     *prom_rules.Manager
+	notifierManager  *notifier.Manager
+	discoveryManager *discovery.Manager
+}
+
+func NewManager(ctx context.Context, r prometheus.Registerer, client *pgclient.Client, opts *Options) (*manager, error) {
```
cevian, line 45: where is this called?

cevian (changes requested):
Needs some changes

pkg/query/query_engine.go
```diff
@@ -12,6 +12,13 @@ import (
 	"github.com/timescale/promscale/pkg/promql"
 )
 
+const (
+	DefaultQueryTimeout         = time.Minute * 2
```
niksajakovljevic, line 16: How did we come up with these numbers?

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/prometheus/prometheus/storage"
+
+	"github.com/timescale/promscale/pkg/pgmodel/cache"
+	"github.com/timescale/promscale/pkg/pgmodel/ingestor"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+	"github.com/timescale/promscale/pkg/pgmodel/model"
+	"github.com/timescale/promscale/pkg/prompb"
+)
+
+var samplesIngested = metrics.IngestorItems.With(map[string]string{"type": "metric", "kind": "sample", "subsystem": "ruling"})
```
niksajakovljevic, line 21: Nit: consider `rules` as subsystem name

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (a ingestAdapter) Appender(_ context.Context) storage.Appender {
+	return &appenderAdapter{
+		data:     make(map[string][]model.Insertable),
+		ingestor: a.ingestor,
+	}
+}
+
+func (app *appenderAdapter) Append(_ storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	series, metricName, err := cache.CreateSeries(labelToPrompbLabels(l))
```
niksajakovljevic, line 48: I am a bit surprised that cache package is responsible for creating series especially because it doesn't cache  but only returns new instance.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/timescale/promscale/pkg/prompb"
+)
+
+var samplesIngested = metrics.IngestorItems.With(map[string]string{"type": "metric", "kind": "sample", "subsystem": "ruling"})
+
+type ingestAdapter struct {
+	ingestor *ingestor.DBIngestor
+}
+
+func NewIngest(ingestor *ingestor.DBIngestor) *ingestAdapter {
```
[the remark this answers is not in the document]
Harkishen-Singh: I thought `adapter` name should not be in the function since that's the package name, recalling some go-vet convention. But now I get that the package is `adapters` and `ingestadapter` is one of them. So it makes sense. Thanks.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return 0, nil
+	}
+	return storage.SeriesRef(id), nil
+}
+
+func (app *appenderAdapter) AppendExemplar(_ storage.SeriesRef, l labels.Labels, e exemplar.Exemplar) (storage.SeriesRef, error) {
+	// We do not support appending exemplars in recording rules since this is not yet implemented upstream.
+	// Once upstream implements this feature, we can modify this function.
+	return 0, nil
```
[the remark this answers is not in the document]
Harkishen-Singh: I do not agree. The moment this is implemented upstream, our ruling process will start erroring out if I return an error here. But the truth is, exemplars are only **a part of rule evaluation**, not the whole of it. Hence, it only makes sense to warn log this saying we need to implement this.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (a ingestAdapter) Appender(_ context.Context) storage.Appender {
+	return &appenderAdapter{
+		data:     make(map[string][]model.Insertable),
+		ingestor: a.ingestor,
+	}
+}
+
+func (app *appenderAdapter) Append(_ storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
```
niksajakovljevic, line 47: I think the idea of this interface is to reuse Series cache and group samples based on seriesRef. Here we are not doing that. We are grouping by metric name - however Ingestor is doing the same thing so I'd recommend just storing prom Timeseries and avoid duplicate work. Later on we can improve as we go.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+type ingestAdapter struct {
+	ingestor *ingestor.DBIngestor
+}
+
+func NewIngest(ingestor *ingestor.DBIngestor) *ingestAdapter {
+	return &ingestAdapter{ingestor}
+}
+
+type appenderAdapter struct {
+	// Map allows us to group similar samples into a single prompb.TimeSeries
```
niksajakovljevic, line 32: This would need some limits (Eg max number of samples?)? Otherwise we can blow up the memory here.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	// We can technically just maintain a list and have a single sample in each
+	// []prompb.TimeSeries like how Prometheus writes, but grouping samples gives
```
niksajakovljevic, line 34: I am not understanding this comment because here we are just grouping things by metric name and we are still packing one sample in each series because we are not using series cache to find and reference series.

niksajakovljevic (changes requested):
I've reviewed only Ingest part - let's discuss more offline.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	// us advantage in our ingestor.
+	data     map[string][]model.Insertable
```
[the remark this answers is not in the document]
niksajakovljevic: If we use `GetSeriesFromProtos` than caching would help and this code would be preferable. So let's try to use caching.

pkg/rules/rules.go
```diff
@@ -0,0 +1,139 @@
...
+	promqlEngine, err := query.NewEngineWithDefaults(log.GetLogger())
```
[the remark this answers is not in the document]
Harkishen-Singh: Prometheus takes the cli options and uses all enablefeatures. We should be good with cli options and if user wants enable features, he can always enable in the cli. updating the same.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+	samples := model.NewPromSamples(series, []prompb.Sample{{Timestamp: t, Value: v}})
+
+	if _, found := app.data[metricName]; !found {
+		app.data[metricName] = make([]model.Insertable, 0)
+	}
+	app.data[metricName] = append(app.data[metricName], samples)
+	id, _, err := series.GetSeriesID()
+	if err != nil {
+		return 0, nil
+	}
```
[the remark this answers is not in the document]
Harkishen-Singh: Sorry, i didnt know it will be error always. The aim here was that series id can be returned the next time, avoiding to create a series. If we have an error doing this optimization, simply don't do it. Which is why returning 0, nil.
But, this optimization is prometheus dependent which right now does not happen. So we can remove this part.

pkg/rules/custom.go
```diff
@@ -0,0 +1,91 @@
...
+		default:
+			return nil, fmt.Errorf("rule result is not a vector or scalar")
+		}
+	}
+}
+
+// My guess is this should be way faster than looping through individual samples
+// and converting into Prometheus Vector type. This lets us convert the type with
+// very less processing.
+func yoloVector(v *promscale_promql.Vector) prometheus_promql.Vector {
```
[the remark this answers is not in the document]
Harkishen-Singh: Added

pkg/rules/custom.go
```diff
@@ -0,0 +1,91 @@
...
+			res = append(res, a)
+		}
+
+		if len(alerts) > 0 {
+			s.Send(res...)
+		}
+	}
+}
+
+func do(ctx context.Context, client *http.Client, req *http.Request) (*http.Response, error) {
```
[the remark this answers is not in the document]
Harkishen-Singh: Used while initializing the notifier manager.

pkg/rules/custom.go
```diff
@@ -0,0 +1,91 @@
+// This file contains code copied from
```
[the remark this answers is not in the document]
Harkishen-Singh: changed to `upstream.go`

pkg/rules/rules.go
Harkishen-Singh: Prometheus allows it, so we should.

pkg/rules/rules.go
Harkishen-Singh: It's not. That's done knowingly. Please see the PR description for reason.

pkg/rules/adapters/ingest.go
Harkishen-Singh: resolved as per discussion

pkg/rules/adapters/ingest.go
Harkishen-Singh: No, this isn't an issue. Technically, the engine has a limit of 50M samples, but a single promql instance query should not return even that much.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,74 @@
...
+	return &appenderAdapter{
+		data:     make(map[string][]model.Insertable),
+		ingestor: a.ingestor,
+	}
+}
+
+func (app *appenderAdapter) Append(_ storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	series, metricName, err := app.ingestor.SCache.GetSeriesFromProtos(util.LabelToPrompbLabels(l))
+	if err != nil {
+		return 0, fmt.Errorf("error creating series: %w", err)
```
niksajakovljevic, line 46: Nit: above line might also get series from cache so let's change the message

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -31,7 +31,7 @@ type Cfg struct {
 
 // DBIngestor ingest the TimeSeries data into Timescale database.
 type DBIngestor struct {
-	sCache     cache.SeriesCache
+	SCache     cache.SeriesCache
```
niksajakovljevic, line 34: Since `SeriesCache` is passed to Ingestor I'd recommend passing it into Rule adapter as well. That would eliminate the need of making it public here.

pkg/query/config.go
```diff
@@ -0,0 +1,73 @@
...
+	fs.Int64Var(&cfg.MaxPointsPerTs, "metrics.promql.max-points-per-ts", 11000, "Maximum number of points per time-series in a query-range request. "+
+		"This calculation is an estimation, that happens as (start - end)/step where start and end are the 'start' and 'end' timestamps of the query_range.")
+	return cfg
+}
+
+func Validate(cfg *Config) error {
+	cfg.EnabledFeatureMap = make(map[string]struct{})
+	for _, f := range cfg.PromscaleEnabledFeatureList {
+		switch f {
+		case "promql-at-modifier", "promql-negative-offset":
```
niksajakovljevic, line 64: This feature flags are showing in few places so I'd suggest turning them into consts.

pkg/api/router.go
```diff
@@ -66,16 +66,13 @@ func GenerateRouter(apiConf *Config, client *pgclient.Client) (*mux.Router, erro
...
-		return nil, fmt.Errorf("creating query-engine: %w", err)
-	}
+	queryEngine := client.QueryEngine()
 
 	apiV1 := router.PathPrefix("/api/v1").Subrouter()
 	queryHandler := timeHandler(metrics.HTTPRequestDuration, "query", Query(apiConf, queryEngine, queryable, updateQueryMetrics))
 	apiV1.Path("/query").Methods(http.MethodGet, http.MethodPost).HandlerFunc(queryHandler)
 
-	queryRangeHandler := timeHandler(metrics.HTTPRequestDuration, "query_range", QueryRange(apiConf, queryEngine, queryable, updateQueryMetrics))
+	queryRangeHandler := timeHandler(metrics.HTTPRequestDuration, "query_range", QueryRange(apiConf, promqlConf, queryEngine, queryable, updateQueryMetrics))
```
niksajakovljevic, line 75: Nit: Seems like queryEngine is used in only one place so why just not inline it in function call

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,74 @@
...
+func (app *appenderAdapter) Commit() error {
+	numInsertablesIngested, err := app.ingestor.Dispatcher().InsertTs(context.Background(), model.Data{Rows: app.data, ReceivedTime: time.Now()})
+	if err == nil {
+		samplesIngested.Add(float64(numInsertablesIngested))
+	}
+	return errors.WithMessage(err, "rules: error ingesting data into db-ingestor")
+}
+
+func (app *appenderAdapter) Rollback() error {
+	app.data = nil
```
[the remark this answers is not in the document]
Harkishen-Singh: Appender must never be used after a rollback. Hence the panic will never be the case. That said, it is a good measure.

https://github.com/prometheus/prometheus/blob/685ce9964da2bcfdfa72a89db48d889a9457574e/storage/interface.go#L223

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/timescale/promscale/pkg/util"
+)
+
+var samplesIngested = metrics.IngestorItems.With(map[string]string{"type": "metric", "kind": "sample", "subsystem": "rules"})
+
+type ingestAdapter struct {
+	ingestor *ingestor.DBIngestor
+}
+
+func NewIngestAdapter(ingestor *ingestor.DBIngestor) *ingestAdapter {
```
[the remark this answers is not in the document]
Harkishen-Singh: I will add some comments but it's an adapter pattern, so it should convey its messageitself.

pkg/pgmodel/ingestor/ingestor.go
Harkishen-Singh: I want to keep the code in rules minimalistic. If we don't want to make it public, lets go with exposing a function to access ingestor's series cache.

pkg/query/config.go
Harkishen-Singh: I want to avoid those changes in this PR since I want the refactoring commit to being just moving stuff.

pkg/api/router.go
Harkishen-Singh: Its used in line 72 too.

pkg/rules/adapters/ingest.go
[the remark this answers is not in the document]
niksajakovljevic: I see. Based on Prometheus docs `appender` shouldn't be reused after either Commit or Rollback.
In that context I think we should have a state in the appender (eg. `closed`) that will be set to true after calling `Commit` or `Rollback`. Obviously we will check if `closed == true` before running any of these two operations. Alternatively adding some code comments explaining the lifecycle could also work: it seems that  each rule evaluation will create new appender. That makes me believe that we should have a metric/gauge to track number of active appenders.

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (app *appenderAdapter) AppendExemplar(_ storage.SeriesRef, l labels.Labels, e exemplar.Exemplar) (storage.SeriesRef, error) {
+	// We do not support appending exemplars in recording rules since this is not yet implemented upstream.
+	// Once upstream implements this feature, we can modify this function.
+	return 0, nil
+}
+
+func (app *appenderAdapter) Commit() error {
+	numInsertablesIngested, err := app.ingestor.Dispatcher().InsertTs(context.Background(), model.Data{Rows: app.data, ReceivedTime: time.Now()})
```
niksajakovljevic, line 65: Please document that the `InsertTS` is not fully following transactional semantics: eg. we can commit series but than fail on another transaction when writing metrics data..

pkg/rules/adapters/ingest.go
```diff
@@ -0,0 +1,75 @@
...
+	if err == nil {
+		samplesIngested.Add(float64(numInsertablesIngested))
+	}
+	return errors.WithMessage(err, "rules: error ingesting data into db-ingestor")
+}
+
+func (app *appenderAdapter) Rollback() error {
+	app.data = map[string][]model.Insertable{}
```
niksajakovljevic, line 73: Why not setting `ingestor` to nil as well?

pkg/rules/adapters/ingest.go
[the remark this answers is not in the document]
Harkishen-Singh: Added the docs. I don't think we need any metric here as I am not sure why we wanna track the number of active appenders.

pkg/rules/upstream.go
```diff
@@ -0,0 +1,94 @@
...
+				Metric: labels.Labels{},
+			}}, nil
+		default:
+			return nil, fmt.Errorf("rule result is not a vector or scalar")
+		}
+	}
+}
+
+// Compile-time test to make sure that sizes of both vectors are the same.
+var _ = [1]bool{}[unsafe.Sizeof(promscale_promql.Vector{})-unsafe.Sizeof(prometheus_promql.Vector{})] // #nosec
```
niksajakovljevic, line 52: This is brittle as it depends on data alignment (this test can be false positive). Since this is a key component I'd rather like us adding a test where we test for equality of two structs using reflection.

antekresic (approved):
:tada:

pkg/rules/adapters/ingest.go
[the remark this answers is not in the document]
cevian: Well we should at-least log a WARNING which this doesn't do. But, I do not agree with this approach. If something changes in the code this drastically it should be an error. We should be safe wrt not loosing data and fail-fast is always safer than fail-open. Pretending you stored something when you didn't is really bad.

pkg/rules/adapters/query.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (q querierAdapter) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
+	// Pushdowns are not supported here. This is fine as Prometheus rule-manager only uses queryable to know
+	// the previous state of the alert. This function is not used in recording/alerting rules evaluation.
+	seriesSet, _ := q.qr.Select(sortSeries, hints, nil, nil, matchers...)
+	return seriesSet
+}
+
+func (q querierAdapter) LabelValues(name string, _ ...*labels.Matcher) ([]string, storage.Warnings, error) {
+	// Weak TODO: We need to implement the matchers.
```
cevian, line 41: is this function ever actually used? Does the lack of matching effect correctness?

pkg/rules/rules.go
cevian: Should we create an issue for this then? Or is this for an upcoming PR?

pkg/rules/rules.go
```diff
@@ -0,0 +1,133 @@
...
+func NewManager(ctx context.Context, r prometheus.Registerer, client *pgclient.Client, opts *Options) (*manager, error) {
+	discoveryManagerNotify := discovery.NewManager(ctx, log.GetLogger(), discovery.Name("notify"))
+
+	notifierManager := notifier.NewManager(&notifier.Options{
+		QueueCapacity: opts.QueueCapacity,
+		Registerer:    r,
+		Do:            do,
+	}, log.GetLogger())
+
+	// For the moment, we do not have any external UI url, hence we provide an empty one.
```
cevian, line 53: Let's create an issue to make such a UI and update this

pkg/rules/rules.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Run runs the managers and blocks on either a graceful exit or on error.
+func (m *manager) Run() error {
+	var g run.Group
+
+	g.Add(func() error {
+		log.Info("msg", "starting discovery manager...")
+		return errors.WithMessage(m.discoveryManager.Run(), "error running discovery manager")
+	}, func(err error) {})
```
cevian, line 112: shouldn't we log `log.Info("msg", "stopping discovery manager")` in the stop func?

cevian (changes requested):
My only strong comment is about AppendExemplars. We really can't promise to store something if we aren't storing it. This is potential data loss.

pkg/rules/rules.go
Harkishen-Singh: Upcoming PR

.github/workflows/gosec.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       - name: Run Gosec Security Scanner
         uses: securego/gosec@v2.10.0
         with:
-          args: -exclude-dir=pkg/promql -exclude-dir=pkg/prompb ./...
+          args: -exclude-dir=pkg/promql -exclude-dir=pkg/prompb -exclude-dir=migration-tool ./...
```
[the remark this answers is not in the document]
Harkishen-Singh: Because it was giving false positives for all imports, which made me feel it is not updated for local mod files. This needs now investigation but not worth at this moment.

pkg/rules/adapters/ingest.go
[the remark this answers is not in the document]
Harkishen-Singh: Ok will add warning.

pkg/rules/adapters/query.go
Harkishen-Singh: In alerting, no. I wrote todo because we need to update ourselves with prometheus, irrelevant of this PR

pkg/rules/rules.go
Harkishen-Singh: Sure will add that

.github/workflows/gosec.yml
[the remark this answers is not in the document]
cevian: ok please create an issue so we don't forget

pkg/rules/adapters/query.go
cevian: I'd prefer this to be a panic if ever called then. I have significant concerns about what's implemented in terms of correctness

cevian (approved):

pkg/rules/rules.go
Harkishen-Singh: https://github.com/timescale/o11y-team-applications/issues/196

pkg/rules/rules.go
Harkishen-Singh: Well, I didn't implement that because the discovery manager does not have any `Stop()` function. I guess is it depends on the context of the notifier manager (which is used to create the discovery manager). So stopping the notifier manager should stop the discovery manager. But for log, we can add to make the lifecycle cleaner.

PR #1294, opened 2022-04-12
Add test case for `reset_metric_retention_period` with two-step agg

JamesGuthrie:
## Description

In [1] we fixed some typos in the `reset_metric_retention_period`
function which caused the function to break when used with two-step
continuous aggregates.

This change adds tests cases to exercise `reset_metric_retention_period`
with two-step aggregates.

[1]: https://github.com/timescale/promscale_extension/pull/180

## Merge requirements

- [] ~~Updated the relevant documentation~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ We use the following categories for changes:
 - Log warning when failing to write response to remote read requests [#1180]
 - Fix Promscale running even when some component may fail to start [#1217]
 - Fix `promscale_ingest_max_sent_timestamp_milliseconds` metric for tracing [#1270]
+- Fix `prom_api.reset_metric_retention_period` on two-step continuous aggregates [#1294]
 
 ### Removed
 - Remove deprecated flags. More info can be found [here](docs/configuration.md#old-flag-removal-in-version-0.11.0) [#1229]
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ We use the following categories for changes:
 - Log warning when failing to write response to remote read requests [#1180]
 - Fix Promscale running even when some component may fail to start [#1217]
 - Fix `promscale_ingest_max_sent_timestamp_milliseconds` metric for tracing [#1270]
+- Fix `prom_api.reset_metric_retention_period` on two-step continuous aggregates [#1294]
```
niksajakovljevic, line 36: How is this a FIX? It's just a test change

niksajakovljevic (commented):
Can you please add description explaining the problem. It would make it easier for reviewing and it's also better for knowledge sharing.

CHANGELOG.md
JamesGuthrie: https://github.com/timescale/promscale_extension/pull/180

niksajakovljevic: Ok. I've found a related extension change here https://github.com/timescale/promscale_extension/pull/180/files
But still let's add some description explaining the problem and the fix.  From extension PR it seems there was some problem with migration script (a wrong column name) if I am not mistaken

JamesGuthrie: done

niksajakovljevic: btw I am thinking if we should track extension changes in their own CHANGELOG? Obviously the fix in the extension can be released separately from Promscale connector so I am not sure why are we coupling these two? I am assuming that being able to release extension separately from connector is better (obviously in some cases that will not be possible...). Thoughts?

sumerman (approved):

niksajakovljevic (approved):

PR #1295, opened 2022-04-12
Update PromQL from upstream

Harkishen-Singh:
Fixes: https://github.com/timescale/promscale/issues/1296

This PR updates the PromQL module with the latest commits from upstream. It also contains the newly implemented feature for PromQL evaluation statistics.


cevian:
Checks aren't passing so taking myself off the review. Please request re-review when this PR is ready and all checks pass

go.mod
```diff
@@ -50,11 +49,10 @@ require (
...
 	go.uber.org/goleak v1.1.12
 	golang.org/x/sys v0.0.0-20220422013727-9388b58f7150
-	google.golang.org/genproto v0.0.0-20220217155828-d576998c0009 // indirect
 	google.golang.org/grpc v1.46.0
 	google.golang.org/protobuf v1.28.0
 	gopkg.in/yaml.v2 v2.4.0
 )
 
 // Make sure Prometheus version is pinned as Prometheus semver does not include Go APIs.
-replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v1.8.2-0.20220117154355-4855a0c067e2
```
paulfantom, line 60: Do you think we can update to latest prometheus and not use git hash anymore?

pkg/api/common.go
```diff
@@ -12,11 +12,12 @@ import (
 	"io/ioutil"
 	"math"
 	"net/http"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/grafana/regexp"
```
paulfantom, line 19: Why using this package instead of default golang one? I mean what are the benefits?

go.mod
Harkishen-Singh: If you mean `@latest`, I think we should stick to git hash (monthly wise). Otherwise, I have seen incompatibility with `@latest` mod version, which is not good, as anyone who updates the mod next might have to fix the incompatibilities.

pkg/api/common.go
Harkishen-Singh: Well, the point is to make outputs similar to what Prometheus does. Recently, Prometheus shifted from go regex to grafana regex. I don't know the reason, will try to see the godoc APIs. But, there might be an edge case in response comparison if we shifted back to the original.

Harkishen-Singh: `This repo is a fork of the upstream Go regexp package, with some code optimisations to make it run faster.`

[Source](https://pkg.go.dev/github.com/grafana/regexp#section-readme)

antekresic (commented):
Looks good in general. 

I'd like us to verify that these traces that are now being generated in the query engine are compatible with our setup and that we can actually use them ourselves.

Harkishen-Singh:
@antekresic yes, it does. We can apply jaeger traces endpoint and get the input. We can also provide otel-collector and direct the traces there and then back to promscale.

image

pkg/promql/fuzz.go
```diff
@@ -57,7 +57,13 @@ const (
 )
 
 func fuzzParseMetricWithContentType(in []byte, contentType string) int {
-	p := textparse.New(in, contentType)
+	p, warning := textparse.New(in, contentType)
+	if warning != nil {
+		// An invalid content type is being passed, which should not happen
+		// in this context.
+		panic(warning)
```
niksajakovljevic, line 64: Should we really panic here because of invalid contentType? I am not sure where this is being used

pkg/api/query_range.go
```diff
@@ -89,6 +89,7 @@ func queryRange(promqlConf *query.Config, queryEngine *promql.Engine, queryable
 
 		qry, err := queryEngine.NewRangeQuery(
 			queryable,
+			&promql.QueryOpts{EnablePerStepStats: true},
```
niksajakovljevic, line 92: Can you explain why we want stats enabled? Do we use these stats?

pkg/promql/fuzz.go
Harkishen-Singh: This is a direct copy from Prometheus 

https://github.com/prometheus/prometheus/blob/957092451175e9c4b37c52b6d3c8444c1d3eebc4/promql/fuzz.go#L64

But I agree, we can replace that with a warning log.

pkg/api/query_range.go
Harkishen-Singh: This allows the stats feature to work.

https://github.com/timescale/promscale/blob/343bfa83afcc05f263ef1a39b4d5cc6db4593325/pkg/promql/engine.go#L479

If you see how the feature works, there is a logical AND condition between
1. Input of `-enable-feature=promql-per-step-stats` that is supplied in the CLI (`ng.enablePerStepStats` part)
2. Input from the caller creating query (range & instant) (`opts.EnablePerStepStats` part). This one is the line you commented on

So, the feature will work only when `1 && 2 == true`. `1` is based on user, `2` is Promscale enabling the feature. Hence, we pass in `&promql.QueryOpts{EnablePerStepStats: true}` so that the stats feature is now enabled.

> why we want stats enabled?

It's a PromQL feature that we were lacking. Prometheus provides it, but we don't, as it was recently merged into Prometheus master. With this PR, we are again up to date with the master.

Harkishen-Singh: Note: We do not enable this, the enabling will happen only when the user passes in `-enable-feature=promql-per-step-stats`

pkg/promql/fuzz.go
niksajakovljevic: Yeah I've noticed it's borrowed code. Just wanted to make sure we have full understanding and doing the right thing in our context.

niksajakovljevic (approved):

niksajakovljevic:
@Harkishen-Singh Btw since you do mention that this brings in new features. Do we miss some tests for new features? I've only seen `fuzz_test` as a new test file, but missed to find one on statistics?

antekresic (approved):

PR #1302, opened 2022-04-17
Remove CHECK constraints on trace tables

niksajakovljevic:
It seems that CHECK constrains come with a huge CPU
cost for the database. We can instead move checks to Promscale code.
Benchmark shows around 35% improvement in CPU usage.

We also calculate span duration in connector.


pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -82,6 +82,9 @@ func (t *traceWriterImpl) addSpanLinks(linkRows *[][]interface{}, tagsBatch tagB
 	for i := 0; i < links.Len(); i++ {
 		link := links.At(i)
 		linkedSpanID := getSpanID(link.SpanID().Bytes())
+		if linkedSpanID.Status != pgtype.Present {
```
Harkishen-Singh, line 85: Why not use

```suggestion
		if linkedSpanID.Status == pgtype.Null {
```

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -296,7 +302,7 @@ func (t *traceWriterImpl) InsertTraces(ctx context.Context, traces pdata.Traces)
 				}
 
 				spanRows = append(spanRows, []interface{}{traceID, spanID, parentSpanID,
-					operationID, start, end, getTraceStateValue(span.TraceState()), jsonTags, span.DroppedAttributesCount(), eventTimeRange, span.DroppedEventsCount(),
+					operationID, start, end, float64(end.Sub(start).Nanoseconds()) / float64(1e6), getTraceStateValue(span.TraceState()), jsonTags, span.DroppedAttributesCount(), eventTimeRange, span.DroppedEventsCount(),
```
Harkishen-Singh, line 305: You can also do this if the aim is to covert in milliseconds.

```suggestion
					operationID, start, end, float64(timestamp.FromFloatSeconds(end.Sub(start).Seconds())), getTraceStateValue(span.TraceState()), jsonTags, span.DroppedAttributesCount(), eventTimeRange, span.DroppedEventsCount(),
```

Harkishen-Singh (approved):

pkg/pgmodel/ingestor/trace/writer.go
```diff
@@ -82,6 +82,9 @@ func (t *traceWriterImpl) addSpanLinks(linkRows *[][]interface{}, tagsBatch tagB
 	for i := 0; i < links.Len(); i++ {
```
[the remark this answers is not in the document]
jgpruitt: just in case a refactor removes the check on line 262. Probably add a check in `addSpanEvents` too for paranoia's sake

jgpruitt (changes requested):
small paranoid asks

jgpruitt (approved):

PR #1324, opened 2022-05-02
Rename tracing enums

cevian:
This makes the enum names much more user-friendly on the SQL level.
We made them lowercase and removed the prefix.

See https://github.com/timescale/promscale_extension/pull/225 for
more details.

## Description

## Merge requirements


niksajakovljevic (approved):

pkg/pgmodel/ingestor/trace/operation_batch.go
```diff
@@ -77,3 +78,22 @@ func (o operationBatch) GetID(serviceName, spanName, spanKind string) (pgtype.In
...
+	case pdata.SpanKindInternal:
+		return "internal", nil
+	case pdata.SpanKindConsumer:
+		return "consumer", nil
+	case pdata.SpanKindProducer:
+		return "producer", nil
+	case pdata.SpanKindUnspecified:
+		return "unspecified", nil
+	default:
+		return "unspecified", fmt.Errorf("unknown span kind: %v", pk)
```
antekresic, line 97: Should we return empty value here? I guess that is safer since "unspecified" is a valid value but we are erroring here so if somebody ignores the error, we could potentially have wrong data.

antekresic (approved):
Approved with a minor nit.

niksajakovljevic:
@cevian This also needs respected changes in our mixins https://github.com/timescale/promscale/tree/master/docs/mixin/dashboards

PR #1327, opened 2022-05-03
Prepare for the 0.11.0-alpha release

cevian:
## Description

## Merge requirements

### DIFF
--- a/deploy/helm-chart/Chart.yaml
+++ b/deploy/helm-chart/Chart.yaml
@@ -1,6 +1,6 @@
 name: promscale
-version: 0.10.0
-appVersion: 0.10.0
+version: 0.11.0-alpha
+appVersion: 0.11.0-alpha
 apiVersion: v2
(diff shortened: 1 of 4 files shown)


cevian:
Note this is a testing-only release in alpha. We will NOT support upgrades from 0.11.0-alpha to 0.11.0. Upgrading to this version will drop your tracing data.

paulfantom (changes requested):
Tests are failing due to:
```
level=error ts=2022-05-03T20:33:22.505Z caller=runner.go:116 msg="aborting startup due to error" err="migration error: error while checking for supported and available promscale extension versions: The promscale extension is not available at the right version, need version: >=0.5.0-alpha <0.5.99 and the default version is 0.5.0-alpha "
```

paulfantom (commented):
helm chart changes lgtm

paulfantom (approved):

PR #1328, opened 2022-05-04
Fixes for new tag_map type

cevian:
## Description

## Merge requirements

### DIFF
--- a/pkg/jaeger/query/trace_query.go
+++ b/pkg/jaeger/query/trace_query.go
@@ -31,12 +31,12 @@ const (
 		s.trace_state 			trace_states,
 		s_url.url 				schema_urls,
 		o.span_name     		span_names,
-	   	ps_trace.jsonb(s.resource_tags) 	resource_tags,
-	   	ps_trace.jsonb(s.span_tags) 		span_tags,
+		_ps_trace.tag_map_denormalize(s.resource_tags) 	resource_tags,
+		_ps_trace.tag_map_denormalize(s.span_tags) 		span_tags,
 	   	array_agg(e.name ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)			event_names,
 		array_agg(e.time ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)			event_times,
 	   	array_agg(e.dropped_tags_count ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_dropped_tags_count,
-	   	jsonb_agg(ps_trace.jsonb(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_tags,
+		array_agg(_ps_trace.tag_map_denormalize(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_tags,
 	   	inst_lib.name 				library_name,
 	   	inst_lib.version 			library_version,
 		inst_lib_url.url 			library_schema_url,
@@ -44,7 +44,7 @@ const (
 		array_agg(lk.linked_span_id ORDER BY lk.link_nbr)  FILTER(WHERE lk IS NOT NULL)	links_linked_span_ids,
 		array_agg(lk.trace_state ORDER BY lk.link_nbr)     FILTER(WHERE lk IS NOT NULL)	links_trace_states,
 		array_agg(lk.dropped_tags_count ORDER BY lk.link_nbr) FILTER(WHERE lk IS NOT NULL) 	links_dropped_tags_count,
-		jsonb_agg(lk.tags ORDER BY lk.link_nbr) FILTER(WHERE lk IS NOT NULL)			links_tags
+		array_agg(_ps_trace.tag_map_denormalize(lk.tags) ORDER BY lk.link_nbr) FILTER(WHERE lk IS NOT NULL)			links_tags
 	FROM
 		_ps_trace.span s
 	INNER JOIN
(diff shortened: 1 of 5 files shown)


pkg/pgmodel/ingestor/trace/tag_batch.go
```diff
@@ -94,34 +94,34 @@ func (t tagBatch) SendBatch(ctx context.Context, conn pgxconn.PgxConn) (err erro
...
 		byteVal, err := json.Marshal(v)
 		if err != nil {
-			return nil, err
+			return pgtype.JSONB{}, err
 		}
 		t := tag{k, string(byteVal), typ}
 		ids, err := tb.b.Get(t)
 		if err != nil {
-			return nil, fmt.Errorf("error getting tag from batch %v: %w", t, err)
+			return pgtype.JSONB{}, fmt.Errorf("error getting tag from batch %v: %w", t, err)
```
niksajakovljevic, line 107: Imho it makes more sense to return `nil` for error cases.

pkg/pgmodel/ingestor/trace/tag_batch_test.go
```diff
@@ -121,23 +121,23 @@ func TestTagBatch(t *testing.T) {
 			getTagMapJSONCheck: func(t *testing.T, batch tagBatch) {
 				tagMap, err := batch.GetTagMapJSON(map[string]interface{}{"test": ""}, SpanTagType)
 				require.Nil(t, err)
-				require.Equal(t, `{"7":8}`, string(tagMap))
+				require.Equal(t, `{"7":8}`, string(tagMap.Bytes))
 
 				tagMap, err = batch.GetTagMapJSON(map[string]interface{}{"nonexistant": ""}, SpanTagType)
 				require.EqualError(t, err, "error getting tag from batch {nonexistant \"\" 1}: error getting item from batch")
-				require.Equal(t, []byte(nil), tagMap)
+				require.Equal(t, []byte(nil), tagMap.Bytes)
```
niksajakovljevic, line 128: If we return nil tagMap pointer (instead of default struct) this check would be simpler: `require.Nil(t, tagMap)`

niksajakovljevic (approved):
Looks good. My only comment is to return `nil` for error cases. It's more efficient and makes code simpler. We seem to never process a struct returned on error.

pkg/jaeger/query/trace_query.go
```diff
@@ -31,20 +31,20 @@ const (
...
 		o.span_name     		span_names,
-	   	ps_trace.jsonb(s.resource_tags) 	resource_tags,
-	   	ps_trace.jsonb(s.span_tags) 		span_tags,
+		_ps_trace.tag_map_denormalize(s.resource_tags) 	resource_tags,
+		_ps_trace.tag_map_denormalize(s.span_tags) 		span_tags,
 	   	array_agg(e.name ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)			event_names,
 		array_agg(e.time ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)			event_times,
 	   	array_agg(e.dropped_tags_count ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_dropped_tags_count,
-	   	jsonb_agg(ps_trace.jsonb(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_tags,
+		array_agg(_ps_trace.tag_map_denormalize(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS NOT NULL)	event_tags,
```
pgnickb, line 39: I don't understand this. The old thing returns a `jsonb` array:
```JSON
[{"a":1}, {"a":2}, ...]
```
The new thing is actually array of `jsonb`s, which is not really the same thing (former is `jsonb` latter is `jsonb[]`)

sumerman: ps_trace.jsonb is pretty much _ps_trace.tag_map_denormalize

pgnickb: Sorry, I didn't realize that we actually change the type as well. This change of type is intended.

pkg/pgmodel/ingestor/trace/tag_batch.go
cevian: we can't since the return type is not a pointer

pgnickb (approved):

pkg/jaeger/query/trace_query.go
```diff
@@ -181,8 +181,9 @@ func buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (string, []interfac
 
 	if len(q.Tags) > 0 {
 		for k, v := range q.Tags {
-			params = append(params, k, v)
-			qual := fmt.Sprintf(`s.span_tags ? ($%d == $%d)`, len(params)-1, len(params))
+			//TODO make sure this is optimized correctly
```
sumerman, line 184: The following seems to work in PSQL. Should be good to go. I'll double-check with `pgx` driver and we can fix post-RC if there's an issue. 

```SQL
/* Equals operator returns correct result */
PREPARE eq_test AS
SELECT trace_id
    FROM ps_trace.span
    WHERE
            span_tags -> $1 = $2::text::jsonb
        AND resource_tags -> $3 = $4;

EXECUTE eq_test('pwlen', 25, 'service.name', '"generator"');

/* Equals uses support function and produces correct plan */
SELECT 
    x::text LIKE '%InitPlan%',
    x::text LIKE '% @> %' 
FROM explain_jsonb($$EXECUTE eq_test('pwlen', 25, 'service.name', '"generator"')$$) AS f(x);
```

sumerman (approved):

PR #1335, opened 2022-05-07
Optimize tracing query through max trace duration

cevian:
The config object will be needed to specify the max
trace duration the query can expect.

## Description

## Merge requirements

### DIFF
--- a/pkg/jaeger/query/trace_query.go
+++ b/pkg/jaeger/query/trace_query.go
@@ -108,42 +108,50 @@ const (
 	`
 )
 
-func buildCompleteTraceQuery(traceIDClause string) string {
+type Builder struct {
+	cfg *Config
+}
+
+func NewBuilder(cfg *Config) *Builder {
+	return &Builder{cfg}
+}
+
+func (b *Builder) buildCompleteTraceQuery(traceIDClause string) string {
 	return fmt.Sprintf(
 		completeTraceSQLFormat,
 		traceIDClause)
 }
 
-func findTracesQuery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
-	subquery, params := buildTraceIDSubquery(q)
+func (b *Builder) findTracesQuery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
+	subquery, params := b.buildTraceIDSubquery(q)
 	traceIDClause := "s.trace_id = trace_ids.trace_id"
-	completeTraceSQL := buildCompleteTraceQuery(traceIDClause)
+	completeTraceSQL := b.buildCompleteTraceQuery(traceIDClause)
 	return fmt.Sprintf(findTraceSQLFormat, subquery, completeTraceSQL), params
 }
 
-func findTraceIDsQuery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
-	subquery, params := buildTraceIDSubquery(q)
+func (b *Builder) findTraceIDsQuery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
+	subquery, params := b.buildTraceIDSubquery(q)
 	return subquery, params
 }
 
-func getTraceQuery(traceID model.TraceID) (string, []interface{}, error) {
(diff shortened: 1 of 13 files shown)


pkg/jaeger/query/trace_query.go
```diff
@@ -222,9 +243,11 @@ func buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (string, []interfac
 
 	query := ""
 	if len(clauses) > 0 {
-		query = fmt.Sprintf(subqueryFormat, strings.Join(clauses, " AND "))
+		params = append(params, b.cfg.MaxTraceDuration)
+		query = fmt.Sprintf(subqueryFormat, len(params), strings.Join(clauses, " AND "))
 	} else {
-		query = fmt.Sprintf(subqueryFormat, "TRUE")
+		params = append(params, b.cfg.MaxTraceDuration)
+		query = fmt.Sprintf(subqueryFormat, len(params), "TRUE")
```
antekresic, line 250: This can be simplified, lots of duplication.

pkg/jaeger/query/trace_query.go
```diff
@@ -108,42 +108,63 @@ const (
...
+
+func NewBuilder(cfg *Config) *Builder {
+	return &Builder{cfg}
+}
+
+func (b *Builder) buildCompleteTraceQuery(traceIDClause string, timeLowRef string, timeHighRef string) string {
+	spanTimeClause := ""
+	eventTimeClause := ""
+	linkTimeClause := ""
+	if timeLowRef != "" {
```
sumerman, line 123: A nit: the way `timeLowRef` and `timeHightRef` are used I'd rather prefer a single boolean flag in the `buildCompleteTraceQuery` signature.

sumerman (changes requested):

pkg/jaeger/query/trace_query.go
```diff
@@ -220,12 +241,14 @@ func buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (string, []interfac
...
+	clauseString := ""
 	if len(clauses) > 0 {
-		query = fmt.Sprintf(subqueryFormat, strings.Join(clauses, " AND "))
+		clauseString = strings.Join(clauses, " AND ")
 	} else {
-		query = fmt.Sprintf(subqueryFormat, "TRUE")
+		clauseString = "TRUE"
 	}
+	params = append(params, b.cfg.MaxTraceDuration)
+	query := fmt.Sprintf(subqueryFormat, len(params), clauseString)
```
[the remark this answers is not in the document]
cevian: so I think that is correct since its a single parameter with an interval. If it's parameter 4 this would become 
```
start_time_max - $4::interval as time_low, start_time_max + $4::interval as time_high
```
another words we are adding and subtracting the same interval

pkg/jaeger/query/trace_query.go
cevian: If there was a boolean then buildCompleteTraceQuery would need to know the names " "trace_ids.time_low", "trace_ids.time_high" but I don't think that function should know anything about the trace_ids table. That's why I think passing in the names is cleaner

pkg/jaeger/query/trace_query.go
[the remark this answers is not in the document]
sumerman: Oh, I see. Could you add a comment? Splicing a numbered prepared parameter with a numbered Sprintf parameter is too terse for me 😅

sumerman (approved):
I'm approving this to unblock while I'm off. But I have stong feelings about this one https://github.com/timescale/promscale/pull/1335#discussion_r868291464

pkg/jaeger/query/trace_query.go
cevian: Good point! I realized I can add clauses independently for timeLowRef and timeHighRef. This allows the api for one to be set but not the other and prevents only one from being used at the sentinel. Makes the code more robust. Thanks!

PR #1368, opened 2022-05-20
Adds E2E test for alerting in Promscale.

Harkishen-Singh:

This commit mimics a case of actual alerting + alertmanager.
It starts the rules-manager with a alert that is always firing.

Then, in E2E test, we start a mock alertmanager server in the `alertsReceiver()`
which listens for alerts just as the alertmanager does. We send down a channel
to this server that signals whether we received an expected alert or not, which
decides the fate of the test.

All references are linked in the commit.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,8 @@ require (
 	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
 	github.com/peterbourgon/ff/v3 v3.1.2
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.12.2
+	github.com/prometheus/alertmanager v0.23.1-0.20210914172521-e35efbddb66a
+	github.com/prometheus/client_golang v1.12.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.34.0
 	github.com/prometheus/prometheus v1.8.2-0.20220308163432-03831554a519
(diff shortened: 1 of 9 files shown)


pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -45,9 +45,10 @@ type pgxDispatcher struct {
 	copierReadRequestCh    chan<- readRequest
 	seriesEpochRefresh     *time.Ticker
 	doneChannel            chan struct{}
-	doneWG                 sync.WaitGroup
-	labelArrayOID          uint32
 	closed                 *uber_atomic.Bool
+	metricBatchers         sync.WaitGroup
```
niksajakovljevic, line 49: I am totally confused here. Can you please explain why do we need to change batcher and disaptcher for alerting e2e tests? I'm quite against this change especially as part of this PR which imho should have nothing to do with how we do batching...

pkg/pgmodel/ingestor/metric_batcher.go
```diff
@@ -139,8 +140,7 @@ func runMetricBatcher(conn pgxconn.PgxConn,
 	completeMetricCreationSignal chan struct{},
 	metricTableNames cache.MetricCache,
 	copierReadRequestCh chan<- readRequest,
-	labelArrayOID uint32) {
```
niksajakovljevic, line 142: Ohh I see this has never been used. Good catch!

niksajakovljevic (approved):

go.mod
```diff
@@ -58,6 +58,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/prometheus/alertmanager v0.24.0
```
antekresic, line 61: Why is this not grouped together with above?

pkg/tests/end_to_end_tests/alerts_test.go
```diff
@@ -0,0 +1,216 @@
...
+	conf, err := amConfig.LoadFile(amConfigPath)
+	require.NoError(t, err)
+	apiAM.Update(conf, func(model.LabelSet) {})
+
+	u, err := url.Parse("http://localhost:8080")
+	require.NoError(t, err)
+
+	router := route.New()
+	mux := apiAM.Register(router, u.Path)
+	srv := &http.Server{Addr: ":8080", Handler: mux}
```
antekresic, line 153: Lets use httptest server here if possible.

antekresic (changes requested):
Lets try updating the test to use httptest server and try to reduce the amount of synchronization there (I feel like it can be simplified quite a bit).

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -230,7 +231,6 @@ func (p *pgxDispatcher) InsertTs(ctx context.Context, dataTS model.Data) (uint64
 				maxt = ts
 			}
 		}
-		// the following is usually non-blocking, just a channel insert
```
antekresic, line 233: Why is this removed?

antekresic (approved):
Approved with a question.

PR #1370, opened 2022-05-23
Manage stopping of ingest components

niksajakovljevic:
We make sure to keep the state of ingest components and to block
race condition where component might be stopped but we still try
to ingest.

We also don't allow for closing of already closed components as that
can cause panic.

For rule manager we  wait for all group actors to stop before returning. Otherwise
a race condition might happen as sub components might continue running.

### DIFF
--- a/pkg/pgmodel/ingestor/dispatcher.go
+++ b/pkg/pgmodel/ingestor/dispatcher.go
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	uber_atomic "go.uber.org/atomic"
 
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgmodel/cache"
@@ -28,6 +29,8 @@ const (
 	finalizeMetricCreation = "CALL _prom_catalog.finalize_metric_creation()"
 	getEpochSQL            = "SELECT current_epoch FROM _prom_catalog.ids_epoch LIMIT 1"
 )
+
+var ErrDispatcherClosed = fmt.Errorf("dispatcher is closed")
 
 // pgxDispatcher redirects incoming samples to the appropriate metricBatcher
 // corresponding to the metric in the sample.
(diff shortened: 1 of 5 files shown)


pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -167,6 +170,9 @@ func (p *pgxDispatcher) getServerEpoch() (model.SeriesEpoch, error) {
 }
 
 func (p *pgxDispatcher) CompleteMetricCreation(ctx context.Context) error {
+	if p.closed.Load() {
+		return fmt.Errorf("dispatcher is closed")
```
Harkishen-Singh, line 174: This error is repeated. I think we should declare as follows. This allows the caller to use the error in some future case.
```golang
var ErrDispatcherClosed = fmt.Errorf("dispatcher is closed")
```

pkg/pgmodel/ingestor/ingestor_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/timescale/promscale/pkg/pgmodel/common/errors"
 	"github.com/timescale/promscale/pkg/pgmodel/model"
 	"github.com/timescale/promscale/pkg/prompb"
+	"go.uber.org/atomic"
```
Harkishen-Singh, line 16: Can we place this above promscale imports?

Harkishen-Singh (changes requested):

Harkishen-Singh (approved):

pkg/pgmodel/ingestor/dispatcher.go
```diff
@@ -256,6 +271,9 @@ func (p *pgxDispatcher) InsertTs(ctx context.Context, dataTS model.Data) (uint64
 }
 
 func (p *pgxDispatcher) InsertMetadata(ctx context.Context, metadata []model.Metadata) (uint64, error) {
+	if p.closed.Load() {
+		return 0, fmt.Errorf("dispatcher is closed")
```
Harkishen-Singh, line 275:
```suggestion
		return 0, ErrDispatcherClosed
```

antekresic (approved):
lgtm

PR #1371, opened 2022-05-24
Retrospectively documenting an unforeseen effect of HTTP router change.

sumerman:
## Description

A user reported change in behaviour of `/healthz` endpoint which affected their load balancing.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ We use the following categories for changes:
 - The Promscale extension is now required, while the Timescaledb extension remains optional. The minimum Timescaledb version supported is now 2.6.1 [#1132], [#1297]
 - Disable running Promscale in HA and read-only simultaneously [#1254]
 - Metric tables and views are now owned by prom_admin [#1283]
+- The only HTTP verb `/healthz` endpoint supports is `GET`, because HTTP router was changed in [#1224].
 
 ### Fixed
 - Register `promscale_ingest_channel_len_bucket` metric and make it a gauge [#1177]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ We use the following categories for changes:
 - The Promscale extension is now required, while the Timescaledb extension remains optional. The minimum Timescaledb version supported is now 2.6.1 [#1132], [#1297]
 - Disable running Promscale in HA and read-only simultaneously [#1254]
 - Metric tables and views are now owned by prom_admin [#1283]
-- The only HTTP verb `/healthz` endpoint supports is `GET`, because HTTP router was changed in [#1224].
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ We use the following categories for changes:
 - The Promscale extension is now required, while the Timescaledb extension remains optional. The minimum Timescaledb version supported is now 2.6.1 [#1132], [#1297]
 - Disable running Promscale in HA and read-only simultaneously [#1254]
 - Metric tables and views are now owned by prom_admin [#1283]
+- HTTP OPTION verb is no longer supported on `/healthz` endpoint because HTTP router was changed in [#1224].
```
JamesGuthrie, line 36: I would even phrase this the other way around: The new behaviour is that _only_ GET is allowed. Previously HEAD/OPTIONS/PUT/POST would all work.

JamesGuthrie (changes requested):

JamesGuthrie (approved):

niksajakovljevic:
I believe this can be easily fixed. This was obviously not a planned change but a missed bug so I am not sure it should go retroactively into CHANGELOG. So maybe we can just create a bug issue instead?

sumerman:
> I believe this can be easily fixed. This was obviously not a planned change but a missed bug so I am not sure it should go retroactively into CHANGELOG. So maybe we can just create a bug issue instead?

Whether or not we consider this a bug, it needs to be in the changelog and release notes because it already affects and confuses existing users. (See https://timescaledb.slack.com/archives/C011FC0PPC5/p1653384295441989?thread_ts=1652775715.500199&cid=C011FC0PPC5)

CHANGELOG.md
niksajakovljevic: I believe all paths are affected not only `/healthz`

niksajakovljevic:
@sumerman Imho change log should contain notable changes and not bugs we discover later on. This is not something that we wanted for our users. Btw also note that this problem probably affects all endpoints

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ We use the following categories for changes:
 - The Promscale extension is now required, while the Timescaledb extension remains optional. The minimum Timescaledb version supported is now 2.6.1 [#1132], [#1297]
 - Disable running Promscale in HA and read-only simultaneously [#1254]
 - Metric tables and views are now owned by prom_admin [#1283]
+- The only HTTP verb `/healthz` endpoint supports is `GET`, because HTTP router was changed in [#1224].
```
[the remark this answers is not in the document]
Harkishen-Singh: But i agree what Niksa said

niksajakovljevic:
A fix was done in https://github.com/timescale/promscale/pull/1378 . We will make a bugfix release

PR #1375, opened 2022-05-25
Fix panic in prom-migrator proto-enum registration

Harkishen-Singh:

This commit fixes the duplicate proto enum registration panic
in prom-migrator. This is done by removing all imports from Promscale
which prevents the duplicate registration of the commonly imported
`prompb` package by both Promscale and Prom-migrator.

## Description

## Merge requirements


migration-tool/cmd/prom-migrator/main.go
```diff
@@ -35,6 +35,7 @@ const (
 	defaultStartTime       = "1970-01-01T00:00:00+00:00" // RFC3339 based time.Unix from 0 seconds.
 	defaultMaxReadDuration = time.Hour * 2
 	defaultLaIncrement     = time.Minute
+	version                = "0.0.3"
```
niksajakovljevic, line 38: Shouldn't we update the version number?

migration-tool/pkg/log/log.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+var (
```
niksajakovljevic, line 17: Can you please add a comment that this is a copy/paste of the same file from Promscale and that hopefully in the future we find some better way to handle this.

migration-tool/pkg/utils/utils.go
```diff
@@ -6,28 +6,23 @@ package utils
...
 )
 
 var (
-	labelsCache *clockcache.Cache
+	labelsCache sync.Map
 	seps        = []byte{'\xff'}
 )
 
-func init() {
-	labelsCache = clockcache.WithMax(maxLabelsCacheSize)
```
niksajakovljevic, line 28: Why are we changing labels cache in this PR?

niksajakovljevic: Ahh I see that we can't import Promscale. Iirc clockcache is sort of  LRU cache so this change can lead to different performance of prom-migrator? Maybe consider using some 3rd party LRU cache impl with limits in cache size to not run oom.

Harkishen-Singh: The perf won't be a big deal, since now we are caching everything. Plus, a single-label entry is pretty light.

migration-tool/cmd/prom-migrator/main.go
Harkishen-Singh: Right, I missed that. Thank you.

migration-tool/pkg/utils/utils.go
niksajakovljevic: We are currently limiting cache size to 50K and using LRU policy to evict, so it's much different from proposed change. I'd recommend going for LRU cache with the same size: eg https://github.com/hashicorp/golang-lru

niksajakovljevic (approved):

antekresic (approved):
:+1:

PR #1384, opened 2022-05-27
Fix Jaeger trace querying by converted Jaeger tags

antekresic:
## Description

When translating Jaeger traces to OTEL format, some Jaeger tags get
converted to the to OTEL specific fields and gets removed from the tag map. This
change takes this conversion into account when querying traces using these
tags.

## Merge requirements


pkg/jaeger/query/trace_query.go
```diff
@@ -206,6 +215,14 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 	}
 
 	if len(q.Tags) > 0 {
+		if tag, code := checkForStatusCodeTags(q.Tags); tag != "" {
+			// Tag needs to be removed so we don't create an additional qual later on.
+			delete(q.Tags, tag)
```
niksajakovljevic, line 220: I'd not recommend modifying trace query parameters as that's a source of truth

pkg/jaeger/query/trace_query.go
```diff
@@ -108,6 +109,14 @@ const (
 	`
 )
 
+var (
+	statusCodeTagValueToDBValue = map[string]string{
+		"OK":    "ok",
```
niksajakovljevic, line 114: I bet otel defines these so let's maybe use defined values from the library? We can uppercase/downcase if needed as well.

pkg/jaeger/query/trace_query.go
```diff
@@ -262,3 +279,20 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 	}
 	return query, params
 }
+
+// checkForStatusCodeTags checks the tags map for tags which contain status code values
+// and returns the tag that was found with the value of the status code as represented
+// in the database.
+// This was introduced as part of Jaeger to OTEL conversion logic:
+// https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a13030ca6208fa7d4477f0805f3b95e271f32b24/pkg/translator/jaeger/jaegerproto_to_traces.go#L218
+func checkForStatusCodeTags(tags map[string]string) (string, string) {
```
niksajakovljevic, line 288: I am not sure we are doing the right thing here. It seems that otel always convert `tracetranslator.TagError` into span status, so I believe we should check for map if `tracetranslator.TagError` is present and make respective qual.

niksajakovljevic: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/tracetranslator/protospan_translation.go#L21 seem to contain all the translations otel is doing and I believe we should make sure we got it all covered.

antekresic: That particular case is not even checked because its only translated to status code if you use a boolean value but the query parameters are always strings here. 

Not sure if the assumption to treat the string `true` as a boolean value here holds up.

niksajakovljevic: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a13030ca6208fa7d4477f0805f3b95e271f32b24/pkg/translator/jaeger/jaegerproto_to_traces.go has few of transalator functions as well (eg. `translateHostnameAttr`) I guess we should make sure that is taken into account when querying (maybe it is but I am not fully confident)

pkg/tests/end_to_end_tests/ingest_trace_test.go
```diff
@@ -207,6 +207,35 @@ func findTraceTest(t testing.TB, q *query.Query) {
 	require.NoError(t, err)
 	require.Equal(t, 1, len(traces))
 
+	// Verify that we find traces by status code based on translated tag.
+	request = &spanstore.TraceQueryParameters{
+		ServiceName:   "service-name-0",
+		OperationName: "operationA",
+		Tags: map[string]string{
+			"error": "true",
```
niksajakovljevic, line 215: Consider testing other tags like `kind` and `space`. Also worth testing `error = false`

pkg/jaeger/query/trace_query.go
```diff
@@ -207,6 +211,25 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 
 	if len(q.Tags) > 0 {
 		for k, v := range q.Tags {
+			switch k {
```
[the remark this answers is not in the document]
antekresic: Yes, let me add those transformations as well.

niksajakovljevic (approved):

pkg/jaeger/query/trace_query.go
```diff
@@ -192,6 +200,54 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 		operation_clauses = append(operation_clauses, qual)
 	}
 
+	if len(q.Tags) > 0 {
```
Harkishen-Singh, line 203: I also wonder, this function is getting long. Should we have a different function for tag work? The fund name `buildTraceIDSubquery` seems verbose with tags block so long

PR #1385, opened 2022-05-30
Fix trace query issues

harkishen:

This commit fixes 2 issues in trace querying. They are:
1. Query returns no result when queried with a numeric value tag like `http.status_code=200`
2. Query returns no result when queried with a tag that is from Process table in Jaeger UI

## Merge requirements


Harkishen-Singh:
> Does this also fix booleans? For example error=true

This is fixed by https://github.com/timescale/promscale/pull/1384

> won't this break attribute values that are strings, but match the integer regex?

This fix will treat the input as numeric only if the entire value part in the tag queried is numeric. If a user sets a tag value as numeric through a string type in Otel traces, then I think the database will still store it as numeric. Hence, this fix should work.

pkg/jaeger/query/trace_query.go
```diff
@@ -237,8 +240,22 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
...
+					// generic and respects integers, booleans, etc.
+					val = v
+				}
+				params = append(params, k, val)
+				// Note: We do not need to check resource tags in above cases, since they
+				// come from Jaeger conversion that are specific to span_tags.
+				qual := fmt.Sprintf(
+					`(_ps_trace.tag_map_denormalize(s.span_tags)->$%d = $%d OR _ps_trace.tag_map_denormalize(s.resource_tags)->$%d = $%d)`,
+					len(params)-1, len(params),
+					len(params)-1, len(params))
```
antekresic, line 258:
```suggestion
					`(_ps_trace.tag_map_denormalize(s.span_tags)->$%[1]d = $%[2]d OR _ps_trace.tag_map_denormalize(s.resource_tags)->$%[1]d = $%[2]d)`,
					len(params)-1, len(params))
```

antekresic (approved):
Approved with a small suggestion.

pkg/jaeger/query/trace_query.go
```diff
@@ -237,8 +240,22 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 				clauses = append(clauses, qual)
 			default:
 				//TODO make sure this is optimized correctly
-				params = append(params, k, "\""+v+"\"")
-				qual := fmt.Sprintf(`_ps_trace.tag_map_denormalize(s.span_tags)->$%d = $%d`, len(params)-1, len(params))
+				val := "\"" + v + "\""
```
niksajakovljevic, line 243: Nit:
```
if digitCheck.MatchString(v) 
  val = v
else
 val = "\"" + v + "\""

File: pkg/jaeger/query/trace_query.go [hunk 3]
```diff
@@ -237,8 +240,22 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
...
+				if digitCheck.MatchString(v) {
```
niksajakovljevic, line 244: Do we make sure that value is trimmed? Otherwise we will not match: ` 1`, `1 ` or ` 1 `

niksajakovljevic: Also it seems we will not match `float` so I believe regex should be fixed

niksajakovljevic: Have you maybe tested what happens with booleans?

niksajakovljevic (commented):
Overall looks good. Check comment about float and value trim.
It would be nice to add some e2e test (or tweak existing one) where we can make sure that this works as expected.

pkg/jaeger/query/trace_query.go
Harkishen-Singh: Subjective. The current one is less blocks so will be happy to continue with that.

pkg/jaeger/query/trace_query.go
Harkishen-Singh: I don't think trimming will be good. We should use it as we get from jaeger.

sumerman (approved):

niksajakovljevic (approved):
LGTM. For completeness I'd recommend changing regex to something like `^\d*\.?\d+$` to support floats

PR #1388, opened 2022-05-31
Add documentation of Promscale metric data model

JamesGuthrie:
## Description

On a number of occasions I have wanted concise documentation of the metric data storage. This is it.

## Merge requirements

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -1,0 +1,39 @@
+# Promscale Data Storage
+
+This document provides the foundations to understand how Promscale stores data.
+
+## Metrics
+
+The most important components of the metric data model are as follows:
+
+```
+                            _prom_catalog
+      ┌───────────────────────────────────────────────────────┐
+      │                                                       │
+      │   ┌──────────┐       ┌──────────┐      ┌──────────┐   │
+      │   │  metric  │◄──────┤  series  ├─────►│  label   │   │
+      │   └──────────┘       └────▲─────┘      └──────────┘   │
+      │                           │                           │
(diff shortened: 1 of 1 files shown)


docs/data-model.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+The `_prom_catalog` schema contains metadata tables which describe `metrics`,
+`label`s, and `series`. Each metric stored in the system is represented through
+an entry in the `metric` table. The `label` table contains (key, value) pairs
+with identity. The `series` table represents the combination of a `metric`, and
+a list of `label`s.
+
+The actual metric data is stored in a TimescaleDB hypertable (one per metric)
+with the same name as the metric. These hypertables are in the `prom_data`
```
Harkishen-Singh, line 33: The name is same as metric only if less than some 64 chars limit that Postgres has. In long metric names, it can be some other name. Maybe write this as a note?

Harkishen-Singh (approved):
Just one point, otherwise, this doc gives a good high-level idea.

VineethReddy02:
Do we need to explain the views offered for querying? I mean what's the preferred source of querying data from the data model as a heads-up?

docs/data-model.md
JamesGuthrie: Done.

JamesGuthrie:
> Do we need to explain the views offered for querying? I mean what's the preferred source of querying data from the data model as a heads-up?

I don't think so. The purpose of this is purely to describe how the data is stored.

PR #1414, opened 2022-06-06
Cache label IDs

niksajakovljevic:
Add inverted cache ( (metric + label pair) -> (id, pos)) to avoid DB calls for
fetching label IDs in cases when series ID is not cached.
This cache is only used on metric ingestion.
Benchmarks are showing around 5-10% gains in ingest performance and
about 25% less DB calls for fetching label IDs (note that these numbers
depend a lot on a shape of the dataset).


cevian:
@niksajakovljevic have you considered using (label pair)=> (id, map[metric_name]=>pos) instead? That would allow reusing the existing cache and thus improve overall cache hit ratio?

pkg/pgclient/config.go
```diff
@@ -82,6 +83,7 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 	fs.BoolVar(&cfg.EnableStatementsCache, "db.statements-cache", defaultDbStatementsCache, "Whether database connection pool should use cached prepared statements. "+
 		"Disable if using PgBouncer")
 	fs.BoolVar(&cfg.AsyncAcks, "metrics.async-acks", false, "Acknowledge asynchronous inserts. If this is true, the inserter will not wait after insertion of metric data in the database. This increases throughput at the cost of a small chance of data loss.")
+	fs.IntVar(&cfg.InvertedLabelsCacheSize, "metrics.inverted-lables-cache-size", 500000, "Maximum number of labelIDs to cache. Inverted lables cache reduces number of DB calls when ingesting new series.")
```
Harkishen-Singh, line 86:
```suggestion
	fs.IntVar(&cfg.InvertedLabelsCacheSize, "metrics.cache.label-ids.size", 500000, "Maximum number of label-ids to cache. This reduces number of DB calls when ingesting new series.")
```
We should abstract the invert indexing details from the naming. Plus, going with the convention of other cache nomenclature, it should follow.

The details regarding inverted index can be added as code comments for others to follow.

pkg/pgmodel/cache/inverted_labels_cache.go
```diff
@@ -0,0 +1,63 @@
+package cache
+
+import (
+	"fmt"
+
+	"github.com/dgraph-io/ristretto"
+)
+
+type LabelInfo struct {
```
Harkishen-Singh, line 9:
```suggestion
type LabelPos struct {
```

pkg/pgmodel/cache/inverted_labels_cache.go
```diff
@@ -0,0 +1,63 @@
...
+	MetricName, Name, Value string
+}
+
+func NewLabelKey(metricName, name, value string) LabelKey {
+	return LabelKey{MetricName: metricName, Name: name, Value: value}
+}
+
+// Cache hash function needs string to work with
+func (lk LabelKey) String() string {
+	return fmt.Sprintf("%s#%s=%s", lk.MetricName, lk.Name, lk.Value)
```
Harkishen-Singh, line 24: I think we should use string builder here. This will be a perf hot path.

pkg/pgmodel/cache/inverted_labels_cache.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func NewLabelInfo(lableID, pos int32) LabelInfo {
+	return LabelInfo{LabelID: lableID, Pos: pos}
+}
+
+// Label key-pair -> (id,pos) cache
+// Used when creating series to avoid DB calls for labels
+type InvertedLabelsCache struct {
+	cache *ristretto.Cache
```
Harkishen-Singh, line 34: Why a new cache library? Why not use the clock cache?

pkg/pgmodel/cache/inverted_labels_cache.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Cache is thread-safe
+func NewInvertedLablesCache(size int) (*InvertedLabelsCache, error) {
+	if size == 0 {
+		return nil, fmt.Errorf("labels cache size must be > 0")
+	}
+	cache, _ := ristretto.NewCache(&ristretto.Config{ // TODO: pass cache config
```
Harkishen-Singh, line 42: If we are returning error, its good to check the error here too. This will protect from future implementation changes in the new cache lib.

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -68,9 +63,10 @@ func (h *seriesWriter) WriteSeries(ctx context.Context, sv SeriesVisitor) error
 			info, ok := infos[metricName]
```
Harkishen-Singh, line 63: `WriteSeries` is getting complex to understand. I feel we should break down this function at some point.

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -212,7 +214,7 @@ func (h *seriesWriter) fillLabelIDs(ctx context.Context, infos map[string]*perMe
 	infoBatches := make([]*perMetricInfo, 0)
 	items := 0
 	for metricName, info := range infos {
-		names, values := info.labelList.Get()
+		names, values := info.labelsToGetOrCreate.Get()
```
Harkishen-Singh, line 217: This looks weird. The `.Get()` of `..GetOrCreate` returns `names` and `values`? I think the function needs renaming.

Harkishen-Singh (changes requested):
1. Why are we not using clockcache?
2.

pkg/pgmodel/cache/inverted_labels_cache.go
antekresic: I agree here, we should either stick to the one we use everywhere or evaluate if we need a different cache (unlikely).

pkg/pgclient/config.go
niksajakovljevic: Inverted is intentionally there because we already have labels cache ( `id` -> `label`).

pkg/pgmodel/cache/inverted_labels_cache.go
niksajakovljevic: Imho `labelInfo` is more appropriate as it contains both label id + position. Also it is in line with previous naming.

pkg/pgmodel/cache/inverted_labels_cache.go
niksajakovljevic: I was actually thinking to replace our own caching impl with some of more popular caching libs. I've did some benchmarking and found that performance are pretty much the same so yeah we can keep using our own caching...

pkg/pgmodel/ingestor/series_writer.go
niksajakovljevic: True. Renamed to `labelsToFetch` - this is essentially list of labels we haven't found in our cache and need to fetch from DB

niksajakovljevic:
>@niksajakovljevic have you considered using (label pair)=> (id, map[metric_name]=>pos) instead? That would allow reusing the existing cache and thus improve overall cache hit ratio?

Labels reader cache is actually inverse `id -> label pair` so we can't reuse it.

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -100,12 +96,18 @@ func (h *seriesWriter) WriteSeries(ctx context.Context, sv SeriesVisitor) error
...
+						labelInfo, cached := h.labelsCache.GetLabelsId(cache.NewLabelKey(series.MetricName(), names[i], values[i]))
+						if cached {
+							info.cachedLabels = append(info.cachedLabels, key)
+						}
+						labelMap[key] = labelInfo
+						if !cached {
+							if err := info.labelsToFetch.Add(names[i], values[i]); err != nil {
+								return fmt.Errorf("failed to add label to labelList: %w", err)
+							}
 						}
```
antekresic, line 111:
```suggestion
						labelInfo, cached := h.labelsCache.GetLabelsId(cache.NewLabelKey(series.MetricName(), names[i], values[i]))
						labelMap[key] = labelInfo
						if cached {
							info.cachedLabels = append(info.cachedLabels, key)
							continue
						}
						if err := info.labelsToFetch.Add(names[i], values[i]); err != nil {
							return fmt.Errorf("failed to add label to labelList: %w", err)
						}
```

pkg/pgmodel/ingestor/series_writer.go
```diff
@@ -235,7 +237,16 @@ func (h *seriesWriter) fillLabelIDs(ctx context.Context, infos map[string]*perMe
 			infoBatches = append(infoBatches, info)
 			items += len(namesSlice.Elements)
 		}
+		// since info.maxPos is only updated with fetched labels we need to iterate through cached as well
+		for _, cachedLabel := range info.cachedLabels {
+			if val, ok := labelMap[cachedLabel]; ok {
+				if int(val.Pos) > info.maxPos {
+					info.maxPos = int(val.Pos)
```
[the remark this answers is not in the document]
niksajakovljevic: Yeah we have to do it. Since some labels are cached and we need `maxPos` to be correct (meaning it contains the max for both fetched and cached labels)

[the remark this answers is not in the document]
antekresic: No, I meant, can you start calculating the max position when fetching cached entries, so you don't have to iterate again through the cached labels.

antekresic (approved):

pkg/pgmodel/cache/inverted_labels_cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewLabelInfo(lableID, pos int32) LabelInfo {
+	return LabelInfo{LabelID: lableID, Pos: pos}
+}
+
+func (li LabelInfo) len() int {
+	return 8
+}
+
+// Label key-pair -> (id,pos) cache
```
Harkishen-Singh, line 34: What is this pos? We should add some comments explaining that id, and pos means. is id from the label_key table, or is it from label table, similarly for pos.

Harkishen-Singh (changes requested):

pkg/pgmodel/cache/flags.go
```diff
@@ -47,11 +49,12 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
...
 	fs.Uint64Var(&cfg.SeriesCacheInitialSize, "metrics.cache.series.initial-size", DefaultSeriesCacheSize, "Maximum number of series to cache.")
-	fs.Uint64Var(&cfg.LabelsCacheSize, "metrics.cache.labels.size", 10000, "Maximum number of labels to cache.")
+	fs.Uint64Var(&cfg.LabelsCacheSize, "metrics.cache.labels.size", 100000, "Maximum number of labels to cache.")
 	fs.Uint64Var(&cfg.ExemplarKeyPosCacheSize, "metrics.cache.exemplar.size", DefaultExemplarKeyPosCacheSize, "Maximum number of exemplar metrics key-position to cache. "+
 		"It has one-to-one mapping with number of metrics that have exemplar, as key positions are saved per metric basis.")
 	fs.Var(&cfg.seriesCacheMemoryMaxFlag, "metrics.cache.series.max-bytes", "Initial number of elements in the series cache. "+
 		"Specified in bytes or as a percentage of the memory-target (e.g. 50%).")
+	fs.Uint64Var(&cfg.InvertedLabelsCacheSize, "metrics.cache.inverted-labels.size", 500000, "Maximum number of label-ids to cache. Inverted lables cache reduces the number of DB calls when ingesting new series.")
```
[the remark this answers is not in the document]
niksajakovljevic: The default is pretty solid so users will most probably not need to touch it, however we need to allow changing it.

pkg/pgmodel/cache/inverted_labels_cache.go
niksajakovljevic: Sure, will add some comments

Harkishen-Singh (approved):

PR #1421, opened 2022-06-08
Fix upgrade tests around version 0.11.0

JamesGuthrie:
## Description

In version 0.11.0 we introduced a hard dependency on the promscale
extension at version 0.5.0. Additionally, we renamed some flags in
0.10.0, and dropped support for the old flags in 0.11.0.

This change adds logic to the upgrade tests to work around these two
facts.

## Merge requirements

### DIFF
--- a/pkg/tests/upgrade_tests/upgrade_test.go
+++ b/pkg/tests/upgrade_tests/upgrade_test.go
@@ -71,7 +71,7 @@ func TestMain(m *testing.M) {
 }
 
 /* Prev image is the db image with the old promscale extension. We do NOT test timescaleDB extension upgrades here. */
-func getDBImages(extensionState testhelpers.TestOptions) (prev string, clean string, err error) {
+func getDBImages(extensionState testhelpers.TestOptions, prevPromscaleVersion *semver.Version) (prev string, clean string, err error) {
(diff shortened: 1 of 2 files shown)


antekresic (approved):

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -86,6 +86,11 @@ func getDBImages(extensionState testhelpers.TestOptions) (prev string, clean str
 		panic("Only use pg12 for upgrade tests")
 	}
 
+	// From Promscale 0.11.0 onwards the minimum extension version is 0.5.0
+	if prevPromscaleVersion != nil && prevPromscaleVersion.GE(semver.MustParse("0.11.0")) {
```
niksajakovljevic, line 90: I don't understand why can't we use `2.6` with `0.10` ? Is that because `ts2.6` always packs `0.5.0` promscale extension?

JamesGuthrie: We want to test the upgrade from the oldest promscale extension version, which is in the `ts2.1` ha docker image.

niksajakovljevic (approved):

PR #1422, opened 2022-06-08
Tweaks for local go bench

niksajakovljevic:
Allow loading the whole test data set into memory which can be useful when
benchmarking memory allocations. It also helps avoid any i/o related issues 
when loading chunk data in streaming fashion (Prometheus uses mmap files), although
I haven't see any so far (the numbers are just slightly better when ingesting from memory)

Don't block on sending ingest requests while creating batches which improves
data load speed.

I've also tried using sync Pool to ease GC, however it slowed down the benchmark so I gave up on that (which makes sense since sync.Pool requres more CPU work then just creating new instance)

### DIFF
--- a/pkg/tests/testsupport/metric_loader.go
+++ b/pkg/tests/testsupport/metric_loader.go
@@ -21,14 +21,36 @@ type PromLoader interface {
 }
 
 type promLoader struct {
-	db     *tsdb.DBReadOnly
-	blocks []tsdb.BlockReader
+	db       *tsdb.DBReadOnly
+	blocks   []tsdb.BlockReader
+	inMemory bool
 }
 
 // PromIterator allows us to iterate over Prometheus data
 type PromIterator interface {
 	Next() bool
 	Get() TimeSeries
+}
+
+type inMemoryIterator struct {
+	data   []TimeSeries
+	curIdx int
+}
+
+func (s *inMemoryIterator) Next() bool {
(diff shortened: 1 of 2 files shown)


pkg/tests/testsupport/metric_loader.go
```diff
@@ -200,29 +232,40 @@ func (si *sampleIngestor) shardSamples() {
...
 			}
 		}
 	}()
 }
 
 func (si *sampleIngestor) ingestSamples(ingest IngestFunc) {
-	var wg sync.WaitGroup
+	var shardWg sync.WaitGroup
+	var ingestWg sync.WaitGroup
+	reqCh := make(chan prompb.WriteRequest, 100)
```
antekresic, line 245: So I'm not sure why the shards are in separate goroutines... IMO this channel should be the size of the number of shards and you spin up the ingestor goroutines and then load shards into the channel sequentially. Close the channel to stop the ingestors and then you need only one waitgroup to wait on.

niksajakovljevic: We need parallel workers on `reqCh` because otherwise things will run sequentially meaning that one shard will block another. Also we want to prepare batches for each shard in their own routines. I hope this makes it more clear or am I missing something?

Harkishen-Singh (approved):
On high level, looks good. Will leave the depth review to Ante.

antekresic (approved):

PR #1424, opened 2022-06-09
Adds support for telemetry for alerting and recording rules.

Harkishen-Singh:

Fixes: https://github.com/orgs/timescale/projects/30/views/38

## Merge requirements

### DIFF
--- a/pkg/rules/config.go
+++ b/pkg/rules/config.go
@@ -31,20 +31,11 @@ type Config struct {
 }
 
 func (cfg *Config) ContainsRules() bool {
-	if cfg.PrometheusConfig == nil {
-		return false
-	}
-	promCfg := cfg.PrometheusConfig
-	if len(promCfg.RuleFiles) == 0 {
-		log.Info("msg", "Rules files not found in the given Prometheus configuration file. Both rule-manager and alerting will not be initialized")
-		return false
-	}
+	return cfg.PrometheusConfig != nil && len(cfg.PrometheusConfig.RuleFiles) != 0
+}
 
-	if promCfg.AlertingConfig.AlertRelabelConfigs == nil && promCfg.AlertingConfig.AlertmanagerConfigs == nil {
-		// This is true when alerting config is not applied.
-		log.Info("msg", "Alerting configuration not present in the given Prometheus configuration file. Alerting will not be initialized")
-	}
-	return true
+func (cfg *Config) ContainsAlertingConfig() bool {
+	return cfg.PrometheusConfig != nil && len(cfg.PrometheusConfig.AlertingConfig.AlertmanagerConfigs) != 0
 }
 
 func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -222,7 +222,18 @@ func Run(cfg *Config) error {
 		},
 	)
 
+	pendingTelemetry := map[string]string{
+		"rules_enabled":    "false", // Will be written in telemetry table as `promscale_rules_enabled: false`
+		"alerting_enabled": "false", // `promscale_alerting_enabled: false`
+	}
 	if cfg.RulesCfg.ContainsRules() {
+		pendingTelemetry["rules_enabled"] = "true"
+		if cfg.RulesCfg.ContainsAlertingConfig() {
+			pendingTelemetry["alerting_enabled"] = "true"
+		} else {
+			log.Debug("msg", "Alerting configuration not present in the given Prometheus configuration file. Alerting will not be initialized")
+		}
+
 		rulesCtx, stopRuler := context.WithCancel(context.Background())
 		defer stopRuler()
 
@@ -245,7 +256,16 @@ func Run(cfg *Config) error {
 				stopRuler()
 			},
 		)
+	} else {
+		log.Debug("msg", "Rules files not found in the given Prometheus configuration file. Both rule-manager and alerting will not be initialized")
 	}
+
+	// Asynchronously update the telemetry information.
+	go func() {
+		for k, v := range pendingTelemetry {
+			telemetryEngine.Write(k, v)
+		}
+	}()
 
 	mux := http.NewServeMux()
 	mux.Handle("/", router)
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -27,7 +27,11 @@ import (
 // that will be monitored every hour and filled into _ps_catalog.promscale_instance_information table and then into
 // the _timescaledb_catalog.metadata table.
 type Engine interface {
+	// RegisterMetric registers a Prometheus metric with a column name. This metric is
+	// monitored every telemetrySync and updated in the telemetry table.
 	RegisterMetric(columnName string, gaugeOrCounterMetric ...prometheus.Metric) error
+	// Write writes a telemetry information into the database instantly.
+	Write(key, value string)
 	Start()
 	Stop()
 }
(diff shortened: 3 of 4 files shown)


pkg/telemetry/telemetry.go
```diff
@@ -121,6 +125,10 @@ func (t *engineImpl) writeToTimescaleMetadataTable(m Metadata) {
 	}
 }
 
+func (t *engineImpl) Write(key, value string) {
+	t.writeToTimescaleMetadataTable(map[string]string{key: value})
```
niksajakovljevic, line 129: Write to DB can fail so I'd recommend returning and checking error

pkg/telemetry/telemetry.go
```diff
@@ -27,7 +27,11 @@ import (
 // that will be monitored every hour and filled into _ps_catalog.promscale_instance_information table and then into
 // the _timescaledb_catalog.metadata table.
 type Engine interface {
+	// RegisterMetric registers a Prometheus metric with a column name. This metric is
+	// monitored every telemetrySync and updated in the telemetry table.
 	RegisterMetric(columnName string, gaugeOrCounterMetric ...prometheus.Metric) error
+	// Write writes a telemetry information into the database instantly.
+	Write(key, value string)
```
niksajakovljevic, line 34: Think about propagating error if write fails

pkg/rules/config.go
```diff
@@ -31,20 +31,17 @@ type Config struct {
...
+		return true
 	}
+	return false
+}
 
-	if promCfg.AlertingConfig.AlertRelabelConfigs == nil && promCfg.AlertingConfig.AlertmanagerConfigs == nil {
-		// This is true when alerting config is not applied.
-		log.Info("msg", "Alerting configuration not present in the given Prometheus configuration file. Alerting will not be initialized")
+func (cfg *Config) ContainsAlertingConfig() bool {
+	if cfg.PrometheusConfig != nil && len(cfg.PrometheusConfig.AlertingConfig.AlertmanagerConfigs) != 0 {
```
niksajakovljevic, line 41: If `cfg.PrometheusConfig.AlertingConfig` is nil this will crash

pkg/runner/runner.go
```diff
@@ -245,8 +257,17 @@ func Run(cfg *Config) error {
 				stopRuler()
 			},
 		)
+	} else {
+		log.Info("msg", "Rules files not found in the given Prometheus configuration file. Both rule-manager and alerting will not be initialized")
 	}
 
+	// Asynchronously update the telemetry information.
+	go func() {
```
niksajakovljevic, line 265: How will this work with hot-reloading of rules? Or maybe you plan to refactor later on?

pkg/rules/config.go
```diff
@@ -31,20 +31,17 @@ type Config struct {
...
-		return false
-	}
-	promCfg := cfg.PrometheusConfig
-	if len(promCfg.RuleFiles) == 0 {
-		log.Info("msg", "Rules files not found in the given Prometheus configuration file. Both rule-manager and alerting will not be initialized")
-		return false
+	if cfg.PrometheusConfig != nil && len(cfg.PrometheusConfig.RuleFiles) != 0 {
+		return true
 	}
+	return false
```
antekresic, line 37:
```suggestion
	return cfg.PrometheusConfig != nil && len(cfg.PrometheusConfig.RuleFiles) > 0 
```

pkg/runner/runner.go
```diff
@@ -222,7 +222,19 @@ func Run(cfg *Config) error {
 		},
 	)
 
+	pendingTelemetry := map[string]string{
+		"recording_enabled": "false", // Will be written in telemetry table as `promscale_recording_enabled: false`
+		"alerting_enabled":  "false", // `promscale_alerting_enabled: false`
+	}
 	if cfg.RulesCfg.ContainsRules() {
+		pendingTelemetry["recording_enabled"] = "true"
+		pendingTelemetry["alerting_enabled"] = "false"
```
antekresic, line 231: Is this line necessary since we do this 4 lines above it?

pkg/rules/config.go
Harkishen-Singh: No it won't. In AND conditions, go does not move ahead with the next logic if the first one is false.

pkg/telemetry/telemetry.go
Harkishen-Singh: There is no error returned for metadata. Plus, writing a telemetry if errors should not be reported ideally.

pkg/telemetry/telemetry.go
Harkishen-Singh: The child functions do not return error as its telemetry. Hence, no propagation.

pkg/runner/runner.go
```diff
@@ -222,7 +222,19 @@ func Run(cfg *Config) error {
 		},
 	)
 
+	pendingTelemetry := map[string]string{
+		"recording_enabled": "false", // Will be written in telemetry table as `promscale_recording_enabled: false`
+		"alerting_enabled":  "false", // `promscale_alerting_enabled: false`
+	}
 	if cfg.RulesCfg.ContainsRules() {
```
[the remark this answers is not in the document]
Harkishen-Singh: It should be `ContainsRules()`, as the rules can be either recording rules or alerting rules.

> is it possible that given file exists but does not contain recording rules?

We never open the rules files. It is handled by the upstream modules of rules manager that we import. Hence, none of those checks apply on us.

pkg/runner/runner.go
```diff
@@ -222,7 +222,19 @@ func Run(cfg *Config) error {
...
+		pendingTelemetry["recording_enabled"] = "true"
```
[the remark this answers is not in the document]
Harkishen-Singh: Again, we cannot check if it is a recording rules or alerting rules. We just know it is some rules. And from telemetry's perspective, we want to know if users are using rules or not. If there exists an alerting config, then there will obviously be a alerting rule, otherwise it makes no sense to have alerting config.

pkg/runner/runner.go
Harkishen-Singh: Right now, we are not thinking of telemetry with hot-reloading. Hot-reloading will be an architectural change, hence thinking right now is not good.

pkg/runner/runner.go
Harkishen-Singh: Right, we just need recording_enabled. alerting_enabled is already set to false.

pkg/runner/runner.go
```diff
@@ -222,7 +222,18 @@ func Run(cfg *Config) error {
...
+	pendingTelemetry := map[string]string{
+		"rules_enabled":    "false", // Will be written in telemetry table as `promscale_rules_enabled: false`
+		"alerting_enabled": "false", // `promscale_alerting_enabled: false`
+	}
 	if cfg.RulesCfg.ContainsRules() {
+		pendingTelemetry["rules_enabled"] = "true"
+		if cfg.RulesCfg.ContainsAlertingConfig() {
+			pendingTelemetry["alerting_enabled"] = "true"
+		} else {
+			log.Info("msg", "Alerting configuration not present in the given Prometheus configuration file. Alerting will not be initialized")
```
antekresic, line 234: I'm kinda wondering if this log line should be INFO or maybe a DEBUG line? We are so noisy as it is when starting up, we might want to hide this unless there is a problem with configuration detection.

pkg/runner/runner.go
```diff
@@ -245,8 +256,17 @@ func Run(cfg *Config) error {
 				stopRuler()
 			},
 		)
+	} else {
+		log.Info("msg", "Rules files not found in the given Prometheus configuration file. Both rule-manager and alerting will not be initialized")
```
antekresic, line 260: Same comment as above.

antekresic (approved):
Couple of thoughts on logging but looks good to me.

pkg/telemetry/telemetry.go
niksajakovljevic: I think we should at least log an error.

pkg/telemetry/telemetry.go
niksajakovljevic: Not sure I follow your logic. Imho any error should be at least logged. If our current api ignores error I think we should change it.

niksajakovljevic (approved):

pkg/runner/runner.go
Harkishen-Singh: Yeah, I think we should change this to debug. Will push this change.

Issue #1194, opened 2022-02-24
.../api/v1/label/__tenant__/values shows all tenants even though -multi-tenancy-valid-tenants=XYZ is set


oscni:
Is this some configuration issue, a bug or by design. I certainly don't want different tenants to be able to know about each other.
series/queries seems to work as expected.

Configuration options:
PROMSCALE_MULTI_TENANCY=true
PROMSCALE_MULTI_TENANCY_ALLOW_NON_TENANTS=false
PROMSCALE_MULTI_TENANCY_VALID_TENANTS=XYZ

Harkishen-Singh:
Hey @oscni, they look like the right options. Are you facing any error? The mentioned way will keep data of XYZ separate from other tenants (which are not present in this case).

oscni:
ok great then we know my fat fingers are not the problem :)
No errors is shown, and series queries seems to work in that sense that it only return data from the allowed tenant.
Problem in this issue is that when one customers lists all __tenant__ labels it return all the tenant/customer names and not only the one that is specified in valid_tenants option.

Harkishen-Singh:
Yes, if labels like '__tenant__=".*"' is applied, then it will return all valid tenants since for that promscale instance, it has been asked to have complete authority over the mentioned valid tenants.

If you want the customer to only see their part of tenant, you should run different promscale instances with the valid tenant being the ones you want to be shared with them the maximum.

oscni:
>If you want the customer to only see their part of tenant, you should run different promscale instances with the valid tenant being the ones you want to be shared with them the maximum.
Yepp read that in some documentation and that's what I've done. But it does only work the the actual data APIs, not when using the labels API and listing all the __tenant__ labels. In that case it shows all tenant names and not only the configured one.

>Seeing your recent comment, you mean that a promql query is returning more tenants than the ones specified in the valid tenants. Is this your point?
yes, data belonging to another tenant is correctly filtered out. But when listing all __tenant__ labels it show incorrectly all tenants names and not only the one configured.

Harkishen-Singh:
Ah, I think you are right. The labels API does not respect the tenants and that is a bug.

Thanks for pointing it out. I will mark this as bug and track in the sprint plannings.

oscni:
:+1: 
Thanks for the awesome work with promscale :)

ramonguiu:
[...]

PR #1427, opened 2022-06-13
List label values of allowed tenants only

niksajakovljevic:
If allowed tenants are configured we only list those when showing `__tenant__` values.

Fixes https://github.com/timescale/promscale/issues/1194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ We use the following categories for changes:
   - Tags from process table in Jaeger UI [#1385]
   - Tags that have a numeric value, like `http.status_code=200` [#1385]
   - Tags that involve status code [#1384]
+- List label values of allowed tenants only [#1427] 
 
 ## [0.11.0] - 2022-05-11
 
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -251,8 +251,9 @@ func Run(cfg *Config) error {
 	mux.Handle("/", router)
 
 	server := http.Server{
-		Addr:    cfg.ListenAddr,
-		Handler: mux,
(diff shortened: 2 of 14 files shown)


antekresic (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ We use the following categories for changes:
   - Tags from process table in Jaeger UI [#1385]
   - Tags that have a numeric value, like `http.status_code=200` [#1385]
   - Tags that involve status code [#1384]
+  - List label values of allowed tenants only [#1427]
```
Harkishen-Singh, line 27:
```suggestion
- List label values of allowed tenants only [#1427] 
```
This should have a different indentation as it is not part of jaeger fix.

pkg/runner/runner.go
```diff
@@ -251,8 +251,9 @@ func Run(cfg *Config) error {
 	mux.Handle("/", router)
 
 	server := http.Server{
-		Addr:    cfg.ListenAddr,
-		Handler: mux,
+		Addr:              cfg.ListenAddr,
+		Handler:           mux,
+		ReadHeaderTimeout: time.Second * 30, // To mitigate Slowloris DDoS attack. Value is arbitrary picked
```
Harkishen-Singh, line 256: Will this have any effect on write/read requests that take > 30s to execute?

Harkishen-Singh (changes requested):

pkg/runner/runner.go
niksajakovljevic: Btw this is a server setting, not a client. This means that client needs to send HTTP header within 30 seconds

Harkishen-Singh (approved):

PR #1430, opened 2022-06-14
Change `db.num-writer-connections` to set absolute number of writing connections

antekresic:
## Description

Previously, this flag was used to set a multiplier of Promscale CPUs as number
of writing connections. Changing this to set an absolute number of writing
connections so its more predictable and easier to reason about.

## Merge requirements


Harkishen-Singh (approved):
One nit, otherwise lgtm.

niksajakovljevic (approved):

PR #1435, opened 2022-06-16
Automatically update the timescaledb extension in tests

### LINKED ISSUE #1433: Handle TimescaleDB version jumps in `upgrade_test`
Our upgrade tests validate that upgrading the Promscale Connector and Extension result in the same DB state as a fresh install.

To do this, the test starts a database with an old version of the Promscale Extension and Promscale Connector, installs promscale, and then (using the same data volume) starts a database with the most recent version of the Promscale Extension and Connector, and runs through the upgrade. Note: The current old versions used are: extension version 0.1.2, connector version 0.3.0, and the `timescale/timescaledb-ha:pg<PGVERSION>-ts2.1-latest` docker image.

This test implicitly relies on having the same version of TimescaleDB available in the old docker image as the new. If this is not the case, the new database is unable to start because it can't find the TimescaleDB versioned shared object from the previous version. (The reason that this happens is that the TimescaleDB extension is preloaded, so it tries to load the shared object on start).

To a certain extent, this is a side effect of us being reliant on the `timescaledb-ha` docker image.

We see the following options to solve this problem:

1. Don't use TimescaleDB in the upgrade test
    This would cut off a large amount of (important) SQL from the upgrade test, which is undesirable.
2. Increase our "old versions" to support to avoid this problem
    This would reduce the surface area that we're testing, and we would still be dependent on which TimescaleDB versions are available in the `timescaledb-ha` docker image.
3. Build a custom Docker image for this test with the versions in that we want
    This should make the `upgrade_test` more stable, but puts us in the position of building another docker image, which we probably don't want, and which is going to a lot of work.
4. Make the upgrade test upgrade through intermediate containers to arrive at the final version
    While this option is also a fair amount of work, it does actually simulate a scenario that our users may find themselves in (e.g. I'm currently using Promscale 0.10.0 with TimescaleDB 2.5.0, how do I upgrade to Promscale 0.11.0?), and makes our upgrade test more robust (if somewhat slower).

After discussing the options above, we would opt for option 4 as it should allow us to better test the reliability of the upgrade process, and illustrates how the timescaledb upgrade process should work (when using `timescaledb-ha` docker images).

JamesGuthrie:
## Description

When we run the upgrade tests, we purposefully initialize a database
with an old timescale docker image, containing an old timescaledb
version. We then start a new version of the timescale docker image with
the same data.

If the old timescaledb version is not present in the new container, then
this can cause problems if we don't immediately run
`ALTER EXTENSION timescaledb UPDATE`.

This change updates timescaledb in all databases of the timescale
container immediately on startup.

Closes #1433

## Merge requirements

### DIFF
--- a/pkg/internal/testhelpers/postgres_container.go
+++ b/pkg/internal/testhelpers/postgres_container.go
@@ -432,6 +432,24 @@ func dropTimescaleAndToolkitExtensionsInDB(dbname string) error {
 	return nil
 }
 
+// updateTimescaleDBInDatabase updates the timescaledb version in a named database
+func updateTimescaleDBInDatabase(database string) error {
+	ctx := context.Background()
+	db, err := pgx.Connect(ctx, PgConnectURL(database, Superuser))
+	if err != nil {
+		return fmt.Errorf("could not connect to %s db: %w", database, err)
+	}
+	defer db.Close(ctx)
+
+	_, err = db.Exec(context.Background(), "ALTER EXTENSION timescaledb UPDATE")
+
+	if err != nil {
+		return fmt.Errorf("unable to update timescaledb: %w", err)
+	}
+
+	return nil
+}
+
 func removeTimescaleAndToolkitExtensions(container testcontainers.Container) error {
 	err := dropTimescaleAndToolkitExtensionsInDB("template1")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


JamesGuthrie:
> Is it safe to do the same in production environment? If so, we can implement [this](https://github.com/thaum-xyz/ankhmorpork/blob/master/apps/paperless/manifests/db/statefulSet.yaml#L57-L59) in tobs.

I'm not sure. I was thinking that we could perhaps bake this functionality into the docker container, and optionally enable it with an env var (like `TIMESCALEDB_AUTO_UPGRADE=true`). IMO it would make sense to discuss this with some cloud people.

pkg/internal/testhelpers/postgres_container.go
```diff
@@ -432,6 +432,24 @@ func dropTimescaleAndToolkitExtensionsInDB(dbname string) error {
...
+// updateTimescaleDBInDatabase updates the timescaledb version in a named database
+func updateTimescaleDBInDatabase(database string) error {
+	ctx := context.Background()
+	db, err := pgx.Connect(ctx, PgConnectURL(database, Superuser))
+	if err != nil {
+		return fmt.Errorf("could not connect to %s db: %w", database, err)
+	}
+	defer db.Close(ctx)
+
+	_, err = db.Exec(context.Background(), "ALTER EXTENSION timescaledb UPDATE")
```
antekresic, line 444: Is this safe to do if the extension is not installed previously? Maybe we need a check for that.

JamesGuthrie: It is not safe to do this if the extension is not installed. For our tests, the extension is always installed though.

I did experiment with this (checking if the extension is installed, and conditionally updating it), unfortunately querying a table to check if the extension is installed triggers loading the timescaledb extension, meaning that you can no longer update it in the same connection. At that point I gave up on that approach and went for the simpler, but less robust code.

JamesGuthrie: I kind of addressed this in a code comment in the file below:

```
// Note: We don't actually know which databases have timescaledb installed
// in them. Instead of spending time trying to find out, we assume that
// they all do. This is probably mostly correct.
```

antekresic (approved):

PR #1436, opened 2022-06-16
Fix reload related issues in rules-manager

Harkishen-Singh:

### Why reloading didn't work in certain situations?

[The implementation in master](https://github.com/timescale/promscale/blob/90289d828b89b91cf6c8dda90f2cf8f26945cb60/pkg/runner/runner.go#L148) starts the rules-manager only if **on-start**, rules files exists. But, if there are no rules files on start, then manager is not at all started. So even when we get reload signal later on, nothing happens, but rather, we panic since the `reloadRules()` func is nil as it never got initialized.

### Updating telemetry
Telemetry used to be updated only during the start, not on reload. With this PR, telemetry will be updated on reload.

**Note**: Now we start rules-manager even if rules are not present. But, this is fine as without rules, it does not cause any issue.

## Merge requirements

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -137,48 +137,33 @@ func Run(cfg *Config) error {
 		}
 	}
 
-	var (
-		group       run.Group
-		reloadRules func() error
+	telemetryEngine := initTelemetryEngine(client)
+	telemetryEngine.Start()
+	defer telemetryEngine.Stop()
+
+	rulesCtx, stopRuler := context.WithCancel(context.Background())
+	defer stopRuler()
+	manager, reloadRules, err := rules.NewManager(rulesCtx, prometheus.DefaultRegisterer, client, &cfg.RulesCfg, telemetryEngine)
+	if err != nil {
+		return fmt.Errorf("error creating rules manager: %w", err)
+	}
+	cfg.APICfg.Rules = manager
+
+	var group run.Group
+	group.Add(
+		func() error {
+			// Reload the rules before starting the rules-manager to ensure all rules are healthy.
+			// Otherwise, we block the startup.
+			if err = reloadRules(); err != nil {
+				return fmt.Errorf("error reloading rules: %w", err)
+			}
+			log.Info("msg", "Started Rule-Manager")
+			return manager.Run()
+		}, func(error) {
+			log.Info("msg", "Stopping Rule-Manager")
+			stopRuler()
+		},
 	)
-	pendingTelemetry := map[string]string{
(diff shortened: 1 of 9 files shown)


pkg/runner/runner.go
```diff
@@ -137,48 +137,33 @@ func Run(cfg *Config) error {
...
+	manager, reloadRules, err := rules.NewManager(rulesCtx, prometheus.DefaultRegisterer, client, &cfg.RulesCfg, telemetryEngine)
+	if err != nil {
+		return fmt.Errorf("error creating rules manager: %w", err)
 	}
+	cfg.APICfg.Rules = manager
+
+	var group run.Group
+	group.Add(
+		func() error {
+			// Reload the rules before starting the rules-manager to ensure all rules are healthy.
```
niksajakovljevic, line 155: I'd actually not bound these two together. If reloading of rules fail we should not just stop rule engine but rather stop Promscale Otherwise things can just stop and user not being aware.

pkg/runner/runner.go
```diff
@@ -137,48 +137,33 @@ func Run(cfg *Config) error {
...
+			// Otherwise, we block the startup.
+			if err = reloadRules(); err != nil {
+				return fmt.Errorf("error reloading rules: %w", err)
+			}
+			log.Info("msg", "Started Rule-Manager")
+			return manager.Run()
+		}, func(error) {
+			log.Info("msg", "Stopping Rule-Manager")
+			stopRuler()
```
niksajakovljevic, line 164: Let's actually stop Promscale here

niksajakovljevic (approved):

pkg/telemetry/telemetry.go
```diff
@@ -30,8 +32,10 @@ type Engine interface {
 	// RegisterMetric registers a Prometheus metric with a column name. This metric is
 	// monitored every telemetrySync and updated in the telemetry table.
 	RegisterMetric(columnName string, gaugeOrCounterMetric ...prometheus.Metric) error
-	// Write writes a telemetry information into the database instantly.
-	Write(key, value string)
+
+	// RegisterDynamicMetadata is a Prometheus metric that changes regularly. This is monitored
+	// every telemetrySync and updated in the telemetry table.
+	RegisterDynamicMetadata(columnName string, gaugeOrCounter prometheus.Metric) error
```
Harkishen-Singh, line 38:
```suggestion
	RegisterDynamicMetadata(columnName string, gauge prometheus.Metric) error
```

antekresic (approved):

Harkishen-Singh (commented):
A nit

PR #1439, opened 2022-06-17
Enhance check for schema migration table

jgpruitt:
## Description

Check that the table is in the public schema. We don't want to consider tables named prom_schema_migrations in schemas other than public.

## Merge requirements


JamesGuthrie (approved):

JamesGuthrie (changes requested):
We don't want to merge this, see: https://github.com/timescale/promscale_extension/issues/371

pkg/pgmodel/new_migrate.go
```diff
@@ -23,7 +23,7 @@ var (
 )
 
 func doesSchemaMigrationTableExist(db *pgx.Conn) (exists bool, err error) {
-	const stmt = "SELECT count(*) FILTER (WHERE tablename = 'prom_schema_migrations') > 0 FROM pg_tables"
+	const stmt = "SELECT count(*) FILTER (WHERE schemaname = 'public' AND tablename = 'prom_schema_migrations') > 0 FROM pg_tables"
```
pgnickb, line 26: Any reason to use `filter` over simple `where` clause? Imo it's less readable like this

jgpruitt:
see https://github.com/timescale/promscale/pull/1452

PR #1441, opened 2022-06-20
Add support for PCRE regex matching on the database side

### LINKED ISSUE #920: query with regex flag to ignore case does not work and produces an error
Running a query via prometheus GUI with promscale as a remote_read endpoint does not provide results from timescale if I use regex flag to ignore case sensitivity. It also produces an error in promscale logs:

- Query exmaple: ```netrounds_delay_ms{source=~"((?i)(a)).*"}```
- Error message produced: ```level=warn ts=2021-11-25T12:35:29.036Z caller=read.go:70 msg="Error executing query" query="unsupported value type" storage=PostgreSQL err="appending sample rows: ERROR: invalid regular expression: quantifier operand invalid (SQLSTATE 2201B)" ```

I get data from prometheus TSDB, but not from timescale with such regex.

Any help appreciated

antekresic:
## Description
When using features supported by PCRE-style regexes, matching would fail
because Postgres only supports POSIX-style regex matching. This change
makes use of the newly introduced PCRE-style regex matching via the extension.

Closes #920 

## Merge requirements


pkg/pgmodel/querier/query_builder.go
```diff
@@ -33,11 +35,19 @@ const (
...
+	subQueryRE2MatchEmpty  = "NOT labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $%d and not _prom_ext.re2_match(l.value, $%d))"
+	subQueryNRE2           = "labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $%d and not _prom_ext.re2_match(l.value, $%d))"
+	subQueryNRE2MatchEmpty = "NOT labels && (SELECT COALESCE(array_agg(l.id), array[]::int[]) FROM _prom_catalog.label l WHERE l.key = $%d and _prom_ext.re2_match(l.value, $%d))"
 )
 
 var (
 	minTime = timestamp.FromTime(time.Unix(math.MinInt64/1000+62135596801, 0).UTC())
 	maxTime = timestamp.FromTime(time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC())
+
+	re2Regex = regexp.MustCompile(`\(\?`)
```
JamesGuthrie, line 50: I would appreciate a comment here on why this marker is useful to identify a regex that should be treated as RE2.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -52,39 +62,52 @@ func BuildSubQueries(matchers []*labels.Matcher) (*clauseBuilder, error) {
...
 			}
-			err = cb.addClause(sq, m.Name, m.Value)
 		case labels.MatchNotEqual:
 			sq := subQueryNEQ
 			if matchesEmpty {
 				sq = subQueryNEQMatchEmpty
 			}
 			err = cb.addClause(sq, m.Name, m.Value)
 		case labels.MatchRegexp:
+			re2 := re2Regex.MatchString(m.Value)
```
JamesGuthrie, line 86: Why do we try to determine if the regex should be treated as an RE2 regex? Why not always use the `_prom_ext.re2_match` function?

antekresic: We could and it would simplify the code... I guess this is trying to minimize the exposure to the new regex matcher until proven its stable enough.

Maybe the best way forward is to leave as is and add a TODO to revisit this once this is out.

pkg/pgmodel/querier/query_builder.go
antekresic: Right, will add.

pkg/pgmodel/querier/query_builder.go
antekresic: Added a TODO for the future to reduce complexity.

JamesGuthrie (commented):
In general this looks good to me. I will withhold from approving for now because the extension side is still in review, and that may take a while still.

pkg/pgmodel/querier/query_builder.go
alejandrodnm: How do you usually handle this kind of things? Would it make sense to release with the current engine as default and have the new one under a feature flag or config option? At a latter release we could make the new one the default. 

That way there's a clearer action for customers if there's an issue with the new engine that we haven't spotted.

antekresic: No, we want to fix this for all the users ASAP and it makes no sense to run with the current engine since it doesn't cover all the necessary use cases. 

Keep in mind, the new engine handles a minority of the cases and vast majority will end up just using the current engine without noticing anything different.

pkg/pgmodel/querier/query_builder.go
```diff
@@ -6,11 +6,13 @@ package querier
 
 import (
 	"fmt"
-	"github.com/timescale/promscale/pkg/log"
 	"math"
+	"regexp"
```
Harkishen-Singh, line 10: We no longer use regex across the code. So, I think this should be

```suggestion
	github.com/grafana/regex
```

Harkishen-Singh (approved):
A nit, otherwise LGTM.

alejandrodnm (approved):

JamesGuthrie (approved):

PR #1448, opened 2022-06-21
Prepare for the next development cycle

niksajakovljevic:


antekresic (approved):

pkg/version/version.go
```diff
@@ -38,8 +38,8 @@ var (
 	// It is customary to bump the version by incrementing the numeral after
 	// the `dev` tag. The SQL migration script name must correspond to the /new/ version.
 
-	Promscale                  = "0.12.0"
-	PrevReleaseVersion         = "0.11.0"
+	Promscale                  = "0.13.0-dev.0"
```
JamesGuthrie, line 41: I guess that the `.0` after `-dev` is no longer necessary, since Promscale isn't dealing with SQL migrations anymore.

JamesGuthrie (approved):

Harkishen-Singh (approved):

PR #1451, opened 2022-06-22
Separate connection pools for reader and writer.

Harkishen-Singh:

This commit separates the reader and writer connection pools.

By default, reader gets 0.3 and writer gets 0.5 times of
the allowed connection from the db. But, in case of readonly mode,
reader gets 0.5 times the connections from the db.

## Merge requirements


pkg/pgclient/client.go
```diff
@@ -33,84 +34,108 @@ type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
...
+	seriesCache  cache.SeriesCache
+	connPools    []pgxconn.PgxConn
+	sigClose     chan struct{}
+	haService    *ha.Service
 }
 
 // NewClient creates a new PostgreSQL client
-func NewClient(cfg *Config, mt tenancy.Authorizer, schemaLocker LockFunc, readOnly bool) (*Client, error) {
-	pgConfig, numCopiers, err := getPgConfig(cfg)
+func NewClient(r prometheus.Registerer, cfg *Config, mt tenancy.Authorizer, schemaLocker LockFunc, readOnly bool) (*Client, error) {
```
alejandrodnm, line 52: Would it make sense to refactor this into 2 different functions? Something like

- NewClient
- NewReadOnlyClient

And moving the code that's inside the `!readOnly` into a `getWriterPool` function and the other part into a `getReaderPool` function then the Client constructors will just call this functions and pass the values to `NewClientWithPool`

pkg/pgclient/client.go
```diff
@@ -33,84 +34,108 @@ type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
...
+	readerPgConfig.AfterConnect = schemaLocker
+	readerPool, err := pgxpool.ConnectConfig(context.Background(), readerPgConfig)
 	if err != nil {
-		log.Error("msg", "err creating connection pool for new client", "err", err.Error())
-		return nil, err
+		return nil, fmt.Errorf("err creating reader connection pool: %w", err)
 	}
 
-	client, err := NewClientWithPool(cfg, numCopiers, connectionPool, mt, readOnly)
+	client, err := NewClientWithPool(r, cfg, numCopiers, writerPool, readerPool, mt, readOnly)
```
alejandrodnm, line 114: You can just `return` here

pkg/pgclient/config.go
```diff
@@ -135,70 +139,57 @@ func (cfg *Config) GetConnectionStr() string {
...
+	}
+	return int(poolSize), nil
+}
 
-		//In HA setups
-		if cfg.UsesHA {
-			max = max / 2
-		}
+func (cfg *Config) maxConn() (int, error) {
+	conn, err := pgx.Connect(context.Background(), cfg.GetConnectionStr())
```
alejandrodnm, line 166: In `GetNumCopiers` you call `defer func() { _ = conn.Close(context.Background()) }()` but not here.

pkg/pgclient/client.go
```diff
@@ -33,84 +34,108 @@ type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
...
-	connectionStr := cfg.GetConnectionStr()
-	pgConfig, err := pgxpool.ParseConfig(connectionStr)
+func (cfg *Config) getPgConfig(poolSize int) (*pgxpool.Config, error) {
+	min := MinPoolSize
+	max := poolSize
+	connStr := cfg.GetConnectionStr()
+
+	pgConfig, err := pgxpool.ParseConfig(connStr)
 	if err != nil {
 		log.Error("msg", "configuring connection", "err", err.Error())
```
alejandrodnm, line 128: Should we log the error here? Wouldn't it be also logged by the `runner.Run` function?

pkg/pgclient/client.go
```diff
@@ -143,40 +167,49 @@ func NewClientWithPool(cfg *Config, numCopiers int, connPool *pgxpool.Pool, mt t
...
+	dbQuerierConn := pgxconn.NewQueryLoggingPgxConn(readerPool)
 	dbQuerier := querier.NewQuerier(dbQuerierConn, metricsCache, labelsReader, exemplarKeyPosCache, mt.ReadAuthorizer())
 	queryable := query.NewQueryable(dbQuerier, labelsReader)
 
 	var dbIngestor *ingestor.DBIngestor
 	if !readOnly {
 		var err error
-		dbIngestor, err = ingestor.NewPgxIngestor(dbConn, metricsCache, seriesCache, exemplarKeyPosCache, &c)
+		writerConnPool := pgxconn.NewPgxConn(writerPool)
+		dbIngestor, err = ingestor.NewPgxIngestor(writerConnPool, metricsCache, seriesCache, exemplarKeyPosCache, &c)
```
alejandrodnm, line 183: Isn't this a little bit dangerous in that it might panic if someone calls `client.Ingest()` on a readonly client?

pkg/pgclient/client.go
```diff
@@ -33,84 +34,108 @@ type LockFunc = func(ctx context.Context, conn *pgx.Conn) error
...
+	readerConn   pgxconn.PgxConn
+	ingestor     *ingestor.DBIngestor
+	querier      querier.Querier
+	promqlEngine *promql.Engine
+	healthCheck  health.HealthCheckerFn
+	queryable    promql.Queryable
+	metricCache  cache.MetricCache
+	labelsCache  cache.LabelsCache
+	seriesCache  cache.SeriesCache
+	connPools    []pgxconn.PgxConn
```
niksajakovljevic, line 46: Minor: Since there are two (reader and writer) + we have `readerConn` field already, it might be easier to just store `readerPool` and `writerPool`. Also, imho `PgxConn` (and using conn across the code) is missleading since this is actually a connection pool. I know this is not part of this change but just bringing it out there (so consider renaming `readerConn` to `readerPool`. We can rename`PgxConn` some other time.

pkg/pgclient/metrics.go
```diff
@@ -5,73 +5,65 @@
...
-	statementCacheLen = createStatementCacheLengthHistogramMetric(client)
-	prometheus.MustRegister(
-		statementCacheEnabled,
-		statementCacheCap,
-		statementCacheLen,
+	readerAcquired := prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Namespace:   util.PromNamespace,
+			Subsystem:   "sql_database",
+			Name:        "acquired_connections",
```
niksajakovljevic, line 49: This are actually active connections, maybe better naming `active_connections`?

pkg/pgclient/metrics.go
```diff
@@ -5,73 +5,65 @@
...
-	statementCacheEnabled = prometheus.NewGauge(
+	writerActive := prometheus.NewGaugeFunc(
 		prometheus.GaugeOpts{
 			Namespace:   util.PromNamespace,
-			Subsystem:   "cache",
-			Name:        "enabled",
-			Help:        "Cache is enabled or not.",
-			ConstLabels: map[string]string{"type": "metric", "name": "statement_cache"},
+			Subsystem:   "sql_database",
+			Name:        "active_connections",
```
niksajakovljevic, line 34: This connections are all the connection in the pool. Can be active + idle. So maybe better naming `total_connections`?

pkg/pgclient/metrics.go
```diff
@@ -5,73 +5,65 @@
...
+	readerActive := prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
 			Namespace:   util.PromNamespace,
-			Subsystem:   "cache",
-			Name:        "elements_histogram",
-			Help:        "Number of elements in cache in terms of elements count.",
-			Buckets:     prometheus.ExponentialBuckets(histogramStartBucket, histogramBucketFactor, totalBuckets),
-			ConstLabels: map[string]string{"type": "metric", "name": "statement_cache"},
+			Subsystem:   "sql_database",
+			Name:        "active_connections",
```
niksajakovljevic, line 60: Same here. Maybe better name `total_connections` since it includes Active + Idle

pkg/pgclient/config.go
Harkishen-Singh: Thank you for catching. This might fix the CI that is complaining for some unclosed connection.

pkg/pgclient/client.go
Harkishen-Singh: well, adding a check of readOnly seems anti-pattern. Instead, we should make 2 interfaces: one for read and one for write, and then implement as said [here](https://github.com/timescale/promscale/pull/1451#discussion_r911999677). That will be a good design.

niksajakovljevic (approved):

pkg/pgclient/client.go
Harkishen-Singh: This is a good idea. But, given the current PR, it will be out of scope plus a lot of changes plus the existing stuff in this PR. Would you like to work on this refactoring once we merge this PR?

alejandrodnm (approved):

pkg/api/delete.go
```diff
@@ -71,7 +71,7 @@ func deleteHandler(config *Config, client *pgclient.Client) http.HandlerFunc {
 			if client == nil {
 				continue
 			}
-			pgDelete := deletePkg.PgDelete{Conn: client.Connection}
```
arajkumar, line 74: Should we use writerPool for series deletion?

pkg/pgclient/client.go
antekresic: Let's create an issue for the refactoring. @alejandrodnm would you mind making it since it is your proposal? :smile:

pkg/pgclient/client.go
```diff
@@ -27,90 +28,114 @@ import (
...
+	var (
+		err            error
+		writerPoolSize int
+		numCopiers     int
+		writerPool     *pgxpool.Pool
+		readerFraction = defaultConnFraction
+		writerFraction = defaultConnFraction
+	)
+	if !readOnly {
+		readerFraction = 0.3 // Since readerFraction + writerFraction should be 0.8 or 80% of allowed database connections.
```
antekresic, line 66: Constant?

pkg/pgclient/client.go
alejandrodnm: I knew it was a trap 😄 Yeah, I'll do it

pkg/pgclient/config.go
```diff
@@ -135,70 +139,65 @@ func (cfg *Config) GetConnectionStr() string {
...
+	defer func() { _ = conn.Close(context.Background()) }()
 
+	numCopiers := 0
 	err = conn.QueryRow(context.Background(), "SELECT _prom_ext.num_cpus()").Scan(&numCopiers)
 	if err != nil {
-		return 0, 0, 0, fmt.Errorf("error fetching number of CPUs from extension: %w", err)
+		return 0, fmt.Errorf("error fetching number of CPUs from extension: %w", err)
 	}
 
 	if cfg.WriteConnections > 0 {
```
antekresic, line 199: I would do this first and just return the cfg value. No need to connect to the DB and fetch the number in this case.

pkg/pgclient/config.go
```diff
@@ -135,70 +139,65 @@ func (cfg *Config) GetConnectionStr() string {
...
-			log.Warn("err", err, "msg", "invalid value from postgres max_connections")
-			max = 100
-		}
-
-		//In HA setups
-		if cfg.UsesHA {
-			max = max / 2
-		}
+	var maxConns int
+	err = conn.QueryRow(context.Background(), "SELECT current_setting('max_connections')::int").Scan(&maxConns)
```
antekresic, line 179: Now there is no way to manually configure Promscale to use max connections? It will always use the max from the DB? I'm not sure that's what we want. We can't assume every DB Promscale connects is dedicated to it IMO.

antekresic (changes requested):
I'm not sure removing a configuration flag like max conns is what we want or that it was decided like that in the design doc.

pkg/pgclient/config.go
Harkishen-Singh: Right. Now, we revive the `db.connections-max` flag, which will be used as a check. If total connections (reader max pool size + writer max pool size) > max, then error. So, the user will be forced to either increase the max-conns or decrease the pool sizes.

pkg/pgclient/config.go
```diff
@@ -135,70 +142,66 @@ func (cfg *Config) GetConnectionStr() string {
...
-			return 0, 0, 0, err
-		}
-		max, err = strconv.Atoi(maxStr)
-		if err != nil {
-			log.Warn("err", err, "msg", "invalid value from postgres max_connections")
-			max = 100
-		}
 
-		//In HA setups
+	if inputPoolSize == defaultPoolSize {
```
antekresic, line 153: I think all of these cases would read nicer in a switch statement.

alejandrodnm (approved):

docs/configuration.md
```diff
@@ -151,56 +153,6 @@ With version 0.11.0, we are removing old versions of flag names and enviromental
...
--tls-key-file
--tput-report
--upgrade-extensions
--upgrade-prerelease-extensions
--use-schema-version-lease
--web-cors-origin
--web-enable-admin-api
--web-listen-address
--web-telemetry-path
+-app -async-acks -auth-password -auth-password-file -auth-username -bearer-token -bearer-token-file -db-connection-timeout -db-connections-max -db-host -db-name -db-password -db-port -db-ssl-mode -db-statements-cache -db-uri -db-user -db-writer-connection-concurrency -exemplar-cache-size -high-availability -ignore-samples-written-to-compressed-chunks -install-extensions -labels-cache-size -log-format -log-level -memory-target -metrics-cache-size -migrate -multi-tenancy -multi-tenancy-allow-non-tenants -multi-tenancy-valid-tenants -otlp-grpc-server-listen-address -promql-default-subquery-step-interval -promql-lookback-delta -promql-max-points-per-ts -promql-max-samples -promql-query-timeout -read-only -series-cache-initial-size -series-cache-max-bytes -thanos-store-api-listen-address -tls-cert-file -tls-key-file -tput-report -upgrade-extensions -upgrade-prerelease-extensions -use-schema-version-lease -web-cors-origin -web-enable-admin-api -web-listen-address -web-telemetry-path
```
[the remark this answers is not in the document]
Harkishen-Singh: ah true. I don't know why Goland always does this bad thingy. Will undo the docs stuff except the cli flags.

pkg/pgclient/config.go
```diff
@@ -135,70 +142,65 @@ func (cfg *Config) GetConnectionStr() string {
...
+			poolSize /= 2
 		}
+		return int(poolSize), nil
+	case inputPoolSize < MinPoolSize:
+		return 0, fmt.Errorf("%s pool size canot be less than %d: received %d", poolName, MinPoolSize, inputPoolSize)
+	case inputPoolSize > maxConns:
+		return 0, fmt.Errorf("%s pool size canot be greater than the 'max_connections' allowed by the database", poolName)
+	default:
+	}
+	if cfg.UsesHA {
```
antekresic, line 167: Nit: doing this check two times and you can do it just once. Improve if you can but lets not waste time on it.

antekresic (approved):

PR #1452, opened 2022-06-22
Refine check for existence of `prom_schema_migrations` table

JamesGuthrie:
## Description

The table could exist in any schema. We expect there to be either 0 or 1
tables. If there are more then something really weird has happened, and
we can't safely proceed.

Part of: https://github.com/timescale/promscale_extension/issues/371

## Merge requirements

### DIFF
--- a/pkg/pgmodel/new_migrate.go
+++ b/pkg/pgmodel/new_migrate.go
@@ -23,12 +23,31 @@ var (
 )
 
 func doesSchemaMigrationTableExist(db *pgx.Conn) (exists bool, err error) {
-	const stmt = "SELECT count(*) FILTER (WHERE tablename = 'prom_schema_migrations') > 0 FROM pg_tables"
-	err = db.QueryRow(
-		context.Background(),
-		stmt,
-	).Scan(&exists)
-	return
+	// We expect there to be either 0 or 1 tables. If there are more then
+	// something really weird has happened, and we can't safely proceed.
+	const stmt = "SELECT schemaname FROM pg_catalog.pg_tables WHERE tablename = 'prom_schema_migrations'"
+	var schemas []string
+	rows, err := db.Query(context.Background(), stmt)
+	if err != nil {
+		return false, err
+	}
+	for rows.Next() {
+		var schema string
+		err = rows.Scan(&schema)
+		if err != nil {
+			return false, err
+		}
+		schemas = append(schemas, schema)
+	}
+
+	switch len(schemas) {
+	case 0:
+		return false, nil
+	case 1:
+		return true, nil
+	default:
+		return false, fmt.Errorf("`prom_schema_migrations` exists in multiple schemas: %s", schemas)
+	}
 }
 
 func removeOldExtensionIfExists(db *pgx.Conn) error {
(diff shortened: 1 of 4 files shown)


jgpruitt (approved):

pkg/tests/end_to_end_tests/new_migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/stretchr/testify/require"
+	"github.com/timescale/promscale/pkg/pgmodel"
+	"github.com/timescale/promscale/pkg/pgmodel/common/extension"
+	"github.com/timescale/promscale/pkg/version"
+	"testing"
+)
+
+func TestMigrationFailesIfMultiplePromSchemaMigrationsTables(t *testing.T) {
```
antekresic, line 16: Nit: test would be easier to read/understand if you were to run migrate once, add an additional schema table and check that migrate fails with the correct error.

niksajakovljevic:
Why don't we define `prom_schema_migrations` in a specific schema? That could help us avoid this code completly

pkg/pgmodel/new_migrate.go
```diff
@@ -23,12 +23,31 @@ var (
...
+	}
+
+	switch len(schemas) {
+	case 0:
+		return false, nil
+	case 1:
+		return true, nil
+	default:
+		return false, fmt.Errorf("`prom_schema_migrations` exists in multiple schemas: %s", schemas)
+	}
```
antekresic, line 50:
```suggestion
        if len(schemas) > 1 {
            return false, fmt.Errorf("`prom_schema_migrations` exists in multiple schemas: %s", schemas)
        }
        return len(schemas) == 1, nil
```

antekresic (changes requested):

pkg/tests/end_to_end_tests/new_migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		extOptions := extension.ExtensionMigrateOptions{Install: true, Upgrade: true, UpgradePreRelease: true}
+
+		err = pgmodel.Migrate(conn.Conn(), pgmodel.VersionInfo{Version: version.Promscale, CommitHash: "azxtestcommit"}, nil, extOptions)
+
+		require.Equal(t, err.Error(), "failed to determine whether the prom_schema_migrations table existed: `prom_schema_migrations` exists in multiple schemas: [public myschema]")
```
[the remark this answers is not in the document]
JamesGuthrie: As discussed, I will ignore this suggestion.

pkg/pgmodel/new_migrate.go
JamesGuthrie: To be honest I prefer the way it is. To me it's simpler to read. I don't really mind either way though.

pkg/tests/end_to_end_tests/new_migrate_test.go
JamesGuthrie: I've been fiddling with this for a while, and to be honest there's no easy way to do what you're asking for. As you brought it up as a nit, I'm not sure if it's worth sinking more time into this. I did find a couple of small improvements to this test, which I will push.

JamesGuthrie: I did end up figuring out how to do this and implemented it as suggested.

antekresic (approved):

PR #1457, opened 2022-06-28
Quick-fix: Do not use quotes when tag value is bool.

Harkishen-Singh:
## Description

## Merge requirements


pkg/jaeger/query/trace_query.go
```diff
@@ -240,8 +242,11 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
 				clauses = append(clauses, qual)
 			default:
 				//TODO make sure this is optimized correctly
-				val := "\"" + v + "\""
-				if digitCheck.MatchString(v) {
+				val := interface{}("\"" + v + "\"")
```
niksajakovljevic, line 245: Curious why did we move from string to interface here?

pkg/jaeger/query/trace_query.go
```diff
@@ -240,8 +242,11 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
...
+				if _, err := strconv.ParseBool(v); err == nil {
```
niksajakovljevic, line 246: ` _, err := strconv.ParseBool(v); err == nil || digitCheck.MatchString(v) `

antekresic (approved):

mlvzk:
ParseBool will have err == nil for `1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False`, so you might wanna do the digit check first

PR #1463, opened 2022-06-30
Merge 0.12.1 release

alejandrodnm:
## Description

Merges into master the changes from the patch release 0.12.1 

## Merge requirements


JamesGuthrie (approved):

arajkumar (approved):
lgtm

alejandrodnm:
@antekresic raised the concern that since we only support Rebase and merge and this PR was taking the merge commit approach we won't be able to merge it. An alternative would be to cherry-pick the commits so they are applied to the HEAD of master, the trade-off is that the hash of the commits will change.


Any objections?

alejandrodnm:
We decided to cherry-pick the commits, because of issues with the repo config

PR #1475, opened 2022-07-06
Remove helm upgrade logic from promscale connection secret

raphaelzoellner:
## Description

This PR adds the `preserveConnectionSecret` value to allow for a removal of the upgrade logic from the promscale connection secret as suggested in #1474.

By default `preserveConnectionSecret` is set to true.

If `preserveConnectionSecret` is set to false this will allow to upgrade the connection secret by a `helm upgrade` for the release of the promscale helm chart.


raphaelzoellner:
@paulfantom sorry I wasn't aware on the dependency in tobs to this mechanism. 

How about introducing a switch to the helm values instead to disable upgrade logic? I've updated the PR and added a value named `preserveConnectionSecret` to switch between `helm upgrade` modes of the secret.

PR #1480, opened 2022-07-10
fix pass []any as any in `logQueryStats`

alingse:
## Description

fix pass []any as any in `Next`, call `logQueryStats` need with `args...`

and fix logQueryStats `fmt.Sprintf("%v", args...)` if args has many this will not format as expected.

see github actions result here https://github.com/alingse/asasalint/runs/7270916179?check_suite_focus=true

## Merge requirements


niksajakovljevic:
@alingse Thanks for contributing. To merge this PR we need you to sign CLA. Thanks

niksajakovljevic (approved):

PR #1494, opened 2022-07-15
Fix double counting while populating `promscale_sql_database_chunks_count` metric

arajkumar:
### Problem statment
`PromscaleCompressionLow` alert fires regardless of a good compression ratio.

### Root cause

SQL query which populates `promscale_sql_database_chunks_count` is a sum of the below metrics,
1) Uncompressed Chunks
2) Proxy Chunks which points to Compressed Chunks
3) Compressed Chunks

However the total chunk count should be just (1) + (3), because (2) is already pointing to (3) and which leads to double counting.

### Solution

Fix the SQL query to consider entries which has `compressed_chunk_id` as null and this will be true for both (1) and (3).

## Merge requirements

### DIFF
--- a/docs/runbooks/PromscaleCompressionLow.md
+++ b/docs/runbooks/PromscaleCompressionLow.md
@@ -14,7 +14,7 @@ High disk usage by Promscale database
 3. Open psql
 4. Check total number of chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id is null;`
 5. Check total number of compressed chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id is not null;`
-6. Check number of maintenancec jobs: `select count(*) from timescaledb_information.jobs where proc_name = 'execute_maintenance_job'`
+6. Check number of maintenance jobs: `select count(*) from timescaledb_information.jobs where proc_name = 'execute_maintenance_job'`
 7. Run the following debugging query:
 
 ```postgresql
--- a/docs/runbooks/PromscaleCompressionLow.md
+++ b/docs/runbooks/PromscaleCompressionLow.md
@@ -12,9 +12,10 @@ High disk usage by Promscale database
 1. Open Grafana and navigate to Promscale dashboard
 2. Go to Database section and see `Compressesd chunks ratio`. If you see a ratio of < 10% then compression is not adequate in your system
 3. Open psql
-4. Check number of uncompressed chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id=null;`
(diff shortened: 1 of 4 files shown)


pkg/tests/end_to_end_tests/database_metrics_test.go
```diff
@@ -58,6 +66,59 @@ func TestDatabaseMetrics(t *testing.T) {
 
 		chunksCreated = getMetricValue(t, "chunks_created")
 		require.Equal(t, chunksCreated, float64(3))
+		chunksCount = getMetricValue(t, "chunks_count")
+		require.Equal(t, chunksCount, float64(3))
```
alejandrodnm, line 70: nit:
```suggestion
		require.Equal(t, float64(3), chunksCount)
```

alejandrodnm (approved):

pkg/tests/end_to_end_tests/database_metrics_test.go
arajkumar: What a blooper :P

docs/runbooks/PromscaleCompressionLow.md
```diff
@@ -12,9 +12,10 @@ High disk usage by Promscale database
 1. Open Grafana and navigate to Promscale dashboard
 2. Go to Database section and see `Compressesd chunks ratio`. If you see a ratio of < 10% then compression is not adequate in your system
 3. Open psql
-4. Check number of uncompressed chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id=null;`
-5. Check number of maintenancec jobs: `select count(*) from timescaledb_information.jobs where proc_name = 'execute_maintenance_job'`
-6. Run the following debugging query:
+4. Check total number of chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id is null;`
+5. Check total number of compressed chunks: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id is not null;`
+6. Check number of maintenancec jobs: `select count(*) from timescaledb_information.jobs where proc_name = 'execute_maintenance_job'`
```
JamesGuthrie, line 17:
```suggestion
6. Check number of maintenance jobs: `select count(*) from timescaledb_information.jobs where proc_name = 'execute_maintenance_job'`
```

JamesGuthrie (changes requested):
minor nit

JamesGuthrie (approved):

antekresic (approved):

pkg/tests/end_to_end_tests/database_metrics_test.go
alejandrodnm: It has happened to me soooo many times.

PR #1500, opened 2022-07-18
Add BenchmarkCacheFalseSharing

JamesGuthrie:
## Description

It seems as though the existing benchmarks cannot be used to measure the
effect of false sharing, so I added a new benchmark.

## Merge requirements

### DIFF
--- a/pkg/clockcache/cache_bench_test.go
+++ b/pkg/clockcache/cache_bench_test.go
@@ -3,6 +3,7 @@ package clockcache
 import (
 	"fmt"
 	"math/rand"
+	"sync"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/clockcache/cache_bench_test.go
```diff
@@ -236,6 +237,74 @@ func BenchmarkInsertConcurrent(b *testing.B) {
...
+//
+// Note: This benchmark _must_ be run with the `-cpu=2` argument, to ensure
+// that each goroutine ends up on a different CPU.
+func BenchmarkCacheFalseSharing(b *testing.B) {
+	keys := make([]int, 2)
+	vals := make([]int, 2)
+
+	for i := 0; i < 2; i++ {
+		keys[i], vals[i] = i, i
+	}
```
arajkumar, line 265: Are these lines are added to avoid compiler optimising this routine?

JamesGuthrie: No, actually this was kinda just "noise" left over from copy-pasting another benchmark. I've taken another pass at this benchmark and simplified it a little.

arajkumar (approved):
lgtm

pkg/clockcache/cache_bench_test.go
```diff
@@ -236,6 +237,62 @@ func BenchmarkInsertConcurrent(b *testing.B) {
...
+	b.ReportAllocs()
+
+	// define waitgroup, so we can wait until background incrementing tasks are finished
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+
+	key1 := 0
+	key2 := 1
+	times := b.N
+	b.ResetTimer()
```
antekresic, line 273: Nit: Could we reset timer in the stressor? That way we ensure the last goroutine to start will reset the timer and just measure after the cache trashing has started.

antekresic (approved):
Approved with optional suggestion to improve timing the benchmark.

pkg/clockcache/cache_bench_test.go
JamesGuthrie: I guess we could, but I suspect that could end up distorting the results too. I've done what you suggested, but with an additional waitgroup to coordinate the goroutine start.

PR #1507, opened 2022-07-20
Prepare for the 0.13.0 release

arajkumar:
## Description

## Merge requirements


JamesGuthrie (approved):

antekresic (approved):

.goreleaser.yml
```diff
@@ -165,7 +165,7 @@ dockers:
   - pkg
   - migration-tool
   image_templates:
-  - "timescale/prom-migrator:0.0.4"
+  - "timescale/prom-migrator:0.0.5"
```
Harkishen-Singh, line 168: I don't think we had any discussion on bumping migrator version, without code changes (irrespective of dependencies)

arajkumar: As mentioned in the DM, I updated the migrator version because there was a [dependency upgrade PR](https://github.com/timescale/promscale/pull/1353) merged after 0.12.x.

AFAIK, any change which lands up in the final binary would need a version bump. Incase if a problem arises due to the dependency upgrade, we might end up in a situation where we will have 2 binaries with same version but different behaviour.

I could think of 2 approaches to mitigate this confusion,

1) Apply same version for migrator-tool and promscale (e.g. prometheus and promtool has same version all the time)
2) Move migrator-tool out of promscale repo and let it follow it's own release cadance

I'm not sure how feasible it is to go with (2), I vote for (1).

Harkishen-Singh (approved):

PR #1505, opened 2022-07-19
Add support for running and comparing benchmarks in CI.

Harkishen-Singh:
Signed-off-by: Harkishen-Singh <harkishensingh@hotmail.com>

## Description

This is a test PR ATM.

## Merge requirements

Please take into account the following non-code changes that you may need to make with your PR:

- [ ] CHANGELOG entry for user-facing changes
- [ ] Updated the relevant documentation

PR #1510, opened 2022-07-21
Adds support to run and compare benchmarks on CI.

Harkishen-Singh:
## Description

This PR adds the support for running benchmarks in Promscale on CI. It also compares with the benchmark results of the master branch.

By default, benchmarks will not be active when a PR (or draft PR) is opened. A dev has to assign `action:benchmarks` label, which will automatically trigger the benchmark workflow.

In order to restart the benchmarks, just remove and reassign the label, and it will trigger the benchmark workflow again.

Note:
1. After the benchmark comparison with benchstat is done, the results will be pasted as a PR comment, tagging the author. Each PR comment with a result contains the code hash corresponding to which the benchmark was executed.2. 
2. The commenting on PR will happen only if the PR is opened from a branch in Promscale repo, not from your fork. This is due to some permission reason.

You can refer to https://github.com/timescale/promscale/pull/1505 to see the testing in verbose level.

## Merge requirements

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,100 @@
+name: Benchmarks
+
+on:
+  pull_request:
+    branches: [ '**' ]
+
+env:
+  PROMSCALE_LOGGING: false
+  PROMSCALE_BENCHMARK: true
+  golang-version: 1.18.4
+
+jobs:
+  benchmark_current:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current branch
+        uses: actions/checkout@v3
+
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Benchmark
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 ./... | tee current.txt
+
+      - name: Upload result
+        uses: actions/upload-artifact@v3
+        with:
+          name: current
+          path: current.txt
+
+  benchmark_master:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


Harkishen-Singh:
Now, lets attach the `action:benchmarks` to trigger the `Benchmarks` workflow.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Benchmark current PR
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 ./... | tee current.txt
```
JamesGuthrie, line 28: The only concern I have is that with only 1 run, we may not get concrete enough results to be able to determine if there was a difference or not.

JamesGuthrie (approved):

.github/workflows/benchmark.yaml
Harkishen-Singh: Do you mean, with `-cpu=1` or with `-benchtime=<some fixed iterations>`?

Harkishen-Singh:
Closing this as just realized

> The commenting on PR will happen only if the PR is opened from a branch in Promscale repo, not from your fork. This is due to some permission reason.

.github/workflows/benchmark.yaml
JamesGuthrie: I meant the `-count` parameter. Implicitly we're setting `-count=1`. I'm not sure that I understand it well yet, but I've had situations where count was low, and benchstat would basically say that it didn't have enough data to determine if the outcome was statistically significant or not.

PR #1511, opened 2022-07-21
Adds support to run and compare benchmarks on CI

Harkishen-Singh:
## Description

This PR adds the support for running benchmarks in Promscale on CI. It also compares with the benchmark results of the master branch.

By default, benchmarks will not be active when a PR (or draft PR) is opened. A dev has to assign `action:benchmarks` label, which will automatically trigger the benchmark workflow.

In order to restart the benchmarks, just remove and reassign the label, and it will trigger the benchmark workflow again.

Note:
1. After the benchmark comparison with benchstat is done, the results will be pasted as a PR comment, tagging the author. Each PR comment with a result contains the code hash corresponding to which the benchmark was executed.2. 
2. The commenting on PR will happen only if the PR is opened from a branch in Promscale repo, not from your fork. This is due to some permission reason.

You can refer to https://github.com/timescale/promscale/pull/1505 to see the testing in verbose level.

## Merge requirements

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,100 @@
+name: Benchmarks
+
+on:
+  pull_request:
+    branches: [ '**' ]
+
+env:
+  PROMSCALE_LOGGING: false
+  PROMSCALE_BENCHMARK: true
+  golang-version: 1.18.4
+
+jobs:
+  benchmark_current:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current branch
+        uses: actions/checkout@v3
+
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Benchmark
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 ./... | tee current.txt
+
+      - name: Upload result
+        uses: actions/upload-artifact@v3
+        with:
+          name: current
+          path: current.txt
+
+  benchmark_master:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout master branch
+        uses: actions/checkout@v3
+        with:
+          ref: 'master'
+
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Benchmark
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 ./... | tee master.txt
+
+      - name: Upload result
+        uses: actions/upload-artifact@v3
+        with:
+          name: master
+          path: master.txt
+
+  compare_and_publish_results:
+    runs-on: ubuntu-latest
+    needs: [benchmark_current, benchmark_master]
+    steps:
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Download result for benchmark_current
+        uses: actions/download-artifact@v3
+        with:
+          name: current
+
+      - name: Download result for benchmark_master
+        uses: actions/download-artifact@v3
+        with:
+          name: master
+
+      - name: Install benchstat
+        run: go install -a golang.org/x/perf/cmd/benchstat@latest
+
+      - name: Compare
+        id: compare
+        run: |
+          OUTPUT=$(benchstat -delta-test=none master.txt current.txt)
+          echo "${OUTPUT}"
+          OUTPUT="${OUTPUT//$'\n'/'%0A'}"
+          OUTPUT="${OUTPUT//$'\r'/'%0D'}"
+          echo "::set-output name=result::$OUTPUT"
+
+      - name: Publish results as comment
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          header: compare
+          message: |
+            @${{ github.event.pull_request.user.login }}
+            ## Benchmark Result for **${{ github.sha }}**
+            ### Master vs PR
+            ```
+            ${{ steps.compare.outputs.result }}
+            ```
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -3,14 +3,16 @@ name: Benchmarks
 on:
   pull_request:
     branches: [ '**' ]
+    types: [opened, synchronize, reopened, labeled]
 
 env:
   PROMSCALE_LOGGING: false
-  PROMSCALE_BENCHMARK: true
+  SKIP_PROMQL_BENCHMARKS: true
   golang-version: 1.18.4
 
 jobs:
-  benchmark_current:
+  benchmark:
+    if: contains( github.event.pull_request.labels.*.name, 'action:benchmarks')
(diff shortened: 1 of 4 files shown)


Harkishen-Singh:
Oops, clicked `Assign` instead of `Review`

Harkishen-Singh:
Now, lets attach the `action:benchmarks` to trigger the `Benchmarks` workflow.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Go ${{ env.golang-version }}
+        uses: actions/setup-go@v3.2.1
+        with:
+          go-version: ${{ env.golang-version }}
+        id: go
+
+      - name: Benchmark current PR
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 ./... | tee current.txt
```
JamesGuthrie, line 28: Continuing the [discussion from the previous PR](https://github.com/timescale/promscale/pull/1510/files#r926485351). I've had a bit more opportunity to fiddle around with this, and I think that we should increase the number of runs that we do (with `-count=X`), and we should not use `-delta-test=none` in `benchstat`. The delta test gives an indication of the statistical significance of the result, which is important in evaluating whether the numbers that we're looking at mean something or not.

pkg/promql/bench_test.go
```diff
@@ -26,7 +27,15 @@ import (
 	"github.com/prometheus/prometheus/storage"
 )
 
+func ignore() bool {
```
alejandrodnm, line 30: Why do we ignore some benchmarks?

Harkishen-Singh: We do not touch the files in promql packages, hence running benchmarks in this package is trivial.

alejandrodnm: Is it useful to keep them if we are going to exclude them?

Also, could we change `PROMSCALE_BENCHMARK` for something more meaninful? Maybe `SKIP_PROMQL_BENCHMARKS` or use a build constraint to exclude the benchmark tests from this package.

alejandrodnm (approved):

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,72 @@
...
+on:
+  pull_request:
+    branches: [ '**' ]
+    types: [opened, synchronize, reopened, labeled]
+
+env:
+  PROMSCALE_LOGGING: false
+  SKIP_PROMQL_BENCHMARKS: true
+  golang-version: 1.18.4
+
```
arajkumar, line 12: nit:

You can use top level permission attribute to gain write access to pull-request., this will enable forks to run benchmark. For more info https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

```
permissions:
  pull-requests: write
```

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Benchmark current PR
+        run: go test -bench=. -run=^$ -benchmem -cpu=1 -count=10 -timeout=6h ./... | tee current.txt
+
+      - name: Upload results
+        uses: actions/upload-artifact@v3
+        with:
+          name: current
+          path: current.txt
+
+      - name: Checkout master branch
```
arajkumar, line 36: I'm not sure how easy it is, but can we run PR & master branch benchmarking in parallel?

JamesGuthrie: If you define the benchmarks in separate jobs, then they run on different github runners (in parallel). But if they run on different runners they have different performance characteristics, so we can't compare them. If they run on parallel on the same runner, they will contend for resources, so we won't be able to compare the results.

.github/workflows/benchmark.yaml
JamesGuthrie: I'm not sure that's true. From what I understand, the problem is that the workflow from the fork doesn't have access to the secrets from the source repository. From [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks):

> workflows from forks do not have access to sensitive data such as secrets

arajkumar: That is partly correct, but GITHUB_TOKEN is a transient/generated token who's scope can be altered. However if you define your own secret, it won't be available in the context of PR created from a fork.

JamesGuthrie: You're right, thank you for correcting me! I found this document which explains how it works quite well: https://docs.github.com/en/actions/security-guides/automatic-token-authentication.

Although that document goes on to say: 

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the **Send write tokens to workflows from pull requests** option in the GitHub Actions settings. For more information, see "[Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks)."

.github/workflows/benchmark.yaml
arajkumar: > If you define the benchmarks in separate jobs, then they run on different github runners (in parallel). But if they run on different runners they have different performance characteristics, so we can't compare them
I get your point. Isn't a noisy neighbour can still affect sequential steps? IIUC, in github's terms runner refers to a VM not a physical machine.

> If they run on parallel on the same runner, they will contend for resources, so we won't be able to compare the results.
I'm not aware that we can run parallel jobs in the same runner.

Harkishen-Singh: That was the first approach. The problem was that, parallel workflows lead to being scheduled on different machines, which introduced throttling results.
To see, please refer to this comment https://github.com/timescale/promscale/pull/1505#issuecomment-1190040727 and see the comment history towards the start, when benchmarks were running in parallel.

Hence, now the jobs run serially, but on same machine, which gives some reduction of noise in the results.

JamesGuthrie: > Isn't a noisy neighbour can still affect sequential steps

Yes, definitely. I think that we're going to have to see how reliable these results are in practice. Based on [this](https://github.com/timescale/promscale/pull/1511#issuecomment-1191349375), there is still some noise... If the results aren't consistent enough then we might need to use a dedicated machine for this.

.github/workflows/benchmark.yaml
Harkishen-Singh: I made a test [PR](https://github.com/timescale/promscale/pull/1517) with this, and it seems to give the [same error](https://github.com/timescale/promscale/runs/7466734734?check_suite_focus=true). Did I mess up something?

JamesGuthrie: The quote I made above is relevant here:

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the **Send write tokens to workflows from pull requests** option in the GitHub Actions settings.

Note that it's not possible to enable `Send write tokens to workflows from pull requests` for public repos (according to this: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories.

.github/workflows/benchmark.yaml
Harkishen-Singh: I think the easy way to make a decision on parallel jobs vs serial jobs, is to see the throttling of results. I saw that throttling was reduced when the jobs ran on serially (due to same runner), than when in parallel

arajkumar (approved):
/lgtm

alejandrodnm (approved):

PR #1512, opened 2022-07-21
Remove multinode e2e tests from the e2e target

alejandrodnm:
## Description

Multinode is not supported and running the e2e tests with the flag
causes the following error when creating the extension:

```
CREATE EXTENSION IF NOT EXISTS promscale VERSION '0.5.4';

ERROR:  [dn0]: schema "ps_trace" does not exist
```
## Merge requirements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ e2e: DOCKER_IMAGE?=$(shell ./scripts/fallback-docker.sh $(LOCAL_DOCKER_BASE):hea
 e2e: pkg/tests/testdata/traces-dataset.sz generate
 	go test -v ./pkg/tests/end_to_end_tests/ -timescale-docker-image=$(DOCKER_IMAGE)
 	go test -v ./pkg/tests/end_to_end_tests/ -use-timescaledb=false -timescale-docker-image=$(DOCKER_IMAGE)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,7 +31,6 @@ e2e: DOCKER_IMAGE?=$(shell ./scripts/fallback-docker.sh $(LOCAL_DOCKER_BASE):hea
 e2e: pkg/tests/testdata/traces-dataset.sz generate
 	go test -v ./pkg/tests/end_to_end_tests/ -timescale-docker-image=$(DOCKER_IMAGE)
 	go test -v ./pkg/tests/end_to_end_tests/ -use-timescaledb=false -timescale-docker-image=$(DOCKER_IMAGE)
-	go test -v ./pkg/tests/end_to_end_tests/ -use-multinode -timescale-docker-image=$(DOCKER_IMAGE)
```
JamesGuthrie, line 34: Could we comment this out instead, and add a note about why we've commented it out? See https://github.com/timescale/promscale/blob/e7745aa1b95ac4d58f18beef9d06924062f5ee4c/.github/workflows/go-e2e.yml#L83

JamesGuthrie (approved):

arajkumar (approved):
lgtm

PR #1522, opened 2022-07-25
Adds support for benchmarking new series creation.

Harkishen-Singh:

## Description

This PR adds benchmarks for new series creation case. This happens when Promscale sees a new series for the first time, like in case of a startup.

The `NewSeriesGenerator` creates batches of distinct series based on the given `numMetrics` and `numSeries`, with each series containing labels equal to `labels` arg of the function. The label key in the series remain constant, and the value always changes to create a new series.

## Merge requirements

### DIFF
--- a/pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
+++ b/pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
@@ -5,16 +5,19 @@
 package end_to_end_tests
 
 import (
+	"context"
 	"fmt"
 	"testing"
+
+	"github.com/walle/targz"
 
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/stretchr/testify/require"
 	"github.com/timescale/promscale/pkg/pgmodel/cache"
 	"github.com/timescale/promscale/pkg/pgmodel/ingestor"
 	"github.com/timescale/promscale/pkg/pgxconn"
+	"github.com/timescale/promscale/pkg/prompb"
 	"github.com/timescale/promscale/pkg/tests/testsupport"
-	"github.com/walle/targz"
 )
 
 var prometheusDataGzip = "../testdata/prometheus-data.tar.gz"
@@ -72,5 +75,29 @@ func BenchmarkMetricIngest(b *testing.B) {
 		b.StartTimer()
 		sampleLoader.Run(metricsIngestor.Ingest)
 		b.StopTimer()
+	})
+}
+
+func BenchmarkNewSeriesIngestion(b *testing.B) {
+	seriesGen, err := testsupport.NewSeriesGenerator(10, 100, 4)
+	require.NoError(b, err)
+
+	ts := seriesGen.GetTimeseries()
+
+	withDB(b, "bench_e2e_new_series_ingest", func(db *pgxpool.Pool, t testing.TB) {
+		metricsIngestor, err := ingestor.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), &ingestor.Cfg{
+			NumCopiers:              8,
+			InvertedLabelsCacheSize: cache.DefaultConfig.InvertedLabelsCacheSize,
+		})
+		require.NoError(b, err)
+		defer metricsIngestor.Close()
+
+		b.ResetTimer()
+		b.ReportAllocs()
+		_, _, _ = metricsIngestor.Ingest(context.Background(), &prompb.WriteRequest{Timeseries: ts})
+
+		numSeries := 0
+		require.NoError(b, db.QueryRow(context.Background(), "SELECT count(*) FROM _prom_catalog.series").Scan(&numSeries))
+		require.Equal(b, 1000, numSeries)
 	})
 }
--- a/pkg/tests/testsupport/series_gen.go
+++ b/pkg/tests/testsupport/series_gen.go
@@ -1,0 +1,69 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package testsupport
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/timescale/promscale/pkg/prompb"
+)
+
+type seriesGen struct {
+	ts []prompb.TimeSeries
+}
+
+const (
+	labelKeyPrefix   = "test_k_"
+	labelValuePrefix = "test_v_"
+	metricNamePrefix = "test_metric_"
+)
+
+var samples = []prompb.Sample{{Timestamp: 1, Value: 1}}
+
+// NewSeriesGenerator generates distinct timeseries equal to numMetrics * numSeriesPerMetric, such that
+// each timeseries has `labels` count of labels. The label keys are common across the series, but their
+// values are random, thereby creating a new series.
+func NewSeriesGenerator(numMetrics, numSeriesPerMetric, labels int) (*seriesGen, error) {
+	if labels < 2 {
+		return nil, fmt.Errorf("minLabels cannot be less than 2")
+	}
+	totalSeries := numMetrics * numSeriesPerMetric
+	ts := make([]prompb.TimeSeries, 0, totalSeries)
+
+	labels -= 1 // Since 1 label will be occupied by __name__
+	labelKeys := make([]string, labels)
+	for i := 0; i < labels; i++ {
+		labelKeys[i] = fmt.Sprintf("%s%d", labelKeyPrefix, i)
+	}
+
+	for i := 0; i < numMetrics; i++ {
+		metricName := randomText(metricNamePrefix)
+		metric := prompb.Label{Name: "__name__", Value: metricName}
+
+		for j := 0; j < numSeriesPerMetric; j++ {
+			serie := []prompb.Label{metric}
+
+			for k := 0; k < labels; k++ {
+				// The keys remain the same across the series, but the values change, everytime creating a new series.
+				serie = append(serie, prompb.Label{Name: labelKeys[k], Value: randomText(labelValuePrefix)})
+			}
+			ts = append(ts, prompb.TimeSeries{
+				Labels:  serie,
+				Samples: samples,
+			})
+		}
+	}
+	return &seriesGen{ts}, nil
+}
+
+func (s *seriesGen) GetTimeseries() []prompb.TimeSeries {
+	return s.ts
+}
+
+func randomText(prefix string) string {
+	suffix := rand.Int() // #nosec
+	return fmt.Sprintf("%s%d", prefix, suffix)
+}


pkg/tests/testsupport/series_gen.go
```diff
@@ -0,0 +1,82 @@
...
+	if labels < 2 {
+		return nil, fmt.Errorf("minLabels cannot be less than 2")
+	}
+	totalSeries := numMetrics * numSeriesPerMetric
+	ts := make([]prompb.TimeSeries, 0, totalSeries)
+
+	labels -= 1 // Since 1 label will be occupied by __name__
+	labelKeys := make([]string, labels)
+	for i := 0; i < labels; i++ {
+		labelKeys[i] = randomText(labelKeyPrefix)
```
arajkumar, line 36: Can we use index as a suffix to generate unique values instead of random? Just to avoid the probability of collision which is any way very low :)

arajkumar: Other option would be setting up a [constant seed](https://pkg.go.dev/math/rand#Seed), so that there we can bring a predictability. Having a real random value in the test is not a good thing.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
```diff
@@ -74,3 +77,30 @@ func BenchmarkMetricIngest(b *testing.B) {
...
+
+	withDB(b, "bench_e2e_new_series_ingest", func(db *pgxpool.Pool, t testing.TB) {
+		metricsIngestor, err := ingestor.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), &ingestor.Cfg{
+			NumCopiers:              8,
+			InvertedLabelsCacheSize: cache.DefaultConfig.InvertedLabelsCacheSize,
+		})
+		require.NoError(b, err)
+		defer metricsIngestor.Close()
+
+		b.Run("New Series Ingestion", func(b *testing.B) {
```
arajkumar, line 95: Looks like benchmark iteration count (`b.N`) is ignored. We should honer that right?

niksajakovljevic: Since we are writing to DB benchmark iterations would result in duplicates. Ingesting fairly large amount of data should help getting more stable results but obviously it's not ideal. Do you maybe have some ideas?

pkg/tests/testsupport/series_gen.go
niksajakovljevic: I believe `math/rand` by default is providing deterministic number generator so we should be good

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
```diff
@@ -74,3 +77,30 @@ func BenchmarkMetricIngest(b *testing.B) {
...
+			b.ReportAllocs()
+			for i := 0; i < len(batches); i++ {
+				_, _, _ = metricsIngestor.Ingest(context.Background(), &prompb.WriteRequest{Timeseries: batches[i]})
```
niksajakovljevic, line 98: Imho it would be useful to also add a benchmark for only metadata ingest, meaning calling `WriteSeries` directly.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
Harkishen-Singh: we anyways run the benchmark 10 time due to `-count` before calculating the difference with the master branch. Hence, we don't have to care much about `b.N` here. That said, I think `b.Run()` takes care internally of the benchmark iteration.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
Harkishen-Singh: If we want to benchmark `WriteSeries`, then it should be written in `pgmodel/ingestor/series_writer_test.go` and there, we cannot use the `WithDB`. If we want to do this way, then it will require a lot of changes in the `testhelpers` package to make WithDB accessible to external modules.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
arajkumar: In that case, can we consider generating number of series based on b.N?

pkg/tests/testsupport/series_gen.go
arajkumar: No, by default it generates pseudo random number.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
Harkishen-Singh: That is a good point, but N is a non-static value, right? So during the benchmarks, the value of N will change, meaning generating series on runtime might not be a good thing.

arajkumar: > Since we are writing to DB benchmark iterations would result in duplicates.

How about using different DB instance for each iteration to avoid duplication? Also this would reproduce the same scenario on each iteration.

The current approach is,
```
1. Create DB
2. Ingest batch 1,
3. ....
4. Ingest batch n
```
Each iteration being adds more data to the existing db which is not deterministic.

Recommended approach is,
```
for i :=0; i < b.N; i++ {
1. Create DB
2. Ingest batch 1,
3. ....
4. Ingest batch n
}
```

arajkumar: I misinterpreted that `b.N` is based on `-count` flag. Ignore my previous comment.

arajkumar: I'm bit confused here, When I run this benchmark, I see the following report.

```
$ go test -test.bench NewSeriesIngestion -run=^$ -cpu=1 -timeout=6h -count=10 ./pkg/tests/end_to_end_tests

BenchmarkNewSeriesIngestion/New_Series_Ingestion                       1        17720620125 ns/op    62112104 B/op     520193 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001638 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001365 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                level=info ts=2022-07-26T11:28:29.708Z caller=throughput.go:76 msg="ingestor throughput" samples/sec=2000 metrics-max-sent-ts=1970-01-01T00:00:00.001Z
1000000000               0.0001356 ns/op               0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001402 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001354 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001322 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001254 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001434 ns/op            0 B/op          0 allocs/op
BenchmarkNewSeriesIngestion/New_Series_Ingestion                1000000000               0.0001302 ns/op            0 B/op          0 allocs/op
```
IIUC, only first time it ran successfully and the rest of the 9 times it gives number of iteration as 1000000000. Am I missing something here?

pkg/tests/testsupport/series_gen.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	labelKeyPrefix   = "test_k_"
+	labelValuePrefix = "test_v_"
+	metricNamePrefix = "test_metric_"
+)
+
+var samples = []prompb.Sample{{Timestamp: 1, Value: 1}}
+
+func NewSeriesGenerator(numMetrics, numSeriesPerMetric, labels int) (*seriesGen, error) {
```
niksajakovljevic, line 26: Let's add some comments explaining how data is generated

pkg/tests/testsupport/series_gen.go
niksajakovljevic:
```
for i := 0; i < 10; i++ {
		fmt.Printf("%d. %d\n", i, rand.Int())
	}
```
Always generates the same number sequence so it is deterministic.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
niksajakovljevic: > How about using different DB instance for each iteration to avoid duplication?

We could try that. That benchmark might run for quite a long time since in each iteration we need to insert a lots of data.

niksajakovljevic: Interesting. I wonder if that's because of test caching

arajkumar: I'm yet to validate, but I think withDB fails and returns silently on the second iteration.

pkg/tests/testsupport/series_gen.go
arajkumar: Ah, you are right. Looks like I assumed it is a crypto/rand, but actually it is a math/rand.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
niksajakovljevic, line 95: Why do we create a subtest here?

arajkumar: @niksajakovljevic Looks like you are right. When running a benchmark inside `b.Run`, it seem to be catching the result because on each iteration benchmark name is same(first arg). Removing b.Run fixes the issue.

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
```diff
@@ -74,3 +77,30 @@ func BenchmarkMetricIngest(b *testing.B) {
 		b.StopTimer()
 	})
 }
+
+func BenchmarkNewSeriesIngestion(b *testing.B) {
+	seriesGen, err := testsupport.NewSeriesGenerator(100, 100, 4)
```
niksajakovljevic, line 82: Shouldn't we restart timer/allocs after generating data

niksajakovljevic (approved):

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
Harkishen-Singh: b.Run is now removed :)

arajkumar (approved):
/lgtm

pkg/tests/end_to_end_tests/metric_ingest_bench_test.go
```diff
@@ -74,3 +76,29 @@ func BenchmarkMetricIngest(b *testing.B) {
...
+		metricsIngestor, err := ingestor.NewPgxIngestorForTests(pgxconn.NewPgxConn(db), &ingestor.Cfg{
+			NumCopiers:              8,
+			InvertedLabelsCacheSize: cache.DefaultConfig.InvertedLabelsCacheSize,
+		})
+		require.NoError(b, err)
+		defer metricsIngestor.Close()
+
+		b.ResetTimer()
+		b.ReportAllocs()
+		for i := 0; i < len(requests); i++ {
```
arajkumar, line 96: @Harkishen-Singh BTW what are the advantage when we ingest in batches? Don't we already batch in our implementation?

Harkishen-Singh: I was trying to make it like what Prometheus does, sending new series in batches. But we could as well ingest all series in a single batch of 10000 like in here.

PR #1525, opened 2022-07-26
Helm pin version

nhudson:
## Description

In the Helm chart we are not forcibly setting a container image version, which will always pull the `latest` tag.  The changes here will set the image tag from `Chart.yaml` but allow override from `values.yaml` if the end user wants to use a different tag.

- Change how the container image information is structures in
  values.yaml
- Take the default container image tag from Chart.yaml, but allow
  override.

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -71,6 +71,7 @@ jobs:
           helm repo update
           helm install \
             --set replicaCount=1 \
+            --set image.tag=pg14.4-ts2.7.2-p0 \
             timescaledb timescale/timescaledb-single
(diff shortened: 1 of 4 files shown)


arajkumar (approved):
/lgtm.

I would recommend to cherry-pick this change to release-0.13 branch as well. Though it might not need a new patch release, still having it under release-0.13 would help because t[imescale docs points to that.](https://docs.timescale.com/promscale/latest/installation/kubernetes/#install-promscale-with-a-manifest-file)

nhudson:
@arajkumar cherry-pick is done in #1532

PR #1526, opened 2022-07-27
Fix upgrade test by pinning to timescaledb 2.7.1

JamesGuthrie:
## Merge requirements

### DIFF
--- a/pkg/tests/upgrade_tests/upgrade_test.go
+++ b/pkg/tests/upgrade_tests/upgrade_test.go
@@ -89,7 +89,7 @@ func getDBImages(extensionState testhelpers.TestOptions, prevPromscaleVersion *s
 
 	// From Promscale 0.13.0 onwards the minimum extension version is 0.5.4
 	if prevPromscaleVersion != nil && prevPromscaleVersion.GE(semver.MustParse("0.13.0")) {
-		return "timescale/timescaledb-ha:pg" + pgVersion + "-ts2.7-latest", dockerImageName, nil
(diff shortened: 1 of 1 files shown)


arajkumar (approved):
/lgtm

pkg/tests/upgrade_tests/upgrade_test.go
```diff
@@ -89,7 +89,7 @@ func getDBImages(extensionState testhelpers.TestOptions, prevPromscaleVersion *s
 
 	// From Promscale 0.13.0 onwards the minimum extension version is 0.5.4
 	if prevPromscaleVersion != nil && prevPromscaleVersion.GE(semver.MustParse("0.13.0")) {
-		return "timescale/timescaledb-ha:pg" + pgVersion + "-ts2.7-latest", dockerImageName, nil
+		return "timescale/timescaledb-ha:pg" + pgVersion + ".11-ts2.7.1-latest", dockerImageName, nil
```
niksajakovljevic, line 92: Why we added `.11` to PG version?

JamesGuthrie: We want to pin the image to timescaledb 2.7.1. The way that the `timescaledb-ha` docker images are tagged means that the only way to specify the patch version of timescaledb is to _also_ specify the minor version of postgres.

The upgrade tests are only run against postgres 12, and the patch version for pg12 is `.11`. We should probably hardcode `pgVersion` in this whole function with `"12"` to make it even more clear what's going on here.

PR #1529, opened 2022-07-28
Prometheus metrics has to be registered after all services are added

niksajakovljevic:
Prometheus metrics has to be registered after all services are added

Enabling server histograms should be done before registering metrics.

Bumping tag cache size since 10K seems pretty low.


arajkumar (approved):
/lgtm. Hope bumping up `tagCacheSize` won't result in a dramatic memory usage.

Harkishen-Singh (approved):

niksajakovljevic:
> Hope bumping up tagCacheSize won't result in a dramatic memory usage.

It really should not. Even if key=value pair takes 1KB (btw never seen that size) it's only 1MB of memory.

PR #1554, opened 2022-08-10
Add batching for traces

niksajakovljevic:
Batching will help to achieve better ingest performance, especially if
traces are sent one by one.

A batch is produced when one of two conditions is met: batch size or timeout.

This PR also adds `async` support for traces meaning that client doesn't need to wait for DB write. This increases ingest performance with a small risk of data loss. 

Added 3 new CLI flags:
- Two flags to control batch size: `tracing.max-batch-size` and `tracing.batch-timeout`.
- Flag for async writes `tracing.async-acks`


niksajakovljevic:
The local benchmarks that I've added are pointing that batching should provide better performance, however I still need to run full blown benchmark with more realistic data loads.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+const (
+	defaultReqBufferSize      = 10000                  // tweaking this does not result in significant perf impact
+	defaultBatchSize          = 5000                   // this is >= (we might produce bigger batches sometimes)
+	defaultBatchTimeout       = 200 * time.Millisecond // arbitrary picked. We don't want client to wait too long
+	defaultMaxBufferedBatches = 200                    // arbitrary picked. We want to avoid wait on batches
+)
+
+var defaultBatchWorkers = runtime.NumCPU() / 2 // we only take half so other half can be used for writers
```
arajkumar, line 20: Since this is primarily IO driven, is it possible to size this based on number of db connections instead of no.of CPUs?

arajkumar, line 20: This would cause a trouble when the promscale is configured with cpu limit of < 1.

niksajakovljevic: Right. Great catch. Will fix.

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
```diff
@@ -0,0 +1,50 @@
+package trace
```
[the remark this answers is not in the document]
niksajakovljevic: Sure. Will look into that.

pkg/pgmodel/ingestor/trace/trace_batcher.go
niksajakovljevic: If you look closer batching and writing batches is separated thus creating batches is mostly CPU driven - batches are only written to a channel and picked up by batch writers.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -46,10 +49,16 @@ func NewPgxIngestor(conn pgxconn.PgxConn, cache cache.MetricCache, sCache cache.
...
+	batcherConfg := trace.BatcherConfig{
+		MaxBatchSize: cfg.TracingMaxBatchSize,
+		BatchTimeout: cfg.TracingBatchTimeout,
+		Writers:      cfg.NumCopiers,
+	}
 	return &DBIngestor{
 		sCache:     sCache,
 		dispatcher: dispatcher,
-		tWriter:    trace.NewWriter(conn),
+		tWriter:    trace.NewDispatcher(trace.NewWriter(conn), cfg.AsyncAcksTraces, batcherConfg),
```
Harkishen-Singh, line 61: I think `AsyncAcksTraces` should be part of batcherConfig, rather than sending it as separate param.

Is it because of 1 each for reading and writing?

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -58,7 +67,7 @@ func NewPgxIngestor(conn pgxconn.PgxConn, cache cache.MetricCache, sCache cache.
 // with an empty config, a new default size metrics cache and a non-ha-aware data parser
 func NewPgxIngestorForTests(conn pgxconn.PgxConn, cfg *Cfg) (*DBIngestor, error) {
 	if cfg == nil {
-		cfg = &Cfg{InvertedLabelsCacheSize: cache.DefaultConfig.InvertedLabelsCacheSize}
+		cfg = &Cfg{InvertedLabelsCacheSize: cache.DefaultConfig.InvertedLabelsCacheSize, NumCopiers: 2}
```
Harkishen-Singh, line 70: I wonder why not keep this as `1`?

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
+package trace
```
Harkishen-Singh, line 1: license header

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+			defer b.wg.Done()
+			b.batch()
+		}()
+	}
+	b.batchWriter.run()
+}
+
+func (b *Batcher) batch() {
+	_, batcherSpan := tracer.Default().Start(context.Background(), "create-batches")
+	ticker := time.NewTicker(b.config.BatchTimeout)
```
Harkishen-Singh, line 130: Let's defer the `ticker.Stop()` here, than in `<-b.stop`. This will avoid any future problem when more than one paths exit the function.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+	batch := NewBatch(b.config.MaxBatchSize)
+	batcherSpan.AddEvent("New Batch")
+	flushBatch := func(batch *Batch) *Batch {
+		batchCp := *batch
+		b.bufferedBatches <- batchCp
+		batcherSpan.AddEvent("Batch sent to buffer")
+		batcherSpan.AddEvent("New Batch")
+		return NewBatch(b.config.MaxBatchSize)
```
Harkishen-Singh, line 138: I don't think `flushBatch` should create a NewBatch(). It feels wrong that a flush function is also creating and returning something. Instead, we should call the `NewBatch()` after `if batch.isFull()`.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+	Batchers     int
+	Writers      int
+	MaxBatchSize int
+	BatchTimeout time.Duration
+}
+
+type WriteTraces func(context.Context, ptrace.Traces) error
+
+func NewBatcher(config BatcherConfig, writer Writer) *Batcher {
+	if config.Batchers == 0 {
```
Harkishen-Singh, line 95: It might be better to move all these checks/validity to `verifyConfig(&config)`. Keeps the constructor clean.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+			if batch.isFull() {
+				batch = flushBatch(batch)
+				ticker.Reset(b.config.BatchTimeout)
+			}
+			batch.add(item)
+		case <-ticker.C:
+			batcherSpan.AddEvent("Batch timeout reached")
+			if !batch.isEmpty() {
+				batch = flushBatch(batch)
+			}
```
Harkishen-Singh, line 158: This won't work as expected.
It will tick even if the very previous moment we did the `item := <-b.in` case and then flushed the batch. This is because the ticker is just created once, so it does not care about most recent event.

We should use `<-time.After(config.BatchTimeout)` in this case.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (b *Batcher) batch() {
+	_, batcherSpan := tracer.Default().Start(context.Background(), "create-batches")
+	ticker := time.NewTicker(b.config.BatchTimeout)
+	batch := NewBatch(b.config.MaxBatchSize)
+	batcherSpan.AddEvent("New Batch")
+	flushBatch := func(batch *Batch) *Batch {
+		batchCp := *batch
+		b.bufferedBatches <- batchCp
```
Harkishen-Singh, line 135: Why not send in the channel by reference? that is more performant than sending direct values.
```suggestion
		b.bufferedBatches <- batch or batchCp
```
Also, I am unable to find how we are reusing the batch for using a copy.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,248 @@
...
+		go func() {
+			defer bw.wg.Done()
+			for {
+				select {
+				case b, ok := <-bw.batches:
+					if !ok {
+						return
+					}
+					bw.flush(b)
+				case <-bw.stopCh:
```
Harkishen-Singh, line 204: I think we should handle these things via context as per convention?

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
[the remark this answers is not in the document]
Harkishen-Singh: I think we also need some Prometheus metrics here, to keep track of num batch active, etc., that will be helpful in debugging/knowing the state of trace ingestion.

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
```diff
@@ -0,0 +1,58 @@
...
+	async   bool // 'true' means that we don't wait for request to be persisted
+}
+
+func NewDispatcher(writer Writer, async bool, batchConfig BatcherConfig) *Dispatcher {
+	batcher := NewBatcher(batchConfig, writer)
+	batcher.Run()
+	return &Dispatcher{batcher: batcher, async: async}
+}
+
+type InsertTracesReq struct {
```
Harkishen-Singh, line 24: Are we using this guy out of `trace` package? Wondering if we need this public or not.

Harkishen-Singh (changes requested):
Would be good to see perf difference with and without batching in case of Jaeger ingestion.

pkg/pgmodel/ingestor/ingestor.go
niksajakovljevic: It is not part of Batcher configuration since Batcher does not care if it's async or not. It will just batch and write batches. The dispatcher is actually affaeted by async flag - so it will either return immediatelly or wait until request is persisted.

pkg/pgmodel/ingestor/ingestor.go
niksajakovljevic: I thought having it set to 2 would hopefully make tests complete sooner.

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
niksajakovljevic: Yeah we can make it private

pkg/pgmodel/ingestor/trace/trace_batcher.go
niksajakovljevic: I hear what you are saying. The reason I did it is to avoid calling `NewBatch()` in multiple places. I am not strongly opinionated here sine the code involved is trivial and `flushBatch` is anonymous function that helps avoid repetition. Let me know if you feel strongly and you want me to change it.

pkg/pgmodel/ingestor/trace/trace_batcher.go
niksajakovljevic: I believe this is more about semantics, not performance. Semantically sending copy more clearly communicates that batch is immutable. I've run quick benchmark for `pointer vs struct` and they are almost identical:
```
BenchmarkSendBatchStruct-10       153936          8596 ns/op       41552 B/op          4 allocs/op
BenchmarkSendBatchPointer-10      143161          8936 ns/op       41619 B/op          6 allocs/op
```
I do agree that when sending large structs it's better to use pointers however not the case here...

pkg/pgmodel/ingestor/trace/trace_batcher.go
niksajakovljevic: Not sure I follow. Isn't closing a channel a common way to stop goroutines?

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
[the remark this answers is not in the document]
niksajakovljevic: Yep. Added metric to keep track of reasons for flushing a batch: ideally batch should be flushed by size and not timeout. Also added one to track batch queue size.

niksajakovljevic:
> Would be good to see perf difference with and without batching in case of Jaeger ingestion.

Yes I will be soon publishing benchmark numbers in this PR.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,261 @@
...
+	if defaultBatchWorkers == 0 {
+		// in case there <=1 CPU
+		defaultBatchWorkers = 1
+	}
+}
+
+// We batch individual insertTracesReq
+type Batch struct {
+	traces    ptrace.Traces
+	spanCount int
```
Harkishen-Singh, line 41: I wonder if `spanCount` is needed. We are using it for `tb.spanCount >= tb.maxSize` which we can get by `tb.traces.SpanCount() >= tb.maxSize`.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,261 @@
...
+		batchWriter:     newBatchWriter(config.Writers, writer, bufferedBatches),
+		in:              make(chan insertTracesReq, defaultReqBufferSize),
+		stop:            make(chan struct{}),
+		bufferedBatches: bufferedBatches,
+		config:          config,
+	}
+}
+
+func validateConfig(config *BatcherConfig) {
+	if config.Batchers == 0 {
```
Harkishen-Singh, line 113: Nit: I think we can simplify this using `switch {}`.

pkg/pgmodel/ingestor/trace/trace_batcher.go
Harkishen-Singh: Can you share the batch size in this benchmark please? Ah, nvm, `ptrace.Traces` in itself is a slice.

Harkishen-Singh:
```suggestion
		b.bufferedBatches <- *batch
```
If we want to use a copy, then how about this? This will save the stack allocation, which would otherwise be the size of the batch.

pkg/pgmodel/ingestor/trace/trace_batcher.go
Harkishen-Singh: No strong suggestion. My aim was more towards readability. But, let's ignore this.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,261 @@
...
+func init() {
+	defaultBatchWorkers = runtime.NumCPU() / 2
+	if defaultBatchWorkers == 0 {
+		// in case there <=1 CPU
+		defaultBatchWorkers = 1
+	}
+}
+
+// We batch individual insertTracesReq
+type Batch struct {
```
Harkishen-Singh, line 39:
```suggestion
// Batch individual insertTracesReq.
type Batch struct {
```
More like godoc :)

pkg/pgmodel/metrics/ingest.go
```diff
@@ -175,6 +175,30 @@ var (
...
+			Name:      "batch_buffer_queue_size",
+			Help:      "Number of batches waiting to be written",
+		}, []string{"type"},
+	)
+	IngestorReqBufferQueueSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Subsystem: "ingest",
+			Name:      "req_buffer_queue_size",
+			Help:      "Number of user requests in queue",
```
Harkishen-Singh, line 199:
```suggestion
	IngestorRequestsQueued = prometheus.NewGaugeVec(
		prometheus.GaugeOpts{
			Namespace: util.PromNamespace,
			Subsystem: "ingest",
			Name:      "requests_queued",
			Help:      "Number of active user requests in queue.",
```

pkg/pgmodel/metrics/ingest.go
```diff
@@ -175,6 +175,30 @@ var (
...
+			Name:      "batch_flush_total",
+			Help:      "Number of batch flushes by reason (size or timeout).",
+		}, []string{"type", "reason"},
+	)
+	IngestorBatchBufferSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: util.PromNamespace,
+			Subsystem: "ingest",
+			Name:      "batch_buffer_queue_size",
+			Help:      "Number of batches waiting to be written",
```
Harkishen-Singh, line 191:
```suggestion
	IngestorPendingBatches = prometheus.NewGaugeVec(
		prometheus.GaugeOpts{
			Namespace: util.PromNamespace,
			Subsystem: "ingest",
			Name:      "pending_batches",
			Help:      "Number of batches waiting to be written",
```

Harkishen-Singh (changes requested):
Overall looks good. However, some changes are needed on the metrics side.

pkg/pgmodel/ingestor/trace/trace_batcher.go
Harkishen-Singh: About the benchmark, its right that difference wouldn't be huge, since `ptrace.Traces` is internally a slice :). Hence, we can ignore the pointer stuff, but see above comment on stack size.

pkg/pgmodel/ingestor/trace/trace_batcher.go
niksajakovljevic: It's there to avoid recomputations since `traces.SpanCount()` has a nested `for` loops.

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -46,10 +49,16 @@ func NewPgxIngestor(conn pgxconn.PgxConn, cache cache.MetricCache, sCache cache.
 	if err != nil {
 		return nil, err
 	}
+
+	batcherConfg := trace.BatcherConfig{
+		MaxBatchSize: cfg.TracingMaxBatchSize,
+		BatchTimeout: cfg.TracingBatchTimeout,
+		Writers:      cfg.NumCopiers,
```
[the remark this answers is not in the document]
niksajakovljevic: Right, as you've noticed as it is now traces and metrics share the same pool of writers. Since not many users use ingest metrics/traces at the same time I haven't been focusing much on this. I'll create an issue to keep a track of this.

cevian (changes requested):
Overall looks good but needs some changes

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	flushBatch := func(batch *Batch) *Batch {
+		b.bufferedBatches <- *batch
+		batcherSpan.AddEvent("Batch sent to buffer")
+		batcherSpan.AddEvent("New Batch")
+		metrics.IngestorPendingBatches.With(prometheus.Labels{"type": "trace"}).Inc()
+		return NewBatch(b.config.MaxBatchSize)
+	}
+	processReq := func(req insertTracesReq) {
+		metrics.IngestorRequestsQueued.With(prometheus.Labels{"type": "trace"}).Dec()
+		if batch.isFull() {
```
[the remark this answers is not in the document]
niksajakovljevic: Sure

pkg/pgmodel/ingestor/trace/trace_batcher.go
Harkishen-Singh: This remains.

Harkishen-Singh (approved):
One nit remaining, otherwise LGTM 👍🏻

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+			batch = flushBatch(batch)
+			ticker.Reset(b.config.BatchTimeout)
+		}
+		batch.add(req)
+	}
+	for {
+		select {
+		case item := <-b.in:
+			processReq(item)
+		case <-ticker.C:
```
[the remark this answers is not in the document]
niksajakovljevic: The ticker is only reset when we flush on size. Imho the reset on first request is matter of semantics - I feel that in practice it makes almost no difference.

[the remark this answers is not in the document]
niksajakovljevic: Not sure I understand why it would be reset on tick every tick?
Currently we only reset when we flush full batch.  To make any decent performance we need to flush by size - timeouts are just for worst case scenarios... Anyhow we can add reset on the first request in a batch to make sure that batch waits the timeout.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	b.once.Do(func() {
+		close(b.stop)
+		b.wg.Wait()
+		close(b.bufferedBatches)
+		b.batchWriter.stop()
+	})
+}
+
+// batchWriter writes batches using a writer.
+type batchWriter struct {
```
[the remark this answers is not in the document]
niksajakovljevic: Another naming discussion `writer` vs `copier` :)  Imho`writer` is more appropriate b/c it actually calls `writer` which writes to DB. `copier` seems tied to PG `COPY` command?  Note that we already have trace `writer` that implement DB logic so it's also inline with that

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+				}
+			}
+		}()
+	}
+}
+
+func (bw *batchWriter) flush(b Batch) {
+	_, flushSpan := tracer.Default().Start(b.ctx, "flush-batch")
+	defer flushSpan.End()
+	if b.spanCount != 0 {
```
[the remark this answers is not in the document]
niksajakovljevic: It's just extra safety check, but not really needed. We only flush non-empty batches.

pkg/pgmodel/ingestor/trace/trace_batcher.go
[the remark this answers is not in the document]
niksajakovljevic: It's only 60-70 lines of code and it's coupled with batcher (the actually write logic is sitting in writer.go).
I am not feeling strongly about this so let me know.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/otel/trace"
+	_ "go.uber.org/automaxprocs"
+)
+
+const (
+	defaultReqBufferSize      = 100000                 // buffer for incoming requests, especially important for async acks
+	DefaultBatchSize          = 1000                   // this is soft limit as we might produce bigger batches in some cases
+	DefaultBatchTimeout       = 100 * time.Millisecond // we should aways aim at reching size limits, not timeout
+	defaultMaxBufferedBatches = 2000                   //  We want to avoid wait on batches
```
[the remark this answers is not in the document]
niksajakovljevic: Indeed. I was experimenting a bit with these settings. I made it now `5 x NumCopiers`.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgmodel/metrics"
+	"github.com/timescale/promscale/pkg/tracer"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/otel/trace"
+	_ "go.uber.org/automaxprocs"
+)
+
+const (
+	defaultReqBufferSize      = 100000                 // buffer for incoming requests, especially important for async acks
```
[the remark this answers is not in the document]
niksajakovljevic: Hmm...not sure about this one. My idea was to have enough buffer for spikes. `100K` might be too much. In the recent benchmark runs I did I've never seen this going above 1K. Maybe we can set it to 5K for now and tune as we go. Frankly fine tuning these requires time and a lots of benchmark runs... Please let me know if you have a better idea.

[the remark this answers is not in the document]
niksajakovljevic: I actually tweak this to be MaxBatchSize * 3. I also split incoming request into multiple queues (we have a separate queue for each batcher). I added a separate commit for easier review.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,283 @@
...
+	b.batchWriter.run()
+}
+
+func (b *Batcher) batch(batchIdx int) {
+	_, batcherSpan := tracer.Default().Start(context.Background(), "create-batches")
+	ticker := time.NewTicker(b.config.BatchTimeout)
+	defer ticker.Stop()
+	batch := NewBatch(b.config.MaxBatchSize)
+	batcherSpan.AddEvent("New Batch")
+	flushBatch := func(batch *Batch) *Batch {
```
cevian, line 146: During a flush, I'd consider stopping the ticker or resetting it to a really long period. It shouldn't really fire until reset again on first insert.

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,283 @@
...
+		loop:
+			for {
+				select {
+				case item, ok := <-b.in[batchIdx]:
+					if !ok {
+						break loop
+					}
+					processReq(item)
+				case <-timeoutCtx.Done():
+					log.Warn("msg", "Forced batcher shutdown due to timeout. Some request might not be persisted.", "batcher", batchIdx)
```
cevian, line 189: can we check len(b.in) to see if "Some request might not be persisted" and print that only if some requests were dropped?

pkg/pgmodel/ingestor/trace/trace_batcher.go
```diff
@@ -0,0 +1,283 @@
...
+func (bw *batchWriter) drainBuffer(ctx context.Context) {
+	for {
+		select {
+		case b, ok := <-bw.batches:
+			if !ok {
+				return
+			}
+			bw.flush(b)
+		case <-ctx.Done():
+			log.Warn("msg", "Forced batchWriter shutdown due to timeout. Some batches might not be persisted.")
```
cevian, line 272: again can we actually check if some batches weren't persisted before printing this msg

pkg/pgmodel/ingestor/trace/trace_dispatcher.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		return err
+	}
+}
+
+// This is to reduce lock contention for DB when spans are coming one by one (eg from Jaeger collector).
+// If it's only one span we shard by it's TraceID so spans with the same TraceID end up in the same batcher.
+// Otherwise we roun-robin between batchers
+func (td *Dispatcher) getBatcherIdx(ctx context.Context, traces ptrace.Traces) (int, error) {
+	numberOfBatchers := td.batcher.config.Batchers
```
cevian, line 88: I don't love that the dispatcher needs to care about the number of batchers but this may be unavoidable...

cevian (approved):

PR #1561, opened 2022-08-16
Reset inverted labels cache on epoch change

JamesGuthrie:
## Description

The database cache epoch advances when series are deleted from the
database. When connector detects a cache epoch change, it invalidates
the series cache, as it may contain values which reference the deleted
rows in the series table.

During the database series deletion, unused labels may also be deleted.
This means that we must also invalidate the inverted label cache, as it
could also reference deleted rows in the `_prom_catalog.label` table.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ We use the following categories for changes:
 
 ### Fixed
 - Make Jaeger Event queryable using name and tags [#1553]
+- Reset inverted labels cache on epoch change [#1561]
 
 ## [0.13.0] - 2022-07-20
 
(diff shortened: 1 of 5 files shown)


sumerman (approved):

niksajakovljevic (approved):
If unused labels are dropped I think we also need to take care of labels cache (not just inverted labels cache). Check `labels_reader.go`

JamesGuthrie:
> If unused labels are dropped I think we also need to take care of labels cache (not just inverted labels cache). Check `labels_reader.go`

That is correct, but stale entries in the labels cache are much less problematic so we decided to only address the inverted labels cache for now.

Issue #1547, opened 2022-08-04
Metric views registered with `register_metric_view` are not shown in Grafana metrics browser


oliora:
Scenario:
1. Create a continuous aggregate metric view `test_metric_delete_me_1m` based on Promscale metric
2. Register the view with `SELECT register_metric_view('public', 'test_metric_delete_me_1m');`

Expected behavior:
Metric `test_metric_delete_me_1m` is shown in the list of available metrics in Grafana

Actual behavior:
Metric is not shown in the list even after reloading Grafana page. Note that the issue is only with exposing the metric name, the data itself is properly exposed i.e. I get proper data from a query like `test_metric_delete_me_1m{__column__="min"}`.

oliora:
Somehow related: metrics deleted with `drop_metric` are still visible in the Grafana metrics browser.

Harkishen-Singh:
@oliora thanks for reporting the issue!

> Metric is not shown in the Grafana metrics browser after reloading the page

I can confirm that with no `__column__` provided, no values are returned in Grafana. My guess is that this is because Promscale Caggs always expect a column, as there is no default column. With column provided, it works as expected. I will look into this more and try to get a fix or some reason.

> metrics deleted with drop_metric are still visible in the Grafana metrics browser

I tried to reproduce this by deleting a metric using `select prom_api.drop_metric('go_gc_duration_seconds');`. After deleting, I could not find any data in Grafana. Did you do the same, simply dropping the metric using this function, or something else?

oliora:
@Harkishen-Singh yes, I’ve just called `prom_api.drop_metric` as you said. I’ll try to reproduce it one more time and come back.
Btw, before you’ve deleted the metric have you seen it in Grafana metrics explorer?

Harkishen-Singh:
> I can confirm that with no __column__ provided, no values are returned in Grafana

@oliora, to get values in PromQL from Caggs without using `__column__`, you should create a column with name `value` while creating Caggs. This is because, when Promscale gets a query without any `__column__` label, it looks for this `value` column. If not found, it returns empty result.

> Btw, before you’ve deleted the metric have you seen it in Grafana metrics explorer?

Yes. I first visualised `go_gc_duration_seconds` in Grafana, then dropped using `select prom_api.drop_metric('go_gc_duration_seconds');` (had to restart Promscale because drop_metric needs all Promscale connectors off) and then I couldn't find the metric next.

oliora:
[...]

PR #1563, opened 2022-08-17
WIP: Fix: Use schema and column from CLI as defaults

Harkishen-Singh:
## Description

Fixes: #1547 

This PR uses the `defaultSchema` and `defaultColumn` from CLI if a Caggs PromQL query does not return any data. This may be because the user missed applying a `__colum__` information, which case we should return a default column. This default column defaults to `avg` since most downsampling systems use/expect `avg`. This features behaves just like `metrics.promql.default-subquery-step-interval` which is applied if the user forgets to use a `step` value in PromQL subquery.

Note: DO NOT REVIEW. Needs more work based on some quick discussion.

## Merge requirements


Harkishen-Singh:
Based on the team discussion, we have decided to return all the `__column__` as individual series if the `__column__` label was not found. This will be in line with the multi-columnar PromQL queries.

Hence, closing this PR as per the above plan.

PR #1501, opened 2022-07-18
WIP: Clean up cache interfaces

JamesGuthrie:
## Description

The clockcache is used in 9 caches, but understanding what each of those
caches actually does is quite difficult, because few of them expose a
clean interface.

This makes the clockcache generic, and propagates type parameters to all
usages of the cache, exposing a clear interface.

## Merge requirements

Please take into account the following non-code changes that you may need to make with your PR:

- [ ] ~~CHANGELOG entry for user-facing changes~~
- [ ] ~~Updated the relevant documentation~~

PR #1564, opened 2022-08-17
Bump to go 1.18

JamesGuthrie:
## Merge requirements

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/timescale/promscale
 
-go 1.17
+go 1.18
 
 require (
 	github.com/NYTimes/gziphandler v1.1.1


alejandrodnm:
How do we decide when to upgrade to golang version? 

BTW 1.19 is already out we got the following PR

- https://github.com/timescale/promscale/pull/1502/files
- https://github.com/timescale/promscale/pull/1539

JamesGuthrie:
> How do we decide when to upgrade to golang version?

I don't know. I don't think that we have any process. I opened this PR because I wanted to see the result of the benchmark compared to another PR which I have open which would also bump the golang version (https://github.com/timescale/promscale/pull/1501).

I would suggest that once those other PRs are merged, we bump to 1.19, but I like being on the bleeding edge 😂

alejandrodnm:
But this is just the minimum Go version supported, right? Because all the build references I found use go 1.18

- https://github.com/timescale/promscale/blob/8ccdc0f9992cfe8210134027d7be905354fc4488/.github/workflows/go-releaser.yml#L15
- https://github.com/timescale/promscale/blob/8ccdc0f9992cfe8210134027d7be905354fc4488/build/Dockerfile#L2 ->
https://github.com/timescale/promscale/blob/8ccdc0f9992cfe8210134027d7be905354fc4488/.goreleaser.yml#L129

JamesGuthrie:
> But this is just the minimum Go version supported, right? Because all the build references I found use go 1.18

Yes.

PR #1566, opened 2022-08-18
Fix tag matching correctness and optimization

### LINKED ISSUE #1467: Can't filter tags with values that are strings, but look like numbers.
If you try to search for tag `exampleTag=1234` in jaeger-query, and the value of `exampleTag` is stored as a string in spans, filtering won't work.

### LINKED ISSUE #1557: Improve tag filtering for traces in jaeger
Tag filtering for traces in jaeger queries need improvement
- indexes are often not used
- searches should cover all possible textual representations #1467 
- tag queries for non-existent tags should be quick

cevian:
Previously, tag matching did not correctly handle
numeric types that were written as jsonb strings (#1467).
This commit fixes that by using the new _ps_trace.text_matches
function.

This also fixes the fact that the gin index was not used
on trace searches. This is because the query contained
OR conditions between the span tags on the span table
and the event tags on the event table. Postgres does
not support using indexes for such OR conditions. (
and this would be very hard to add support for)

We also tried using UNION ALL to optimize this,
but that approach falls apart when considering
matches on multiple tags (e.g. tag1=val1,tag2=val2)

In the end we switched to a 2 round protocol:
round 1: fetch tags and their types
round 2: add quals for tags only for the types present.

This avoids the case above for tags that don't have
the same key, values on both spans and events. As far
as we know, this is a rare case.

The other downside is that these queries now require 2
database calls. In the future, this can be optimized by
moving the query generation logic into PLPGSQL functions.

Other optimizations:
- if no tags are found in round 1 we can return an empty
  result early
- avoid join to the event table when not needed
- add quals on event times to the query for chunk exclusion

Fixes #1557, Fixes #1467 

## Merge requirements


niksajakovljevic:
Good catch!
I am confused how the tests are passing b/c I see no extension version change and this change utilizes new function introduced in https://github.com/timescale/promscale_extension/pull/461 ?

pkg/jaeger/store/tag_info.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	params = append(params, keys, values)
+	c := fmt.Sprintf(`
+		SELECT result.*
+		FROM unnest ($%d::text[], $%d::text[]) with ordinality as tag_qual(key, value, nr)
+		INNER JOIN LATERAL
+		(
+			SELECT
+				array_agg(pg_catalog.jsonb_build_object(tag.key_id, tag.id)) json_pair_array,
+				(bit_or(tag.tag_type) OPERATOR(pg_catalog.&) ps_trace.span_tag_type() != 0) is_span_type,
```
niksajakovljevic, line 118: Wondering why can't we use PG bitwise operator but `OPERATOR(pg_catalog.&)`?

pkg/tests/end_to_end_tests/ingest_trace_test.go
```diff
@@ -79,9 +80,65 @@ func TestQueryTraces(t *testing.T) {
 		getOperationsTest(t, q)
 		findTraceTest(t, q)
 		getDependenciesTest(t, q)
+		findTracePlanTest(t, q, db)
 	})
 }
 
+func findTracePlanTest(t testing.TB, q *store.Store, db *pgxpool.Pool) {
```
niksajakovljevic, line 87: Seeing this test here makes me believe that it's more appropriate to generate SQL in the extension and have it tested there... We already moved some pure SQL things into the extension.

arajkumar: +1, What is the contract defined between connector and extension?

pkg/jaeger/store/tag_info.go
```diff
@@ -0,0 +1,168 @@
...
+			(
+					SELECT v, _prom_ext.jsonb_digest(v) as digest
+					FROM unnest(_ps_trace.text_matches(tag_qual.value)) as v
+			) as matchers
+			INNER JOIN LATERAL (
+				SELECT t.*
+				FROM _ps_trace.tag t
+				WHERE t.key OPERATOR(pg_catalog.=) tag_qual.key
+				AND _prom_ext.jsonb_digest(t.value) OPERATOR(pg_catalog.=) matchers.digest
+				AND t.value OPERATOR(pg_catalog.=) matchers.v
```
arajkumar, line 131: Isn't value comparison and value digest comparison would be always true?

pkg/jaeger/store/tag_info.go
```diff
@@ -0,0 +1,168 @@
...
+
+type tagsInfo struct {
+	spanClauses      []string
+	operationClauses []string
+	eventClauses     []string
+	generalTags      []*tag
+	params           []interface{}
+}
+
+func NewTagInfo(ctx context.Context, q *spanstore.TraceQueryParameters, conn pgxconn.PgxConn) (*tagsInfo, error) {
```
arajkumar, line 34: nit:

Should we can it `FindTagInfo`?

pkg/jaeger/store/trace_query.go
```diff
@@ -206,86 +197,84 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
...
+			   FROM _ps_trace.operation op
+			   WHERE
+			      %s
+		`, strings.Join(operation_clauses, " AND "))
+		return subquery, params
+	}
+	return "", params
+}
+
+func (b *Builder) buidTagClauses(tInfo *tagsInfo, params []interface{}) (string, []interface{}, bool) {
```
arajkumar, line 213: nit: typo

```suggestion
func (b *Builder) buildTagClauses(tInfo *tagsInfo, params []interface{}) (string, []interface{}, bool) {
```

pkg/jaeger/store/trace_query.go
```diff
@@ -183,11 +177,8 @@ func (b *Builder) getTraceQuery(traceID model.TraceID) (string, []interface{}, e
 	return b.buildCompleteTraceQuery(traceIDClause, "", ""), params, nil
 }
 
-func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (string, []interface{}) {
-	clauses := make([]string, 0, 15)
-	params := make([]interface{}, 0, 15)
-
-	operation_clauses := make([]string, 0, 2)
+func (b *Builder) buildOperationSubquery(q *spanstore.TraceQueryParameters, tInfo *tagsInfo, params []interface{}) (string, []interface{}) {
+	operation_clauses := tInfo.operationClauses
```
arajkumar, line 181: nit: go style
```suggestion
	operationClauses := tInfo.operationClauses
```

pkg/jaeger/store/trace_query.go
```diff
@@ -206,86 +197,84 @@ func (b *Builder) buildTraceIDSubquery(q *spanstore.TraceQueryParameters) (strin
...
+		return subquery, params
+	}
+	return "", params
+}
+
+func (b *Builder) buidTagClauses(tInfo *tagsInfo, params []interface{}) (string, []interface{}, bool) {
+	clauses := make([]string, 0, len(tInfo.generalTags))
+	hasEventTags := false
+	for _, tag := range tInfo.generalTags {
+		tag_clauses := make([]string, 0, 3)
```
arajkumar, line 217: nit: go style
```suggestion
		tagClauses := make([]string, 0, 3)
```

arajkumar:
@cevian 

This implementation fails when the query tags spread across different events. For example when running hot-rod app, tag query "event=baggage customer_id=123" fails to yield result however both in memory store and our master works fine.

cevian:
@niksajakovljevic

cevian:
@niksajakovljevic 
> Good catch!
I am confused how the tests are passing b/c I see no extension version change and this change utilizes new function introduced in https://github.com/timescale/promscale_extension/pull/461 ?

This uses our same-branch-names between extension and connector test against one-another convention to test this pr against 461

pkg/jaeger/store/tag_info.go
cevian: This is just a slightly safer way to write `&` since it's schema-qualified

pkg/tests/end_to_end_tests/ingest_trace_test.go
cevian: While re-writing the logic in a database function may have many benefits, I think that's a huge change that is beyond the scope of this PR.

pkg/jaeger/store/tag_info.go
cevian: This is to future-proof against possible collisions

cevian:
@niksajakovljevic @arajkumar I Updated the EXTENSION_VERSION file as I believe we'll need this change after the PR is merged and the same-branch-name stuff in CI no longer works. I also think we should modify ExtVersionRangeString as the current version (0.5.4) is no longer valid. So, you should change `>=` to `>`. Is this correct?

arajkumar (approved):
lgtm

Issue #1567, opened 2022-08-19
column "exemplar_label_values" does not exist


Yamakaky:
I have an issue when ingesting metrics with exemplars. Unless I'm mistaken, I do not get exemplars when doing promql queries. What additional info would you need?

```
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-508] postgres@tsdb,app=promscale@0.13.0 [23505] ERROR:  duplicate key value violates unique constraint "ei_latency_bucket"
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-509] postgres@tsdb,app=promscale@0.13.0 [23505] DETAIL:  Key (series_id, "time")=(17372, 2022-08-19 12:44:02.235+00) already exists.
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-510] postgres@tsdb,app=promscale@0.13.0 [23505] CONTEXT:  COPY latency_bucket, line 7
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-511] postgres@tsdb,app=promscale@0.13.0 [23505] STATEMENT:  copy "prom_data_exemplar"."latency_bucket" ( "time", "series_id", "exemplar_label_values", "value" ) from stdin binary;
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-512] postgres@tsdb,app=promscale@0.13.0 [42703] ERROR:  column "exemplar_label_values" does not exist at character 29
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-513] postgres@tsdb,app=promscale@0.13.0 [42703] STATEMENT:  select "time", "series_id", "exemplar_label_values", "value" from "s1_latency_bucket"
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-514] postgres@tsdb,app=promscale@0.13.0 [42703] ERROR:  column "exemplar_label_values" does not exist at character 29
2022-08-19 12:44:02 UTC [85]: [62ff83ec.55-515] postgres@tsdb,app=promscale@0.13.0 [42703] STATEMENT:  select "time", "series_id", "exemplar_label_values", "value" from "s0_latency_bucket"
2022-08-19 12:44:03 UTC [85]: [62ff83ec.55-516] postgres@tsdb,app=promscale@0.13.0 [23505] ERROR:  duplicate key value violates unique constraint "ei_latency_bucket"
[...]

PR #1574, opened 2022-08-22
Fix: Exemplar ingestion due to conflicting temporary tables.

Harkishen-Singh:

Fixes: #1567 

This PR fixes the `ERROR: column \"exemplar_label_values\" does not exist (SQLSTATE 42703)` error seen in exemplar ingestion. Based on what I studied, the error occurred due to temporary tables created for samples being used for exemplars as well. Exemplars have a different column requirements and hence the error.

## Merge requirements


niksajakovljevic (approved):
I don't quite understand how this fixes the problem reported in https://github.com/timescale/promscale/issues/1547 ?

Issue #1573, opened 2022-08-22
Regression of #1176


oliora:
There was a bug https://github.com/timescale/promscale/issues/1176 that was fixed by upgrading PGX from **v4.15.0** to **v4.15.1-0.20220219175125-b6b24f9e8a5d**. Later if was upgraded further to **v4.16.0** (I haven't followed the reason though). Later the PGX upgrades were reverted by https://github.com/timescale/promscale/commit/c7acfc8f897bc2c72b39db73584ab3c16ceb7eb6 which most probably the reason why I see crashes similar to #1176 happening to me.

jgpruitt:
We have upgraded the pgx libraries. It ought to go out with the next release.

Issue #1176, opened 2022-02-21
Promscale crashing on trace ingest


niksajakovljevic:
While ingesting trace data with `otel-collector` Promscale would randomly crash with stack trace:

```
panic: BUG: unexpected error from stmtCache

goroutine 15280 [running]:
github.com/jackc/pgx/v4.(*Conn).SendBatch(0xc059743f00, 0x1792860, 0xc0c2c5d980, 0xc00b71c300, 0xc05c11ac30, 0xc05c130280)
	/go/pkg/mod/github.com/jackc/pgx/v4@v4.15.0/conn.go:801 +0x1734
github.com/jackc/pgx/v4/pgxpool.(*Conn).SendBatch(0xc03878bb50, 0x1792860, 0xc0c2c5d980, 0xc00b71c300, 0x0, 0x0)
	/go/pkg/mod/github.com/jackc/pgx/v4@v4.15.0/pgxpool/conn.go:66 +0x6f
github.com/jackc/pgx/v4/pgxpool.(*Pool).SendBatch(0xc00097e150, 0x1792860, 0xc0c2c5d980, 0xc00b71c300, 0x0, 0x0)
	/go/pkg/mod/github.com/jackc/pgx/v4@v4.15.0/pgxpool/pool.go:543 +0x8c
github.com/timescale/promscale/pkg/pgxconn.(*connImpl).SendBatch(0xc000788148, 0x1792860, 0xc0c2c5d980, 0x17777e0, 0xc00b71c300, 0x4da486, 0x620e2e18, 0x3f3bcae1323d4e60, 0x1db12002a0b7b)
	/promscale/pkg/pgxconn/pgx_conn.go:136 +0x65
github.com/timescale/promscale/pkg/pgmodel/ingestor/trace.(*traceWriterImpl).sendBatch(0xc000bf23c0, 0x1792860, 0xc0c2c5d980, 0x17777e0, 0xc00b71c300, 0x0, 0x0)
	/promscale/pkg/pgmodel/ingestor/trace/writer.go:367 +0x5f
github.com/timescale/promscale/pkg/pgmodel/ingestor/trace.(*traceWriterImpl).InsertTraces(0xc000bf23c0, 0x1792860, 0xc0c2c5d980, 0xc0cd016620, 0x0, 0x0)
	/promscale/pkg/pgmodel/ingestor/trace/writer.go:352 +0x2c14
[...]

PR #1575, opened 2022-08-22
upgrade pgx dependencies

jgpruitt:
## Description

A user has reported a regression and it is suspected that upgrading pgx may solve it.
https://github.com/timescale/promscale/issues/1573

## Merge requirements


Harkishen-Singh:
Yesterday in team meeting, we discussed about this issue. The plan was to ask the user how he produced the issue and then check on our side. This is because we haven't seen that issue in recent installations in Promscale.

oliora:
@Harkishen-Singh I can post a callstack of the crash. I don’t have a scenario how to reproduce it. I’ve just got it sometimes (about once per week) under conditions that matches https://github.com/timescale/promscale/issues/1176

Issue #1428, opened 2022-06-13
/api/v1/labels should only return metric labels for the specified tenants in the Promscale multi-tenancy configuration


ramonguiu:
In a multi-tenant setup where Promscale has been configured to only return data for a specific set of tenants, the labels API (/api/v1/labels) doesn't respect that setting. It currently returns label data for all labels in the system which is wrong.

Currently all labels for all tenants are stored together. Separating them by tenant requires a change in the data model to make the API call execution performant.

JosefPihrt:
@ramonguiu ping

Harkishen-Singh:
I think we could simply filter the tenants not permitted in multi-tenancy after fetching from the db before replying to the Api request. That should work as a quick fix to this issue.

@JosefPihrt sorry for the delay. I will give this issue a try sometime today and see if it can be fixed without major changes. If so, then we can target this in the next release.

JosefPihrt:
@Harkishen-Singh Thanks in advance!

ramonguiu:
@JosefPihrt so sorry I missed your mentions in this ticket. I would love to understand more about your use case. If you would be open to a discussion please drop me an email (ramon at timescale.com).

PR #1577, opened 2022-08-23
Fix: labels API to respond with labels of authorized tenants only when in multi-tenancy.

Harkishen-Singh:

Fixes: #1428 

This PR fixes the `/api/v1/labels` & `/api/v1/label/:name/values` in multi-tenancy mode.

Example query for `/api/v1/labels`
```postgresql
SELECT
   array_agg(distinct l.key) 
FROM
   _prom_catalog.label l 
WHERE
   EXISTS
   (
      SELECT
         1 
      FROM
         _prom_catalog.series 
      WHERE
         (
            labels @> array[l.id] 
            AND labels @> ( 
            SELECT
               array_agg(id :: INTEGER) 
            FROM
               _prom_catalog.label 
            WHERE
               key = '__tenant__' 
               AND value = 'tenant-a' )::int[] 
         )
   )
```

Example query for `/api/v1/label/:name/values`
```postgresql
SELECT
   array_agg(distinct a.value) 
FROM
   (
      SELECT
         * 
      FROM
         _prom_catalog.label l 
      WHERE
         EXISTS
         (
            SELECT
               1 
            FROM
               _prom_catalog.series 
            WHERE
               (
                  labels @> array[l.id] 
                  AND labels @> ( 
                  SELECT
                     array_agg(id :: INTEGER) 
                  FROM
                     _prom_catalog.label 
                  WHERE
                     key = '__tenant__' 
                     AND value = 'tenant-a' )::int[] 
               )
         )
   )
   a 
WHERE
   a.key = '<some_label_name>'
```

Update: The PR now uses the optimised query (still slower than the non-multi-tenancy case) for getting label names or label values corresponding to authorised tenants only. Using this query is optional and can be controlled via -metrics.multi-tenancy.experimental.label-queries flag. This allows enables the users with authorized labels results with some cost on performance.

## Merge requirements


pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -78,9 +80,79 @@ func (lr *labelsReader) LabelValues(labelName string) ([]string, error) {
 	return labelValues, nil
 }
 
+const (
+	getLabelIDofValidTenant           = `SELECT id::INTEGER FROM _prom_catalog.label WHERE key = '__tenant__' AND value = $1::TEXT`
+	labelCompareClause                = ` $%d::INTEGER = any(labels) `
+	getLabelNamesBelongingToTenantIDs = `
+SELECT ARRAY_AGG(a.key) FROM
```
arajkumar, line 87: Can we use the following query which hopefully gives the expected result in a single query.?

```
select distinct key from "_prom_catalog".label where id in (select unnest(labels) as valid_label_ids from "_prom_catalog".series where labels && (SELECT array_agg(id::INTEGER) FROM _prom_catalog.label WHERE key = '__tenant__' and value=any(array['foo', 'bar'])));
```
'foo' and 'bar' are the tenants here.

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -78,9 +79,49 @@ func (lr *labelsReader) LabelValues(labelName string) ([]string, error) {
...
+	if lr.authConfig != nil {
+		// Multi-tenancy is enabled. Hence, we have to use a different LabelNames() query
+		// only when some tenants are authorized, i.e., when using NewSelectiveTenancyConfig().
+		//
+		// Note: Label names of non-tenants will not be sent. Only label names belonging to
+		// authorized tenants will be sent. This means, if -metrics.multi-tenancy.allow-non-tenants
+		// is enabled, the non-tenant labels will not be answered ATM.
+		validTenants, allTenantsValid := lr.authConfig.ValidTenants()
+		if len(validTenants) == 0 {
+			return []string{}, nil
```
niksajakovljevic, line 111: Maybe log a warning here that no valid tenants are found. It would make it easier for debugging if no labels are retutned

niksajakovljevic:
I am a bit skeptical about how performant will this be b/c we are scanning `series` table which usually have millions of rows... I'd recommend either benchmarking this with decent amount of series before merging. I am aware that with current model we need to scan series but if it performs really bad we should consider some other approach.... Fetching of labels is mostly used by UI tools so it might have huge impact on user experience

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -736,6 +736,99 @@ func TestMultiTenancyWithValidTenantsAsLabels(t *testing.T) {
 	})
 }
 
+func TestMultiTenancyLabelNames(t *testing.T) {
+	ts, _ := generateSmallMultiTenantTimeseries()
+	withDB(t, *testDatabase, func(db *pgxpool.Pool, t testing.TB) {
+		getClient := func(auth tenancy.Authorizer) *pgclient.Client {
```
arajkumar, line 742: Is it possible to convert this into test table ? I see quite a bit redundancy and might possible to generalise into TT.

Harkishen-Singh:
Note: After some discussion, we should benchmark this PR against few million series and see the effect on `/api/v1/labels` endpoint. This will be the next task in roadmap of this PR.

Harkishen-Singh:
Update: The PR now uses the optimised query (still slower than the non-multi-tenancy case) for getting label names or label values corresponding to authorised tenants only. Using this query is optional and can be controlled via `-metrics.multi-tenancy.experimental.label-queries` flag. This allows enables that need accurate results with some cost on performance.

niksajakovljevic (approved):

pkg/tests/end_to_end_tests/multi_tenancy_test.go
```diff
@@ -736,6 +751,103 @@ func TestMultiTenancyWithValidTenantsAsLabels(t *testing.T) {
...
+				validTenants:                      []string{"tenant-b"},
+				allowNonTenants:                   true,
+				expectedLabelNames:                []string{"__name__", "__tenant__", "common", "empty", "foo", "ins", "job", "shared"},
+				expectedLabelValuesForTenantLabel: []string{"tenant-b"},
+				expectedLabelValuesForSharedLabel: []string{"first", "second"},
+			},
+		}
+
+		for _, tc := range tcs {
+			func() {
```
arajkumar, line 818: Use t.Run here
```suggestion
			t.Run(tc.name, func(t *testing.T) {
```

arajkumar (approved):
lgtm with a small nit. 
Disclaimer: I didn't really tested the new query

PR #1581, opened 2022-08-25
Allow using version 0.6.0 of the Promscale extension

JamesGuthrie:
## Description

We must relax the extension version constraints in order for the
extension's release tests to pass. The constraints can be tightened
once the new version of the extension is release.

## Merge requirements


jgpruitt (approved):
approved. might want to get rid of the comment on line 54 though

jgpruitt (approved):

PR #1590, opened 2022-08-29
Filter by label key before Exists() for LabelValues() in multi-tenancy mode

Harkishen-Singh:
## Description

Same as the heading.

## Merge requirements

### DIFF
--- a/pkg/pgmodel/lreader/labels_reader.go
+++ b/pkg/pgmodel/lreader/labels_reader.go
@@ -49,7 +49,7 @@ const (
 EXISTS(
     SELECT 1 FROM _prom_catalog.series WHERE (
         labels @> array[l.id] AND 
-          labels @> 
+          labels && 
           (
              SELECT array_agg(id :: INTEGER) 
              FROM _prom_catalog.label 
@@ -59,7 +59,7 @@ EXISTS(
 )
 `
 	getLabelNamesForTenant  = `SELECT array_agg(distinct l.key) FROM _prom_catalog.label l WHERE %s`
-	getLabelValuesForTenant = `SELECT array_agg(distinct a.value) FROM (SELECT * FROM _prom_catalog.label l WHERE %s) a WHERE a.key = %s`
+	getLabelValuesForTenant = `SELECT array_agg(distinct value) FROM (SELECT * FROM _prom_catalog.label l WHERE key = $1 AND ( %s )) a`
 )
 
 type labelsReader struct {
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -59,7 +59,7 @@ EXISTS(
 )
 `
 	getLabelNamesForTenant  = `SELECT array_agg(distinct l.key) FROM _prom_catalog.label l WHERE %s`
-	getLabelValuesForTenant = `SELECT array_agg(distinct a.value) FROM (SELECT * FROM _prom_catalog.label l WHERE %s) a WHERE a.key = %s`
+	getLabelValuesForTenant = `SELECT array_agg(distinct value) FROM _prom_catalog.label l WHERE key = $1 AND %s`
```
cevian, line 62:
```suggestion
	getLabelValuesForTenant = `SELECT array_agg(distinct value) FROM _prom_catalog.label l WHERE key = $%d AND %s`
```

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -78,13 +78,12 @@ func (lr *labelsReader) LabelValues(labelName string) ([]string, error) {
...
-		existsClause, args := []string{}, []interface{}{}
+		existsClause, args := []string{}, []interface{}{labelName}
 		for i, tenantName := range validTenants {
-			existsClause = append(existsClause, fmt.Sprintf(tenantExists, i+1))
+			existsClause = append(existsClause, fmt.Sprintf(tenantExists, i+2)) // i+2 since $1 is reserved for labelName.
 			args = append(args, tenantName)
 		}
-		args = append(args, labelName)
-		labelValuesQuery := fmt.Sprintf(getLabelValuesForTenant, strings.Join(existsClause, " OR "), fmt.Sprint("$", len(args)))
+		labelValuesQuery := fmt.Sprintf(getLabelValuesForTenant, strings.Join(existsClause, " OR "))
```
cevian, line 86:
```suggestion
		labelValuesQuery := fmt.Sprintf(getLabelValuesForTenant, len(args), strings.Join(existsClause, " OR "))
```

pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -78,13 +78,12 @@ func (lr *labelsReader) LabelValues(labelName string) ([]string, error) {
 			log.Debug("msg", "no tenants found for LabelValues()")
 			return []string{}, nil
 		}
-		existsClause, args := []string{}, []interface{}{}
+		existsClause, args := []string{}, []interface{}{labelName}
 		for i, tenantName := range validTenants {
-			existsClause = append(existsClause, fmt.Sprintf(tenantExists, i+1))
+			existsClause = append(existsClause, fmt.Sprintf(tenantExists, i+2)) // i+2 since $1 is reserved for labelName.
 			args = append(args, tenantName)
```
cevian, line 84:
```suggestion
                         args = append(args, tenantName)
			existsClause = append(existsClause, fmt.Sprintf(tenantExists,len(args)))
			
```

cevian (approved):
Approved if the comments are implemented.

pkg/pgmodel/lreader/labels_reader.go
Harkishen-Singh: Why `%d` here, when it will always be `1`?

pkg/pgmodel/lreader/labels_reader.go
Harkishen-Singh: This will be wrong, as here, it will apply `$3` which is the size of args, but the requirement is for `$1`.

PR #1591, opened 2022-08-29
Rework ingest pipeline documentation

JamesGuthrie:
## Description

When working on some changes to the caches, I found that the existing
documentation of the write path is somewhat out of date, and not 1:1
comparable to the existing code.

## Merge requirements

### DIFF
--- a/pkg/pgmodel/Readme.md
+++ b/pkg/pgmodel/Readme.md
@@ -3,7 +3,7 @@
 This package contains the core Promscale functionality: the code to translate Read and Write requests into things the database can understand. Logically this code is divided into two mostly separate paths:
 
 1. The Read path, which is entered through the `Reader` interface, and translates `ReadRequest`s into SQL requests the DB can understand.
-2. The Write path, entered via `Ingest()`, inserts data from `WriteRequests`
+2. The Write path, entered via `ingestor.IngestMetrics()`, inserts data from `WriteRequests`
 
 The code paths for these two areas of functionality are large independent (I believe the only things shared are the schema names and the metric cache) and may be split into separate packages at some point.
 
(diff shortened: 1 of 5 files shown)


pkg/pgmodel/Readme.md
```diff
@@ -36,46 +80,51 @@ type Sample {
...
-                    \ /
- +----------------+  |  +----------------+       +------------------+
- | insert handler | ... | insert handler | ====> | DB (metric data) |
- +----------------+     +----------------+       +------------------+
-```
+If asyncAcks=true, the dispatcher returns immediately, otherwise it waits for the metric batcher goroutine to return a result, and returns that in response.
+
+### Metric batcher
+
+The second stage consists of the metric batcher, one goroutine for each metric name. These goroutines are created in `dispatcher.InsertTs`, but are pooled for later re-use.
```
niksajakovljevic, line 106: `one batcher running in it's own goroutine for each metric name`

niksajakovljevic, line 106: I think this might be a bit misleading as someone might think there is a pool of batchers which are re-used. Essentially we just have a batcher per metric name running in it's own goroutine and communicating with dispatcher and copiers using channels.

pkg/pgmodel/Readme.md
```diff
@@ -36,46 +80,51 @@ type Sample {
...
+
+On the first new insert request, the metric batcher creates a metric table in the database, if it doesn't exist yet.
+
+From then onwards, when the metric batcher receives a new insert request on its channel it: 
+1. Creates an unbuffered data-transfer channel for this batcher
+2. Initiates a read request to the copier on the copier read-request channel, sending a handle to the batcher's data-transfer channel
+3. Continues to (in a loop over `select`):
+   1. accept data for the metric (from the dispatcher) on its channel, writing them into a `model.Batch`, until its maximum batch size is reached (`metrics.FlushSize`, currently 2000)
```
niksajakovljevic, line 114: It will even try sending a batch before reaching the max size if possible (if copier channel is not full), otherwise it will keep batching until max size and block until it sends the batch.

niksajakovljevic, line 114: Worth mentioning that Batcher is using additional buffer structure to gather/batch requests that also keeps a track about completed requests. The same buffer is being sent over to copier which updates task statuses so we know which requests have completed successfully.

pkg/pgmodel/Readme.md
```diff
@@ -36,46 +80,51 @@ type Sample {
...
+
+### Copier
+
+The final stage consists of a fixed number of goroutines, by default equal to the number of database CPU cores. The copiers are started by the dispatcher, and each hold a handle to the read-request channel, which the batcher writes requests into.
+
+The copier:
 
-`TODO` metric table creation
+1. Reads up to `metrics.MaxInsertStmtPerTxn` (currently 100) read requests from its handle to the read-request channel, timing out after 20 milliseconds.
+2. Reads all of the `ingestor.copyRequest` from all of the read requests' data-transfer channels into its own batch (note: each `copyRequest` must be for a different metric).
```
niksajakovljevic, line 125: Worth mentioning that each `copyRequest` is basically one buffer(one batch).

niksajakovljevic (changes requested):
👍  for improving documentation! Just a few small changes to batching explanation (note that batch can be sent out even before it's full)

pkg/pgmodel/Readme.md
JamesGuthrie: done

JamesGuthrie: done

pkg/pgmodel/Readme.md
JamesGuthrie: done

JamesGuthrie: I've kind of mentioned this now, but not in as much detail as you describe.

pkg/pgmodel/Readme.md
JamesGuthrie: I haven't mentioned this. I'm not sure how relevant it is at this level.

niksajakovljevic (approved):
Approving but I haven't seen changes for `Metric batcher` section  `3.1.` 
```
accept data for the metric (from the dispatcher) on its channel, writing them into a `model.Batch`. We are batching until copier is ready to receive data or maximum batch size is reached (`metrics.FlushSize`, currently 2000).
```

pkg/pgmodel/Readme.md
```diff
@@ -36,46 +80,54 @@ type Sample {
...
+### Metric batcher
+
+The second stage consists of the metric batcher goroutines, one for each metric name. These goroutines are created in `dispatcher.InsertTs`.
+
+On the first new insert request, the metric batcher creates a metric table in the database, if it doesn't exist yet.
+
+From then onwards, when the metric batcher receives a new insert request on its channel it: 
+1. Creates an unbuffered data-transfer channel for this batcher.
+2. Initiates a read request to the copier on the copier read-request channel, sending a handle to the batcher's data-transfer channel.
+3. Continues to (concurrently, in a loop over `select`):
```
cevian, line 113:
```suggestion
3. Tries to batch as much data as possible while waiting for a free copier. As soon as a copier frees up, it flushes the batch (even if not full). In particular the batcher will (concurrently, in a loop over `select`):
```

cevian (approved):
Great work. One (optional) suggestion

pkg/pgmodel/Readme.md
JamesGuthrie: done

PR #1599, opened 2022-08-30
Fix upgrade tests for 0.14.0

JamesGuthrie:
## Merge requirements


JamesGuthrie:
This PR is destined for failure. I managed to get a conflict in branch name for a branch from a month ago, so the docker image that this was tested against was a month old. Deleting the docker image from the registry didn't help because it was cached in the runner.

PR #1603, opened 2022-08-30
Fix broken cache eviction in clockcache

JamesGuthrie:
## Description

The clockcache is a simple approximation of an LRU cache. Instead of
maintaining an exact ordering of which elements were least recently
used, the cache simply keeps track of whether each element is "used".

When an item is inserted into the cache, it is counted as unused. When
an item is fetched from the cache, it is counted as used.

The eviction algorithm looks through the elements of the cache one by
one, trying to find one to evict. If the item is used, it marks it as
unused. If it is unused, it can evict the item. The next time an
eviction is required, it can continue with the next element.

Through a bug in index manipulation, if the next item to be evicted was
at index 1 of the storage array, the next eviction would not continue to
index 2. Instead, after successfully evicting an item from index 1, it
would attempt (on the next eviction) to evict the item at index 1
(again).

The above, combined with the fact that items just inserted are marked as
unused, meant that it was possible to end up in a situation where each
insert evicted the _previously inserted_ item.

This caused much woe, and thrashing.

## Merge requirements

### DIFF
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -174,7 +174,7 @@ func (self *Cache) evict() (insertPtr *element) {
 			}
 
 			if insertPtr != nil {
-				self.next = next + 1
+				self.next = (startLoc + next + 1) % self.Len()
 				return
 			}
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,12 @@ We use the following categories for changes:
 ## [Unreleased]
 
 ### Added
-
 - prom-migrator: Support for passing custom HTTP headers via command line arguments for both
   reader and writer [#1020].
 - Run timescaledb-tune with the promscale profile [#1615]
+
+### Fixed
+- Fix broken cache eviction in clockcache [#1603]
 
 ## [0.14.0] - 2022-08-30
 
--- a/pkg/clockcache/cache_test.go
+++ b/pkg/clockcache/cache_test.go
@@ -6,6 +6,7 @@ package clockcache
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"math/rand"
 	"reflect"
 	"sync"
@@ -319,5 +320,36 @@ func TestElementCacheAligned(t *testing.T) {
 	}
 	if elementSize != 64 {
 		t.Errorf("unexpected element size: %d", elementSize)
+	}
+}
+
+func TestCacheEvictionOnWraparound(t *testing.T) {
+	elems := 3
+	cache := WithMax(uint64(elems))
+	cachedItems := make([]int, elems)
+
+	// Prepopulate cache
+	for i := 0; i < 3; i++ {
+		cache.Insert(i, i, 1)
+		cachedItems[i%3] = i
+	}
+
+	assertCacheContains := func(cache *Cache, pseudoCache []int) {
+		cacheKeys := make([]int, elems)
+		i := 0
+		for k := range cache.elements {
+			cacheKeys[i] = k.(int)
+			i += 1
+		}
+		require.ElementsMatch(t, cacheKeys, pseudoCache)
+	}
+
+	// Let's cause a bunch of evictions
+	for i := 0; i < 100; i++ {
+		cache.Insert(i, i, 1)
+		// Track what we expect the cache to contain
+		cachedItems[i%3] = i
+		// Assert that our expectations are met
+		assertCacheContains(cache, cachedItems)
 	}
 }


pkg/clockcache/cache.go
```diff
@@ -174,7 +174,7 @@ func (self *Cache) evict() (insertPtr *element) {
 			}
 
 			if insertPtr != nil {
-				self.next = next + 1
+				self.next = (startLoc + next + 1) % self.Len()
```
cevian, line 177: are we sure we don't need `% self.Len()` on line 189 too? Even if startLoc is the last element?

JamesGuthrie: Assuming that `self.storage` has length 10, and `startLoc == 9` (i.e. it points to the last element).

When we slice `self.storage` into `postStart` and `preStart`, the contents of `postStart` will be one item, the last item, and the contents of `preStart` will be the first 9 items.

This is because go's slices are half-open (i.e. exclusive of the last element).

JamesGuthrie:
The commit I added to simplify the `evict` method has very different performance characteristic on AMD Ryzen processors (AWS cloud instance):

```
name            old time/op    new time/op    delta
Eviction/500       647ns ± 1%    1304ns ± 0%  +101.48%  (p=0.008 n=5+5)
Eviction/5000     7.58µs ±11%   11.56µs ± 2%   +52.53%  (p=0.008 n=5+5)
Eviction/50000    64.8µs ±10%   102.8µs ± 5%   +58.61%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
Eviction/500       0.00B          0.00B           ~     (all equal)
Eviction/5000      0.00B          0.00B           ~     (all equal)
Eviction/50000    0.40B ±150%     3.00B ± 0%  +650.00%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
Eviction/500        0.00           0.00           ~     (all equal)
Eviction/5000       0.00           0.00           ~     (all equal)
Eviction/50000      0.00           0.00           ~     (all equal)
```

cevian (approved):

niksajakovljevic (approved):

niksajakovljevic:
Just wondering if you managed to benchmark version of code that does not split array?

Harkishen-Singh:
Is https://github.com/timescale/promscale/pull/1603#issuecomment-1232919086 benchmark before the fix or after the fix. If its after the fix, then we are loosing perf in `Eviction`, right?

JamesGuthrie:
> Is [#1603 (comment)](https://github.com/timescale/promscale/pull/1603#issuecomment-1232919086) benchmark before the fix or after the fix. If its after the fix, then we are loosing perf in `Eviction`, right?

This benchmark is after the fix. I was curious why we're losing performance, so I took a deeper look at this to better understand it.

I _think_ that I understand it now. What follows is a bit of a brain dump into understanding what's going on. The bottom line is: buggy code can look fast, but if it's buggy, it doesn't matter how fast it is.

My first hypothesis was: maybe the code being emitted is different enough to be causing a problem (because we broke some "complier magic" like bounds check elimination), so I took to the disassembler (using `go tool objdump -s 'clockcache.*evict().*' -S dist/promscale` after running `make build`).

This was before the fix:

```
...
				self.next = next + 1
  0x1007ace64		910004e1		ADD $1, R7, R1		
  0x1007ace68		f9002c01		MOVD R1, 88(R0)		
...  
```

And after the fix:

```
...
				self.next = (startLoc + next + 1) % self.Len()
  0x1007ace74		940003ff		CALL github.com/timescale/promscale/pkg/clockcache.(*Cache).Len(SB)	
  0x1007ace78		f94017e1		MOVD 40(RSP), R1							
  0x1007ace7c		f94013e2		MOVD 32(RSP), R2							
  0x1007ace80		8b020021		ADD R2, R1, R1								
  0x1007ace84		b4000480		CBZ R0, 36(PC)								
  0x1007ace88		91000421		ADD $1, R1, R1								
  0x1007ace8c		9ac00c3b		SDIV R0, R1, R27							
  0x1007ace90		9b008761		MSUB R0, R1, R27, R1							
  0x1007ace94		f94027e2		MOVD 72(RSP), R2							
  0x1007ace98		f9002c41		MOVD R1, 88(R2)								
...
```

That function call is probably not helping, so I replaced `self.Len()` with `len(self.storage)`:
```
				self.next = (startLoc + next + 1) % len(self.storage)
  0x1007ace64		8b0200e1		ADD R2, R7, R1		
  0x1007ace68		f9401802		MOVD 48(R0), R2		
  0x1007ace6c		b4000462		CBZ R2, 35(PC)		
  0x1007ace70		91000421		ADD $1, R1, R1		
  0x1007ace74		9ac20c3b		SDIV R2, R1, R27	
  0x1007ace78		9b028761		MSUB R2, R1, R27, R1	
  0x1007ace7c		f9002c01		MOVD R1, 88(R0)		
```

There are more instructions (7 vs 2), but in a branch of code which is not hit very frequently, so I don't think that it could be exactly here that is causing the problem, but there's only one way to find out.

So I used `-cpuprofile` on `BenchmarkEviction` (`go test -bench='BenchmarkEviction' -run='^$' -cpu=1 -count=5 -cpuprofile master.out`.

Then, using pprof's `list` on the `evict` function, we can see where time is spent in the function:

Before fix:
```
(pprof) list evict
Total: 47.50s
ROUTINE ======================== github.com/timescale/promscale/pkg/clockcache.(*Cache).evict in /home/ubuntu/promscale/pkg/clockcache/cache.go
    22.28s     22.28s (flat, cum) 46.91% of Total
         .          .    161:	//     based off an induction variable e.g. `next := range slice`,
         .          .    162:	//     if the value is merely guarded by e.g. `if next >= len(slice) { next = 0 }`
         .          .    163:	//     the bounds check will not be elided. Doing the walk like this lowers
         .          .    164:	//     eviction time by about a third
         .          .    165:	startLoc := self.next
      20ms       20ms    166:	postStart := self.storage[startLoc:]
         .          .    167:	preStart := self.storage[:startLoc]
         .          .    168:	for i := 0; i < 2; i++ {
      30ms       30ms    169:		for next := range postStart {
    20.91s     20.91s    170:			elem := &postStart[next]
     620ms      620ms    171:			old := atomic.SwapUint32(&elem.used, 0)
     680ms      680ms    172:			if old == 0 {
         .          .    173:				insertPtr = elem
         .          .    174:			}
         .          .    175:
      20ms       20ms    176:			if insertPtr != nil {
         .          .    177:				self.next = next + 1
         .          .    178:				return
         .          .    179:			}
         .          .    180:		}
         .          .    181:		for next := range preStart {
```

After fix:
```
Total: 50.66s
ROUTINE ======================== github.com/timescale/promscale/pkg/clockcache.(*Cache).evict in /home/ubuntu/promscale/pkg/clockcache/cache.go
    26.45s     26.45s (flat, cum) 52.21% of Total
         .          .    161:	//     based off an induction variable e.g. `next := range slice`,
         .          .    162:	//     if the value is merely guarded by e.g. `if next >= len(slice) { next = 0 }`
         .          .    163:	//     the bounds check will not be elided. Doing the walk like this lowers
         .          .    164:	//     eviction time by about a third
         .          .    165:	startLoc := self.next
      40ms       40ms    166:	postStart := self.storage[startLoc:]
         .          .    167:	preStart := self.storage[:startLoc]
         .          .    168:	for i := 0; i < 2; i++ {
      10ms       10ms    169:		for next := range postStart {
    12.51s     12.51s    170:			elem := &postStart[next]
     370ms      370ms    171:			old := atomic.SwapUint32(&elem.used, 0)
     400ms      400ms    172:			if old == 0 {
         .          .    173:				insertPtr = elem
         .          .    174:			}
         .          .    175:
      50ms       50ms    176:			if insertPtr != nil {
         .          .    177:				self.next = (startLoc + next + 1) % self.Len()
         .          .    178:				return
         .          .    179:			}
         .          .    180:		}
      10ms       10ms    181:		for next := range preStart {
    12.09s     12.09s    182:			elem := &preStart[next]
     520ms      520ms    183:			old := atomic.SwapUint32(&elem.used, 0)
     430ms      430ms    184:			if old == 0 {
         .          .    185:				insertPtr = elem
         .          .    186:			}
         .          .    187:
      20ms       20ms    188:			if insertPtr != nil {
         .          .    189:				self.next = next + 1
         .          .    190:				return
         .          .    191:			}
         .          .    192:		}
         .          .    193:	}
```

The first thing to note is that practically no time is spent on the line of code that we were looking at in the disassembler. What is also interesting is that _where_ time is spent in the function is different. The explanation for this is relatively simple: the bug that we had would effectively ensure that `self.next` is 1 very often. If `self.next` is small, then we will predominantly be looking at items after the split point in `self.storage`, so we spend all of our time in `for next := range postStart`, and (practically) none in `for next := range preStart`.

With this insight, I had another theory: because the location we check for an entry to evict is different in the broken and fixed versions, maybe we spend more time looking for an item to evict in the fixed code. I instrumented the code to tell me how many items in `self.storage` we looked at during eviction. That turned out not to be it. (The explanation is simple. In the eviction test, whenever we insert an item, we mark all items as used, so we have to do a full pass through `self.storage` anyway).

The final attempt, which yielded fruit, was the realisation that the memory location that we're looking at is always the same in the broken version. In the broken code, most of the time `self.next` is 1. What I did next was to modify the fixed code to set `startLoc` to always be 1 (instead of `self.next`). Then I ran the `BenchmarkEviction` code, and saw that it is the same speed as the broken version.

The patch:
```diff
diff --git a/pkg/clockcache/cache.go b/pkg/clockcache/cache.go
index 43448ca7..1a36818f 100644
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -162,7 +162,7 @@ func (self *Cache) evict() (insertPtr *element) {
 	//     if the value is merely guarded by e.g. `if next >= len(slice) { next = 0 }`
 	//     the bounds check will not be elided. Doing the walk like this lowers
 	//     eviction time by about a third
-	startLoc := self.next
+	startLoc := 1
 	postStart := self.storage[startLoc:]
 	preStart := self.storage[:startLoc]
 	for i := 0; i < 2; i++ {
```

The result when comparing between current master and this PR with the above patch:
```
name          old time/op    new time/op    delta
Eviction/500     336ns ± 1%     338ns ± 1%  +0.64%  (p=0.043 n=10+10)

name          old alloc/op   new alloc/op   delta
Eviction/500     0.00B          0.00B         ~     (all equal)

name          old allocs/op  new allocs/op  delta
Eviction/500      0.00           0.00         ~     (all equal)
```

(If you're curious, the `+0.64%` is because we're using `self.Len()` instead of `len(self.storage)`).

I believe that this explains what changed in our code to cause this change in performance. What it doesn't explain is _why_ this change causes a ~20% change in eviction performance. I suspect that the answer has to do with "magical CPU stuff", like branch prediction, or cache access patterns. I would like to be able to answer this, but I cannot (yet).

JamesGuthrie:
> Just wondering if you managed to benchmark version of code that does not split array?

This was the benchmark result: https://github.com/timescale/promscale/pull/1603#issuecomment-1232979484

PR #1611, opened 2022-09-01
Fix table indentation in configuration.md

VineethReddy02:
## Description

Fix table indentation. The table indentation is broken when viewed as GH file.


alejandrodnm:
The CI for the docs is failing and the entry for `tracing.batch-workers` is not being rendered as part of the table.

PR #1614, opened 2022-09-02
Fix possible goroutine leak

### LINKED ISSUE #1604: Close method in Proxy may have a hanging go routine bug
[pkg/jaeger/proxy/proxy.go:81](https://github.com/timescale/promscale/blob/master/pkg/jaeger/proxy/proxy.go)

Receiving unbuffered `errChan` in a select block. That may cause a [hanging go routine](https://github.com/trailofbits/semgrep-rules/blob/main/go/hanging-goroutine.yml). Maybe should change `errChan` to a buffered one. 
```
func (p *Proxy) Close(ctx context.Context, _ *storage_v1.CloseWriterRequest) (*storage_v1.CloseWriterResponse, error) {
	errChan := make(chan error)
	go func() {
		errChan <- p.conn.Close()
		close(errChan)
	}()
	var err error
	select {
	case err = <-errChan:
	case <-ctx.Done():
		err = ctx.Err()
	}
	if err != nil {
		err = fmt.Errorf("error closing connection to Promscale GRPC server: %w", err)
	}
	return nil, err
}
```
If no influence, please ignore this.

alejandrodnm:
## Description

Using unbuffered channels to receive in a select block might lead to a
goroutine leak if another branch of the select is selected instead.

By using a buffered channel the writer doesn't have to wait for another
goroutine to unblock it with a read. Instead it just uses the buffered
space to send the message and continue its execution.

fixes #1604 

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,6 @@ We use the following categories for changes:
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [Unreleased]
-
 ### Added
 - prom-migrator: Support for passing custom HTTP headers via command line arguments for both
   reader and writer [#1020].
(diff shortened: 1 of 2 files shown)


arajkumar (approved):
lgtm

PR #1620, opened 2022-09-05
Compare for context.Canceled error with `Is`

alejandrodnm:
Since `context.Caceled` is declared as:

```
var Canceled = errors.New("context canceled")
```

We should use `errors.Is` instead of `errors.As` because we are
comparing against and error value instead of the type of the error.

golangci-lint reports the current usage of `errors.As` with the
following error:

```
golangci-lint run
pkg/runner/runner.go:61:20: errorsas: second argument to errors.As should not be *error (govet)
        if err != nil && !errors.As(err, &context.Canceled) {
                          ^
pkg/runner/runner.go:69:20: errorsas: second argument to errors.As should not be *error (govet)
        if err != nil && !errors.As(err, &context.Canceled) {
```

## Merge requirements

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -58,7 +58,7 @@ func init() {
 
 func loggingUnaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	m, err := handler(ctx, req)
-	if err != nil && !errors.As(err, &context.Canceled) {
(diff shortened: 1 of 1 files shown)


arajkumar (approved):
I'm wondering why this is not caught in the CI, probably we are not running the go linter?

alejandrodnm:
> I'm wondering why this is not caught in the CI, probably we are not running the go linter?

I got it when I updated to the latest version of golangci to fix some issues in the PR that updates the version of the tool

https://github.com/timescale/promscale/pull/1502/files

PR #1623, opened 2022-09-07
statement timeout

alejandrodnm:
## Merge requirements

### DIFF
--- a/pkg/pgxconn/helpers.go
+++ b/pkg/pgxconn/helpers.go
@@ -1,0 +1,178 @@
+// This file and its contents are licensed under the Apache License 2.0.
+// Please see the included NOTICE for copyright information and
+// LICENSE for a copy of the license.
+
+package pgxconn
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v4"
+)
+
+const (
+	timeoutDisabled = 0
+	minTimeoutMs    = 1
+)
+
+type rowsFromSingleQueryBatch struct {
+	PgxRows
+	batchResults pgx.BatchResults
+}
+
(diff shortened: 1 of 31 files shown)


pkg/pgxconn/helpers.go
```diff
@@ -0,0 +1,178 @@
...
+) (rows PgxRows, commit func() error, err error) {
+	noopCommit := func() error { return nil }
+	if timeout < timeoutDisabled {
+		timeout = timeoutDisabled
+	}
+	b := conn.NewBatch()
+	b.Queue("BEGIN")
+	b.Queue(fmt.Sprintf("set local statement_timeout = %d", timeout))
+	b.Queue(sql, args...)
+	b.Queue("COMMIT")
```
sumerman, line 94: I don't think COMMIT is required here. Am I missing something?

pkg/pgxconn/helpers.go
```diff
@@ -0,0 +1,178 @@
...
+	ctx context.Context,
+	conn PgxConn,
+	timeout int64,
+	sql string,
+	args ...interface{},
+) (rows PgxRows, commit func() error, err error) {
+	noopCommit := func() error { return nil }
+	if timeout < timeoutDisabled {
+		timeout = timeoutDisabled
+	}
```
jgpruitt, line 89: In postgres, a statement_timeout = 0 disables the timeout, but here we use < 0. I would prefer that we match postgres so that it's less confusing to people familiar with how statement_timeout works. Is there a reason we used < 0 here instead of = 0?

jgpruitt (changes requested):

alejandrodnm:
@jgpruitt @sumerman I'm trying some things out still. This is actually based from a non-master branch so it has a lot of changes from another PR.

PR #1624, opened 2022-09-07
Allow configuring HA replica and cluster label names

### LINKED ISSUE #1578: Allow configuring replica and cluster label required in HA mode
Currently when running in high availability mode, Promscale requires two external labels, `cluster` to identify a Prometheus cluster, and `__replica__` to identify a Prometheus replica within that cluster.

There is one problem with the current replica label. Since, any label starting with `__` is reserved for internal use in Prometheus, anything that uses prometheus/client_golang (or similar) cannot add the `__replica__` label (such as [avalanche](https://githib.com/prometheus-community/avalanche)).

The other argument is that Prometheus Operator adds `prometheus_replica` as the replica label by default and it'd be nice to have that as our default as well.

### What I'd like to see

We have a flag/config option to configure the required replica/cluster label names. And optionally we switch to `prometheus_replica` as replica label by default.

onprem:
## Description
Fixes #1578 .

This PR adds two new flags to allow configuring the replica and cluster label names. Currently both of the label names are not configurable but are required for HA to work.

## Merge requirements

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -20,6 +20,7 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/util/httputil"
+	"github.com/timescale/promscale/pkg/ha"
 	"github.com/timescale/promscale/pkg/log"
 	pgmodel "github.com/timescale/promscale/pkg/pgmodel/model"
 	"github.com/timescale/promscale/pkg/promql"
@@ -35,7 +36,7 @@ var (
 type Config struct {
 	AllowedOrigin    *regexp.Regexp
 	ReadOnly         bool
-	HighAvailability bool
+	HighAvailability *ha.Config
 	AdminAPIEnabled  bool
 	TelemetryPath    string
 
@@ -44,8 +45,9 @@ type Config struct {
 }
 
 func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
+	cfg.HighAvailability = ha.ParseFlags(fs)
+
 	fs.BoolVar(&cfg.ReadOnly, "db.read-only", false, "Read-only mode for the connector. Operations related to writing or updating the database are disallowed. It is used when pointing the connector to a TimescaleDB read replica.")
-	fs.BoolVar(&cfg.HighAvailability, "metrics.high-availability", false, "Enable external_labels based HA.")
 	fs.BoolVar(&cfg.AdminAPIEnabled, "web.enable-admin-api", false, "Allow operations via API that are for advanced users. Currently, these operations are limited to deletion of series.")
 	fs.StringVar(&cfg.TelemetryPath, "web.telemetry-path", "/metrics", "Web endpoint for exposing Promscale's Prometheus metrics.")
 
@@ -53,7 +55,7 @@ func ParseFlags(fs *flag.FlagSet, cfg *Config) *Config {
 }
 
 func Validate(cfg *Config) error {
-	return nil
+	return cfg.HighAvailability.Validate()
 }
 
 func corsWrapper(conf *Config, f http.HandlerFunc) http.HandlerFunc {
--- a/pkg/ha/filter.go
+++ b/pkg/ha/filter.go
@@ -13,19 +13,20 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
-const ReplicaNameLabel = "__replica__"
-const ClusterNameLabel = "cluster"
-
 // Filter is a HA filter which filters data based on lease information it
 // gets from the lease service.
 type Filter struct {
-	service *Service
+	service          *Service
+	replicaLabelName string
+	clusterLabelName string
 }
 
 // NewFilter creates a new Filter based on the provided Service.
-func NewFilter(service *Service) *Filter {
+func NewFilter(service *Service, replicaLabelName, clusterLabelName string) *Filter {
 	return &Filter{
-		service: service,
+		service:          service,
+		replicaLabelName: replicaLabelName,
+		clusterLabelName: clusterLabelName,
 	}
 }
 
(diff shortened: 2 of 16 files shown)


arajkumar (commented):
Actually we assume `cluster` label for telemetry, I guess we need to propagate the user provided value.

https://github.com/timescale/promscale_extension/blob/master/migration/idempotent/010-telemetry.sql#L85

@Harkishen-Singh

Harkishen-Singh:
@arajkumar thanks for catching that. Yes, this will trouble the telemetry.

> I guess we need to propagate the user provided value.

We should not do that. This will be breaking, if multiple Promscale are running. Also, we should remember that the reason for this change is limited to testing/benchmarking purpose.

I think we need to discuss this on joint standup. Maybe instead of `__`, we just use `_`, like `_replica_`, which should be allowed. Please correct me @onprem if I am wrong.

arajkumar:
How about if we translate any configured value into __cluster__ and translate it back to configured value when queried?  Just shooting in the dark :)

VineethReddy02:
@onprem The `0.15.0` release will happen this week. If you wish to include this PR in the release. Can you follow up with the reviewers accordingly?

Harkishen-Singh:
@VineethReddy02 we need to make changes to the telemetry engine to propagate the telemetry information.

@arajkumar how about if we do not touch the telemetry implementation. In that way, `metrics_ha_cluster_count` works only when someone is using `__cluster__` label. Otherwise, we do not get the `"count"` thingy.

For knowing if `ha` feature is used or not, we add another telemetry `metrics_ha_enabled` which is just a boolean. It is true if `-metrics.high-availability` flag is applied, otherwise false. This information will be filled as we start promscale, in our very first write of telemetry. WDYT?

Note: This means once ha is started, it will always remain true in telemetry. But that said, I don't think users will stop using ha if they started to use it.

alejandrodnm:
@Harkishen-Singh in terms of effort what would you say to make this configurable? 

Is it something like the following?

- Add the entry in the configuration for the label 
- Pass the configurable label to the telemetry engine
- Pass down the label to the extension. Maybe for this part we could create a new version of the functions in the extention that accepts the label, that way we could make it backwards compatible. 

Maybe something like:

_ps_catalog.promscale_sql_telemetry_with_cluster_label(label TEXT)

_ps_catalog.promscale_sql_telemetry() would just return promscale_sql_telemetry_with_custom_cluster_label("cluster")

BTW I haven't looked into the telemetry engine and don't know the implications of a change like this.

Harkishen-Singh:
>_ps_catalog.promscale_sql_telemetry() would just return promscale_sql_telemetry_with_custom_cluster_label("cluster")

Do you mean `_ps_catalog.promscale_sql_telemetry()` will call the `.._with_custom_cluster`?

In anyways, we will need a change in extension for this.

On the telemetry engine side, we need to pass down the cluster label, and the engine will specifically call another func (we will need to create this), that will look for the new cluster label name and update in the metadata table. This will be 30-40 lines of code at max.

**What we should do right now?**
Since telemetry is not a priority at the moment and most users will use the default values in the flags, I suggest we leave the telemetry specific to `__cluster__` only and make an issue for this. We will fix this issue in the next iteration of telemetry improvements, which will happen with the doc that Vineeth created for new telemetry requirements.
Thought? @arajkumar @alejandrodnm

pkg/api/common.go
```diff
@@ -84,10 +85,16 @@ func (a *Auth) Validate() error {
 	return nil
 }
 
+type HighAvailability struct {
```
Harkishen-Singh, line 88: This seems a configuration for `ha` package. Based on what we do for other packages, we have a `config.go` for all modules and we should do the same here. Essentially, All changes here goes in `ha/config.go`. Earlier, it was simply a bool so we didn't have in the `ha` package.

You can take `rules/config.go` for reference.

harkishen (changes requested):
The code looks good, except for one refactoring. Then we can merge.

harkishen (commented):
You will also need to change `defaultAPIConfig()` in `promql_endpoint_integration_tests.go` to

```golang
func defaultAPIConfig() *api.Config {
	return &api.Config{
		AllowedOrigin:    regexp.MustCompile(".*"),
		TelemetryPath:    "/metrics",
		HighAvailability: &api.HighAvailability{}, // or &ha.HighAvailability{} after refactoring.
	}
}
```

And then the tests should pass.

Harkishen-Singh:
ping @onprem

pkg/api/common.go
onprem: Done!

onprem:
Heya @Harkishen-Singh, sorry for the delay on this, finally found some time to work on this. I have moved the configuration into `ha` package as you suggested, as well as rebased the PR to resolve some merge conflicts. This PR should be ready for another round of reviews.

pkg/ha/filter.go
```diff
@@ -13,35 +13,39 @@ import (
 	"github.com/timescale/promscale/pkg/prompb"
 )
 
-const ReplicaNameLabel = "__replica__"
-const ClusterNameLabel = "cluster"
+const DefaultReplicaLabelName = "__replica__"
+const DefaultClusterLabelName = "cluster"
```
arajkumar, line 17: I think the better place for these constants is pkg/ha/config.go.

pkg/ha/filter.go
```diff
@@ -208,33 +212,33 @@ func findDataTimeRange(tts []prompb.TimeSeries) (minTUnix int64, maxTUnix int64)
...
 		)
 	} else if cluster == "" {
 		return fmt.Errorf("HA enabled, but %s label is empty; %s set to: %s",
-			ClusterNameLabel,
-			ReplicaNameLabel,
+			h.clusterLabelName,
+			h.replicaLabelName,
 			replica,
 		)
 	} else if replica == "" {
```
arajkumar, line 238: Should we also check `cluster == replica`?

harkishen (approved):

pkg/ha/filter.go
Harkishen-Singh: Agreed.

pkg/ha/filter.go
Harkishen-Singh: We can add, which also makes me think that we should convert this block of if-else-if to switch. It will be easier to read.

onprem: I don't see a problem in `cluster` and `replica` label values being same. Someone can have external labels like 

```yaml
cluster: "one"
replica: "one"
---
cluster: "one"
replica: "two"
```

which should work perfectly fine.

arajkumar (approved):

harkishen (approved):

Harkishen-Singh:
@onprem we should also add an entry to the changelog before merging this PR.

PR #1625, opened 2022-09-07
Manually Vacuum Freeze Compressed Chunks

jgpruitt:
## Description

Adding an "engine" to the connector that will periodically look for compressed chunks that have not been vacuumed and frozen. The engine will grab a batch of them and manually `vacuum (freeze, analyze)` them. A configurable amount of parallelism will be employed in working through the batch.

The goal of this engine is to help prevent txid wraparound.

## Merge requirements

### DIFF
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	sqlStart            = "SELECT pg_try_advisory_lock(1982010619820106)"
-	sqlListChunks       = "SELECT format('%I.%I', schema_name, table_name) from _ps_catalog.chunks_to_vacuum_freeze LIMIT 1000"
+	sqlListChunks       = "SELECT format('%I.%I', schema_name, table_name) from _ps_catalog.chunks_to_freeze LIMIT 1000"
 	sqlVacuumFmt        = "VACUUM (FREEZE, ANALYZE) %s"
 	sqlStop             = "SELECT pg_advisory_unlock(1982010619820106)"
 	delay               = 10 * time.Second
@@ -101,7 +101,6 @@ func every(every time.Duration, task func()) (execute, kill func()) {
 	kill = func() {
 		once.Do(func() {
 			ticker.Stop()
-			die <- struct{}{}
 			close(die)
 		})
 	}
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -1,0 +1,223 @@
+package vacuum
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const (
+	sqlStart            = "SELECT pg_try_advisory_lock(1982010619820106)"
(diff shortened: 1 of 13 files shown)


pkg/vacuum/vacuum.go
```diff
@@ -0,0 +1,234 @@
...
+	"time"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const (
+	sqlStart            = "SELECT pg_try_advisory_lock(1982010619820106)"
+	sqlListChunks       = "SELECT format('%I.%I', schema_name, table_name) from _ps_catalog.chunks_to_vacuum_freeze"
```
sumerman, line 17: I would add `LIMIT 1000` here and, perhaps, a random `ORDER BY`, just in case.

jgpruitt: the view itself has an order by, but yeah the limit makes a lot of sense

jgpruitt: 2dda9cbd732cb5c13e78e0405618fa4e9fe21063
`LIMIT 1000` and continue to process batches while we get results

pkg/vacuum/vacuum.go
```diff
@@ -0,0 +1,234 @@
...
+	if e.runFreq <= 0 {
+		return fmt.Errorf("run frequency must be positive: %d", e.runFreq)
+	}
+	return nil
+}
+
+// Start starts the Engine
+// Blocks forever unless Stop is called
+func (e *Engine) Start() error {
+	if err := e.validate(); err != nil {
```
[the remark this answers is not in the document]
jgpruitt: I think I'll just remove it from `Start`. We already validate the config elsewhere

pkg/vacuum/vacuum.go
```diff
@@ -0,0 +1,223 @@
...
+	"sync"
+	"time"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const (
+	sqlStart            = "SELECT pg_try_advisory_lock(1982010619820106)"
```
cevian, line 16: I wonder if it's worth defining the id in the extension code like `_prom_catalog.get_advisory_lock_prefix_maintenance` or at least using a similar prefix (12379?)

pkg/vacuum/vacuum.go
```diff
@@ -0,0 +1,223 @@
...
+// terminate the execution function if called
+func every(every time.Duration, task func()) (execute, kill func()) {
+	var ticker = time.NewTicker(every)
+	var once sync.Once
+	var die = make(chan struct{}, 1)
+
+	kill = func() {
+		once.Do(func() {
+			ticker.Stop()
+			die <- struct{}{}
```
cevian, line 104: you shouldn't need to send anything, just close. `After calling close, and after any previously sent values have been received, receive operations will return the zero value for the channel's type without blocking.`

cevian (commented):
Looks really good. two small comments

pkg/vacuum/vacuum.go
jgpruitt: https://github.com/timescale/promscale/pull/1625/commits/ec4cd25a431cba7e06bd082650d5b43738aa3c20

jgpruitt:
have to switch to a new branch because of CI issues.
see https://github.com/timescale/promscale/pull/1648

PR #1627, opened 2022-09-08
Import jaeger storage integration tests

arajkumar:
## Description

This PR imports jaeger storage integration tests from https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/integration and adopts those for promscale environment.

May be we can remove the imported tests once we enable promscale integration tests in upstream jaeger.

This PR also fixes `TestJaegerStorageIntegration/*/GetOperations` test which was failing due to a typo.

## Merge requirements

- [X] CHANGELOG entry for user-facing changes

### DIFF
--- a/pkg/jaeger/store/translation.go
+++ b/pkg/jaeger/store/translation.go
@@ -6,6 +6,7 @@ package store
 import (
 	"fmt"
 
+	"github.com/jaegertracing/jaeger/storage/spanstore"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
@@ -34,6 +35,15 @@ var (
 		"consumer": "consumer",
 		"producer": "producer",
 	}
+	// Map of internal span kind to Jaeger.
+	internalToJaegerSpanKind = map[string]string{
+		"client":      "client",
+		"server":      "server",
+		"internal":    "internal",
+		"consumer":    "consumer",
+		"producer":    "producer",
+		"unspecified": "",
+	}
 )
 
 func internalToStatusCode(s string) ptrace.StatusCode {
@@ -51,6 +61,17 @@ func statusCodeToInternal(sc ptrace.StatusCode) string {
 		}
 	}
 	panic(fmt.Sprintf("status code %s does not have internal representation", sc.String()))
+}
+
+func internalToJaegerOperation(name, kind string) spanstore.Operation {
+	spanKind, ok := internalToJaegerSpanKind[kind]
+	if !ok {
+		return spanstore.Operation{}
+	}
+	return spanstore.Operation{
+		Name:     name,
+		SpanKind: spanKind,
+	}
 }
 
 func internalToSpanKind(s string) ptrace.SpanKind {
--- a/pkg/tests/end_to_end_tests/jaeger_store_integration_tests/trace_compare.go
+++ b/pkg/tests/end_to_end_tests/jaeger_store_integration_tests/trace_compare.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2019 The Jaeger Authors.
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file is copied from upstream jaeger[1] to run storage integration
+// tests as part of promscale e2e.
+// [1] https://github.com/jaegertracing/jaeger/blob/4fc291568d8ac59a1c67cc47ee1d91ab20dd06c4/plugin/storage/integration/trace_compare_test.go
+
+// TODO: Remove this file after adding promscale integration storage changes
+// into upstream jaeger.
+
+package jaeger_integration_tests
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/kr/pretty"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+// CompareSliceOfTraces compares two trace slices
+func CompareSliceOfTraces(t *testing.T, expected []*model.Trace, actual []*model.Trace) {
+	require.Equal(t, len(expected), len(actual), "Unequal number of expected vs. actual traces")
+	model.SortTraces(expected)
+	model.SortTraces(actual)
+	for i := range expected {
+		checkSize(t, expected[i], actual[i])
+	}
+	if diff := pretty.Diff(expected, actual); len(diff) > 0 {
+		for _, d := range diff {
+			t.Logf("Expected and actual differ: %s\n", d)
+		}
+		out, err := json.Marshal(actual)
+		out2, err2 := json.Marshal(expected)
+		assert.NoError(t, err)
+		assert.NoError(t, err2)
+		t.Logf("Actual traces: %s", string(out))
+		t.Logf("Expected traces: %s", string(out2))
+		t.Fail()
+	}
+}
+
+// CompareTraces compares two traces
+func CompareTraces(t *testing.T, expected *model.Trace, actual *model.Trace) {
+	if expected.Spans == nil {
+		require.Nil(t, actual.Spans)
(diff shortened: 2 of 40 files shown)


pkg/tests/end_to_end_tests/integration_test.go
```diff
@@ -0,0 +1,390 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
alejandrodnm, line 1: Should we add a prefix to the file name to specify that's a jaeger integration test?

arajkumar: Done.

alejandrodnm (approved):

alejandrodnm (approved):

pkg/jaeger/store/translation.go
```diff
@@ -44,8 +45,30 @@ var (
...
+	// integration tests validates the same[2]. Hence we replace empty span name
+	// with unspecifiedOperationName to make database happy.
+	//
+	// [1] https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L110-L121
+	// [2] https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/integration/fixtures/traces/default.json#L6
+	unspecifiedOperationName = "unspecified"
 )
 
+func fillDefaultsForJaeger(span *model.Span) {
+	if span.OperationName == "" {
```
arajkumar, line 61: [Jaeger author thinks](https://github.com/jaegertracing/jaeger/pull/3922) that having an empty span name is a valid edge cause and it shouldn't really trouble the span ingestion.

May be we should think of removing the constraint [enforced in the db](https://github.com/timescale/promscale_extension/blob/e6aaa0f3292f48caad326ff0de2ed1467f17d2d7/migration/incremental/012-tracing.sql#L99)?

pkg/jaeger/store/trace_query.go
```diff
@@ -130,12 +130,10 @@ func (b *Builder) buildCompleteTraceQuery(traceIDClause string, timeLowRef strin
 	linkTimeClause := ""
 	if timeLowRef != "" {
 		spanTimeClause += " AND s.start_time > " + timeLowRef
-		eventTimeClause += " AND e.time > " + timeLowRef
```
cevian, line 133: why do these need to be removed. This will really kill some optimizations

pkg/jaeger/store/translation.go
cevian: I am ok removing the constraint. You want to open a PR against the extension?

cevian (changes requested):
Some questions. Also can we avoid importing the jaeger integration tests like this? Instead can't we (in ci) download the jaeger repo and run the tests directly from there? This would make it easier to keep the tests in-sync for when jaeger changes them. Otherwise, we have to run sync them up every so often. If we do have to copy code like this, I think it should be into a separate subdirectory with a script to sync the code.

arajkumar:
> Some questions. Also can we avoid importing the jaeger integration tests like this? Instead can't we (in ci) download the jaeger repo and run the tests directly from there? This would make it easier to keep the tests in-sync for when jaeger changes them. Otherwise, we have to run sync them up every so often. If we do have to copy code like this, I think it should be into a separate subdirectory with a script to sync the code.

@cevian thats the end goal, We can do it once our promscale integration enablement changes lands in jaeger upstream.

pkg/jaeger/store/trace_query.go
arajkumar: `TestExternalGRPCStorage/FindTraces/Trace_spans_over_multiple_indices` would fail without this fix.

Here is a trace fixture https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/integration/fixtures/traces/multi_index_trace.json

The reason is that event stamp is not with in span time boundary(including upper & lower).

```sql
$ select * from _ps_trace.event;
time                         |trace_id                            |span_id|event_nbr|dropped_tags_count|name|tags|
-----------------------------+------------------------------------+-------+---------+------------------+----+----+
2022-09-26 22:16:31.639 +0530|00000000-0000-0000-0000-000000000005|      1|        0|                 0|    |{}  |
2022-09-26 22:16:31.639 +0530|00000000-0000-0000-0000-000000000005|      1|        1|                 0|    |{}  |
2022-09-26 22:16:31.639 +0530|00000000-0000-0000-0000-000000000005|      2|        0|                 0|    |{}  |
2022-09-26 22:16:31.639 +0530|00000000-0000-0000-0000-000000000005|      2|        1|                 0|    |{}  |


$ select * from _ps_trace.span;
trace_id                            |span_id|parent_span_id|operation_id|start_time                   |end_time                     |duration_ms|instrumentation_lib_id|resource_schema_url_id|event_time|dropped_tags_count|dropped_events_count|dropped_link_count|resource_dropped_tags_count|status_code|trace_state|span_tags|status_message|resource_tags|
------------------------------------+-------+--------------+------------+-----------------------------+-----------------------------+-----------+----------------------+----------------------+----------+------------------+--------------------+------------------+---------------------------+-----------+-----------+---------+--------------+-------------+
00000000-0000-0000-0000-000000000005|      1|              |           1|2022-09-26 05:33:31.639 +0530|2022-09-26 05:33:31.639 +0530|      0.005|                      |                      |empty     |                 0|                   0|                 0|                          0|unset      |           |{}       |              |{"1": 1}     |
00000000-0000-0000-0000-000000000005|      2|              |           2|2022-09-26 05:26:31.639 +0530|2022-09-26 05:26:31.639 +0530|      0.005|                      |                      |empty     |                 0|                   0|                 0|                          0|unset      |           |{}       |              |{"1": 1}     |
```

alejandrodnm:
@cevian I agree with @arajkumar on the approach he wants to take. 

I don’t think they’ll make changes soon to those tests. I’d rather have them now and mess with the CI later. That way we can be sure that we are passing everything reliably before we go the route of automating running the tests by downloading the Jaeger repo and setting up everything.

So the tests here will be only temporary ( We promise :D ).

pkg/jaeger/store/translation.go
arajkumar: @cevian FYI https://github.com/timescale/promscale_extension/pull/520

pkg/jaeger/store/trace_query.go
cevian: Can't we adjust this with the tracing.max-trace-duration CLI parameter instead?

arajkumar: Thanks @cevian , Increasing `tracing.max-trace-duration` seem to work. I will remove this change.

pkg/tests/end_to_end_tests/jaeger_store_integration_tests/trace_compare.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
alejandrodnm, line 1: I also copied this file in my latest PR, I think it's worth keeping https://github.com/timescale/promscale/pull/1678/files#diff-cdda0957b5a9bdfc7d49e834ef343f4850c254f1dfcffc78b2575fe09a7feb3bR1

alejandrodnm (approved):

pkg/tests/end_to_end_tests/jaeger_store_integration_tests/trace_compare.go
```diff
@@ -0,0 +1,85 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/kr/pretty"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
```
niksajakovljevic, line 28: Can you please add a comment why we needed to copy/modify this?

niksajakovljevic (approved):
Approving but unclear what code has been changed. Maybe we can be more expressive about that? 
It would be nice if we could add README into copied folder explaining why are we doing that

pkg/tests/end_to_end_tests/jaeger_store_integration_tests/trace_compare.go
arajkumar: @niksajakovljevic It is just a temporary solution and in the long run we can either import these as packages (after removing `_test` suffix) or download from jaeger repo.

arajkumar:
> Approving but unclear what code has been changed. Maybe we can be more expressive about that? It would be nice if we could add README into copied folder explaining why are we doing that. I also have concerns on how to keep code in sync with Jaeger so maybe we can comment on that as well.

My PR to expose test helpers as go pkg has been merged into upstream jaeger.

https://github.com/jaegertracing/jaeger/pull/3944

PR #1631, opened 2022-09-09
Add statement_timeout helpers

alejandrodnm:
### Add statement_timeout helpers 
 
The helper functions wrap queries in a implicit transaction that
executes a `set local statement_timeout` as its first item. The `set
local` directive scopes the statement to the transaction and not to the
session/connection.

Additional helpers that set the timeout to the remaining time before a
context is deadline is reached are provided. These are suffixed with
`WithTimeoutFromCtx`, for example `QueryWithTimeoutFromCtx` and
`QueryRowWithTimeoutFromCtx` are meant to replace the usage of `Query`
and `QueryRow` respectively.

When working directly with batches `NewBatchWithTimeoutFromCtx` can be
used to handle the process of:

- Creating a new batch.
- Getting the timeout from the context.
- Queue the statement_timeout query.
- Sending the batch and if the batch has a statement_timeout query then
  reading its result.

There's a bug with the use of `set local` and implicit transactions.
Even though the statements are executed correctly and apply the desire
effects the following warning message is logged:

`SET LOCAL can only be used in transaction blocks`

https://www.postgresql.org/message-id/flat/16988-58edba102adb5128@postgresql.org
https://github.com/npgsql/npgsql/issues/3688

For more contect on statement_timeout:
https://www.postgresql.org/docs/current/runtime-config-client.html

## Merge requirements

### DIFF
--- a/pkg/pgmodel/querier/query_sample.go
+++ b/pkg/pgmodel/querier/query_sample.go
@@ -11,6 +11,7 @@ import (
 	"github.com/prometheus/prometheus/storage"
 	"github.com/timescale/promscale/pkg/pgmodel/common/errors"
 	"github.com/timescale/promscale/pkg/pgmodel/common/schema"
+	"github.com/timescale/promscale/pkg/pgxconn"
 )
 
 type querySamples struct {
@@ -74,12 +75,12 @@ func (q *querySamples) fetchSamplesRows(mint, maxt int64, hints *storage.SelectH
 // successfully applied, the new top node is returned together with the metric
 // rows. For more information about top nodes, see `engine.populateSeries`.
 func fetchSingleMetricSamples(ctx context.Context, tools *queryTools, metadata *evalMetadata) ([]sampleRow, parser.Node, error) {
-	sqlQuery, values, topNode, tsSeries, err := buildSingleMetricSamplesQuery(metadata)
+	query, err := buildSingleMetricSamplesQuery(metadata)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	rows, err := tools.conn.Query(ctx, sqlQuery, values...)
+	rows, closeFn, err := pgxconn.QueryWithTimeoutFromCtx(ctx, tools.conn, query.sql, query.values...)
 	if err != nil {
 		if e, ok := err.(*pgconn.PgError); ok {
 			switch e.Code {
@@ -96,21 +97,34 @@ func fetchSingleMetricSamples(ctx context.Context, tools *queryTools, metadata *
 		}
 		return nil, nil, err
 	}
-	defer rows.Close()
+	defer func() {
+		_ = closeFn()
+	}()
 
-	updatedMetricName := ""
-	// If the table name and series table name don't match, this is a custom metric view which
-	// shares the series table with the raw metric, hence we have to update the metric name label.
-	if metadata.timeFilter.metric != metadata.timeFilter.seriesTable {
-		updatedMetricName = metadata.timeFilter.metric
-	}
+	samplesRows, err := appendSampleRows(
+		make([]sampleRow, 0, 1),
+		rows,
+		query.tsSeries,
+		metricNameOverride(metadata),
+		metadata.timeFilter.schema,
+		metadata.timeFilter.column,
+	)
 
-	filter := metadata.timeFilter
-	samplesRows, err := appendSampleRows(make([]sampleRow, 0, 1), rows, tsSeries, updatedMetricName, filter.schema, filter.column)
 	if err != nil {
-		return nil, topNode, fmt.Errorf("appending sample rows: %w", err)
+		return nil, query.topNode, fmt.Errorf("appending sample rows: %w", err)
 	}
-	return samplesRows, topNode, nil
+
+	return samplesRows, query.topNode, nil
+}
+
+// metricNameOverride returns the metric name to use if the table name and
+// series table name don't match. This happens on custom metric views that
+// share the series table with the raw metric.
+func metricNameOverride(metadata *evalMetadata) string {
+	if metadata.timeFilter.metric != metadata.timeFilter.seriesTable {
+		return metadata.timeFilter.metric
+	}
+	return ""
 }
 
 // fetchMultipleMetricsSamples returns all the result rows for across multiple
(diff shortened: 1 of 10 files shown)


pkg/pgxconn/helpers.go
```diff
@@ -0,0 +1,248 @@
...
+//
+// All the queries in a batch are executed in an implicit transaction by using
+// `set local statement_timeout` the timeout it's only applied to the queries
+// belonging to the transaction and not to the connection session.
+//
+// There's a bug with the use of `set local` and implicit transactions.
+// Even though the statements are executed correctly and apply the desire
+// effects the following warning message is logged:
+//
+// `SET LOCAL can only be used in transaction blocks`
```
alejandrodnm, line 47: This will make the DB output a lot of WARNING messages with that message. If that's a problem I can change the implementation to set an explicit transaction per batch.

It would be queueing BEGIN and COMMIT to the batch, and consuming the responses for both statements. Code-wise the downside is that using `NewBatchWithTimeoutFromCtx` and `SendBatch` won't be as pretty as we have to remember to put COMMIT at the end and to consume it.

pkg/pgmodel/querier/row.go
```diff
@@ -164,7 +164,7 @@ func (r *sampleRow) GetAdditionalLabels() (ll labels.Labels) {
 
 // appendSampleRows adds new results rows to already existing result rows and
 // returns the as a result.
-func appendSampleRows(out []sampleRow, in pgxconn.PgxRows, tsSeries TimestampSeries, metric, schema, column string) ([]sampleRow, error) {
+func appendSampleRows(out []sampleRow, in pgxconn.PgxRows, tsSeries TimestampSeries, metricNameOverride, schema, column string) ([]sampleRow, error) {
```
alejandrodnm, line 167: Small refactor

pkg/pgmodel/querier/query_sample.go
```diff
@@ -124,8 +138,14 @@ func fetchMultipleMetricsSamples(ctx context.Context, tools *queryTools, metadat
 
 	// TODO this assume on average on row per-metric. Is this right?
 	results := make([]sampleRow, 0, len(metrics))
+
+	if len(metrics) == 0 {
+		return results, nil
+	}
```
alejandrodnm, line 144: If metrics is empty there's no reason to continue with the code. Before adding the timeout, it created an empty batch and tried to send it. Without this early return, if metrics is empty it will create a batch with just statement_timeout, so we return early and save the request.

pkg/pgmodel/querier/query_sample.go
```diff
@@ -96,21 +97,34 @@ func fetchSingleMetricSamples(ctx context.Context, tools *queryTools, metadata *
...
+
+// metricNameOverride returns the metric name to use if the table name and
+// series table name don't match. This happens on custom metric views that
+// share the series table with the raw metric.
+func metricNameOverride(metadata *evalMetadata) string {
+	if metadata.timeFilter.metric != metadata.timeFilter.seriesTable {
+		return metadata.timeFilter.metric
 	}
-	return samplesRows, topNode, nil
+	return ""
```
alejandrodnm, line 127: Refactor

pkg/pgxconn/helpers.go
cevian: I think such WARNINGS would not be great and pollute the log. I would probably do an explicit transaction then. But before doing so can we do some microbenchmarks and make sure it doesn't negatively affect performance (especially in high-latency cases)

jgpruitt: If I read the postgres mailing list you linked correctly, it seems like prepending `SET LOCAL statement_timeout = x;` to the sql provided may have the intended effect without the warning message.

alejandrodnm: I tried that with:

```go
r, err := conn.Query(context.Background(), "SET LOCAL statement_timeout = 4; SELECT 1, pg_sleep(3)")
```

But got the following error:

```
cannot insert multiple commands into a prepared statement (SQLSTATE 42601)
```

jgpruitt: Ah. Bummer.

niksajakovljevic:
Have you maybe considered alternative approach of canceling PG backend if timeout is reached? It would be more optimal because we would only execute additional statement when the context deadline is exceeded.

alejandrodnm:
> Have you maybe considered alternative approach of canceling PG backend if timeout is reached? It would be more optimal because we would only execute additional statement when the context deadline is exceeded.

You mean using `pg_cancel_backend(PID)`? I don't know how to reliably get the PID if that's the case. I'm going to try to dig a little more.

alejandrodnm:
I went looking into the pgx code, relying on propagating the context and setting a deadline to it might be good enough. If the context is canceled while a query is executing a cancel request is send to the server to terminate the running query. This has the same effect as what @niksajakovljevic is proposing here https://github.com/timescale/promscale/pull/1631#issuecomment-1265397741

I didn't think this worked because when I did the ctx propagation PR https://github.com/timescale/promscale/pull/1610 I didn't realise that the cancel request was being sent asynchronous. My tests ended before the request was being sent so the queries remained. Talking with the maintainer of PGX I realised my mistake https://github.com/jackc/pgx/issues/1313#issuecomment-1266763227

Some notes on canceling a query this way:

> The cancellation signal might or might not have any effect — for example, if it arrives after the backend has finished processing the query, then it will have no effect. If the cancellation is effective, it results in the current command being terminated early with an error message.
>
> The upshot of all this is that for reasons of both security and efficiency, the frontend has no direct way to tell whether a cancel request has succeeded. It must continue to wait for the backend to respond to the query. Issuing a cancel simply improves the odds that the current query will finish soon, and improves the odds that it will fail with an error message instead of succeeding.
>
> Since the cancel request is sent across a new connection to the server and not across the regular frontend/backend communication link, it is possible for the cancel request to be issued by any process, not just the frontend whose query is to be canceled. This might provide additional flexibility when building multiple-process applications. It also introduces a security risk, in that unauthorized persons might try to cancel queries. The security risk is addressed by requiring a dynamically generated secret key to be supplied in cancel requests.

I think we can close this PR and just rely on the context deadline for canceling long running queries WDYT @cevian @jgpruitt @niksajakovljevic @sumerman ? Is there a different use case I'm not aware that might require an explicit `statement_timeout`? At the end these were just helpers and in case we need to use `statement_timeout` for some particular reason we can just do it explicitly with a batch and transactions.

niksajakovljevic:
@alejandrodnm For now I'd rely on PGX canceling the query once the context expires. However if we figure out that cancel is not working well and it's affecting our users (like previously said there are no strong guarantees that query will terminate) we can then go with `statement_timeout` approach. 

> Is there a different use case I'm not aware that might require an explicit statement_timeout?
I don't think so. We do have background jobs that run from the extension but I am not sure we should set any timeout there.

alejandrodnm:
I'm going to close this. I agree with @niksajakovljevic and we should rely on pgx, we can revisit this if we find the cancel requests are not enough.

Issue #1632, opened 2022-09-09
tracing: StreamingSpanWriter doesn't seem to improve ingestion throughput 


arajkumar:
Currently there are 2 ways to get asynchronous ingestion behaviour in jager span ingestion.

1. Launch promscale with `--tracing.async-acks`
2. Use StreamingSpanWriter (enabled by default in latest the jaeger:main for grpc plugin, provided `StreamingImpl` in the server side)

But option 2 doesn't behave like a streaming writer when I benchmarked. Let's revisit the implementation to find the reason and fix it.

cevian:
Related thought: So I think our tracing async-acks stuff is right now a bit suboptimal. Right now tracing.async-acks is a boolean and false by default. But this setting is suboptimal for jaeger in terms of performance and doesn't actually help with data safety at all. I suggest making tracing.asyn-ack a three-valued variable (on, off, auto) - defaults to auto.
Auto mode would keep sync acks for otel, but have async acks for jaeger (both regular and streaming).

ramonguiu:
Why not make it on by default?

PR #1633, opened 2022-09-09
Enable async-acks for tracing by default

arajkumar:
## Description

Tracing ingestion pipeline in jaeger doesn't do any retry incase of span write failure, hence we are enabling async ack for tracing by default.

It also adds a flag to turn off streaming span writer incase if needed. By default streaming span writer is enabled.

Fixes https://github.com/timescale/promscale/issues/1632
Fixes https://github.com/timescale/promscale/issues/1635

## Merge requirements


pkg/jaeger/store/store.go
```diff
@@ -45,7 +49,17 @@ func (p *Store) SpanWriter() spanstore.Writer {
 }
 
 func (p *Store) StreamingSpanWriter() spanstore.Writer {
-	return p
+	return streamingSpanWriter{p}
+}
+
+func (s streamingSpanWriter) WriteSpan(ctx context.Context, span *model.Span) error {
+	go func() {
```
alejandrodnm, line 56: Should we add some back-preasure mechanism? Something like putting it behind a queue with a pool of goroutines that call the `WriteSpan`?

cevian: Yes backpressure is key here. Why isn't it enough to simply turn off async-acks for the streaming stuff

arajkumar: Sure, lets add a back pressure based on the defaults from https://github.com/jaegertracing/jaeger/blob/c1bb2946e670129314bf88d9730d8ed7566766d4/plugin/storage/grpc/shared/streaming_writer.go#L32.

arajkumar: Looks like I missed the point that the existing batching implementation already has the [back-pressure mechanism](https://github.com/timescale/promscale/blob/efec5ae94de49b1afcca4d3abeff13b8b14b2c59/pkg/pgmodel/ingestor/trace/trace_dispatcher.go#L63) and I can completely skip creating new go routine and turn on async ack as mentioned by @cevian .

alejandrodnm (approved):

cevian (approved):

PR #1634, opened 2022-09-09
Maintenance backlog metrics

### LINKED ISSUE #1597: Introduce the "the amount of work left" metrics for the maintenance job
The amount of work that remains to be done:
- The number of chunks to be removed
- The number of chunks to be compressed
- Break down by the job type

sumerman:
## Description

- Introducing metrics for to-be-compressed and to-be-deleted chunks.
- Adding new metrics to the dashboard.
- Adding alerts based on the new metrics

Resolves #1597  &  #1598

## Merge requirements


pkg/pgmodel/metrics/database/metrics.go
```diff
@@ -92,6 +92,90 @@ var metrics = []metricQueryWrap{
...
+			     INNER JOIN _timescaledb_catalog.chunk c ON c.id = chcons.chunk_id
+			     INNER JOIN _timescaledb_catalog.hypertable h ON h.id = c.hypertable_id
+		    WHERE c.dropped IS FALSE
+		      AND h.compression_state = 1 -- compression_enabled = TRUE
+		      AND (c.status & 1) != 1) -- only check for uncompressed chunks
+		SELECT count(*)::BIGINT
+		FROM chunk_candidates cc
+			 INNER JOIN _timescaledb_catalog.dimension_slice ds ON ds.id = cc.dimension_slice_id
+			 INNER JOIN _prom_catalog.metric m ON (m.table_name = cc.table_name AND m.table_schema = cc.schema_name)
+		WHERE (m.delay_compression_until IS NULL OR m.delay_compression_until < now())
```
cevian, line 134: we should probably add a metric for counting chunks that would be compressed but are delayed as well. That's useful for debugging cases where backfill has caused disk increase due to delay in compression

cevian: also applies to traces

cevian (changes requested):
Overall looks good. One change requested. Also, if these queries are more optimized than the maintenance queries, should we change those?

sumerman:
> Also, if these queries are more optimized than the maintenance queries, should we change those?

IMO only the metric retention path has a significant effort/gains ratio. I've created the issue for it https://github.com/timescale/promscale_extension/issues/507 

Traces retention relies on the native `drop_chunks`, and I don't think we should roll our own implementation here.

Both metric and trace compression would require more refactoring, yet the new query is only marginally faster.

pkg/pgmodel/metrics/database/metrics.go
sumerman: I've added it for the metrics data, but I couldn't find it for traces.

pkg/pgmodel/metrics/database/metrics.go
```diff
@@ -45,6 +45,30 @@ type metricQueryWrap struct {
...
+
+	return `WITH chunk_candidates AS MATERIALIZED (
+		    SELECT chcons.dimension_slice_id, h.table_name, h.schema_name
+		    FROM _timescaledb_catalog.chunk_constraint chcons
+			     INNER JOIN _timescaledb_catalog.chunk c ON c.id = chcons.chunk_id
+			     INNER JOIN _timescaledb_catalog.hypertable h ON h.id = c.hypertable_id
+		    WHERE c.dropped IS FALSE
+		      AND h.compression_state = 1 -- compression_enabled = TRUE
+		      AND (c.status & 1) != 1) -- only check for uncompressed chunks
+		SELECT count(*)::BIGINT
```
cevian, line 62: it would be more efficient to execute one query for both the delayed and not delayed case 
```suggestion
		SELECT count(*)::BIGINT FILTER(WHERE m.delay_compression_until IS NULL OR m.delay_compression_until < now()) as chunks_to_compress,
		SELECT count(*)::BIGINT FILTER(WHERE m.delay_compression_until IS NOT NULL AND m.delay_compression_until >= now()) as chunks_delayed
```

but I am not sure how hard it would be to build in multi-column/multi-metric infrastructure or if we already have it

sumerman: done

Harkishen-Singh (approved):
I didn't go through the SQL queries in deep, but I assume they are right. The changes to metrics engine are good (more efficiently, since a single query can now support multiple metrics).

Changes to the dashboard look good. However, it will be good if we can rename the top panel (see screenshot) in this PR from `Samples Ingest Rate` to `Ingest Rate`, since it contains samples, spans and metadata (otherwise please create an issue).

Screenshot 2022-09-15 at 12 44 19 PM

Harkishen-Singh:
I found something unusual.

I have 324 metrics in my system.

```postgresql
postgres=# select count(*) from _prom_catalog.metric;
 count 
-------
   324
(1 row)
```

But, the new panel added for `Maintenance job backlog` says that its 390 in backlog, which is confusing. Can this be a bug?

Screenshot 2022-09-15 at 12 49 42 PM

sumerman:
> I found something unusual.
> 
> 
> 
> I have 324 metrics in my system.
> 
> 
> 
> ```postgresql
> 
> postgres=# select count(*) from _prom_catalog.metric;
> 
>  count 
> 
> -------
> 
>    324
> 
> (1 row)
> 
> ```
> 
> 
> 
> But, the new panel added for `Maintenance job backlog` says that its 390 in backlog, which is confusing. Can this be a bug?
> 
> Screenshot 2022-09-15 at 12 49 42 PM
> 
> 
> 
> Edit: 5 mins after, its saying `425` in the panel.
> 
> Screenshot 2022-09-15 at 12 53 01 PM

Apologies. The naming scheme I chose ended up confusing. The backlog panel always shows the number of chunks and individual series names are meant to indicate what kind of chunks they include. I.e. "metrics uncompressed 400" means that there are 400 chunks, from metric-related hypertables are pending compression. I'll try and come up with a better naming scheme, suggestions are very welcome!

pkg/pgmodel/metrics/database/metrics.go
```diff
@@ -59,41 +77,127 @@ var metrics = []metricQueryWrap{
...
 				Subsystem: "sql_database",
 				Name:      "chunks_compressed_count",
 				Help:      "Total number of compressed chunks in TimescaleDB currently.",
 			},
 		),
-		query: `select count(*)::bigint from _timescaledb_catalog.chunk where dropped=false and compressed_chunk_id is not null`,
+		// Compressed_chunk_id is null for both yet to be compressed and already compressed chunks.
+		query: `SELECT 
+				count(*) FILTER (WHERE dropped=false AND compressed_chunk_id IS NULL)::BIGINT AS chunks_count,
+				count(*) FILTER (WHERE compressed_chunk_id IS NULL)::BIGINT AS chunks_created,
```
cevian, line 103: while not part of the changes introduced in this pr. I just noticed that chunks_created is a pretty meaningless metric. The problem is that in timescaledb chunks are sometimes dropped by just deleting the row and sometimes by updating dropped=false (depends on how/if cont aggs are defined) so just relying on dropped=true is misleading. Can we just drop this metric? cc @Harkishen-Singh .

cevian (approved):
Approved since don't want to block on requested change. But let's see if we can coordinate with @Harkishen-Singh

Harkishen-Singh:
Not a blocker for merging. That said, does 'metric chunks uncompressed 400' sound relatable to context?

sumerman:
> Not a blocker for merging. That said, does 'metric chunks uncompressed 400' sound relatable to context?

It would work, but then we'll have very long labels. My suggestion would be to change the panel's title to "The number of chunks to be processed by maintenance jobs" and leave the labels as they are. WDYT?

Harkishen-Singh:
Sounds good

pkg/pgmodel/metrics/database/metrics.go
Harkishen-Singh: Sorry, I do not understand the problem. The aim of `chunks_created` is to keep track of all chunks created in TimescaleDB since starting and it is meaningful for the user.

This is the query for knowing `promcsale_sql_database_chunks_created` is `select count(*)::bigint from _timescaledb_catalog.chunk where compressed_chunk_id is null` and I don't see where is `dropped=true` here.

Harkishen-Singh: > in timescaledb chunks are sometimes dropped by just deleting the row and sometimes by updating dropped=false (depends on how/if cont aggs are defined) so just relying on dropped=true

Which `row`? Sorry, but the above statement is confusing to me.

cevian: The _timescaledb_catalog.chunk can be deleted. which would then cause chunks_created to decrease according to the current query. That would be weird, no?

Harkishen-Singh: Yes that will be weird. In that case chunks_created metric would be wrong and we should remove it.

I guess, this behaviour is for performance reasons, like preventing the chunks table from getting too big, but at the same time, this is a buggy behaviour from db, right!

Issue #1636, opened 2022-09-09
ignore HTTP AUTH for  /healthz


nbari:
when using options `-web.auth.username & -web.auth.password` is there a way to skip the AUTH for the  endpoint ` /healthz` 

It would simplify the checks when using a load balancer

PR #1637, opened 2022-09-12
Add cmd flag to skip http paths from authentication

arajkumar:

## Description

This PR adds a flag named `--web.auth.ignore-path` which takes a http path and passed path would be ignored from authentication. The flag shall be repeated to pass array of ignored paths.

e.g.
```
# skips authn for heathz and api/query_range endpoints.
./dist/promscale --web.auth.ignore-path=/heathz --web.auth.ignore-path=/api/query_range
```

Fixes https://github.com/timescale/promscale/issues/1636

Code changes are inspired from https://github.com/brancz/kube-rbac-proxy/blob/fc1ca4f969941340a8adb66932bd64dc5773d37a/main.go#L298-L321

## Merge requirements


pkg/api/router.go
```diff
@@ -163,8 +164,22 @@ func authHandler(cfg *Config, handler http.Handler) http.Handler {
 		return handler
 	}
 
+	isIgnoredPath := func(r *http.Request) bool {
```
alejandrodnm, line 167: Maybe we can add this as a method to `Auth`?

alejandrodnm (approved):

antekresic (changes requested):
Overall looks good, a bit of organization would benefit us here with the rise of complexity in the auth code.

Harkishen-Singh:
I will pass this review to other reviewers.

alejandrodnm (approved):

pkg/api/router.go
```diff
@@ -29,7 +28,7 @@ import (
 	"github.com/timescale/promscale/pkg/telemetry"
 )
 
-func GenerateRouter(apiConf *Config, promqlConf *query.Config, client *pgclient.Client, store *jaegerStore.Store, reload func() error) (*mux.Router, error) {
+func GenerateRouter(apiConf *Config, promqlConf *query.Config, client *pgclient.Client, store *jaegerStore.Store, authWrapper mux.MiddlewareFunc, reload func() error) (*mux.Router, error) {
```
antekresic, line 31: This is obviously a smell, having so many input parameters on a function. The function itself is too big as well.

Lets put a TODO on this to refactor it in the future so when somebody goes to add another parameter, they can see it and not do it 😃

antekresic (approved):
Approved with a small ask to add a TODO for some future refactoring.

PR #1646, opened 2022-09-15
Optimize label queries with tenants

cevian:
Prevent multiple series scans when there are multiple valid tenants.

## Merge requirements

### DIFF
--- a/pkg/pgmodel/lreader/labels_reader.go
+++ b/pkg/pgmodel/lreader/labels_reader.go
@@ -48,7 +48,7 @@ const (
 	tenantLabelValueQual   = `value = $%d`
 	getLabelNamesForTenant = `
 	SELECT
-		array_agg(keys.key)
+		array_agg(keys.key order by keys.key)
 	FROM (
 		SELECT
 			distinct l.key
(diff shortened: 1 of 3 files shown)


pkg/pgmodel/lreader/labels_reader.go
```diff
@@ -45,21 +45,53 @@ func NewLabelsReader(conn pgxconn.PgxConn, labels cache.LabelsCache, mt tenancy.
...
-)
+	tenantLabelValueQual   = `value = $%d`
+	getLabelNamesForTenant = `
+	SELECT
+		array_agg(keys.key)
+	FROM (
+		SELECT
+			distinct l.key
+		FROM _prom_catalog.label l
+	) as keys
```
Harkishen-Singh, line 56: I think we need some `ORDER BY key` here to pass the E2E tests.

Harkishen-Singh (changes requested):

Harkishen-Singh (approved):
The test s should pass after updating the SQL query in `TestLabelsReaderLabelsValues` and then we can merge (along with commits clean up)

PR #1649, opened 2022-09-16
Support for Jaeger tags with binary value

alejandrodnm:
### Support for Jaeger tags with binary value 
 
 Since we are using the OTEL translator, when ingesting Jaeger traces all
the tags with `VType=ValueType_BINARY` are being encoded into base64 and
stored as strings. The context of the type of value is lost after this
translation occurs, meaning that, when we return those tags in a Jaeger
query we return the base64 string as `VType=ValueType_STRING`.

To preserve the type of the value, the prefix
`data:application/octet-stream; base64,` will be added to the base64
string before storing in the DB. When retrieving a trace we look for
tags that are string and have this prefix, we remove it and return a
`ValueType_BINARY` with the result of decoding the base64 string.

The prefix was choosen from the `The "data" URL scheme` RFC
https://www.rfc-editor.org/rfc/rfc2397#section-2 .

Jaeger allows these type of tags for Spans, Process and Logs.

## Merge requirements

### DIFF
--- a/pkg/jaeger/store/binary_tags.go
+++ b/pkg/jaeger/store/binary_tags.go
@@ -1,0 +1,95 @@
+package store
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+const MEDIA_TYPE_ENCODED_BINARY = "data:application/octet-stream; base64,"
+const MEDIA_TYPE_ENCODED_BINARY_LEN = len(MEDIA_TYPE_ENCODED_BINARY)
+
+// decodeSpanBinaryTags decodes the tags with binary values that are present in
+// the binaryTags sets for the span, process and logs.
+//
+// When writing binary tags we encode the slice of bytes into a base64 string
+// representation and add the prefix `__ValueType_BINARY__`. Decoding implies
+// removing the prefix and decoding the base64 string.
+func decodeSpanBinaryTags(span *model.Span) {
+	decodeBinaryTags(span.Tags)
+	decodeBinaryTags(span.Process.Tags)
+	for _, log := range span.Logs {
+		decodeBinaryTags(log.Fields)
+	}
+}
+
+func decodeBinaryTags(tags []model.KeyValue) {
+	for i, tag := range tags {
+		if tag.GetVType() != model.ValueType_STRING {
+			continue
+		}
+
+		encoded := tag.VStr
+		if !strings.HasPrefix(encoded, MEDIA_TYPE_ENCODED_BINARY) {
+			continue
+		}
+
+		vBin, err := decodeBinaryTagValue(encoded)
+
+		// If we can't decode it means that we didn't encode it in the
+		// first place, so we should keep it as is.
+		if err != nil {
+			continue
+		}
+		tags[i] = model.KeyValue{
+			Key:     tag.Key,
+			VType:   model.ValueType_BINARY,
+			VBinary: vBin,
+		}
+	}
+}
+
+func decodeBinaryTagValue(encoded string) ([]byte, error) {
+	v := encoded[MEDIA_TYPE_ENCODED_BINARY_LEN:]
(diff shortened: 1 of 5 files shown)


pkg/tests/end_to_end_tests/jeager_binary_tag_test.go.b
```diff
@@ -0,0 +1,123 @@
+package end_to_end_tests
```
alejandrodnm, line 1: I uploaded this by mistake 😄

pkg/jaeger/store/binary_tags.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for i, tag := range span.Process.Tags {
+		if isBinaryTag(tag) {
+			continue
+		}
+		span.Process.Tags[i] = encodeBinaryTagToStr(tag)
+	}
+}
+
+func isBinaryTag(tag model.KeyValue) bool {
```
arajkumar, line 134: This function does exactly opposite of what the name tells. Either fix the name or `if `handler

pkg/jaeger/store/binary_tags.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"github.com/jaegertracing/jaeger/model"
+)
+
+const VALUE_TYPE_BINARY = "__ValueType_BINARY__"
+const VALUE_TYPE_BINARY_LEN = len(VALUE_TYPE_BINARY)
```
arajkumar, line 12: Can we follow data URI spec[1] which is been widely used in the web world? May be following a standard would help us to pitch this change in the opentelemetry upstream without much hassle.

e.g.

```
data:application/octet-stream; base64,AAAwOQ==
```

Might look lengthy, but it follows the standard :)

[1] https://www.rfc-editor.org/rfc/rfc2397#section-2

pkg/tests/end_to_end_tests/fixtures/jaeger_binaray_tag/trace.json
```diff
@@ -0,0 +1,41 @@
+{
```
arajkumar, line 1: I think we should import all tests, let me do it as part of https://github.com/timescale/promscale/pull/1627

arajkumar (changes requested):

pkg/jaeger/store/binary_tags.go
alejandrodnm: OMG xD

pkg/tests/end_to_end_tests/fixtures/jaeger_binaray_tag/trace.json
alejandrodnm: Once we merge https://github.com/timescale/promscale/pull/1627/files in which Arun imports all the Jaeger integration tests we can clean this fixtures and the test related code added by this PR

alejandrodnm: I deleted the e2e test and added some unittests instead

pkg/jaeger/store/find_traces.go
```diff
@@ -36,13 +36,19 @@ func findTraces(ctx context.Context, builder *Builder, conn pgxconn.PgxConn, q *
 
 func scanTraces(rows pgxconn.PgxRows) ([]*model.Trace, error) {
 	traces := ptrace.NewTraces()
+	spansBinaryTags := map[int64]*spanBinaryTags{}
```
arajkumar, line 39: Instead of tracking span with binary tag why can't we just iterate over final jaeger trace and decode if the string type is matching base64 prefix?

alejandrodnm: We are already iterating them when we decode from the database. 

I didn't want to have to do it twice, per every span, span process and span logs.

WDYT? Am I being too paranoid

arajkumar (approved):
lgtm

PR #1650, opened 2022-09-16
First attempt at a fairer copier

cevian:
## Merge requirements

### DIFF
--- a/pkg/pgmodel/ingestor/copier.go
+++ b/pkg/pgmodel/ingestor/copier.go
@@ -193,6 +193,8 @@ func copierGetBatch(ctx context.Context, batch []copyRequest, in <-chan readRequ
 		targetPoints = 1000
 	}
 
+	metrics.IngestorBatchTargetPoints.With(labelsCopier).Observe(float64(targetPoints))
+
 	//we use a small timeout to prevent low-pressure systems from using up too many
 	//txns and putting pressure on system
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/ingestor/copier.go
```diff
@@ -174,30 +174,51 @@ func copierGetBatch(ctx context.Context, batch []readRequest, in <-chan readRequ
...
+	copyRequest, ok := <-req.copySender
+	if !ok {
+		return batch, stats, false
+	}
+	batch = append(batch, copyRequest)
+	stats.Update(copyRequest)
+
+	totalPoints := atomic.LoadInt64(&metrics.BatchingPoints)
+	//try to get a fair number of points across copiers
+	targetPoints := int(totalPoints) / numCopiers
```
cevian, line 189: The key change is this line which specifies how many points a batch should be instead of relying on a constant number of metrics (100 as was the case previously)

pkg/pgmodel/ingestor/copier.go
```diff
@@ -174,30 +174,51 @@ func copierGetBatch(ctx context.Context, batch []readRequest, in <-chan readRequ
...
+	//but don't go too crazy small batches help no one
+	if targetPoints < 1000 {
```
cevian, line 191: this may be too high

PR #1658, opened 2022-09-20
Fix event tag correctness

cevian:
When searching for tags in jaeger, a tag can come from any event attached to a span. Different tags can come from different events on the same span. Previously, this was broken since the events were joined to spans and tags searched one-at-a-time.

Fixed this by searching with EXISTS.

## Merge requirements


arajkumar (approved):
lgtm

PR #1660, opened 2022-09-20
Make TimescaleDB usage mandatory

JamesGuthrie:
## Description

In https://github.com/timescale/promscale_extension/issues/514 we realised that something unexpected had happened because the user installed the promscale extension without installing the timescaledb extension beforehand.

In https://github.com/timescale/promscale_extension/pull/515 we make timescaledb a requirement to install the promscale extension.

This change makes the Promscale connector error out when the timescaledb extension is not installed in the target DB, and performs some cleanup.

## Merge requirements


alejandrodnm (approved):
Would we need to change anything in the official docs regarding the now mandatory nature of timescale? https://github.com/timescale/docs/tree/latest/promscale

And I'm curious, what was the use case for promscale without timescale? Was it to run it against a vanilla PostgreSQL?

JamesGuthrie:
> Would we need to change anything in the official docs regarding the now mandatory nature of timescale? https://github.com/timescale/docs/tree/latest/promscale

I think that we will have to do some minor changes. Looking into it, I realised that we need to do similar changes in this repo too.

> And I'm curious, what was the use case for promscale without timescale? Was it to run it against a vanilla PostgreSQL?

Yes.

Harkishen-Singh (commented):
Did we have any discussion on this before removing it?

JamesGuthrie:
> Did we have any discussion on this before removing it?

The trigger for this change is this issue: https://github.com/timescale/promscale_extension/issues/514.

Harkishen-Singh:
Based on discussion in https://github.com/timescale/promscale_extension/issues/514
Instead of making TimescaleDB usage mandatory, why not never install jobs in pure postgres scenarios?

JamesGuthrie:
> Instead of making TimescaleDB usage mandatory, why not never install jobs in pure postgres scenarios?

This is what we currently do, and that is exactly how the issue in https://github.com/timescale/promscale_extension/issues/514 came about.

As I understand it, our current thesis is that 99.9% of people who want to use Promscale want to use it with TimescaleDB. The issue is that because we support both use-cases of with and without TimescaleDB, it's easy for users to accidentally end up in a situation where they're not actually using TimescaleDB, or not "fully" (as above).

So the question arises: is it worthwhile to continue maintaining a use-case which nobody wants to use, and which makes our intended use-case harder to "get right".

As I understand it, the answer is: No, it is not worthwhile.

Harkishen-Singh:
>it's easy for users to accidentally end up in a situation where they're not actually using TimescaleDB, or not "fully" (as above).

Ah, now i get it. With this PR, we avoid the accidental case and ensure users have a common experience with Promscale. SGTM.

PR #1663, opened 2022-09-22
Add cache benchmarks for struct key and value combinations

JamesGuthrie:
## Description

In our real-world cache usage, we tend to store struct objects, not just
primitives. These benchmarks should help to ascertain performance
differences in the cache for these use-cases.

## Merge requirements

### DIFF
--- a/pkg/clockcache/cache_bench_test.go
+++ b/pkg/clockcache/cache_bench_test.go
@@ -37,6 +37,168 @@ func BenchmarkStringCache(b *testing.B) {
 	}
 }
 
+type Key struct {
+	index int64
+	value string
+}
+
+type Value struct {
+	name, value string
+}
+
+// tests cache speed and memory usage for struct key and value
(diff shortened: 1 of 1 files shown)


pkg/clockcache/cache_bench_test.go
```diff
@@ -37,6 +37,168 @@ func BenchmarkStringCache(b *testing.B) {
...
+	index int64
+	value string
+}
+
+type Value struct {
+	name, value string
+}
+
+// tests cache speed and memory usage for struct key and value
+func BenchmarkStructKeyValueCacheEviction(b *testing.B) {
```
alejandrodnm, line 50: I saw that for the other benchmarks we do 

```
	for _, n := range []int{500, 5000, 50000} {
		b.Run(fmt.Sprintf("%d", n), func(b *testing.B) {
```

Is there a reason why you are not doing it like this for these benchmarks?

PR #1665, opened 2022-09-22
Wrap extension upgrades in a transaction

jgpruitt:
## Description

In some cases, an extension upgrade is a multi-step process. This change wraps some of those steps in an explicit transaction. Specifically, we don't want a failure to result in the extension being left at version 0.0.0

## Merge requirements

### DIFF
--- a/pkg/pgmodel/new_migrate.go
+++ b/pkg/pgmodel/new_migrate.go
@@ -141,23 +141,64 @@ func Migrate(conn *pgx.Conn, appVersion VersionInfo, leaseLock *util.PgAdvisoryL
 		return fmt.Errorf("failed to determine whether the prom_schema_migrations table existed: %w", err)
 	}
 	if schemaMigrationTableExists {
-		mig := NewMigrator(conn, migrations.MigrationFiles, TableOfContents)
-		if err = mig.Migrate(appSemver); err != nil {
-			return fmt.Errorf("Error encountered during migration: %w", err)
+		if err = upgradeThroughAllBalls(conn, appSemver, extOptions); err != nil {
+			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/pgmodel/new_migrate.go
```diff
@@ -145,19 +146,40 @@ func Migrate(conn *pgx.Conn, appVersion VersionInfo, leaseLock *util.PgAdvisoryL
 		if err = mig.Migrate(appSemver); err != nil {
```
alejandrodnm, line 146: Is it necessary for this to be inside the explicit transaction you're creating?

alejandrodnm (approved):

antekresic (changes requested):
Would it work if you wrapped the whole migration path into a single transaction? Would make this change so much better.

If not, I suggest wrapping this extension upgrade section into a single function so it makes the code cleaner.

pkg/pgmodel/new_migrate.go
jgpruitt: It is not strictly necessary though it would be nice. Unfortunately, the Migrate function starts its own explicit transactions somewhere in its path. So, it's not as straightforward to wrap it as I had hoped. I really don't want to touch that old code unless i absolutely have to.

jgpruitt: The Migrate function is the old way of doing database migrations. It would be nice if it were inside the transaction, but it's not strictly necessary. It creates its own explicit transactions, so it's not as easy as just wrapping it, unfortunately. What I mainly want to prevent here is a user getting stuck at version 0.0.0 which will be prevented by wrapping the rest in a transaction.

antekresic (approved):

PR #1678, opened 2022-09-29
Fix retrieve traces query data inconsistencies

### LINKED ISSUE #1664: Traces query doesn't return all the related data and sometimes returns duplicates
**Describe the bug**

Problems with find traces query

### `array_agg` returns an empty value when `trace_state` is null.

The query uses a select statement similar to `array_agg(lk.<column> ORDER BY lk.link_nbr) FILTER(WHERE lk IS NOT NULL)` with `lk` defined as `LEFT JOIN _ps_trace.link lk ON lk.trace_id = s.trace_id AND lk.span_id = s.span_id AND lk.span_start_time < X`

https://github.com/timescale/promscale/blob/c231ed2edf0dff2dda850a5cc36441b8f5a2f7f6/pkg/jaeger/store/trace_query.go#L43-L47

The problem is with `lk IS NOT NULL`, as @cevian pointed out, from the [PostgreSQL docs](https://www.postgresql.org/docs/9.2/functions-comparison.html):

```
If the expression is row-valued, then IS NULL is true when the row expression itself is null or when all the row's fields are null, while IS NOT NULL is true when the row expression itself is non-null and all the row's fields are non-null. Because of this behavior, IS NULL and IS NOT NULL do not always return inverse results for row-valued expressions; in particular, a row-valued expression that contains both null and non-null fields will return false for both tests. In some cases, it may be preferable to write row IS DISTINCT FROM NULL or row IS NOT DISTINCT FROM NULL, which will simply check whether the overall row value is null without any additional tests on the row fields.
```

That's one problem that can be solved with `row IS DISTINCT FROM NULL` as the docs say. 

### Duplicate values in `array_agg` results

This a second problem is not going into a separate issue because it is related to the same query statement.

In the query we aggregate with `array_agg` values from 2 different tables `_ps_trace.link` and `_ps_trace.event`. This tables are LEFT JOINED against the `_ps_trace.span` table matching against `trace_id` and `span_id`. When we have multiple values in each table that satisfy the join condition we end up with the cartesian product of both tables, and thus duplicated values.

A simple example:

```sql
promscale=# select * from _ps_trace.event;
             time              │               trace_id               │ span_id │ event_nbr │ dropped_tags_count │ name │               tags
═══════════════════════════════╪══════════════════════════════════════╪═════════╪═══════════╪════════════════════╪══════╪═══════════════════════════════════
 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-000000000012 │       4 │         0 │                  0 │      │ {"1002": 3, "1003": 4, "1004": 5}
 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-000000000012 │       4 │         1 │                  0 │      │ {}

promscale=# select * from _ps_trace.link;
               trace_id               │ span_id │        span_start_time        │           linked_trace_id            │ linked_span_id │ link_nbr │ dropped_tags_count │ trace_state │ tags
══════════════════════════════════════╪═════════╪═══════════════════════════════╪══════════════════════════════════════╪════════════════╪══════════╪════════════════════╪═════════════╪══════
 00000000-0000-0000-0000-000000000012 │       4 │ 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-0000000000ff │            255 │        0 │                  0 │ ∅           │ {}
 00000000-0000-0000-0000-000000000012 │       4 │ 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-000000000001 │              2 │        1 │                  0 │ ∅           │ {}
 00000000-0000-0000-0000-000000000012 │       4 │ 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-000000000001 │              2 │        2 │                  0 │ ∅           │ {}
 00000000-0000-0000-0000-000000000012 │       4 │ 2022-09-20 18:46:31.639875+02 │ 00000000-0000-0000-0000-0000000000ff │            255 │        0 │                  0 │ ∅           │ {}
```

Running the full query we get duplicate values for both event and link values (notice that we are already using `IS DISTINCT FROM NULL`):

```sql
promscale=# 
SELECT
  s.trace_id,
  s.span_id,
  s.parent_span_id,
  s.start_time start_times,
  s.end_time end_times,
  o.span_kind,
  s.dropped_tags_count dropped_tags_counts,
  s.dropped_events_count dropped_events_counts,
  s.dropped_link_count dropped_link_counts,
  s.status_code status_code_string,
  s.status_message,
  s.trace_state trace_states,
  s_url.url schema_urls,
  o.span_name span_names,
  _ps_trace.tag_map_denormalize(s.resource_tags) resource_tags,
  _ps_trace.tag_map_denormalize(s.span_tags) span_tags,
    array_agg(e.name ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL) event_names,
  array_agg(e.time ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL) event_times,
    array_agg(e.dropped_tags_count ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL)  event_dropped_tags_count,
  array_agg(_ps_trace.tag_map_denormalize(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL) event_tags,
    inst_lib.name library_name,
    inst_lib.version library_version,
  inst_lib_url.url library_schema_url,
  array_agg(lk.linked_trace_id ORDER BY lk.link_nbr) FILTER(WHERE lk IS DISTINCT FROM NULL) links_linked_trace_ids,
  array_agg(lk.linked_span_id ORDER BY lk.link_nbr)  FILTER(WHERE lk IS DISTINCT FROM NULL) links_linked_span_ids,
  array_agg(lk.trace_state ORDER BY lk.link_nbr)     FILTER(WHERE lk IS DISTINCT FROM NULL) links_trace_states,
  array_agg(lk.dropped_tags_count ORDER BY lk.link_nbr) FILTER(WHERE lk IS DISTINCT FROM NULL) links_dropped_tags_count,
  array_agg(_ps_trace.tag_map_denormalize(lk.tags) ORDER BY lk.link_nbr) FILTER(WHERE lk IS DISTINCT FROM NULL) links_tags
FROM
  _ps_trace.span s
INNER JOIN
  _ps_trace.operation o ON (s.operation_id = o.id)
LEFT JOIN
  _ps_trace.schema_url s_url ON s.resource_schema_url_id = s_url.id
LEFT JOIN
  _ps_trace.event e ON e.span_id = s.span_id AND e.trace_id = s.trace_id
LEFT JOIN
  _ps_trace.instrumentation_lib inst_lib ON s.instrumentation_lib_id = inst_lib.id
LEFT JOIN
  _ps_trace.schema_url inst_lib_url ON inst_lib_url.id = inst_lib.schema_url_id
LEFT JOIN
  _ps_trace.link lk ON lk.trace_id = s.trace_id AND lk.span_id = s.span_id
GROUP BY
  s.trace_id,
  s.span_id,
  s.parent_span_id,
  s.start_time,
  s.end_time,
  s.resource_tags,
  s.span_tags,
  o.span_name,
  o.span_kind,
  s_url.url,
  inst_lib.name,
  inst_lib.version,
  inst_lib_url.url
promscale-# ;
─[ RECORD 1 ]────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
trace_id                 │ 00000000-0000-0000-0000-000000000012
span_id                  │ 4
parent_span_id           │ ∅
start_times              │ 2022-09-20 18:46:31.639875+02
end_times                │ 2022-09-20 18:46:31.639877+02
span_kind                │ unspecified
dropped_tags_counts      │ 0
dropped_events_counts    │ 0
dropped_link_counts      │ 0
status_code_string       │ unset
status_message           │
trace_states             │ ∅
schema_urls              │ ∅
span_names               │ query12-operation
resource_tags            │ {"service.name": "query12-service"}
span_tags                │ {"blob": "data:application/octet-stream; base64,AAAwOQ==", "sameplacetag1": "sameplacevalue", "sameplacetag2": 123, "sameplacetag3": 72.5, "sameplacetag4": true}
event_names              │ {"","","","","","","",""}
event_times              │ {"2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02","2022-09-20 18:46:31.639875+02"}
event_dropped_tags_count │ {0,0,0,0,0,0,0,0}
event_tags               │ {"{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}",NULL,NULL,NULL,NULL}
library_name             │ ∅
library_version          │ ∅
library_schema_url       │ ∅
links_linked_trace_ids   │ {00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001}
links_linked_span_ids    │ {255,255,255,255,2,2,2,2}
links_trace_states       │ {NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL}
links_dropped_tags_count │ {0,0,0,0,0,0,0,0}
links_tags               │ {NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL}
```

One option here could be aggregating at the JOIN level (suggestion from @cevian):

```sql
LEFT JOIN LATERAL (
     SELECT 
        array_agg(lk.linked_trace_id ORDER BY lk.link_nbr) 	links_linked_trace_ids,
	array_agg(lk.linked_span_id ORDER BY lk.link_nbr) 	links_linked_span_ids,
	array_agg(lk.trace_state ORDER BY lk.link_nbr)      links_trace_states,
	array_agg(lk.dropped_tags_count ORDER BY lk.link_nbr) links_dropped_tags_count,
	array_agg(_ps_trace.tag_map_denormalize(lk.tags) ORDER BY lk.link_nbr) 	links_tags
     FROM  _ps_trace.link
     WHERE lk.trace_id = s.trace_id AND lk.span_id = s.span_id
) as link ON (TRUE)
```

Since the problem is just with these 2 tables, we would only need to aggregate in the JOIN for only one of them, this would give us span(1 row) x link(1 row) x events(x rows).

Since `lk.link_nbr` is the position on the array of links of the span, there's an implicit unique constraint (trace_id, span_id, link_nbr), since we are already matching against the first 2 in the JOIN, then `lk.link_nbr` should be unique per trace_id, span_id pair. Taking advantage of this knowledge, I tried:
 
#### FAILED: adding `DISTINCT` to `array_agg`

`array_agg(DISTINCT lk.<column> ORDER BY lk.link_nbr)`: This would be ideal but postgreSQL doesn't allow it, the column used in `ORDER BY` must be used as input to `DISTINCT`.

```sql
promscale=#
SELECT
  s.trace_id,
  s.span_id,
  _ps_trace.tag_map_denormalize(s.resource_tags)        resource_tags,
  _ps_trace.tag_map_denormalize(s.span_tags)            span_tags,
  array_agg(e.name ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL)                  event_names,
  array_agg(_ps_trace.tag_map_denormalize(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL)   event_tags,
  array_agg(DISTINCT lk.linked_trace_id, lk.linked_trace_id ORDER BY lk.link_nbr) FILTER(WHERE lk IS DISTINCT FROM NULL)        links_linked_trace_ids
FROM
  _ps_trace.span s
LEFT JOIN
  _ps_trace.event e ON e.span_id = s.span_id AND e.trace_id = s.trace_id
LEFT JOIN
  _ps_trace.link lk ON lk.trace_id = s.trace_id AND lk.span_id = s.span_id
GROUP BY
  s.trace_id,
  s.span_id,
  s.resource_tags,
  s.span_tags;
ERROR:  42883: function array_agg(trace_id, trace_id) does not exist
LINE 8:   array_agg(DISTINCT lk.linked_trace_id, lk.linked_trace_id ...
          ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
LOCATION:  ParseFuncOrColumn, parse_func.c:636
Time: 1.119 ms
promscale=# \e
ERROR:  42P10: in an aggregate with DISTINCT, ORDER BY expressions must appear in argument list
LINE 8:   array_agg(DISTINCT lk.linked_trace_id ORDER BY lk.link_nbr...
```

Other option 

```sql
array_agg(DISTINCT ROW(lk.<column>, lk.link_nbr) ORDER BY lk.link_nbr)
``` 

Fails as well for the same reason.

#### SUCCESS: Aggregating as json

Using the json aggregation functions we can do either `json_object_agg(DISTINCT lk.link_nbr, lk.<column>)`. Ex:

```sql
json_object_agg(DISTINCT lk.link_nbr, lk.linked_trace_id)
```

Returns

```json
{ 
"0" : "00000000-0000-0000-0000-0000000000ff", 
"1" : "00000000-0000-0000-0000-000000000001", 
"2" : "00000000-0000-0000-0000-000000000001"
 }
```

It handle null columns well:

`json_object_agg(DISTINCT lk.link_nbr, lk.trace_state) = { "0" : null, "1" : null, "2" : null }`

Another option could be using `jsonb_object_agg` and aggregate the full row:

```sql
jsonb_object_agg(
  lk.link_nbr,
  json_build_object(
    'linked_trace_id', lk.trace_id,
    'lk.linked_span_id', lk.linked_span_id,
    'lk.trace_state', lk.trace_state,
    'lk.dropped_tags_count', lk.dropped_tags_count,
    'tags', _ps_trace.tag_map_denormalize(lk.tags)
  )
)
```

```json
{
  "0": {
    "tags": null,
    "lk.trace_state": null,
    "linked_trace_id": "00000000-0000-0000-0000-000000000012",
    "lk.linked_span_id": 255,
    "lk.dropped_tags_count": 0
  },
  "1": {
    "tags": null,
    "lk.trace_state": null,
    "linked_trace_id": "00000000-0000-0000-0000-000000000012",
    "lk.linked_span_id": 2,
    "lk.dropped_tags_count": 0
  },
  "2": {
    "tags": null,
    "lk.trace_state": null,
    "linked_trace_id": "00000000-0000-0000-0000-000000000012",
    "lk.linked_span_id": 2,
    "lk.dropped_tags_count": 0
  }
}
```

This works without `DISTINCT` because jsonb doesn't support duplicate keys like json and it always keep the last value, in our case `link_nbr` is unique per `trace_id` and `span_id` as stated before:

> Since `lk.link_nbr` is the position on the array of links of the span, there's an implicit unique constraint (trace_id, span_id, link_nbr), since we are already matching against the first 2 in the JOIN, then `lk.link_nbr should be unique per trace_id, span_id pair. 

If we go this route then some changes in the code that deserialises the values should be added to handle the new format. 

Here's a full example with both approaches. 

```sql
promscale=# 
SELECT
  s.trace_id,
  s.span_id,
  _ps_trace.tag_map_denormalize(s.resource_tags)        resource_tags,
  _ps_trace.tag_map_denormalize(s.span_tags)            span_tags,
  array_agg(e.name ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL)                  event_names,
  array_agg(_ps_trace.tag_map_denormalize(e.tags) ORDER BY e.event_nbr) FILTER(WHERE e IS DISTINCT FROM NULL)   event_tags,
  array_agg(lk.linked_trace_id ORDER BY lk.link_nbr) FILTER(WHERE lk IS DISTINCT FROM NULL)     links_linked_trace_ids,
  json_object_agg(DISTINCT lk.link_nbr, lk.linked_trace_id) json_links_linked_trace_ids,
  jsonb_object_agg(lk.link_nbr, json_build_object('linked_trace_id',lk.trace_id,'lk.linked_span_id',lk.linked_span_id,'lk.trace_state',lk.trace_state,'lk.dropped_tags_count',lk.dropped_tags_count,'tags',_ps_trace.tag_map_denormalize(lk.tags))) aggregated_links
FROM
  _ps_trace.span s
LEFT JOIN
  _ps_trace.event e ON e.span_id = s.span_id AND e.trace_id = s.trace_id
LEFT JOIN
  _ps_trace.link lk ON lk.trace_id = s.trace_id AND lk.span_id = s.span_id
GROUP BY
  s.trace_id,
  s.span_id,
  s.resource_tags,
  s.span_tags
promscale-# ;
─[ RECORD 1 ]───────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
trace_id                    │ 00000000-0000-0000-0000-000000000012
span_id                     │ 4
resource_tags               │ {"service.name": "query12-service"}
span_tags                   │ {"blob": "data:application/octet-stream; base64,AAAwOQ==", "sameplacetag1": "sameplacevalue", "sameplacetag2": 123, "sameplacetag3": 72.5, "sameplacetag4": true}
event_names                 │ {"","","","","","","",""}
event_tags                  │ {"{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}","{\"log-sameplacetag1\": \"sameplacevalue\", \"log-sameplacetag2\": 123, \"log-sameplacetag4\": true}",NULL,NULL,NULL,NULL}
links_linked_trace_ids      │ {00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-0000000000ff,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001,00000000-0000-0000-0000-000000000001}
json_links_linked_trace_ids │ { "0" : "00000000-0000-0000-0000-0000000000ff", "1" : "00000000-0000-0000-0000-000000000001", "2" : "00000000-0000-0000-0000-000000000001" }
aggregated_links            │ {"0": {"tags": null, "lk.trace_state": null, "linked_trace_id": "00000000-0000-0000-0000-000000000012", "lk.linked_span_id": 255, "lk.dropped_tags_count": 0}, "1": {"tags": null, "lk.trace_state": null, "linked_trace_id": "00000000-0000-0000-0000-000000000012", "lk.linked_span_id": 2, "lk.dropped_tags_count": 0}, "2": {"tags": null, "lk.trace_state": null, "linked_trace_id": "00000000-0000-0000-0000-000000000012", "lk.linked_span_id": 2, "lk.dropped_tags_count": 0}}

Time: 6.856 ms
```

DISCLAIMER: I'm not really sure about the implications of these approaches, I'll work on getting feedback and setting up a database with significant data to test against.

**To Reproduce**

Execute the query that fetches traces/spans with a span that has a `trace_state=NULL`

**Expected behavior**

The values for links and events should be returned and there shouldn't be any duplicates.

**Screenshots**

**Configuration (as applicable)**
 - Promscale Connector:
 - TimescaleDB:
 - Prometheus:
 - Jaeger:
 - OpenTelemetry:

**Version**
 - Distribution/OS: 
 - Promscale: 
 - TimescaleDB:

**Additional context**
<Add any other context about the problem here.>

alejandrodnm:
Solves 2 issues with the query that's executed for retrieving traces:

- Empty results from aggregating data from `link` table.
- Duplicated data on events when the cartesian product of the tables `link` x `event` was greater than one when joining to the `span` table.

### Empty results:

The first issue occurs when working with the native Jaeger API because `trace_state` in these cases is always null. The query uses select statements similar to:

```sql
array_agg(
  lk.<column> ORDER BY lk.link_nbr
) FILTER (
  WHERE lk IS NOT NULL
)
```

With `lk` defined as:

```sql
LEFT JOIN _ps_trace.link lk ON lk.trace_id = s.trace_id
  AND lk.span_id = s.span_id
  AND lk.span_start_time < X`
```

The problem is the statement `lk IS NOT NULL`, from the PostgreSQL docs for `IS NOT NULL`:

https://www.postgresql.org/docs/9.2/functions-comparison.html

> If the expression is row-valued, then IS NULL is true when the row expression itself is null or when all the row's fields are null, while IS NOT NULL is true when the row expression itself is non-null and all the row's fields are non-null. Because of this behavior, IS NULL and IS NOT NULL do not always return inverse results for row-valued expressions; in particular, a row-valued expression that contains both null and non-null fields will return false for both tests. In some cases, it may be preferable to write row IS DISTINCT FROM NULL or row IS NOT DISTINCT FROM NULL, which will simply check whether the overall row value is null without any additional tests on the row fields.

As per the docs suggestion changing to `lf IS DISTINCT FROM NULL` solves this issue.

### Duplicated data:

The duplicated data problem is because the query generates a Cartesian product with the rows of 2 tables which are both latter aggregated into a single row.

The `_ps_trace.link` and `_ps_trace.event` tables are LEFT JOINED against the `_ps_trace.span` table matching `trace_id` and `span_id`, then for each of the joined table (`_ps_trace.link` and `_ps_trace.event`) an aggregate for each of their columns is generated using `array_agg`. When we have multiple rows in each table that satisfy the join condition we end up with the Cartesian product of the tables, and thus duplicated values.

Since the problem is only with these 2 tables doing the aggregation for one of them (`_ps_trace.link` was chosen) at JOIN level with by moving all the aggregates into a `LEFT JOIN LATERAL` and using the resulting output in the `GROUP BY` clause of the query avoids the Cartesian product and the duplicated values.

fixes #1664

## Merge requirements


arajkumar (approved):
lgtm, I pass this to @cevian for further review.

niksajakovljevic (approved):

pkg/jaeger/store/trace_query.go
```diff
@@ -57,8 +60,16 @@ const (
 		_ps_trace.instrumentation_lib inst_lib ON s.instrumentation_lib_id = inst_lib.id
 	LEFT JOIN
 		_ps_trace.schema_url inst_lib_url ON inst_lib_url.id = inst_lib.schema_url_id
-	LEFT JOIN
-		_ps_trace.link lk ON lk.trace_id = s.trace_id AND lk.span_id = s.span_id %[4]s
+	LEFT JOIN LATERAL (
```
cevian, line 63: is there a reason we don't use this same method for events? It seems to me that we should be using the same method for links and events if possible.

pkg/jaeger/store/trace_query.go
```diff
@@ -74,7 +85,12 @@ const (
 	  s_url.url,
 	  inst_lib.name,
 	  inst_lib.version,
-	  inst_lib_url.url`
+	  inst_lib_url.url,
+		links_linked_trace_ids,
+		links_linked_span_ids,
+		links_trace_states,
+		links_dropped_tags_count,
+		links_tags`
```
cevian, line 93: Using a whole-row reference looks better and probably performs better too
```suggestion
		link`
```

cevian (changes requested):
Overall looks good but have some questions I want answered before approval

cevian:
Actually I just realized if we use the same method for events we are now using for links, we could get rid of the aggregation at the top level, getting rid of the ugly and not very performant GROUP BY too. That would be good I think

alejandrodnm:
@cevian and I agreed to leave the group by since the tests I did showed that the group by improved the query plan. I attempted a couple of things to remove the group by but they all seemed to result in worst performance. 

Since we are going to be tackling performance for these scenario in Q4 we'll leave this for now

ramonguiu:
@alejandrodnm does this mean we can merge this PR?

pkg/jaeger/store/trace_query.go
```diff
@@ -15,66 +15,93 @@ import (
 )
 
 const (
+	// A lateral join is used for the `link` table instead of using directly
```
cevian, line 18: can we please add a note about why we need a group by to the comment. It's a bit counter-intuitive so would be good to document

cevian (approved):

pkg/jaeger/store/trace_query.go
alejandrodnm: I had automerge already enabled, I'll add it in a follow up PR

PR #1679, opened 2022-09-29
Reuse ingestion temp tables across batches

jgpruitt:
## Description

Currently, the temp tables used for ingestion are named using a transient value. This PR substitutes the transient value for the metric_id. This change creates a temp table specific to the metric and database session, allowing it to be reused.

The temp table is currently created with ON COMMIT DROP, which requires that the temp table be recreated with each transaction. This incurs significant pg_catalog churn/bloat. A corresponding PR will be made in the extension to change this to ON COMMIT DELETE ROWS.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ We use the following categories for changes:
 - TimescaleDB is now mandatory [#1660].
 - When querying for Jaeger tags with binary values the binary data will be
   returned instead of the base64 representation of the string [#1649].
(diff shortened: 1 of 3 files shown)


jgpruitt:
corresponding extension PR https://github.com/timescale/promscale_extension/pull/526

alejandrodnm (approved):

JamesGuthrie (approved):

sumerman (approved):

PR #1681, opened 2022-10-03
Preserve Jaeger RefType when storing/retrieving traces.

alejandrodnm:
**NOTE**: This PR is based on https://github.com/timescale/promscale/pull/1678 . I'll update the base once #1678 is merged

Jaeger, like OpenTracing, identifies the type of references between Spans. There are only two types of references `CHILD_OF` and `FOLLOWS_FROM`.

OTEL doesn't have an analogous to Jaeger reference type in their link specification. An OTEL span can have only one parent, and that's set as an attributed of the span instead of an OTEL link.

When using the Jaeger to OTEL translator the first reference with type `CHILD_OF` is used to set the OTEL span parent attribute, all other references are converted to links, and since there is no notion of reference type that information is lost. On the reverse transformation, OTEL to Jaeger, the OTEL Span parent is converted to a reference with `CHILD_OF` type, while all the other OTEL links are converted to `FOLLOWS_FROM` references.

The problem is that Jaeger, unlike OTEL, supports the notion of multiple parents (one use case is fork/join workloads), running this multi-parent through the translator and back returns a span with a single parent with all the other parents turned into `FOLLOWS_FROM` references.

There's an open PR in the translator to keep this information by adding the type as attribute to the links following the semantic convention for OpenTracing compatibility:

https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/compatibility/#OpenTracing https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14463

While that gets merge we are going to be manually setting the reference type as an attribute on the links when writing traces. On the retrieving traces side, we'll be using those attributes to set the corresponding reference type when constructing the Jaeger traces responses.

## Merge requirements


pkg/jaeger/store/translation.go
```diff
@@ -79,3 +82,156 @@ func jSpanKindToInternal(spanKind string) string {
...
+// be the OTEL representation of the given `span.References`.
+//
+// The added attributes follow the OpenTracing to OTEL semantic convention
+// https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/compatibility/#opentracing
+func addRefTypeAttributeToLinks(span *model.Span, links ptrace.SpanLinkSlice) {
+
+	// The reference to the parent span is stored directly as an attribute
+	// of the Span and not as a Link.
+	parentsSpanID := span.ParentSpanID()
+	otherParentsSpanIDs := make(map[model.SpanID]struct{})
```
arajkumar, line 127: nit: `otherParentsSpanID`s can be a stack variable
```suggestion
	otherParentsSpanIDs := map[model.SpanID]struct{}
```

pkg/tests/end_to_end_tests/generate_jaeger_response_test.go
```diff
@@ -14,16 +14,12 @@ import (
 	"fmt"
```
arajkumar, line 14: IIUC, all jaeger related tests which we own can be removed once we get jaeger storage integration tests up & running right? I feel most of the tests which we own are redundant. WDYT?

arajkumar (approved):
Looks super clean :)

pkg/jaeger/store/translation.go
alejandrodnm: I thought that using make or the literal form where basically the same, except that the later lets you specify some initial values while the former lets you specify an initial size.

pkg/tests/end_to_end_tests/generate_jaeger_response_test.go
alejandrodnm: The other point is that some of the tests we have also test the OTEL ingest pipeline. The thing is that the responses to compare against the fixtures are retrieved using the Jaeger endpoint.

Issue #1607, opened 2022-08-31
Cache hit ratio dashboard panel should use rate


JamesGuthrie:
The `Hit ratio (metrics)` panel is driven by the following query:

```
promscale_cache_query_hits_total{type="metric"} / promscale_cache_queries_total{type="metric"}
```

This counts the total number of hits compared to the total queries issued against the cache. This means that the more cache queries are performed, the less "visible" cache misses become.

For example, consult the following graphs:

<img width="705" alt="image" src="https://user-images.githubusercontent.com/1117749/187676502-701f66b5-d778-4758-a739-c95020176d1f.png">

The top graph shows:

```
rate(promscale_cache_query_hits_total{type="metric"}[$__rate_interval]) / rate(promscale_cache_queries_total{type="metric"}[$__rate_interval])
```

Here it is immediately obvious that the cache hit ratio dropped considerably.

The bottom graph shows Promscale's `Hit ratio (metrics)` panel. It is possible to see that _something_ changed, but it's not possible to see the extent of the change in hit ratio.

One wrinkle to using rate, which I have glossed over here, is that not all caches are constantly in use. The following graphs show the difference between using rate (top) and not (bottom). Note the gaps, where the cache is not in use.

<img width="705" alt="image" src="https://user-images.githubusercontent.com/1117749/187677113-a5b79d50-e51a-43d7-954f-8c8c6689fff3.png">

Harkishen-Singh:
I agree, this should have been `rate` Thank you for pointing this out.

PR #1683, opened 2022-10-04
Promscale dashboard: Use rate in `Cache hit ratio` panel

Harkishen-Singh:

Fixes: #1607 

Screenshot 2022-10-04 at 2 46 04 PM

TODO:

- [ ] Update changelog after `0.15.0` is released

## Merge requirements

### DIFF
--- a/docs/mixin/dashboards/promscale.json
+++ b/docs/mixin/dashboards/promscale.json
@@ -2887,7 +2887,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "exemplar": true,
-              "expr": "promscale_cache_query_hits_total{type=\"metric\"} / promscale_cache_queries_total{type=\"metric\"}",
+              "expr": "rate(promscale_cache_query_hits_total{type=\"metric\"}[$__rate_interval]) / rate(promscale_cache_queries_total{type=\"metric\"}[$__rate_interval])",
               "interval": "",
               "legendFormat": "{{ name }}",
               "refId": "A"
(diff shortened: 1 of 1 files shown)


docs/mixin/dashboards/promscale.json
```diff
@@ -2887,7 +2887,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "exemplar": true,
-              "expr": "promscale_cache_query_hits_total{type=\"metric\"} / promscale_cache_queries_total{type=\"metric\"}",
+              "expr": "rate(promscale_cache_query_hits_total{type=\"metric\"}[$__rate_interval]) / rate(promscale_cache_queries_total{type=\"metric\"}[$__rate_interval])",
```
arajkumar, line 2890: Can we use irate for graphs instead of rate?

https://www.robustperception.io/irate-graphs-are-better-graphs/

Harkishen-Singh: We could, but irate is very sensitive to instant spikes. I am not confident that we need such instant spikes for cache hits.

Another reason being, see the graphs, `irate` (top) vs `rate` (bottom) at the drop just after 12:55 (that's where i added more series to replicate a cache miss)

Screenshot 2022-10-06 at 1 01 34 PM

Screenshot 2022-10-06 at 1 01 25 PM

As evident, `irate` is so sensitive, that it just cared about the 1 step evaluation of range query, and just showed a drop denoted by `1 sample`. On the contrary, the hits went down was justified by `rate`, since the averaging logic reflects the drop in _samples evaluated (for `rate`) after the actual drop_.

Given this behaviour, I think we should continue with `rate`.

arajkumar (approved):
lgtm

Harkishen-Singh:
Merging blocked until `0.15.0` is released.

JamesGuthrie (approved):
LGTM, thanks!

paulfantom:
Merging this as I don't see why this cannot be merged now.

PR #1701, opened 2022-10-17
Run Jaeger storage integration tests in separate GH actions job

arajkumar:
Prior to this PR Jaeger storage tests used to run as part of promscale e2e job.

Now it has been modified to run Jaeger tests in a dedicated job for a better visibility and easy auditing.

## Merge requirements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,16 @@ LOCAL_DOCKER_BASE := local/dev_promscale_extension
 GHCR_DOCKER_BASE := ghcr.io/timescale/dev_promscale_extension
 MDOX_BIN=mdox
 MDOX_VALIDATE_CONFIG=
+CURRENT_BRANCH?=$(shell git branch --show-current | sed 's#/#-#')
+EXTENSION_VERSION=$(shell cat EXTENSION_VERSION | tr -d '[:space:]')
+
 
 .PHONY: build
 build: generate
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -41,6 +41,13 @@ upgrade-test: DOCKER_IMAGE?=$(shell ./scripts/fallback-docker.sh $(LOCAL_DOCKER_
 upgrade-test:
 	go test -v ./pkg/tests/upgrade_tests/ -timescale-docker-image=$(DOCKER_IMAGE)
 
+.PHONY: jaeger-storage-test
+jaeger-storage-test: CURRENT_BRANCH?=$(shell git branch --show-current | sed 's#/#-#')
```
niksajakovljevic, line 45: It would be nice to reuse target variables since they are repeated among different targets. Couldn't we just define them on top?

pkg/tests/end_to_end_tests/jaeger_store_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build jaeger_storage_test
```
niksajakovljevic, line 1: It does seem that we will run Jaeger tests two times. (1) when we run e2e tests (2) when we run only Jaeger storage tests with verbose?

niksajakovljevic (approved):

alejandrodnm (approved):

PR #1702, opened 2022-10-17
Run multiple tracegen workers

jgpruitt:
## Description

tracegen.py now runs 4 worker processes generating traces concurrently. Also, fixes an issue with inserts into _ps_trace.span which were missing the duration_ms column.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ We use the following categories for changes:
 ### Added
 
 ### Changed
+- tracegen.py now runs 4 worker processes generating traces concurrently. [#1702]
 
 ### Fixed
 - Fix queries returning no references/links when querying traces that were
--- a/misc/tracegen.py
+++ b/misc/tracegen.py
@@ -7,12 +7,13 @@ import string
 import os
 import json
 import psycopg2
+from multiprocessing import Process
 
-
-MIN_DEPTH = 2
-MAX_DEPTH = 5
-MIN_BREADTH = 1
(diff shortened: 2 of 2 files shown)


misc/tracegen.py
```diff
@@ -7,12 +7,13 @@
 import os
```
arajkumar, line 7: Should we just kill this in favour of https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/tracegen?

jgpruitt: It's nice to be able to load up a database without having to spin up both the open telemetry collector and the promscale connector. I like having this option for the sake of simplicity, and at this point, there's very little upkeep to it.  I don't see much benefit in killing it unless it becomes a big burden.

arajkumar: You should able to tracegen without otel collector, but promscale is needed.
```
tracegen -otlp-endpoint localhost:9202 -otlp-insecure -rate 10000 -duration 10s
```

Since this tool bypasses promscale, I would recommend to move it to extension repo.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use the following categories for changes:
 ### Added
 
 ### Changed
+- tracegen.py now runs 4 worker processes generating traces concurrently. [#1702]
```
arajkumar, line 21: IIUC, we usually add CHANGELOG entry only if the change is user facing.

PR #1703, opened 2022-10-18
Adds support to create/delete metric rollups via dataset-config

Harkishen-Singh:

In relation with https://github.com/timescale/promscale_extension/pull/538 & https://github.com/timescale/o11y-team-applications/issues/228

## Description

This commit adds support for creation or deletion of metric rollups by reading
the dataset-config file. The dataset config file contains a new field `downsample:`
which supports 2 options: `automatic` & `resolution`.

If `automatic` is false, then connector updates this setting in the database (in the `default` table) so that `_prom_catalog.scan_for_new_rollups` does not create any new rollups (Caggs).

`resolution` is a comma separated list of resolutions for downsampling, in a way that
`label:resolution:retention` is the format that is needed.
Example: `short:5m:90d,long:1h:365d`
-> The resolution works in a strict manner. Its aim is to make sure that the database contains
exactly those resolutions that are mentioned in the dataset config. Like, if the database
already contains `short` and `long` downsampled data and the dataset is supplied with `short:5m:90d,very_long:1d:365d`
then `long` downsampled case is deleted and `very_long` is created.
-> `resolution` is applied if `automatic` is `true`. If automatic is not given in dataset config, it defaults to `true`
based on our plan to keep downsampling applied as a default setting.
## Merge requirements

### DIFF
--- a/pkg/dataset/config.go
+++ b/pkg/dataset/config.go
@@ -12,8 +12,8 @@ import (
 	"github.com/jackc/pgx/v4"
 	"gopkg.in/yaml.v2"
 
+	"github.com/timescale/promscale/pkg/internal/day"
 	"github.com/timescale/promscale/pkg/log"
-	"github.com/timescale/promscale/pkg/util"
 )
 
 const (
@@ -44,17 +44,17 @@ type Config struct {
 
 // Metrics contains dataset configuration options for metrics data.
 type Metrics struct {
-	ChunkInterval   util.DayDuration `yaml:"default_chunk_interval"`
-	Compression     *bool            `yaml:"compress_data"` // Using pointer to check if the value was set.
-	HALeaseRefresh  util.DayDuration `yaml:"ha_lease_refresh"`
-	HALeaseTimeout  util.DayDuration `yaml:"ha_lease_timeout"`
-	RetentionPeriod util.DayDuration `yaml:"default_retention_period"`
-	Downsample      *Downsample      `yaml:"downsample,omitempty"`
+	ChunkInterval   day.Duration `yaml:"default_chunk_interval"`
+	Compression     *bool        `yaml:"compress_data"` // Using pointer to check if the value was set.
+	HALeaseRefresh  day.Duration `yaml:"ha_lease_refresh"`
+	HALeaseTimeout  day.Duration `yaml:"ha_lease_timeout"`
+	RetentionPeriod day.Duration `yaml:"default_retention_period"`
+	Downsample      *Downsample  `yaml:"downsample,omitempty"`
 }
 
 // Traces contains dataset configuration options for traces data.
 type Traces struct {
-	RetentionPeriod util.DayDuration `yaml:"default_retention_period"`
+	RetentionPeriod day.Duration `yaml:"default_retention_period"`
 }
 
 // NewConfig creates a new dataset config based on the configuration YAML contents.
@@ -100,21 +100,21 @@ func (c *Config) Apply(conn *pgx.Conn) error {
 
 func (c *Config) applyDefaults() {
 	if c.Metrics.ChunkInterval <= 0 {
-		c.Metrics.ChunkInterval = util.DayDuration(defaultMetricChunkInterval)
+		c.Metrics.ChunkInterval = day.Duration(defaultMetricChunkInterval)
 	}
 	if c.Metrics.Compression == nil {
 		c.Metrics.Compression = &defaultMetricCompressionVar
 	}
 	if c.Metrics.HALeaseRefresh <= 0 {
-		c.Metrics.HALeaseRefresh = util.DayDuration(defaultMetricHALeaseRefresh)
+		c.Metrics.HALeaseRefresh = day.Duration(defaultMetricHALeaseRefresh)
 	}
 	if c.Metrics.HALeaseTimeout <= 0 {
-		c.Metrics.HALeaseTimeout = util.DayDuration(defaultMetricHALeaseTimeout)
+		c.Metrics.HALeaseTimeout = day.Duration(defaultMetricHALeaseTimeout)
 	}
 	if c.Metrics.RetentionPeriod <= 0 {
-		c.Metrics.RetentionPeriod = util.DayDuration(defaultMetricRetentionPeriod)
+		c.Metrics.RetentionPeriod = day.Duration(defaultMetricRetentionPeriod)
 	}
 	if c.Traces.RetentionPeriod <= 0 {
-		c.Traces.RetentionPeriod = util.DayDuration(defaultTraceRetentionPeriod)
+		c.Traces.RetentionPeriod = day.Duration(defaultTraceRetentionPeriod)
 	}
 }
--- a/pkg/dataset/config.go
+++ b/pkg/dataset/config.go
@@ -49,7 +49,7 @@ type Metrics struct {
 	HALeaseRefresh  util.DayDuration `yaml:"ha_lease_refresh"`
 	HALeaseTimeout  util.DayDuration `yaml:"ha_lease_timeout"`
 	RetentionPeriod util.DayDuration `yaml:"default_retention_period"`
-	Downsample      `yaml:"downsample,omitempty"`
+	Downsample      *Downsample      `yaml:"downsample,omitempty"`
 }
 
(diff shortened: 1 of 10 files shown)


pkg/dataset/config.go
```diff
@@ -38,16 +44,17 @@ type Config struct {
...
-	Compression     *bool       `yaml:"compress_data"` // Using pointer to check if the the value was set.
-	HALeaseRefresh  DayDuration `yaml:"ha_lease_refresh"`
-	HALeaseTimeout  DayDuration `yaml:"ha_lease_timeout"`
-	RetentionPeriod DayDuration `yaml:"default_retention_period"`
+	ChunkInterval   util.DayDuration `yaml:"default_chunk_interval"`
+	Compression     *bool            `yaml:"compress_data"` // Using pointer to check if the value was set.
+	HALeaseRefresh  util.DayDuration `yaml:"ha_lease_refresh"`
+	HALeaseTimeout  util.DayDuration `yaml:"ha_lease_timeout"`
+	RetentionPeriod util.DayDuration `yaml:"default_retention_period"`
+	Downsample      `yaml:"downsample,omitempty"`
```
niksajakovljevic, line 52: Why don't we add a field here? It will add more structure thus easier to work with `Metrics` struct

pkg/dataset/downsample.go
```diff
@@ -0,0 +1,76 @@
...
+var (
+	setDefaultDownsampleStateSQL      = "SELECT prom_api.set_automatic_downsample($1)"
+	setDefaultDownsampleResolutionSQL = "SELECT prom_api.set_downsample_resolution($1)"
+
+	defaultDownsampleStateVar = defaultDownsampleState
+)
+
+type Downsample struct {
+	Automatic  *bool  `yaml:"automatic,omitempty"`
+	Resolution string `yaml:"resolution,omitempty"`
```
niksajakovljevic, line 31: I think having a list of resolutions would be easier to navigate. Strings are tedious to parse/understand and more error prone.
```
type Resolution struct {
   Label string
   Interval time.Duration
   Retention time.Duration
}
```

pkg/dataset/downsample.go
```diff
@@ -0,0 +1,76 @@
...
+	for sql, param := range queries {
+		if _, err := conn.Exec(context.Background(), sql, param); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *Downsample) applyDefaults() {
+	if d.Automatic == nil {
```
niksajakovljevic, line 58: Do we validate `automatic: false` and no resolution specified?

pkg/rollup/rollup.go
```diff
@@ -0,0 +1,133 @@
...
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/jackc/pgx/v4"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+type DownsampleResolution struct {
```
niksajakovljevic, line 17: Ok I see you have defined the struct. I'd recommend using it in config struct instead of custom string format.

pkg/rollup/rollup.go
```diff
@@ -0,0 +1,133 @@
...
+	label      string
+	resolution util.DayDuration
+	retention  util.DayDuration
+}
+
+// Parse parses the rollup resolution string and returns true if it respects the format of downsampling resolution,
+// which is a string of comma separated label:resolution:retention.
+// Example: short:5m:90d,long:1h:395d
+func Parse(resolutionStr string) ([]DownsampleResolution, error) {
```
niksajakovljevic, line 26: This code could go away if we use defined structs in the Downsampling configuration.

niksajakovljevic (changes requested):
I haven't been through all the code yet, but given early feedback so you're not blocked on me. My biggest concern so far is using custom string format for configuration. Maybe there is some reason behind but I'd really like to avoid that if possible. Good work.

pkg/runner/client.go
```diff
@@ -170,6 +170,13 @@ func CreateClient(r prometheus.Registerer, cfg *Config) (*pgclient.Client, error
 		if err != nil {
 			return nil, fmt.Errorf("error applying dataset configuration: %w", err)
 		}
+	} else {
+		// We apply downsampling settings even when DatasetConfig is not given, which is the most common case.
```
niksajakovljevic, line 174: Shouldn't we keep this feature off by default until it's battle tested? Otherwise if there is a bug it will affect all the users who upgrade

pkg/rollup/rollup.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, r := range res {
+		_, err := conn.Exec(context.Background(), "CALL _prom_catalog.delete_metric_rollup($1)", r.label)
+		if err != nil {
+			return fmt.Errorf("error deleting rollup for %s: %w", r.label, err)
+		}
+	}
+	return nil
+}
+
+// diff returns the elements of a that are not in b.
```
niksajakovljevic, line 117: Ok, so here are look for a complement of set B. Can you explain more in comments why this is needed.

pkg/rollup/rollup_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+func TestParse(t *testing.T) {
```
niksajakovljevic, line 15: This would be gone if we are using structs for rollup configuration.

pkg/rollup/rollup.go
niksajakovljevic: I think it would be nice mentioning here why we need this.

pkg/rollup/rollup.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error parsing retention %s: %w", items[2], err)
+		}
+		r = append(r, DownsampleResolution{lName, res, retention})
+	}
+	return r, nil
+}
+
+// EnsureRollupWith ensures "strictly" that the given new resolutions are applied in the database.
+// Note: It follows a "strict" behaviour, meaning any existing resolutions of downsampling in
```
niksajakovljevic, line 61: It would be good to describe that we want to preserve existing downsampling data if possible to avoid recalculations and only create the new items.

niksajakovljevic:
Went through the rest of the code. Beside my previous comments I've noticed there are no e2e tests to test DB interaction. I'd recommend including those. Otherwise looks good (left few nits about improving some comments).

pkg/dataset/downsample.go
Harkishen-Singh: If no resolution is specified (which is the default case), we consider it `5m` & `1h` downsampling case. If a user does not want downsampling, then has to set `automatic: false`.

pkg/runner/client.go
Harkishen-Singh: I agree, let's keep it off by default.

pkg/rollup/rollup.go
Harkishen-Singh: This is not in line with the design. In the design discussion I recall that we wanted this "strict" behaviour.
We can do what you suggested (ideally I am for it), but for now, maybe we follow what's discussed in the design doc, for things to move quickly.

Harkishen-Singh:
This PR will be merged after the merge of https://github.com/timescale/promscale_extension/pull/538

Harkishen-Singh:
Oops, I just saw that the PR is opened against the master, not the feature branch https://github.com/timescale/promscale/tree/feature_metric_rollup. Need to change this.

pkg/util/day_duration.go
```diff
@@ -1,4 +1,8 @@
-package dataset
```
alejandrodnm, line 1: NIT: 

Instead of `util` why don't we move to its own package? You can put it in `internal/dayduration`? Additionally you can change the `DayDuration` to `Duration` because it will be referenced as `dayduration.Duration`. Another option could be `day.Duration`.

I don't like stuffing everything into a `util` or `misc` package. I try to follow https://github.com/golang/go/wiki/CodeReviewComments#package-names

Since we already have one I'd try not to add more things there.

Issue #1389, opened 2022-05-31
SeriesCache does not grow when cache is large


cevian:
When the series cache is large, the math in `(t *SeriesCacheImpl) runSizeCheck(sigClose <-chan struct{})` doesn't work. The current check is whether 20% of the cache was evicted in the last minute. But if the cache is large, getting through 20% may take a long time and yet the cache is still too small. I've encountered this with a cache for 4,000,000  elements but 6,000,000 active series.

We need better logic here. Some options:
1. Grow a cache as soon as full if less than it's max size (may waste a bit of memory)
2. Grow cache if you are evicting elements less than a minute old (cache is LRU so that should work but requires tracking time an element was added to the cache)

cevian:
One more possible solution: grow if 20% of all cache access in the last minute involved eviction (as opposed to 20% of all elements). That's probably the right solution

PR #1705, opened 2022-10-19
Fix series cache growth

niksajakovljevic:
Approach is using timestamps to have better understanding on what elements are getting evicted. 
If we see an eviction of recently used/inserted element (defined with `evictionMaxAge`) we will attempt to grow the cache. 
`evictionMaxAge` defines max age where we are doing evictions and not growing the cache.
We don't use additional memory to track timestamps but re-use existing field by bit packing data into `uint32`.

Closes https://github.com/timescale/promscale/issues/1389

### DIFF
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -27,6 +27,8 @@ type Cache struct {
 	dataSize uint64
 	// number of evictions
 	evictions uint64
+	// max eviction timestamp in unix seconds
+	maxEvictionTs int32
 
 	// guards next, and len(storage) and ensures that at most one eviction
 	// occurs at a time, always grabbed _before_ elementsLock
@@ -40,9 +42,13 @@ type element struct {
 	key   interface{}
 	value interface{}
 
-	// CLOCK marker if this is recently used
-	used uint32
-	size uint64
+	// CLOCK marker if this element is recently used together with timestamp
+	// To avoid memory waste we pack together `used` flag with timestamp into 32 bits
+	// Left most bit represents if element was used (1 or 0). The rest (31 bits) represent timestamp
+	// Timestamp is helping us figure out if we should attempt to grow cache or not
+	// Timestamp is stored in Unix seconds (31 bits should be enough for the next 15 years)
+	usedWithTs uint32
+	size       uint64
 
 	// pad Elements out to be cache-aligned, see BenchmarkCacheFalseSharing
 	_ [16]byte
@@ -100,17 +106,18 @@ func (self *Cache) insert(key interface{}, value interface{}, size uint64) (exis
 	elem, present := self.elements[key]
 	if present {
 		// we'll count a double-insert as a hit. See the comment in get
-		if atomic.LoadUint32(&elem.used) != 0 {
-			atomic.StoreUint32(&elem.used, 1)
+		if !extractUsed(atomic.LoadUint32(&elem.usedWithTs)) {
+			atomic.StoreUint32(&elem.usedWithTs, packUsedAndTimestamp(true, int32(time.Now().Unix())))
 		}
 		return elem, false, true
 	}
 
 	var insertLocation *element
+	newElement := element{key: key, value: value, size: size, usedWithTs: packUsedAndTimestamp(false, int32(time.Now().Unix()))}
 	if len(self.storage) >= cap(self.storage) {
 		insertLocation = self.evict()
 		if insertLocation == nil {
-			return &element{key: key, value: value, size: size}, false, false
+			return &newElement, false, false
 		}
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
@@ -118,11 +125,15 @@ func (self *Cache) insert(key interface{}, value interface{}, size uint64) (exis
 		self.dataSize -= insertLocation.size
 		self.dataSize += size
 		self.evictions++
-		*insertLocation = element{key: key, value: value, size: size}
+		evictionTs := extractTimestamp(insertLocation.usedWithTs)
+		if self.maxEvictionTs < evictionTs {
+			self.maxEvictionTs = evictionTs
+		}
+		*insertLocation = newElement
 	} else {
 		self.elementsLock.Lock()
 		defer self.elementsLock.Unlock()
-		self.storage = append(self.storage, element{key: key, value: value, size: size})
+		self.storage = append(self.storage, newElement)
 		self.dataSize += size
 		insertLocation = &self.storage[len(self.storage)-1]
 	}
@@ -168,8 +179,8 @@ func (self *Cache) evict() (insertPtr *element) {
 	for i := 0; i < 2; i++ {
 		for next := range postStart {
 			elem := &postStart[next]
-			old := atomic.SwapUint32(&elem.used, 0)
-			if old == 0 {
+			old := atomic.SwapUint32(&elem.usedWithTs, setUsed(false, elem.usedWithTs))
+			if !extractUsed(old) {
 				insertPtr = elem
 			}
 
@@ -180,8 +191,8 @@ func (self *Cache) evict() (insertPtr *element) {
 		}
 		for next := range preStart {
 			elem := &preStart[next]
-			old := atomic.SwapUint32(&elem.used, 0)
-			if old == 0 {
+			old := atomic.SwapUint32(&elem.usedWithTs, setUsed(false, elem.usedWithTs))
+			if !extractUsed(old) {
 				insertPtr = elem
 			}
 
@@ -252,8 +263,8 @@ func (self *Cache) get(key interface{}) (interface{}, bool) {
 	// this is a read-only operation, and doesn't trash the cache line that used
 	// is stored on. The lack of atomicity of the update doesn't matter for our
 	// use case.
-	if atomic.LoadUint32(&elem.used) == 0 {
-		atomic.StoreUint32(&elem.used, 1)
+	if !extractUsed(atomic.LoadUint32(&elem.usedWithTs)) {
+		atomic.StoreUint32(&elem.usedWithTs, packUsedAndTimestamp(true, int32(time.Now().Unix())))
 	}
 	self.metrics.Inc(self.metrics.hitsTotal)
 
@@ -274,8 +285,8 @@ func (self *Cache) unmark(key string) bool {
 	// this is a read-only operation, and doesn't trash the cache line that used
 	// is stored on. The lack of atomicity of the update doesn't matter for our
 	// use case.
-	if atomic.LoadUint32(&elem.used) != 0 {
-		atomic.StoreUint32(&elem.used, 0)
+	if extractUsed(atomic.LoadUint32(&elem.usedWithTs)) {
+		atomic.StoreUint32(&elem.usedWithTs, setUsed(false, elem.usedWithTs))
 	}
 
 	return true
@@ -296,9 +307,9 @@ func (self *Cache) ExpandTo(newMax int) {
 	for i := range self.storage {
 		elem := &self.storage[i]
 		newStorage = append(newStorage, element{
-			key:   elem.key,
-			value: elem.value,
-			used:  atomic.LoadUint32(&elem.used),
+			key:        elem.key,
+			value:      elem.value,
+			usedWithTs: atomic.LoadUint32(&elem.usedWithTs),
 		})
 	}
 
@@ -313,6 +324,8 @@ func (self *Cache) ExpandTo(newMax int) {
 
 	self.elements = newElements
 	self.storage = newStorage
+	self.maxEvictionTs = 0
+	self.evictions = 0
 }
 
 func (self *Cache) Reset() {
@@ -328,6 +341,8 @@ func (self *Cache) Reset() {
 	self.elements = newElements
 	self.storage = newStorage
 	self.next = 0
+	self.maxEvictionTs = 0
+	self.evictions = 0
 }
 
 func (self *Cache) Len() int {
@@ -340,6 +355,12 @@ func (self *Cache) Evictions() uint64 {
 	self.elementsLock.RLock()
 	defer self.elementsLock.RUnlock()
 	return self.evictions
+}
+
+func (self *Cache) MaxEvictionTs() int32 {
+	self.elementsLock.RLock()
+	defer self.elementsLock.RUnlock()
+	return self.maxEvictionTs
 }
 
 func (self *Cache) SizeBytes() uint64 {
@@ -365,4 +386,32 @@ func (self *Cache) debugString() string {
 		str = fmt.Sprintf("%s%v: %v, ", str, elem.key, elem.value)
 	}
 	return fmt.Sprintf("%s]", str)
+}
+
+// extractTimestamp skips leftmost bit and converts the rest into timestamp (unix seconds)
+func extractTimestamp(in uint32) int32 {
+	return int32(in & ((1 << 31) - 1))
+}
+
+// extractUsed gets the first bit from the left
+// checking if element has been used
+func extractUsed(in uint32) bool {
+	return 1 == (in >> 31)
+}
+
+// setUsed sets first bit from the left. This bit marks if element has been used
+func setUsed(used bool, in uint32) uint32 {
+	if used {
+		return in | uint32(1<<31)
+	}
+	return in &^ uint32(1<<31)
+}
+
+// packUsedAndTimestamp packs bool flag and timestamp into 32 bits
+func packUsedAndTimestamp(used bool, ts int32) uint32 {
+	var usedBit int
+	if used {
+		usedBit = 1
+	}
+	return uint32(usedBit<<31) | uint32(ts)
 }
--- a/pkg/pgmodel/cache/series_cache_test.go
+++ b/pkg/pgmodel/cache/series_cache_test.go
@@ -6,11 +6,13 @@ package cache
 
 import (
 	"bytes"
-	"github.com/stretchr/testify/require"
-	"github.com/timescale/promscale/pkg/prompb"
 	"math"
 	"strings"
 	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"github.com/timescale/promscale/pkg/prompb"
 
 	promLabels "github.com/prometheus/prometheus/model/labels"
 )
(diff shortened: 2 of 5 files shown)


pkg/clockcache/cache.go
```diff
@@ -40,9 +42,13 @@ type element struct {
 	key   interface{}
 	value interface{}
 
-	// CLOCK marker if this is recently used
-	used uint32
-	size uint64
+	// CLOCK marker if this element is recently used together with timestamp
+	// To avoid memory waist we pack together `used` flag with timestamp into 32 bits
```
alejandrodnm, line 46:
```suggestion
	// To avoid memory waste we pack together `used` flag with timestamp into 32 bits
```

pkg/clockcache/cache.go
```diff
@@ -40,9 +42,13 @@ type element struct {
...
+	// Left most bit represents if element was used (1 or 0). The rest (31 bits) represent timestamp
+	// Timestamp is helping us figure out if we should attempt to grow cache or not
+	// Timestamp is stored in Unix seconds (31 bits should be enough for the next 15 years)
+	usedWithTs uint32
```
alejandrodnm, line 50: Can we take from the `_ [16]bytes` that we are using for padding and store the TS as a separate int64?

pkg/pgmodel/cache/series_cache_test.go
```diff
@@ -52,3 +55,43 @@ func TestGenerateKey(t *testing.T) {
...
+			cacheGrowCnt: 1,
+		},
+		{
+			name:         "Growth criteria not satisfied - we can keep evicting old elements",
+			sleep:        time.Second * 2,
+			cacheGrowCnt: 0,
+		},
+	}
+
+	os.Setenv("IS_TEST", "true")
```
alejandrodnm, line 77:
```suggestion
	t.Setenv("IS_TEST", "true")
```

alejandrodnm: It's better to use `t.Setenv` because it cleans the envvar after it finishes while `os.Setenv` will leave the variable set for subsequent tests.

pkg/pgmodel/cache/series_cache_test.go
```diff
@@ -52,3 +55,43 @@ func TestGenerateKey(t *testing.T) {
...
+	evictionMaxAge = time.Second // tweaking it to not wait too long
+	for _, tc := range testCases {
```
alejandrodnm, line 79: Can you change it to use `t.Run`, that way we get better reporting.

pkg/pgmodel/cache/series_cache_test.go
```diff
@@ -52,3 +55,43 @@ func TestGenerateKey(t *testing.T) {
...
+				time.Sleep(tc.sleep)
+			}
+			if cache.shouldGrow() {
+				cache.grow()
+				cacheGrowCounter++
+			}
+		}
+		if cacheGrowCounter != tc.cacheGrowCnt {
+			t.Errorf("series cache unexpectedly grow. grow counter: %d", cacheGrowCounter)
+		}
```
alejandrodnm, line 94: NIT:

```suggestion
		assert.Equal(t, tc.cacheGrowCnt, cacheGrowCounter)
```

Well, we are just importing the require package so we need to add also assert for this.

pkg/pgmodel/cache/series_cache_test.go
```diff
@@ -52,3 +55,43 @@ func TestGenerateKey(t *testing.T) {
...
+
+	os.Setenv("IS_TEST", "true")
+	evictionMaxAge = time.Second // tweaking it to not wait too long
+	for _, tc := range testCases {
+		cache := NewSeriesCache(Config{SeriesCacheInitialSize: 100, SeriesCacheMemoryMaxBytes: DefaultConfig.SeriesCacheMemoryMaxBytes}, nil)
+		cacheGrowCounter := 0
+		for i := 0; i < 200; i++ {
+			cache.cache.Insert(i, i, 1)
+			if i%100 == 0 {
+				time.Sleep(tc.sleep)
```
alejandrodnm, line 85: Another option that wouldn't require sleep is to add an attribute to the cache that would be a time provider, something like:

```
type timeProvider interface {
  now() int64 
```

That way you can control the clock for the tests. But that will require more changes.

pkg/clockcache/cache.go
niksajakovljevic: I don't think that's possible. If I am not mistaken `[16]bytes` are there to guarantee alignment of 16 bytes.

pkg/pgmodel/cache/series_cache_test.go
niksajakovljevic: Sure. I just don't feel it's worth doing it

pkg/clockcache/cache.go
niksajakovljevic: So padding was added to prevent cache-line bouncing that's (potentially) happening on concurrent access. I've just run local benchmark of `BenchmarkCacheFalseSharing` and there was no difference with and without padding. 😕  I think we should consider removing [16] bytes of padding if it provides no CPU gain since it clearly waists memory... I'd leave that out of the scope of this PR but happy to revisit it.

alejandrodnm: Fine by me

alejandrodnm (approved):

pkg/clockcache/cache.go
JamesGuthrie: I think that it should be possible to "steal" this field from the padding bytes, as @alejandrodnm suggested.

> I've just run local benchmark of BenchmarkCacheFalseSharing and there was no difference with and without padding.

What platform are you on? I noticed that on M1 macs the cache alignment doesn't make a difference, but it does make a difference on x86 platforms.

niksajakovljevic: I was testing on M1 and made no difference although the comment stated that there is about 16% difference on Mac M1?! Ideally we could remove padding and save a lots of memory.

>I noticed that on M1 macs the cache alignment doesn't make a difference, but it does make a difference on x86 platforms.

How much is the difference there?

niksajakovljevic: >I think that it should be possible to "steal" this field from the padding bytes, as @alejandrodnm suggested.

Can you explain your reasoning.  Why would we pad on smaller amount of bytes (I assumed there was a reason we wanted to pad on 16 bytes, otherwise it's padded on 8 bytes b/c of int64?

niksajakovljevic: I've just run bench tests on x86. It seems difference between padding and no padding is statistically insignificant. Not sure how did you get 2X difference 😕 

```
root@my-bench-test:/promscale/pkg/clockcache# cat no-padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkCacheFalseSharing-2   	 2510319	       490.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2503368	       489.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2464582	       494.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2555012	       482.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2532994	       470.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2496974	       466.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2564890	       471.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2497118	       482.9 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2585713	       477.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2540541	       470.8 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	16.974s
```

And with padding:
```
root@my-bench-test:/promscale/pkg/clockcache# cat padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkCacheFalseSharing-2   	 2926026	       408.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2917959	       428.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2709094	       423.8 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2362786	       525.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2888036	       493.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2351191	       507.8 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2457351	       516.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2332888	       498.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2971623	       432.9 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2854483	       515.7 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	19.381s
```
I ran: `go test -bench=BenchmarkCacheFalseSharing -cpu=2 -count=10 -run=^$`
I was running tests in K8s environment with 4 CPUs. However I would be really surprised if we would get 2X on bare metal - I can verify it once I get my hands on the old laptop.
I did repeat my runs couple of time without major difference.

JamesGuthrie: Hmm... I'm obviously misremembering that 2x.

I just tested this out on some AWS VMs as well. Maybe the 2x that I had in mind was the difference between how long each operation took on Intel vs. M1 (~200ns vs ~400ns).

Anyway, here are the results on an AMD x86_64 system:
```
ubuntu@ip-172-31-17-88:~/promscale/pkg/clockcache$ go test -bench=BenchmarkCacheFalseSharing -cpu=2 -count=10 | tee padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: AMD EPYC 7R13 Processor
BenchmarkCacheFalseSharing-2   	 5070142	       240.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5058939	       241.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5113359	       224.2 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5607207	       238.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4946768	       219.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5392538	       240.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4942778	       226.3 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5245922	       241.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5132740	       217.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 5445062	       236.0 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	20.516s

ubuntu@ip-172-31-17-88:~/promscale/pkg/clockcache$ go test -bench=BenchmarkCacheFalseSharing -cpu=2 -count=10 | tee no-padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: AMD EPYC 7R13 Processor
BenchmarkCacheFalseSharing-2   	 4700396	       251.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4702579	       252.2 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4681315	       253.8 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4787164	       253.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4723281	       253.2 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4739452	       252.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4748074	       254.3 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4760126	       255.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4706127	       253.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 4706882	       254.5 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	20.693s

ubuntu@ip-172-31-17-88:~/promscale/pkg/clockcache$ ~/go/bin/benchstat no-padding.txt padding.txt
name                 old time/op    new time/op    delta
CacheFalseSharing-2     253ns ± 1%     233ns ± 6%  -8.21%  (p=0.000 n=10+10)

name                 old alloc/op   new alloc/op   delta
CacheFalseSharing-2     15.0B ± 0%     15.0B ± 0%    ~     (all equal)

name                 old allocs/op  new allocs/op  delta
CacheFalseSharing-2      1.00 ± 0%      1.00 ± 0%    ~     (all equal)
```

And on an Intel x86_64 system:

```
ubuntu@ip-172-31-24-242:~/promscale/pkg/clockcache$ go test -bench=BenchmarkCacheFalseSharing -cpu=2 -count=10 | tee padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkCacheFalseSharing-2   	 2494210	       411.9 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2985262	       492.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 3059500	       408.8 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2997885	       412.2 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2947947	       423.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 3034550	       413.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 3029181	       414.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 3026805	       433.5 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 3033912	       442.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2942834	       421.6 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	34.228s

ubuntu@ip-172-31-24-242:~/promscale/pkg/clockcache$ go test -bench=BenchmarkCacheFalseSharing -cpu=2 -count=10 | tee no-padding.txt
goos: linux
goarch: amd64
pkg: github.com/timescale/promscale/pkg/clockcache
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkCacheFalseSharing-2   	ben 2200164	       473.0 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2439961	       453.4 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2496398	       496.9 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2463030	       488.6 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2362767	       491.9 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2445717	       497.1 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2410858	       488.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2452435	       527.7 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2280991	       492.3 ns/op	      15 B/op	       1 allocs/op
BenchmarkCacheFalseSharing-2   	 2286775	       475.9 ns/op	      15 B/op	       1 allocs/op
PASS
ok  	github.com/timescale/promscale/pkg/clockcache	26.548s

ubuntu@ip-172-31-24-242:~/promscale/pkg/clockcache$ ~/go/bin/benchstat no-padding.txt padding.txt
name                 old time/op    new time/op    delta
CacheFalseSharing-2     489ns ± 8%     420ns ± 5%  -14.00%  (p=0.000 n=10+9)

name                 old alloc/op   new alloc/op   delta
CacheFalseSharing-2     15.0B ± 0%     15.0B ± 0%     ~     (all equal)

name                 old allocs/op  new allocs/op  delta
CacheFalseSharing-2      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

Note that the following diff between the `padding` and `no-padding` runs:
```diff
ubuntu@ip-172-31-24-242:~/promscale/pkg/clockcache$ git diff
diff --git a/pkg/clockcache/cache.go b/pkg/clockcache/cache.go
index 43448ca7..f57974c5 100644
--- a/pkg/clockcache/cache.go
+++ b/pkg/clockcache/cache.go
@@ -45,7 +45,7 @@ type element struct {
        size uint64

        // pad Elements out to be cache-aligned, see BenchmarkCacheFalseSharing
-       _ [16]byte
+       //_ [16]byte
 }

 func WithMax(max uint64) *Cache {
diff --git a/pkg/clockcache/cache_test.go b/pkg/clockcache/cache_test.go
index e65f301d..59a99342 100644
--- a/pkg/clockcache/cache_test.go
+++ b/pkg/clockcache/cache_test.go
@@ -314,6 +314,7 @@ func TestReset(t *testing.T) {
 }

 func TestElementCacheAligned(t *testing.T) {
+       t.Skip()
        elementSize := unsafe.Sizeof(element{})
        if elementSize%64 != 0 {
                t.Errorf("unaligned element size: %d", elementSize)
```

Anyway, needless to say there does seem to be a difference. But, is it worth sweating over? Honestly, probably not.

niksajakovljevic: In some of my runs I did see around 5% diff (for padded version), but I feel the same: probably not worth the `cost`. On some average case with around 5M series we end up spending close to 100MB of memory... - only for series cache....
Anyways let's leave this for a separate PR, but I am in favor of removing extra 16 bytes of padding....

PR #1707, opened 2022-10-20
Fix typo (?) in MaintenanceJobRunningTooLong runbook

mikberg:
## Description

When going through this runbook, i found what I assume is a mixup of `_interval` and `_chunks`, since the former would be, of course, an interval, and the latter seems to fit better with the context.

My apologies if I'm wrong, and I really appreciate these well-written runbooks!

## Merge requirements


alejandrodnm (approved):

alejandrodnm:
Thanks for the fix @mikberg you're right that's a typo. Glad you enjoy the runbooks, we plan to add more troubleshooting runbooks soon :)

PR #1718, opened 2022-10-24
Removing a test that is failing upon PGX upgrade and is a duplicate of a test from promscale_extension.

sumerman:
## Description

- A companion PR for https://github.com/timescale/promscale_extension/pull/523
- The removed test are duplicates of 
  - https://github.com/timescale/promscale_extension/blob/30845156be3588100c16b5d48075daa4f35d1a9e/src/aggregates/prom_delta.rs#L132
  - https://github.com/timescale/promscale_extension/blob/30845156be3588100c16b5d48075daa4f35d1a9e/src/aggregates/prom_increase.rs#L110

## Merge requirements


niksajakovljevic (approved):
LGTM. Can we maybe also remove `GapfillIncrease` test? ( I assume that increase function is tested in the extension as well)

PR #1724, opened 2022-10-26
Remove info_view golden file test

jgpruitt:
## Description

The `info_view.sql` golden file test also exists verbatim in the extension repo. It exclusively tests extension code - no Go code. The view tested needs to be changed, and the PR to change it in the extension cannot be merged unless this test is removed from the connector repo. This PR removes the test.

## Merge requirements


arajkumar (approved):
lgtm

PR #1735, opened 2022-11-02
Refactor ingestor flags to be in ingestor package

cevian:
## Description

Cleans up the code a bit

## Merge requirements


arajkumar (approved):

pkg/pgmodel/ingestor/ingestor.go
```diff
@@ -43,33 +31,42 @@ type DBIngestor struct {
 	closed     *atomic.Bool
 }
 
+type Config struct {
+	NumCopiers              int
+	InvertedLabelsCacheSize uint64
+	Flags                   *Flags
```
niksajakovljevic, line 37: Not sure how much I like this. I'd rather keep all Ingestor configuration in one struct. Check my other comment.

pkg/pgmodel/ingestor/flags.go
```diff
@@ -0,0 +1,37 @@
...
+package ingestor
+
+import (
+	"flag"
+	"time"
+
+	"github.com/timescale/promscale/pkg/pgmodel/ingestor/trace"
+)
+
+type Flags struct {
```
niksajakovljevic, line 14: This is not really in line with existing convention. We usually define a config struct that is passed into `ParseFlags` to populate configurable fields.

niksajakovljevic (changes requested):
Great to separate configs so code becomes more clear. I'd only like to follow the existing convention and have one configuration struct for ingestion that is passed into `ParseFlags` (no need for artificial `Flags` struct)

pkg/pgmodel/ingestor/flags.go
cevian: You are right, the naming was off. I renamed this to Config and had the other struct renamed to Parameters. I do not want to combine the CLI-settable fields with the non-cli-settable fields. Thats why I keep two structs...

niksajakovljevic: I did recognize your idea but (1) It's not followed through the rest of the codebase (so this will be a snowflake) (2) The cut is not really clean because `NumCopiers` is configurable but it's not part of the `Config` struct (3) From the code perspective I am not really sure if it matters where the value is coming from (default or user) . Anyhow I'll approve the PR so you are not blocked on me.

niksajakovljevic (approved):

pkg/pgclient/config.go
```diff
@@ -17,13 +17,14 @@ import (
...
 	"github.com/timescale/promscale/pkg/pgmodel/cache"
-	"github.com/timescale/promscale/pkg/pgmodel/ingestor/trace"
+	"github.com/timescale/promscale/pkg/pgmodel/ingestor"
 	"github.com/timescale/promscale/pkg/version"
 )
 
 // Config for the database.
 type Config struct {
 	CacheConfig             cache.Config
+	IngestorFlags           ingestor.Config
```
alejandrodnm, line 27: nit:

```suggestion
	IngestorConfig           ingestor.Config
```

For consistency `Config` is used to refer to almost every other configuration grouping, and we already have the `CacheConfig` above.

alejandrodnm (approved):

PR #1737, opened 2022-11-03
Allow config options in file as yaml mappings

alejandrodnm:
Replaces the `ff` package with a custom implementation that uses Viper to parse the config file and sets the values for the `flag.Flag`s.

By design the ff package doesn't allow flag subsets or nested maps inside the config yaml and all config options must be explicitely defined as `flag.Flag`. This limits the ability to specify more dynamic config options like lists or maps of user defined rules (Ex: metric rollups).

Viper provides the flexibility we require for parsing the config file, and unmarshaling directly into more complex datastructures. To define how to unmarshal a subtree of the configuration into a datastructure via unmarshalRule:

```
withUnmarshalRules(
  []unmarshalRule{{"startup.dataset", &dataset.Config}}
)
```

Given the  the config file contains the subtree:

```
startup:
  dataset:
    metrics:
      ...
    traces:
      ...
```

Viper can access a nested field by passing a `.` delimited path of keys. In this scenario, the subtree `startup.dataset` will be unmarshal into `&dataset.Config` by Viper using the github.com/mitchellh/mapstructure package.

This change also allows us to format our configuration files to use yaml mappings instead of long key names with `.`, thus maintaining backwards compatibility:

```
web.listen-address: localhost:9201
web.auth.password: my-password
web.auth.username: promscale
```

Can be rewritten as:

```
web:
  listen-address: localhost:9201
  auth:
    password: my-password
    username: promscale
```

## Merge requirements


pkg/runner/client.go
```diff
@@ -165,11 +165,13 @@ func CreateClient(r prometheus.Registerer, cfg *Config) (*pgclient.Client, error
 		cfg.APICfg.MultiTenancy = multiTenancy
 	}
 
-	if cfg.DatasetConfig != "" {
+	if (cfg.DatasetCfg != dataset.Config{}) {
+		err = cfg.DatasetCfg.Apply(conn)
+	} else if cfg.DatasetConfig != "" {
```
alejandrodnm, line 170: I'm keeping both `startup.dataset.config` (the old string one) and `startup.dataset` (new one that's a yaml mapping).  We can add a deprecation notice and remove `startup.dataset.config` later.

I think the suffix `config` here is redundant since these are all config options. This gives us an opportunity to fix that and don't  cause a breaking change.

Other options are:

- Change `startup.dataset.config` to accept the yaml mapping directly. 
- Don't add deprecation notice and just remove the old config option `startup.dataset.config`

pkg/dataset/config.go
```diff
@@ -32,22 +32,22 @@ var (
...
+	Compression     *bool       `mapstructure:"compress_data" yaml:"compress_data"` // Using pointer to check if the the value was set.
+	HALeaseRefresh  DayDuration `mapstructure:"ha_lease_refresh" yaml:"ha_lease_refresh"`
+	HALeaseTimeout  DayDuration `mapstructure:"ha_lease_timeout" yaml:"ha_lease_timeout"`
+	RetentionPeriod DayDuration `mapstructure:"default_retention_period" yaml:"default_retention_period"`
 }
 
 // Traces contains dataset configuration options for traces data.
 type Traces struct {
-	RetentionPeriod DayDuration `yaml:"default_retention_period"`
+	RetentionPeriod DayDuration `mapstructure:"default_retention_period" yaml:"default_retention_period"`
```
alejandrodnm, line 50: These can be easily exposed as command line flags, is there a reason why they're not?

pkg/runner/client.go
cevian: This was actually supposed to be startup.dataset_config. another words these options are "dataset configs" but I don't love the name and am up for something else (although I think dataset by itself reads weird)

cevian (approved):
Overall LGTM. I want to ask @antekresic to review as well.

pkg/runner/client.go
antekresic: The only problem I see with keeping both options is that you can use them both at the same time and create confusion for the user on which one is actually being used to configure. We might need to validate that case and notify or maybe even force using just one option at a time.

docs/dataset.md
```diff
@@ -2,28 +2,19 @@
...
-    default_chunk_interval: 6h
-    compress_data: true
-    ha_lease_refresh: 10s
-    ha_lease_timeout: 1m
-    default_retention_period: 90d
-  traces:
-    default_retention_period: 30d
+startup:
+  dataset:
+    metrics:
```
antekresic, line 10: I know its not a part of this PR but could you update the table below to call the section `metrics` instead of `metric`?

pkg/runner/config_parser.go
```diff
@@ -0,0 +1,289 @@
...
+//
+// ```
+// web.listen-address: localhost:9201
+// web:
+//   listen-address: htto://not.going.to.be.used:9201
+// ```
+//
+// Migrating our codebase to take advantage of all the features of Viper will
+// require some extra effort, and might not bring extra benefits since we
+// already consolidate commanline arguments with environment variables. This
```
antekresic, line 95:
```suggestion
// already consolidate commandline arguments with environment variables. This
```

pkg/runner/config_parser.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// First priority: commandline flags (explicit user preference).
+	if err := fSet.Parse(args); err != nil {
+		return fmt.Errorf("error parsing commandline args: %w", err)
+	}
+
+	configFile, err := configFileNameFromFlags(fSet)
+
+	// This error won't be triggered unless the line that defines the config
+	// flag is deleted
```
antekresic, line 110:
```suggestion
	// flag is deleted.
```

pkg/runner/config_parser.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		return err
+	}
+
+	v, err := newViperConfig(configFile)
+
+	if err != nil {
+		// If the file can't be found we don't return the error this is the
+		// same behabiour we had with `ff.WithAllowMissingConfigFile(true)`
```
antekresic, line 119:
```suggestion
		// If the file can't be found, we don't return the error. This is the
		// same behaviour we had with `ff.WithAllowMissingConfigFile(true)`.
```

pkg/runner/config_parser.go
```diff
@@ -0,0 +1,289 @@
...
+	provided := map[string]struct{}{}
+	fSet.Visit(func(f *flag.Flag) {
+		provided[f.Name] = struct{}{}
+	})
+	return provided
+}
+
+// checkUndefinedFlags returns an error if it founds at least one key in the
+// Viper registry that's neither defined in the FlagSet nor is part of an
+// unmarshalRule
```
antekresic, line 191:
```suggestion
// unmarshalRule.
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ We use the following categories for changes:
 
 ### Added
 - Alerts from promscale monitoring mixin are groupped also by namespace label [#1714]
+- Allow config options in file to be set as yaml mappings [#1737]
+- Add `startup.dataset` option in the config file for the dataset
+  configuration. Supersedes `statup.dataset.config` which accepts a string
```
antekresic, line 22:
```suggestion
  configuration. Supersedes `startup.dataset.config` which accepts a string
```

pkg/runner/flags_test.go
```diff
@@ -242,10 +241,97 @@ func TestParseFlagsConfigPrecedence(t *testing.T) {
 			result: func(c Config) Config { return c },
 		},
 		{
-			name:               "Config file only",
-			configFileContents: "web.listen-address: localhost:9201",
+			name: "Config file only",
```
antekresic, line 244: Consider adding a test case for using both `startup.dataset` and `startup.dataset.config` at the same time to make sure we handle it correctly.

antekresic (commented):
Looks good overall, some minor tweak suggestions and a test case to add.

pkg/runner/flags.go
```diff
@@ -325,6 +326,7 @@ func checkForRemovedConfigFlags(fs *flag.FlagSet, args []string) error {
 func handleOldFlags(oldFlag, newFlag string) {
 	if oldFlag != "migrate" {
 		log.Warn("msg", fmt.Sprintf("Update removed flag `%s` to new flag `%s`", oldFlag, newFlag))
+		return
```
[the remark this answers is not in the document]
alejandrodnm: What you are saying is donde by the method `checkForRemovedConfigFlags` which goes through all the old flags, and for each old flag, calls this `handleOldFlags` method that just logs the `Warn` message. 

If we don't return here, then it will go over the `migrate flag` switch case. Since we are in this part of the branch because `oldFlag != "migrate"` it will always print the `default` case.

alejandrodnm:
cc @paulfantom this is the PR to allow a yaml mapping instead of a string for the dataset config

pkg/dataset/config.go
Harkishen-Singh: > These can be exposed as command line flags, is there a reason why they're not?

I think we see dataset config as a place for those flags that are purely related to DB (hence dataset), which is why these are not kept as flags.

pkg/runner/client.go
alejandrodnm: I’ll add warning if both are being used, saying which one has precedence and a deprecation warning for users that have the old one

docs/configuration.md
```diff
@@ -140,15 +144,16 @@ The following subsections cover all CLI flags which promscale supports. You can
...
+| startup.dataset.config                | string  | "" (disabled) | **DEPRECATION NOTICE**: *this flag is going to be removed in a future version, please follow the [upgrade guide](dataset.md#upgrading-from-startupdatasetconfig)*. Dataset configuration in YAML format for Promscale. It is used for setting various dataset configuration like default metric chunk interval. For more information, please consult the following resources: [dataset](dataset.md) |
+| startup.dataset *(config.yaml only)*  |  yaml   | "" (disabled) | Dataset configuration in YAML format for Promscale. It is used for setting various dataset configuration like default metric chunk interval. For more information, please consult the following resources: [dataset](dataset.md)                                                                                                                                                                    |
```
antekresic, line 150: Isn't the type string? This should be set to path to YAML config file, right?

alejandrodnm: No, it's a yaml mapping node. That's why it's only supported in the config.yaml file.

```
startup.dataset:
    metrics:
      default_chunk_interval: 8h  
```

antekresic (approved):

PR #1738, opened 2022-11-03
Move tracegen to extension repo

jgpruitt:
## Description

Removes the tracegen.py utility. It was a simple script to directly insert sample traces in the database and was used during the development of the tracing schema. Since this has nothing to do with the connector, we're moving it to the extension.

## Merge requirements


jgpruitt:
See https://github.com/timescale/promscale_extension/pull/553

PR #1743, opened 2022-11-04
Add DeepCopy method to dataset.Config

alejandrodnm:
## Description

DeepCopy support is added with the deepcopy-gen tool from the k8s.io/code-generator project.

## Merge requirements


paulfantom (approved):

Makefile
```diff
@@ -71,5 +71,15 @@ check-docs:
 	@echo ">> checking formatting and local/remote links"
 	$(MDOX_BIN) fmt --soft-wraps --check -l --links.validate.config-file=$(MDOX_VALIDATE_CONFIG) **/*.md
 
+.PHONY: deepcopy-gen
```
arajkumar, line 74: Should we add this as a dependant of build target?

Makefile
```diff
@@ -71,5 +71,15 @@ check-docs:
...
+deepcopy-gen:
```
arajkumar, line 75:
```suggestion
deepcopy-gen: deepcopy-gen-install
```

alejandrodnm: I thought about it. The thing for me was that every time we run the target it will run the installation, and that's something we don't do for other targets. 

I have zero strong opinions on this, I guess this is something that we will hardly ever touch that we can just add the dependant as you say.

arajkumar (approved):
lgtm

PR #1745, opened 2022-11-08
More maintenance jobs metrics

sumerman:
## Description

Collecting job stats for each maintenance job type and signal.

This is a matching PR for https://github.com/timescale/promscale_extension/pull/555

## Merge requirements


pkg/pgmodel/metrics/database/metrics.go
```diff
@@ -397,6 +397,96 @@ var metrics = []metricQueryWrap{
 			timescaledb_information.jobs jobs
 				on jobs.job_id = stats.job_id
 			where jobs.proc_name = 'execute_maintenance_job' and stats.last_run_status = 'Failed'`,
+	}, {
+		metrics: append(gauges(
```
niksajakovljevic, line 401: Isn't it better to use histograms to represent `duration` so we can see the distribution and compute aggregates? I think overall we always use histograms to track durations

sumerman: Yeah, I remember it wasn't supported, and it isn't https://github.com/timescale/promscale/blob/master/pkg/pgmodel/metrics/database/database.go#L197. 

Would be nice to have in the future, but I don't consider it to be in the scope of this PR.

niksajakovljevic: 1. I guess recording histograms in DB is more work.
2. Not sure I understand what you mean by longer-than-a-day interval? I assume jobs are continuously running in the background and we want to know p95 or avg. For me this is the same analogy as for tracking request duration.

Anyways if it's too much effort to store histograms in DB I am fine with using gauges for now . I just wanted to understand the reasons.

sumerman: 1. It doesn't seem too hard to call `histogram.Observe`, AFAICT there's no need to bake a histogram on the DB side.
2. Jobs run once every 30m at most. Any sort of aggregation would only make sense on sufficiently long time intervals. Otherwise, you'd be looking at p95 or avg of a small number of values.

sumerman: In other words I'll try and play with histograms, but I don't think it should be a blocker.

sumerman: Well, I made them histograms 🎉

niksajakovljevic (commented):
Great for adding histograms! 
I'd only ask for adding e2e tests so we can validate that given query and metrics is working. Checking this e2e test as ref: `database_metrics_test.go`

pkg/pgmodel/metrics/database/metrics.go
niksajakovljevic: Awesome!

pkg/tests/end_to_end_tests/database_metrics_test.go
```diff
@@ -50,7 +52,9 @@ func TestDatabaseMetrics(t *testing.T) {
 		compressionStatus = getMetricValue(t, "compression_status")
 		require.Equal(t, float64(1), compressionStatus)
 		numMaintenanceJobs = getMetricValue(t, "worker_maintenance_job")
-		require.Equal(t, float64(2), numMaintenanceJobs)
+		require.GreaterOrEqual(t, numMaintenanceJobs, float64(1))
+		metricsJobDuration = getMetricValue(t, "worker_maintenance_job_metrics_compression_last_duration_seconds")
```
niksajakovljevic, line 56: Can we check this metric is non zero after running `compress_chunk`. 
 Here https://github.com/timescale/promscale/pull/1745/files#diff-4517fe19b15a219db0b35419034d9c69c1fe5316a8d0a1b9db75579fed381ab6L137

sumerman: But it won't. It's only counting our own jobs and only when they run on schedule. AFAICT `compress_chunk` won't go through it.

niksajakovljevic (approved):

pkg/tests/end_to_end_tests/database_metrics_test.go
niksajakovljevic: I see....If we could do some trick to make it schedule immediatelly? maybe wait a second and check the output it would be great. Anyway if it's too much effort I am fine with this.

sumerman: I've tried, and the obvious way, sadly, doesn't work, I'll leave it as is.

Issue #1754, opened 2022-11-15
Add prometheus metrics for vacuum engine


jgpruitt:
Add prometheus metrics for the vacuum engine so we can more easily observe its behavior

* add a gauge for the number of tables left to vacuum
* add a counter for the number of tables vacuumed

PR #1755, opened 2022-11-15
Add prometheus metrics to vacuum engine

jgpruitt:
## Description

Exposes three new prometheus metrics in the vacuum engine so we can more easily observe its behavior

* adds a gauge for the number of tables left to vacuum
* adds a counter for the number of tables vacuumed
* adds a gauge that roughly tracks the number of database connections currently in use by the vacuum engine

Since we limit our query to 1000 results, the `tables left to vacuum` gauge will never go above 1000 even if there are many more tables to be vacuumed.

 #1754

## Merge requirements

### DIFF
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -8,9 +8,42 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/pgxconn"
+	"github.com/timescale/promscale/pkg/util"
 )
+
+var (
+	tablesVacuumedTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "tables_vacuumed_total",
+		Help:      "Total number of tables vacuumed by the Promscale vacuum engine.",
+	})
+	vacuumErrorsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "vacuum_errors_total",
+		Help:      "Total number of errors encountered by the Promscale vacuum engine while vacuuming tables.",
+	})
+	tablesNeedingVacuum = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "tables_needing_vacuum",
+		Help:      "Number of tables needing a vacuum detected on this iteration of the engine. This will never exceed 1000 even if there are more to vacuum.",
+	})
+	numberVacuumConnections = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "number_vacuum_connections",
+		Help:      "Number of database connections currently in use by the vacuum engine. One taken up by the advisory lock, the rest by vacuum commands.",
+	})
+)
+
+func init() {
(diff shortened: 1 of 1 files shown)


pkg/vacuum/vacuum.go
```diff
@@ -220,5 +254,8 @@ func (e *Engine) worker(ctx context.Context, id int, todo <-chan string) {
 			log.Error("msg", "failed to vacuum chunk", "chunk", chunk, "worker", id, "error", err)
 			// don't return error here. attempt to vacuum other chunks. keep working
 		}
+		tablesVacuumedTotal.Inc()
+		tablesNeedingVacuum.Dec()
```
cevian, line 258: I am torn on this decrement. On one hand its true you just vacummed a table, on the other: 
1) its really weird when against the 1000 limit
2) another table may have been created that needs vacumming

Overall I think there is something nice about this gauge being always set from data in the DB. The "single source of truth"

but not sure and dont feel super strongly

pkg/vacuum/vacuum.go
```diff
@@ -8,10 +8,37 @@ import (
...
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "tables_needing_vacuum",
+		Help:      "Number of tables needing a vacuum. This will never exceed 1000 even if there are more to vacuum.",
+	})
+	numberVacuumConnections = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: util.PromNamespace,
+		Subsystem: "vacuum",
+		Name:      "number_vacuum_connections",
+		Help:      "Number of database connections currently in use by the vacuum engine.",
```
cevian, line 34:
```suggestion
		Help:      "Number of database connections currently in use by the vacuum engine. One taken up by the advisory lock, the rest by vacuum commands.",
```

cevian (changes requested):
some questions/suggestions

pkg/vacuum/vacuum.go
```diff
@@ -212,6 +242,10 @@ func runWorkers(ctx context.Context, parallelism int, chunks []string, worker fu
 
 // worker pulls chunks from a channel and vacuums them
 func (e *Engine) worker(ctx context.Context, id int, todo <-chan string) {
+	// incrementing the number of connections here and decrementing at the end of the function
+	// is a bit of a lie since we are really only using a connection from the pool during
+	// the call to Exec, however conceptually it is probably less confusing to do it this way
+	numberVacuumConnections.Inc()
```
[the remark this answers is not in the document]
jgpruitt: i was afraid it would bounce too much, but given that time taken by the Exec call should far exceed the time of the rest of the function, it probably won't matter.

pkg/vacuum/vacuum.go
jgpruitt: okay. I'll update the help of the gauge to hopefully better describe it

pkg/vacuum/vacuum.go
```diff
@@ -215,10 +245,13 @@ func (e *Engine) worker(ctx context.Context, id int, todo <-chan string) {
...
 		log.Debug("msg", "vacuuming a chunk", "worker", id, "chunk", chunk)
 		sql := fmt.Sprintf(sqlVacuumFmt, chunk)
+		numberVacuumConnections.Inc()
 		_, err := e.pool.Exec(ctx, sql)
+		numberVacuumConnections.Dec()
 		if err != nil {
 			log.Error("msg", "failed to vacuum chunk", "chunk", chunk, "worker", id, "error", err)
 			// don't return error here. attempt to vacuum other chunks. keep working
 		}
+		tablesVacuumedTotal.Inc()
```
niksajakovljevic, line 255: We should only increment this in else statement (err == nil).

niksajakovljevic (approved):
Looks good. One minor thing: to not increment `tablesVacuumedTotal` if we got an error (I assume here we only count successful vacuums)

pkg/vacuum/vacuum.go
arajkumar: May be we can have an error counter?

jgpruitt: good catch and good idea. thanks

PR #1776, opened 2022-12-01
Update rollups creation/deletion/updation using dataset-config

Harkishen-Singh:

## Merge requirements


pkg/dataset/config.go
```diff
@@ -44,12 +45,12 @@ type Config struct {
...
-	HALeaseRefresh  day.Duration `yaml:"ha_lease_refresh"`
-	HALeaseTimeout  day.Duration `yaml:"ha_lease_timeout"`
-	RetentionPeriod day.Duration `yaml:"default_retention_period"`
-	Downsample      *Downsample  `yaml:"downsample,omitempty"`
+	ChunkInterval   day.Duration   `yaml:"default_chunk_interval"`
+	Compression     *bool          `yaml:"compress_data"` // Using pointer to check if the value was set.
+	HALeaseRefresh  day.Duration   `yaml:"ha_lease_refresh"`
+	HALeaseTimeout  day.Duration   `yaml:"ha_lease_timeout"`
+	RetentionPeriod day.Duration   `yaml:"default_retention_period"`
+	Downsample      *rollup.Config `yaml:"downsample,omitempty"`
```
alejandrodnm, line 53: This was changed [here](https://github.com/timescale/promscale/pull/1737/files#diff-b16f8ee862c8f1f5b69f5f1856d60cb38c393083834de0446cdd238b94e94f53R45-R49) so maybe it would be good if you rebase your feature branch, since this might give you conflicts later.

alejandrodnm, line 53: Why do we call it some places rollup and in other downsample? I think using a single name and sticking with it through everything, code, docs, communication leads to less confusion.

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/jackc/pgx/v4"
+
+	"github.com/timescale/promscale/pkg/internal/day"
+	"github.com/timescale/promscale/pkg/log"
+)
+
+type Config struct {
+	Enable               *bool `yaml:"enabled,omitempty"`
```
alejandrodnm, line 15: NIT: let's use the same as the config key 

```suggestion
	Enabled               *bool `yaml:"enabled,omitempty"`
```

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+	UseDefaultResolution *bool `yaml:"use_default_resolution"`
+	Resolution           `yaml:"resolutions,omitempty"`
```
alejandrodnm, line 17: This contains multiple resolutions `Info`, I think the plural form suits it better. Also, it differentiates it from `Info.Resolution`.

```suggestion
	Resolutions           `yaml:"resolutions,omitempty"`
```

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+type (
+	Label      string
+	Resolution map[Label]Info
+)
+
+var (
+	setDefaultDownsampleStateSQL = "SELECT prom_api.set_automatic_downsample($1)"
+	defaultDownsampleState       = false
+	useDefaultResolution         = true
```
alejandrodnm, line 34: Should these 3 be const instead of var?

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+	log.Info("msg", fmt.Sprintf("Setting automatic metric downsample to %t", *d.Enable))
+	if _, err := conn.Exec(context.Background(), setDefaultDownsampleStateSQL, d.Enable); err != nil {
+		return err
+	}
+
+	if *d.Enable {
+		if *d.UseDefaultResolution {
+			// Apply defaults only when user has not applied "short" & "long".
+			if _, exists := d.Resolution["short"]; !exists {
```
alejandrodnm, line 58: Do we need to worry about the key case? Also, `short` and `long` might be good candidates to turn into enums.

pkg/rollup/rollup.go
```diff
@@ -14,93 +14,92 @@ import (
...
+// different retention duration.
+func updateExistingRollups(conn *pgx.Conn, existingRes, newRes Resolution) error {
+	var batch pgx.Batch
+	for labelName, res := range newRes {
+		if oldRes, exists := existingRes[labelName]; exists && oldRes.Retention != res.Retention {
+			batch.Queue("UPDATE _prom_catalog.rollup SET retention = $1 WHERE name = $2", time.Duration(res.Retention), labelName)
+		}
+	}
+	if batch.Len() > 0 {
+		results := conn.SendBatch(context.Background(), &batch)
```
alejandrodnm, line 64: Would it make sense to accept the context from the upstream functions?

alejandrodnm (approved):
Just some minor comments. Good job :)

pkg/rollup/config.go
niksajakovljevic, line 58: Why don't we `block` keys that we use so this can't happen? I think it will make it easier to reason about rollups since we will always know that these two (`short` and `long`) are system rollups.

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/timescale/promscale/pkg/log"
+)
+
+type Config struct {
+	Enable               *bool `yaml:"enabled,omitempty"`
+	UseDefaultResolution *bool `yaml:"use_default_resolution"`
+	Resolution           `yaml:"resolutions,omitempty"`
+}
+
+type Info struct {
```
niksajakovljevic, line 20: I think we can name this more specific: eg `RollupDefinition` or similar.

pkg/rollup/config.go
```diff
@@ -0,0 +1,82 @@
...
+	Resolution day.Duration `yaml:"resolution"`
+	Retention  day.Duration `yaml:"retention"`
+	Delete     bool         `yaml:"delete"`
+}
+
+type (
+	Label      string
+	Resolution map[Label]Info
```
niksajakovljevic, line 28: I'd rather have this very simple: just have a map in `Config` like `map[string]ResolutionDefinition` .

pkg/rollup/rollup.go
```diff
@@ -14,93 +14,92 @@ import (
...
 }
 
-func createRollups(conn *pgx.Conn, res []DownsampleResolution) error {
-	for _, r := range res {
-		_, err := conn.Exec(context.Background(), "CALL _prom_catalog.create_rollup($1, $2, $3)", r.Label, time.Duration(r.Resolution), time.Duration(r.Retention))
-		if err != nil {
-			return fmt.Errorf("error creating rollup for %s: %w", r.Label, err)
+// updateExistingRollups updates the existing rollups retention if the new resolutions with a same name has
+// different retention duration.
+func updateExistingRollups(conn *pgx.Conn, existingRes, newRes Resolution) error {
```
niksajakovljevic, line 56: What is happening if user tweaks resolution for the existing rollup? I guess we should handle that in some way. 
Also we need to fix naming here. `Resolution` type and argument names are misleading since we are passing in Rollup definitions.

niksajakovljevic (changes requested):
The overall logic looks good.
However namings are confusing and make it harder to reason about the code.
There are few other things I commented on that might need some discussion (eg. block our default names, what happens on resolution changes for existing rollups).

pkg/dataset/config.go
Harkishen-Singh: I am not confident of rebasing the branch ATM, given there might be more conflicts than this. I think it will be better to resolve conflicts later on when I dedicate 1 day for it, than doing in between. This way the benchmarks also behave stable for rollups development only. The branch is already behind 29 commits now :)

pkg/rollup/config.go
Harkishen-Singh: We can for `setDefaultDownsampleStateSQL` but not for the rest two, since we use the address of the rest two.

pkg/rollup/config.go
Harkishen-Singh: I forgot about case sensitivity. Yes, we should

Harkishen-Singh: I like this idea.

pkg/rollup/rollup.go
Harkishen-Singh: If the user tweaks resolutions, we cannot do anything and hence should error. I will add this in the code.

pkg/rollup/rollup.go
Harkishen-Singh: We do not pass down context in datasetConfig.Apply() (which we should), but I will add a context field for future reference.

PR #1777, opened 2022-12-01
Add documentation for the series cache invalidation mechanism

JamesGuthrie:
## Merge requirements


pkg/pgmodel/invalidation.readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+The soft-deletion works as follows: When a row is to be removed, we set the
+`delete_epoch` column of `_prom_catalog.series` to the value currently
+contained in the `current_epoch` column of the `_prom_catalog.ids_epoch` table.
+This indicates: it was decided at `current_epoch` to delete this row at some
+point in the future.
+
+The row is removed when the following condition is met:
+`prom_catalog.series.delete_epoch` >= `current_epoch` - `epoch_duration`
+When the row is removed, `_prom_catalog.ids_epoch.delete_epoch` is set to
```
sumerman, line 58: These two sentences aren't wrong and convey the system's intent, but I believe they misrepresent the causality of events taking place during an invalidation. 

We indeed for a row to be removed when the following condition is met: `prom_catalog.series.delete_epoch` >= `current_epoch` - `epoch_duration`. To make the system's state more explicit and allow for reconfigurations we track `current_epoch` - `epoch_duration` as `_prom_catalog.ids_epoch.delete_epoch`. Therefore a row is considered safe to delete when `prom_catalog.series.delete_epoch` >= `_prom_catalog.ids_epoch.delete_epoch`

JamesGuthrie: Yes, you're right - thank you!

sumerman (approved):

PR #1780, opened 2022-12-02
Implement rollup schema decider based on querying range

Harkishen-Singh:

## Merge requirements


alejandrodnm (commented):
I didn't finish my review, so I'll leave this as comments.

I'll give it another pass later

pkg/pgclient/client.go
```diff
@@ -61,7 +61,7 @@ type Client struct {
 }
 
 // NewClient creates a new PostgreSQL client
-func NewClient(r prometheus.Registerer, cfg *Config, mt tenancy.Authorizer, schemaLocker LockFunc, readOnly bool) (*Client, error) {
+func NewClient(r prometheus.Registerer, cfg *Config, mt tenancy.Authorizer, schemaLocker LockFunc, readOnly, useRollups bool) (*Client, error) {
```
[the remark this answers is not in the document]
alejandrodnm: I checked and we can't because rollup comes from dataset

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+		var (
+			schemaName string
+			resolution time.Duration
+		)
+		if err = rows.Scan(&schemaName, &resolution); err != nil {
+			return fmt.Errorf("error scanning rows: %w", err)
+		}
+		h.resolutionsASC = append(h.resolutionsASC, rollupInfo{schemaName: schemaName, resolution: resolution})
+	}
+	sort.Sort(sortRollupInfo(h.resolutionsASC))
```
niksajakovljevic, line 166: Let's order on DB side to avoid sort operation here.

niksajakovljevic: That would also mean we get rid of below struct comparison code

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+		samples := estimateSamples(info.resolution)
+		if samples < upperLimit {
+			// The first highest resolution that is below upper limit is our answer,
+			// since it provides the highest granularity at the expected samples.
+			return info.schemaName
+		}
+	}
+	// All rollups are above upper limit. Hence, send the schema of the lowest resolution
+	// as this is the best we can do.
+	lowestRollup := h.resolutionsASC[len(h.resolutionsASC)-1]
```
niksajakovljevic, line 81: If order is ascending I would expect that the highest resolution is the last.

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+	return lowestRollup.schemaName
+}
+
+func (h *QueryHelper) ContainsMetricType(metricName string) bool {
+	_, present := h.metricType[metricName]
+	return present
+}
+
+func (h *QueryHelper) Refresh() error {
```
niksajakovljevic, line 90: Nit: why is this function public? It seem to only be used by internal go routine

pkg/rollup/resolution.go
Harkishen-Singh: Sorry, but highest resolution is the one which has the least interval, since that will give maximum granularity. The last will be 1 hour (in example of 5mins, 15mins, 1hour) and hence 1 hour is of lowest resolution.

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return helper, nil
+}
+
+// DecideSchema returns the schema name of the rollups that should be used for querying.
+// The returned schema represents a downsampled resolution that should be an optimal
+// resolution for querying.
+//
+// If no rollups exists or if downsampling is disabled, "prom_data" is returned.
+func (h *QueryHelper) DecideSchema(min, max int64) string {
```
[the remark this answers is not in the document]
Harkishen-Singh: The correct answer is a bit complicated. But ideally, yes for a range query.

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/pgxconn"
+)
+
+const (
+	originalSchema          = "prom_data"
+	upperLimit              = 5000 // Maximum samples allowed
+	assumedScrapeInterval   = time.Second * 30
+	refreshRollupResolution = time.Minute * 30
```
niksajakovljevic, line 22: Got confused with this name and what it does. Maybe  just `refreshRollupInterval`? 
Btw naming of the file is `resolution` and the logic in it has mostly to deal with finding the right rollups to use.
I think this points that we need to agree on naming here. It does seem that we use `rollup` term interchangeably with `resolution` and I think it brings confusion.

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+	schemaName string
+	resolution time.Duration
+}
+
+type QueryHelper struct {
+	conn                pgxconn.PgxConn
+	ctx                 context.Context
+	refreshMtx          sync.RWMutex
+	metricType          map[string]string // metric_name: metric_type
+	resolutionsASC      []rollupInfo      // {schemaName, resolution} in ascending order of resolution.
```
niksajakovljevic, line 35: Since this is just an array of rollups, why not name it `rollups`? I also feel that we don't strictly need to make ordering part of a name (at least the most of our code doesn't seem to follow that practice)

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+			return info.schemaName
+		}
+	}
+	// All rollups are above upper limit. Hence, send the schema of the lowest resolution
+	// as this is the best we can do.
+	lowestRollup := h.resolutionsASC[len(h.resolutionsASC)-1]
+	return lowestRollup.schemaName
+}
+
+func (h *QueryHelper) ContainsMetricType(metricName string) bool {
```
niksajakovljevic, line 85: This doesn't seem to be used anywhere so can be removed

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+	_, present := h.metricType[metricName]
```
niksajakovljevic, line 86: Also I haven't seen `metricType` being used anywhere? Is that something planned for some things that are not part of this PR? If so it's better to not have them here

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+type rollupInfo struct {
+	schemaName string
+	resolution time.Duration
+}
+
+type QueryHelper struct {
+	conn                pgxconn.PgxConn
+	ctx                 context.Context
+	refreshMtx          sync.RWMutex
+	metricType          map[string]string // metric_name: metric_type
```
niksajakovljevic, line 34: What is this used for?

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+	return nil
+}
+
+func (h *QueryHelper) refreshRollupResolutions() error {
+	rows, err := h.conn.Query(h.ctx, "SELECT schema_name, resolution FROM _prom_catalog.rollup")
+	if err != nil {
+		return fmt.Errorf("fetching rollup resolutions: %w", err)
+	}
+	h.resolutionsASC = []rollupInfo{}
+	for rows.Next() {
```
niksajakovljevic, line 156: Close rows to avoid connection leak

pkg/rollup/resolution.go
```diff
@@ -0,0 +1,182 @@
...
+type QueryHelper struct {
+	conn                pgxconn.PgxConn
+	ctx                 context.Context
+	refreshMtx          sync.RWMutex
+	metricType          map[string]string // metric_name: metric_type
+	resolutionsASC      []rollupInfo      // {schemaName, resolution} in ascending order of resolution.
+	downsamplingEnabled bool
+}
+
+func NewQueryHelper(ctx context.Context, conn pgxconn.PgxConn) (*QueryHelper, error) {
```
niksajakovljevic, line 39: Nit: naming here seems a bit vague. It seems that the only functionality that it provides is returning the `schema` name of rollups based on time range. Maybe `RollupManager` would be more appropriate?

pkg/rollup/resolution.go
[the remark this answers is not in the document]
Harkishen-Singh: We initialize the rollups information on constructor `NewQueryHelper()`. Resolving this comment.

pkg/rollup/resolution.go
Harkishen-Singh: Since this file mostly helps us decide which rollups to be used for the given query ranges, I am renaming the file and the struct to `Decider{}`.

pkg/rollup/resolution.go
Harkishen-Singh: I did think to name manager, but then I couldn't conclude as what does it manage. So I thought in the direction that it helps to performing querying on rollups, so named `QueryHelper`. Now, I have changed to `Decider` since it helps to decide which `Schema` to be used based on {min, max}. Please let me know if the confusion regarding naming persists.

pkg/rollup/resolution.go
niksajakovljevic: I believe as it is now is counter-intuitive. I think we should first align on what resolution is. In our case we can probably define it as number of data points per time interval. Let' say we take 1h as a time for calculating resolution. So if we pick 5m granularity, that will have maximum resolution since we can fit 12 x 5000 data points in one hour. If we take 1h granularity it will be 5000 data points in 1 hour. That's why I believe that 5m is our highest resolution (not lowest) since resolution is a density of data points.
One potential solution could be to rename all `resolution` terms with interval. That will be more inline with current logic. It would also avoid confusion with terms b/c we are actually storing rollup intervals and not resolution

pkg/tests/end_to_end_tests/rollup_query_helper_test.go
```diff
@@ -0,0 +1,102 @@
...
+		_, err = db.Exec(context.Background(), "CALL _prom_catalog.create_rollup('short', interval '5 minutes', interval '30 days')")
+		require.NoError(t, err)
+		_, err = db.Exec(context.Background(), "CALL _prom_catalog.create_rollup('medium', interval '15 minutes', interval '30 days')")
+		require.NoError(t, err)
+		_, err = db.Exec(context.Background(), "CALL _prom_catalog.create_rollup('long', interval '1 hour', interval '30 days')")
+		require.NoError(t, err)
+		_, err = db.Exec(context.Background(), "CALL _prom_catalog.create_rollup('very_long', interval '1 week', interval '30 days')")
+		require.NoError(t, err)
+
+		var numResolutions int
```
niksajakovljevic, line 31: Nit: `numRollups`

pkg/rollup/decider_test.go
```diff
@@ -0,0 +1,113 @@
...
+			expectedSchemaName: DefaultSchema,
+		},
+		{
+			// DRY RUN on 500 - 5000 logic
+			// --------
+			//
+			// Assumed default scrape interval being 30 secs
+			// raw 		-> 20,160
+			//
+			// And, when using following rollup resolutions, num samples:
```
niksajakovljevic, line 69: Nit: `rollup intervals`

niksajakovljevic (approved):

pkg/rollup/decider_test.go
```diff
@@ -0,0 +1,113 @@
...
+			max:                30 * 24 * time.Hour,
+			expectedSchemaName: "15_minute",
+		}, {
+			// DRY RUN on 500 - 5000 logic
+			// --------
+			//
+			// Assumed default scrape interval being 30 secs
+			// raw 		-> 20,160
+			//
+			// And, when using following rollup resolutions, num samples:
```
niksajakovljevic, line 91: Nit: `resolutions` -> `intervals`

Harkishen-Singh:
Update: We plan to not merge PRs related to querying of rollups ATM.

PR #1785, opened 2022-12-06
Cache invalidation interleaving test

sumerman:
## Description

This PR adds a test triggering the background job and eviction simultaneously with a randomized overlap.

## Merge requirements

- [ ] ~CHANGELOG entry for user-facing changes~
- [ ] ~Updated the relevant documentation~

### DIFF
--- a/pkg/tests/end_to_end_tests/create_test.go
+++ b/pkg/tests/end_to_end_tests/create_test.go
@@ -7,7 +7,9 @@ package end_to_end_tests
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"reflect"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -577,6 +579,119 @@ func TestInsertAbort(t *testing.T) {
...
+			err = ingestor.CompleteMetricCreation(context.Background())
+			require.NoError(t, err)
+
+			// An epoch duration delay
+			time.Sleep(1 * time.Second)
+
+			var wg sync.WaitGroup
+			wg.Add(3)
+
+			bgDelay := time.Duration(rand.Intn(90)) * time.Millisecond
```
niksajakovljevic, line 639: Can you please add comments for delays. I believe I do understand what they mean but still better to be explicit

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -577,6 +579,119 @@ func TestInsertAbort(t *testing.T) {
...
+				time.Sleep(cacheDelay)
+				// Run invalidation
+				ingestor.Dispatcher().(*ingstr.PgxDispatcher).RefreshSeriesEpoch()
+				wg.Done()
+			}()
+
+			insertDelay := time.Duration(rand.Intn(300)) * time.Millisecond
+			go func() {
+				time.Sleep(insertDelay)
+				// This should either appear as a new series or fail due to epoch change
```
niksajakovljevic, line 659: Can you please explain more the idea.
I also got confused that we are not sure about the outcome?

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -577,6 +579,119 @@ func TestInsertAbort(t *testing.T) {
...
+				insertSamples(farFuture-int64(time.Hour/time.Millisecond*24*365*10), "test1", true)
```
niksajakovljevic, line 660: Can we maybe use date and get seconds from it? It would make it much easier to reason about time input.

pkg/tests/end_to_end_tests/create_test.go
```diff
@@ -577,6 +579,119 @@ func TestInsertAbort(t *testing.T) {
...
+func TestEvictionVInsert(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	// Interate until we encounter all acceptable outcomes
+	// (S1, S1 + S2 (ressurection), S1 + S3 (new series id allocation).
+	expectedSeriesIds := map[int64]bool{}
+	for len(expectedSeriesIds) < 3 {
+		withDB(t, *testDatabase, func(dbOwner *pgxpool.Pool, t testing.TB) {
```
niksajakovljevic, line 592: I don't understand why are we re-provisioning the test DB?

niksajakovljevic: If this to test different scenarios I'd recommend using `testCases` pattern

pkg/tests/end_to_end_tests/create_test.go
sumerman: Done

pkg/tests/end_to_end_tests/create_test.go
sumerman: Done. Please read the comment above the test function.

pkg/tests/end_to_end_tests/create_test.go
sumerman: The specific value doesn't matter, I'll be using just `farFuture`.

niksajakovljevic (approved):
Thanks! So much easier to understand now.

PR #1787, opened 2022-12-06
Moving test to extension

jgpruitt:
## Merge requirements


cevian (approved):
Can we also adjust the metrics for chunks to be compressed to use the new function?

PR #1791, opened 2022-12-09
Use common array pool when unmarshalling protobufs

niksajakovljevic:
We reuse one bigger array and avoid allocating smaller arrays for each Timeseries.
This results in less memory allocations.

Before:
`BenchmarkProtoUnmarshall-10           144       8194256 ns/op    22833242 B/op     150342 allocs/op`

After:
`BenchmarkProtoUnmarshall-10          145       8174649 ns/op    32628033 B/op     100650 allocs/op`

So the allocations are down by around 1/3, however memory usage is up for 1/3 due to larger arrays being allocated.

### DIFF
--- a/pkg/api/write_test.go
+++ b/pkg/api/write_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/timescale/promscale/pkg/api/parser"
 	"github.com/timescale/promscale/pkg/log"
 	"github.com/timescale/promscale/pkg/prompb"
+	"github.com/timescale/promscale/pkg/tests/testsupport"
 )
 
 func TestDetectSnappyStreamFormat(t *testing.T) {
@@ -245,6 +246,47 @@ func TestWrite(t *testing.T) {
 				)
 			}
 		})
+	}
+}
+
+// BenchmarkProtoUnmarshall is primarily focused on benchmarking
+// memory usage and allocations when unmarshalling WriteRequests
+func BenchmarkProtoUnmarshall(b *testing.B) {
+	seriesGen, err := testsupport.NewSeriesGenerator(100, 100, 5)
+	if err != nil {
+		b.Error(err)
+	}
+	batches := seriesGen.GetTimeseriesInBatch(1000)
+	mock := &mockInserter{}
+	metrics = &Metrics{LastRequestUnixNano: 0}
+	dataParser := parser.NewParser()
+	numSamplesReceived := &mockMetric{}
+	writeHandler := Write(mock, dataParser, mockUpdaterForIngest(&mockMetric{}, nil, numSamplesReceived, nil))
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		for _, batch := range batches {
+			b.StopTimer()
+			req, err := testsupport.GetHTTPWriteRequest("", &prompb.WriteRequest{
+				Timeseries:           batch,
+				Metadata:             []prompb.MetricMetadata{},
+				XXX_NoUnkeyedLiteral: struct{}{},
+				XXX_unrecognized:     []byte{},
+				XXX_sizecache:        0,
+			})
+			if err != nil {
+				b.Error(err)
+			}
+			recorder := httptest.NewRecorder()
+			b.StartTimer()
+			writeHandler.ServeHTTP(recorder, req)
+			b.StopTimer()
+			res := recorder.Result()
+			if res.StatusCode != 200 {
+				b.Error("unexpected status code", res.StatusCode)
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/api/write_test.go
```diff
@@ -248,6 +249,47 @@ func TestWrite(t *testing.T) {
...
+	dataParser := parser.NewParser()
+	numSamplesReceived := &mockMetric{}
+	writeHandler := Write(mock, dataParser, mockUpdaterForIngest(&mockMetric{}, nil, numSamplesReceived, nil))
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		for _, batch := range batches {
+			b.StopTimer()
+			req, err := testsupport.GetHTTPWriteRequest("", &prompb.WriteRequest{
```
cevian, line 271: We should create and pre-allocate these write requests before we call b.ResetTimer() because othewise these allocations are included in the benchmark results, making them harded to interpret

pkg/prompb/remote.pb.go
```diff
@@ -61,12 +61,27 @@ func (ReadRequest_ResponseType) EnumDescriptor() ([]byte, []int) {
...
+	exemplarsPool []Exemplar
+}
+
 type WriteRequest struct {
 	Timeseries           []TimeSeries     `protobuf:"bytes,1,rep,name=timeseries,proto3" json:"timeseries"`
 	Metadata             []MetricMetadata `protobuf:"bytes,3,rep,name=metadata,proto3" json:"metadata"`
 	XXX_NoUnkeyedLiteral struct{}         `json:"-"`
 	XXX_unrecognized     []byte           `json:"-"`
 	XXX_sizecache        int32            `json:"-"`
+	XXX_arrayPool        *ArrayPool       // gogo proto forces us to have XXX_ prefix for unexported fields
```
cevian, line 78: I wonder why not make this a struct embedding instead of a pointer. This makes it more friendly for GC 
```suggestion
	XXX_arrayPool        ArrayPool       // gogo proto forces us to have XXX_ prefix for unexported fields
```

pkg/prompb/types.pb.go
```diff
@@ -2167,6 +2177,12 @@ func (m *TimeSeries) Unmarshal(dAtA []byte) error {
 	if iNdEx > l {
 		return io.ErrUnexpectedEOF
 	}
+	m.Labels = labelsPool[labelsPos:]
+	m.Samples = samplesPool[samplesPos:]
+	m.Exemplars = exemplarsPool[exemplarsPos:]
+	arrayPool.labelsPool = labelsPool
+	arrayPool.samplesPool = samplesPool
+	arrayPool.exemplarsPool = exemplarsPool
```
cevian, line 2185: nit: I think this is error prone. Let's just use arrayPool.labelsPool in the code directly rather than create intermediate variables

pkg/api/write_test.go
niksajakovljevic: `StopTimer` turns of allocations/mem counter so we should be good I believe. I also tested without `StopTimer` and noticed more allocations.

pkg/prompb/types.pb.go
niksajakovljevic: Yeah I can see that. The reason I did it was to reduce verbosity of having `arrayPool.X` repeating. Anyways I don't feel strongly either way.

pkg/api/write_test.go
niksajakovljevic: Also I had that done initially but since here we iterate `b.N` we might re-use HTTP request which we should not.

pkg/prompb/types.pb.go
```diff
@@ -2228,7 +2238,10 @@ func (m *Label) Unmarshal(dAtA []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			m.Name = string(dAtA[iNdEx:postIndex])
+			tmpNameBytes := dAtA[iNdEx:postIndex]
+			m.NameBytes = make([]byte, len(tmpNameBytes))
```
alejandrodnm, line 2242: Question. We need to keep the bytes so that they don't get GC'd and break `m.Name` or something like that?

alejandrodnm (approved):

pkg/prompb/types.pb.go
niksajakovljevic: Right. Since these bytes are coming from HTTP request which we get rid off after unmarshalling.

niksajakovljevic:
Surprisingly cluster benchmark is showing that this implementation uses more memory.

Screenshot 2022-12-23 at 14 58 55

I'll investigate further.

niksajakovljevic:
Project got deprecated

PR #1793, opened 2022-12-13
Custom metric update interval for internal database metrics

sumerman:
## Merge requirements


pkg/pgmodel/metrics/database/database.go
```diff
@@ -114,11 +114,17 @@ func (e *metricsEngineImpl) Run() error {
 func (e *metricsEngineImpl) Update() error {
 	batch := e.conn.NewBatch()
 	batchMetrics := []metricQueryWrap{}
+	now := time.Now()
```
niksajakovljevic, line 117: Nit: it would be more correct to grab `Now()` before checking customPoll config b/c we do healthCheck call.

pkg/pgmodel/metrics/database/database.go
```diff
@@ -114,11 +114,17 @@ func (e *metricsEngineImpl) Run() error {
 func (e *metricsEngineImpl) Update() error {
```
niksajakovljevic, line 114: Consider adding small unit test of this function to test newly added logic.

sumerman: The existing test tests that metric collection is delayed, but not that it's eventually collected. I did consider it but I couldn't find a good way to mock time intervals and waiting for 10 minutes in a test doesn't sound like a great idea.

niksajakovljevic: I was thinking calling `Update` function directly. Have two metrics defined with customConfigs and some short intervals: eg 200ms and 1s. Then just sleep before renewed Update call?

sumerman: Do you suggest adding `cusromPollConfig` to metrics that don't actually need it? Or is there a way to pass a mocked set of metrics to the `Update`?

niksajakovljevic: So I was thinking something like this:
```
func TestStuff(t *testing.T){
	engine := &metricsEngineImpl{
		conn:    nil, // here we need to pass a mock check comment below
		ctx:     context.Background(),
		metrics: []metricQueryWrap{},
	}
	engine.Update()
	// iterate over engine.metrics and check 
	time.Sleep(time.Second)
	engine.Update()
	// iterate over engine.metrics and check 
}
```
For PGConn mock we can maybe use (worth maybe moving it to `testsupport` :  https://github.com/timescale/promscale/blob/master/pkg/telemetry/telemetry_test.go#L54

niksajakovljevic: I hope this is not too much work

niksajakovljevic: Otherwise maybe we can expose public function to pass in custom metrics with some short interval and have it covered in e2e tests?

niksajakovljevic (approved):

Issue #1796, opened 2022-12-14
In read-only mode shouldn't apply dataset config


zfy0701:
**Describe the bug**
In read-only mode shouldn't apply dataset config

https://github.com/timescale/promscale/blob/9df822a114e08cce67f792e6289a022a08ee7044/pkg/runner/client.go#L168-L178

otherwise it reports:
```
{"caller":"runner.go:94","level":"info","msg":"Migrations disabled for read-only mode","ts":"2022-12-14T21:15:28.032Z"}
{"caller":"client.go:91","level":"info","msg":"Skipping migration","ts":"2022-12-14T21:15:28.032Z"}
{"caller":"client.go:105","level":"warn","msg":"skipping schema version lease","ts":"2022-12-14T21:15:28.032Z"}
{"caller":"config.go:63","level":"info","msg":"Setting metric dataset default chunk interval to 8h0m0s","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"config.go:64","level":"info","msg":"Setting metric dataset default compression to true","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"config.go:65","level":"info","msg":"Setting metric dataset default high availability lease refresh to 10s","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"config.go:66","level":"info","msg":"Setting metric dataset default high availability lease timeout to 1m0s","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"config.go:67","level":"info","msg":"Setting metric dataset default retention period to 2160h0m0s","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"config.go:68","level":"info","msg":"Setting trace dataset default retention period to 720h0m0s","ts":"2022-12-14T21:15:28.102Z"}
{"caller":"runner.go:127","err":"error applying dataset configuration: ERROR: cannot execute INSERT in a read-only transaction (SQLSTATE 25006)","level":"error","msg":"aborting startup due to error","ts":"2022-12-14T21:15:28.104Z"}
```

**To Reproduce**
run promscale connect to timescale replica

**Expected behavior**
should be able to correctly connect to db

**Screenshots**

**Configuration (as applicable)**
 - Promscale Connector:
 - TimescaleDB:
 - Prometheus:
 - Jaeger:
 - OpenTelemetry:

**Version**
 - Distribution/OS: 
 - Promscale: 
 - TimescaleDB:

**Additional context**
<Add any other context about the problem here.>

PR #1797, opened 2022-12-14
skip apply dateset config when read only

zfy0701:
## Description

skip apply dateset config when read only, resolve https://github.com/timescale/promscale/issues/1796

## Merge requirements

### DIFF
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -165,16 +165,12 @@ func CreateClient(r prometheus.Registerer, cfg *Config) (*pgclient.Client, error
 		cfg.APICfg.MultiTenancy = multiTenancy
 	}
 
-	if (cfg.DatasetCfg != dataset.Config{}) {
-		if cfg.DatasetConfig != "" {
-			log.Warn("msg", "Ignoring `startup.dataset.config` in favor of the newer `startup.dataset` config option since both were set.")
+	if !cfg.APICfg.ReadOnly {
+		// Applying the dataset config needs a connection with write permissions.
+		err := applyDatasetConfigIfDefined(conn, cfg)
+		if err != nil {
+			return nil, fmt.Errorf("error applying dataset configuration: %w", err)
 		}
-		err = cfg.DatasetCfg.Apply(conn)
-	} else if cfg.DatasetConfig != "" {
-		err = applyDatasetConfig(conn, cfg.DatasetConfig)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("error applying dataset configuration: %w", err)
 	}
 
 	// client has to be initiated after migrate since migrate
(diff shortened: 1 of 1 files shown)


pkg/runner/client.go
```diff
@@ -165,16 +165,18 @@ func CreateClient(r prometheus.Registerer, cfg *Config) (*pgclient.Client, error
...
+		if (cfg.DatasetCfg != dataset.Config{}) {
+			if cfg.DatasetConfig != "" {
+				log.Warn("msg", "Ignoring `startup.dataset.config` in favor of the newer `startup.dataset` config option since both were set.")
+			}
+			err = cfg.DatasetCfg.Apply(conn)
+		} else if cfg.DatasetConfig != "" {
+			err = applyDatasetConfig(conn, cfg.DatasetConfig)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error applying dataset configuration: %w", err)
```
alejandrodnm, line 178: I think we can move this logic to a different function, something like:

```
func foo(cfg ) error {
  if cfg.APICfg.ReadOnly {
     return nil     
  }
  if (cfg.DatasetCfg != dataset.Config{}) {
    if cfg.DatasetConfig != "" {
	  log.Warn("msg", "Ignoring `startup.dataset.config` in favor of the newer `startup.dataset` config option since both were set.")
    }
    return cfg.DatasetCfg.Apply(conn)
  }
  if cfg.DatasetConfig != "" {
    return applyDatasetConfig(conn, cfg.DatasetConfig)
  }
  return nil
}

err := foo(cfg)
if err != nil {
  return nil, fmt.Errorf("error applying dataset configuration: %w", err)
}
```

I don't have the brain cells right now to come up with a name for the function. But, WDYT?

zfy0701: how about `applyDatasetConfigIfNeeded`

alejandrodnm: How about we take the ReadOnly check and call the function `applyDatasetConfigIfDefined`

```
if !cfg.APICfg.ReadOnly {
  // Applying the dataset config needs a connection with write permissions.
  err := applyDatasetConfigIfDefined(cfg)
  if err != nil {
    return nil, fmt.Errorf("error applying dataset configuration: %w", err)
  }
}

...

func applyDatasetConfigIfDefined(cfg ) error {
  if (cfg.DatasetCfg != dataset.Config{}) {
    if cfg.DatasetConfig != "" {
	  log.Warn("msg", "Ignoring `startup.dataset.config` in favor of the newer `startup.dataset` config option since both were set.")
    }
    return cfg.DatasetCfg.Apply(conn)
  }
  if cfg.DatasetConfig != "" {
    return applyDatasetConfig(conn, cfg.DatasetConfig)
  }
  return nil
}
```

zfy0701: sounds good! adjusted accord to the suggestion

alejandrodnm (approved):

Harkishen-Singh (approved):

Harkishen-Singh:
ping @alejandrodnm

Harkishen-Singh:
Thank you @zfy0701 for the fix

PR #1800, opened 2022-12-16
Aggregate metrics at global level to avoid legend pollution in dashboards

arajkumar:

## Description

Currently we don't aggregate most of the dashboard queries which leads to legend pollution if we deploy Promscale multiple times. 

### Before

Promscale - Dashboards - Grafana

### After

Promscale - Test - Dashboards - Grafana

## Merge requirements


paulfantom (approved):
LGTM! :shipit:

PR #1804, opened 2022-12-19
Vacuum compressed chunks missing stats

jgpruitt:
## Description

We have seen instances in which compresses chunks are missing statistics. Autovacuum ignores tables that are missing statistics. Analyzing these tables does not help since we rarely modify chunks after they are compressed. Therefore, these chunks are ignored until they pass the vacuum_freeze_max_age. This can be bad for performance. So, the vacuum engine makes a second pass looking for these chunks and vacuums them. We only use one worker for these.

See also: https://github.com/timescale/promscale_extension/pull/595

## Merge requirements

### DIFF
--- a/docs/vacuum.md
+++ b/docs/vacuum.md
@@ -35,3 +35,11 @@ We get the autovacuum count when we produce the list. Just before
 vacuuming a chunk, we check the autovacuum count again to see if it
 has increased. If it has, the autovacuum engine beat us to the
 chunk, and we skip it.
+
+Additionally, we have seen instances in which compresses chunks are
+missing statistics. Autovacuum ignores tables that are missing
+statistics. Analyzing these tables does not help since we rarely
+modify chunks after they are compressed. Therefore, these chunks are
+ignored until they pass the vacuum_freeze_max_age. This can be bad
+for performance. So, the vacuum engine also looks for these chunks
+and vacuums them. We only use one worker for these.
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -34,6 +34,14 @@
 // vacuuming a chunk, we check the autovacuum count again to see if it
 // has increased. If it has, the autovacuum engine beat us to the
 // chunk, and we skip it.
+//
+// Additionally, we have seen instances in which compresses chunks are
+// missing statistics. Autovacuum ignores tables that are missing
+// statistics. Analyzing these tables does not help since we rarely
+// modify chunks after they are compressed. Therefore, these chunks are
+// ignored until they pass the vacuum_freeze_max_age. This can be bad
+// for performance. So, the vacuum engine also looks for these chunks
+// and vacuums them. We only use one worker for these.
 package vacuum
 
 import (
@@ -99,16 +107,27 @@ const (
 		current_setting('vacuum_freeze_min_age')::bigint as vacuum_freeze_min_age,
 		current_setting('autovacuum_freeze_max_age')::bigint as autovacuum_freeze_max_age
 	`
-	sqlListChunks = `
+	sqlListChunksToFreeze = `
 	SELECT 
 		format('%I.%I', schema_name, table_name) AS schema_table,
 		pg_stat_get_autovacuum_count(format('%I.%I', schema_name, table_name)::regclass::oid) AS autovacuum_count,
-		pg_catalog.age(relfrozenxid) as age
-	FROM _ps_catalog.chunks_to_freeze 
+		pg_catalog.age(relfrozenxid) AS age
+	FROM _ps_catalog.compressed_chunks_to_freeze 
 	WHERE coalesce(last_vacuum, '-infinity'::timestamptz) < now() - interval '15 minutes' 
 	AND pg_catalog.age(relfrozenxid) > current_setting('vacuum_freeze_min_age')::bigint
 	ORDER BY pg_catalog.age(relfrozenxid) DESC --oldest first
 	LIMIT 1000`
+	sqlListChunksMissingStats = `
+	SELECT 
+		format('%I.%I', schema_name, table_name) AS schema_table,
+		0 AS autovacuum_count,
+		pg_catalog.age(relfrozenxid) AS age
+	FROM _ps_catalog.compressed_chunks_missing_stats
+	-- older than 20% of vacuum_freeze_min_age 
+	WHERE pg_catalog.age(relfrozenxid) > (current_setting('vacuum_freeze_min_age')::bigint / 5)
+	ORDER BY pg_catalog.age(relfrozenxid) DESC --oldest first
+	LIMIT 1000
+	`
 	// finds out how many times the autovacuum engine has vacuumed a chunk
 	sqlGetAutovacuumCount = "SELECT pg_catalog.pg_stat_get_autovacuum_count($1::regclass::oid)"
 	// delay - on each iteration of the engine, we will sleep a bit before querying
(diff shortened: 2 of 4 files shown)


cevian (approved):

docs/vacuum.md
```diff
@@ -35,3 +35,11 @@ We get the autovacuum count when we produce the list. Just before
 vacuuming a chunk, we check the autovacuum count again to see if it
 has increased. If it has, the autovacuum engine beat us to the
 chunk, and we skip it.
+
+Additionally, we have seen instances in which compresses chunks are
+missing statistics. Autovacuum ignores tables that are missing
+statistics. Analyzing these tables does not help since we rarely
+modify chunks after they are compressed. Therefore, these chunks are
+ignored until they pass the vacuum_freeze_min_age. This can be bad
```
cevian, line 43:
```suggestion
ignored until they pass the vacuum_freeze_max_age. This can be bad
```

pkg/vacuum/vacuum.go
```diff
@@ -34,6 +34,14 @@
 // vacuuming a chunk, we check the autovacuum count again to see if it
 // has increased. If it has, the autovacuum engine beat us to the
 // chunk, and we skip it.
+//
+// Additionally, we have seen instances in which compresses chunks are
+// missing statistics. Autovacuum ignores tables that are missing
+// statistics. Analyzing these tables does not help since we rarely
+// modify chunks after they are compressed. Therefore, these chunks are
+// ignored until they pass the vacuum_freeze_min_age. This can be bad
```
cevian, line 42:
```suggestion
// ignored until they pass the vacuum_freeze_max_age. This can be bad
```

pkg/vacuum/vacuum.go
```diff
@@ -275,35 +294,57 @@ func (e *Engine) Run(ctx context.Context) {
...
 		}
 		runWorkers(ctx, numWorkers, chunks, e.worker)
 		log.Debug("msg", "vacuum workers finished. delaying before next iteration...")
 		// in some cases, have seen it take up to 10 seconds for the stats to be updated post vacuum
 		time.Sleep(delay)
 	}
+	// next, we look for compressed chunks that have never been vacuum and seem to be missing
+	// statistics. autovacuum will ignore these until they hit vacuum_freeze_min_age, so let's
+	// catch these early
+	for {
```
cevian, line 325: This for, I believe can starve out the loop above and vice-versa,.This should probably be one for loop that alternates between the two queries. Or maybe the loop above is more important and should starve-out this loop?

cevian (changes requested):

pkg/vacuum/vacuum.go
```diff
@@ -275,24 +294,50 @@ func (e *Engine) Run(ctx context.Context) {
 		}
 		numberVacuumConnections.Set(0)
 	}()
+
+	for {
```
sumerman, line 298: It took me a while to understand what this outer `for` is supposed to achieve — prevent starvation of either workload. Am I right in this assessment? If so, I think, it would be easier to understand if there were just a single top-level loop taking batches of rows in a round-robin fashion from one source and then another.

jgpruitt: Yes, it is intended to prevent either workload from being completely starved. See [the comment above](https://github.com/timescale/promscale/pull/1804#discussion_r1052649470).

We could do round-robin, but that would give equal preference to both workloads. In my opinion, the `vacuumChunksToFreeze` workload is more important because it's more urgent. These chunks are older and closer to wraparound.

You could argue that if the `vacuumChunksToFreeze` starves out the other, then it might cause more chunks to fall into the `vacuumChunksToFreeze` bucket, but my hunch is that this is unlikely. But I could be wrong.

WDYT?

pkg/vacuum/vacuum.go
```diff
@@ -301,9 +346,35 @@ func (e *Engine) Run(ctx context.Context) {
...
+	// we only do a max of 3000 before we give vacuumChunksToFreeze another turn
+	for i := 0; i <= 3; i++ {
+		chunks, err := e.listChunks(ctx, con, sqlListChunksMissingStats)
+		if err != nil {
+			log.Error("msg", "failed to list compressed chunks missing stats", "error", err)
+			return false, err
+		}
+		tablesNeedingVacuum.Set(float64(len(chunks)))
+		if len(chunks) == 0 {
+			log.Info("msg", "zero compressed chunks missing stats")
```
alejandrodnm, line 362: Should we changed this Info to Debug? Do you think it's worth it to report this as telemetry metrics?

pkg/vacuum/vacuum.go
sumerman: I suppose you are right about `vacuumChunksToFreeze` being more important, but my concern is primarily about structure. I think it is possible to allocate more time for `vacuumChunksToFreeze` even if we round-robin between several workloads. I made an attempt at illustrating my suggestion here https://github.com/timescale/promscale/pull/1810/files

pkg/vacuum/vacuum.go
jgpruitt: As configured, you'd get two of these messages every 10 minutes. Is that too frequent?

`tablesNeedingVacuum` is a gauge that exposes it.

pkg/vacuum/vacuum.go
jgpruitt: I like that there is less duplicate code, and I think you're weighted round-robin is likely better too.
I think there are a couple of issues: it will keep trying both workloads when one is finished, and I'm not sure there is logic to exit the loop when both finish. I can fix that though. Thanks!

pkg/vacuum/vacuum.go
jgpruitt: since we expose it in telemetry, i'll make these DEBUG

pkg/vacuum/vacuum.go
jgpruitt: okay, how's this change?

sumerman (approved):
🔥

pkg/vacuum/vacuum.go
jgpruitt: take another look? @cevian

alejandrodnm (approved):

pkg/vacuum/vacuum.go
```diff
@@ -275,25 +305,72 @@ func (e *Engine) Run(ctx context.Context) {
...
+
+		// every 3rd iteration is chunksMissingStats
+		w := &chunksToFreeze
+		if i%3 == 0 {
+			w = &chunksMissingStats
+		}
+
+		// if this workload already finished, don't continue working it, work the other
+		if w.finished {
+			continue
```
cevian, line 340: doesn't this mean chunksMissingStats can indefinitely starve out chunksToFreeze after chunksToFreeze.finish becomes true?

cevian (changes requested):

pkg/vacuum/vacuum.go
jgpruitt: yes. dangit.

pkg/vacuum/vacuum.go
```diff
@@ -275,25 +304,63 @@ func (e *Engine) Run(ctx context.Context) {
...
+		// find some chunks to vacuum
+		chunks, err := e.listChunks(ctx, con, w.query)
 		if err != nil {
-			log.Error("msg", "failed to list chunks for vacuuming", "error", err)
-			return
+			log.Error("msg", fmt.Sprintf("failed to list %s", w.name), "error", err)
+			w.stop = true
+			continue
 		}
 		tablesNeedingVacuum.Set(float64(len(chunks)))
```
cevian, line 344: shouldn't we add separate metrics as well to count the work for each type?

cevian (approved):
I think the metrics could be improved but wont hold up the PR for that

pkg/vacuum/vacuum.go
jgpruitt: yes. I'll do a follow-up PR

jgpruitt: https://github.com/timescale/promscale/pull/1812

PR #1808, opened 2022-12-21
Adds support for new downsampling UX

Harkishen-Singh:

This commit updates the UX for creating downsampled data for Prometheus metrics.

Now, downsampling can be done by
```
metrics:
  downsampling:
  - interval: 5m
    retention: 30d
  - interval: 1h
    retention: 365d
```

The above dataset-config setting will create 2 downsampling, ds_5m (interval 5m & retention 30d) and ds_1h (interval 1h & retention 365d).

If an entry is removed, like
```
metrics:
  downsampling:
  - interval: 1h
    retention: 365d
```

Then we disable the ds_5m downsampling from refreshing. Deletion can only happen via SQL, `CALL _prom_catalog.delete_downsampling(<schema_name>)`. This keeps the UX clean and simple.

If the removed downsampling is applied again later on, we simply enable the refreshing.

This also means that this commit adds ability to enable or disable downsampling as well.

## Merge requirements


pkg/runner/client.go
```diff
@@ -173,13 +172,6 @@ func CreateClient(r prometheus.Registerer, cfg *Config) (*pgclient.Client, error
 		err = cfg.DatasetCfg.Apply(context.Background(), conn)
 	} else if cfg.DatasetConfig != "" {
 		err = applyDatasetConfig(context.Background(), conn, cfg.DatasetConfig)
-	} else {
-		// We apply downsampling settings even when DatasetConfig is not given, which is the most common case.
-		rollupCfg := &rollup.Config{}
-		err = rollupCfg.Apply(context.Background(), conn)
-		if err != nil {
-			return nil, fmt.Errorf("error applying downsampling configuration: %w", err)
-		}
```
Harkishen-Singh, line 182: Please ignore this. This was the result of a bad conflict resolution. We do not plan to apply downsampling if dataset-config and downsampling is explicitly given, in order to be in the safe place.

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,178 @@
...
+	interval, retention time.Duration
+	disabled            bool
+}
+
+// Sync updates the downsampling cfgs in the DB in accordance with the given new cfgs. It:
+//  1. Creates of new downsampling cfgs that are not in the database
+//  2. Updates retention duration of downsampling cfgs that are present in the database but with a different retention duration
+//  3. Disables refreshing of downsampling cfgs in the database that were not found in the new cfgs
+//  4. Enables refreshing of downsampling cfgs in the database that are in the new cfgs but were previously disabled
+func Sync(ctx context.Context, conn *pgx.Conn, cfgs []Config) error {
```
niksajakovljevic, line 51: I feel we can simplify this logic by merging two sets in memory: one is array of new `Config`, another is array of existing `Config`. Upon merge we just UPSERT Configs (insert or update existing). Having a `merge` method defined will also allow for easier unit testing.

pkg/internal/day/duration.go
```diff
@@ -20,7 +20,13 @@ const (
 
 // Duration acts like a time.Duration with support for "d" unit
 // which is used for specifying number of days in duration.
-type Duration time.Duration
+// It stores the text of duration while parsing, which can be retrieved via Text().
+// This can be useful when we need to know the num of days user wanted, since
```
niksajakovljevic, line 24: Is internal `text` being used anywhere? I did a quick scan and didn't find any usage of it.
If not being used I'd recommend avoid creating new Duration type.

niksajakovljevic (commented):
Left few comments after quick pass. I haven't digged deep into tests since there are a lot of linter complaints.

pkg/internal/day/duration.go
Harkishen-Singh: Yes, we use it in `Config.Name()`. It is important, otherwise the downsampling schemas will be created as `ds_24h` if user provided `1d` as input.

Harkishen-Singh:
Note for reviewers: The lint tests fails at the moment. I haven't fixed them since I aim to first be sure that **we are fine with the new way `day.Duration` is implemented**. Once we are fine, I will fix the linting tests. Otherwise, if we plan to change `day.Duration` down the line, it will consume good time updating the tests.

Harkishen-Singh:
Note: Commit [Adds support to query default downsampling views in absence of __column__](https://github.com/timescale/promscale/pull/1808/commits/7464174d04281b2e81652fa8c1c6c08b13378075) is not directly related to this PR, but it was more comfortable and faster for me to include in this PR, than wait for approval and merge and then open a following PR.

pkg/pgmodel/querier/query_builder_samples.go
```diff
@@ -213,6 +213,10 @@ func buildSingleMetricSamplesQuery(metadata *evalMetadata) (string, []interface{
 		start, end = metadata.timeFilter.start, metadata.timeFilter.end
 	}
 
+	if filter.useDownsamplingViews {
```
Harkishen-Singh, line 216: See code around https://github.com/timescale/promscale_extension/pull/596/files#r1060609175 for more info.

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,174 @@
...
+	newCfgs := make(map[string]Config) // These are the new downsampling cfgs that the user provided. Relation => schema_name: definition{}
+	for _, c := range cfgs {
+		newCfgs[c.Name()] = Config{Interval: c.Interval, Retention: c.Retention}
```
alejandrodnm, line 50: Are there constraints regarding multiple configs for the same interval? I'm thinking about what happens if someone does `24h` and `1d`. In this case they are not the same interval because they differ on the text representation which is what's being used as the key here.

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,174 @@
...
+	setDownsamplingStateSQL       = "SELECT prom_api.set_downsampling_state($1)"
+	createDownsamplingSQL         = "CALL _prom_catalog.create_downsampling($1, $2, $3)"
+	updateDownsamplingStateForSQL = "SELECT _prom_catalog.update_downsampling_state($1, $2)"
+	downsamplePrefix              = "ds_" // Stands of downsample_
+)
+
+type Config struct {
+	Interval  day.Duration `yaml:"interval"`
+	Retention day.Duration `yaml:"retention"`
+	disabled  bool
```
alejandrodnm, line 27: I don't see where this `disabled` flag is being used 🙈

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,174 @@
...
+		return fmt.Errorf("error setting downsampling state: %w", err)
+	}
+	return nil
+}
+
+// Sync updates the downsampling cfgs in the DB in accordance with the given new cfgs. It:
+//  1. Creates of new downsampling cfgs that are not in the database
+//  2. Updates retention duration of downsampling cfgs that are present in the database but with a different retention duration
+//  3. Disables refreshing of downsampling cfgs in the database that were not found in the new cfgs
+//  4. Enables refreshing of downsampling cfgs in the database that are in the new cfgs but were previously disabled
```
alejandrodnm, line 46: You could refactor the `Sync` method into multiple functions per each step. Something like:

```
func Sync(newCfgs) {
  existingCfgs, err := getExistingResolutions()
  err = updateExistingConfigs(existingCfgs, newCfgs)
  err = disableRemovedConfigs(existingCfgs, newCfgs)
  err = enableNewConfigs(existingCfgs, newCfgs)
}
```

The same goes if you go the UPSERT route as niksa recommended

alejandrodnm (changes requested):

pkg/internal/day/duration.go
niksajakovljevic: Yeah I was thinking the same. It seems the only way out is to always use `minute` granularity?

niksajakovljevic: Yeah. It seems we are better with just using Duration String(). For example: `time.Second * 3600` and `time.Minute * 60` will both print `1h0m0s` which is what we want.

jgpruitt (dismissed):

pkg/internal/day/duration.go
Harkishen-Singh: > I asked it [here](https://github.com/timescale/promscale/pull/1808/files#r1060615105) but is there a problem if a user creates 2 configs, one with 24h and another with 1d?

Yes, that is a valid point. I am not sure how to work around this. Doing `duration.String()` is not good, since then we will have schemas like `ds_1h0m0s` which is not tidy. We cannot choose to convert everything into hours or minutes, otherwise it hurts readability. So I left this as the user gave input. Any other suggestions please?

pkg/downsample/downsample.go
Harkishen-Singh: This is a good point. I missed it. Maybe I should error after finding a similar resolution in the database.

Harkishen-Singh:
This ([`_prom_catalog.create_or_update_downsampling`](https://github.com/timescale/promscale_extension/blob/7ef9ea79390bcd94716e6fbf4653bdea770a5a6a/migration/idempotent/017-metric-downsampling.sql#L233)) might be helpful while reviewing.

alejandrodnm (approved):
lgtm

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,168 @@
...
+	pgLock, err := util.NewPgAdvisoryLock(lockID, conn.Config().ConnString())
+	if err != nil {
+		return fmt.Errorf("error getting lock for syncing downsampling config")
+	}
+	defer pgLock.Close()
+	got, err := pgLock.GetAdvisoryLock() // To prevent failure when multiple Promscale start at the same time.
+	if err != nil {
+		return fmt.Errorf("error trying pg advisory_lock")
+	}
+	if !got {
```
niksajakovljevic, line 60: I believe this lock is used for other purposes as well. So I'd recommend waiting/blocking until lock is taken an configuration is applied. We can obviously add some timeout and error if we were unable to take lock and apply downsampling configuration.

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/jackc/pgx/v4"
+
+	"github.com/timescale/promscale/pkg/internal/day"
+	"github.com/timescale/promscale/pkg/log"
+	"github.com/timescale/promscale/pkg/util"
+)
+
+const (
+	setDownsamplingStateSQL       = "SELECT prom_api.set_downsampling_state($1)"
```
niksajakovljevic, line 20: Frankly having three different SQL functions to update downsampling seems confusing

pkg/internal/day/duration.go
```diff
@@ -20,7 +20,13 @@ const (
...
+// this information is lost after parsing.
+type Duration struct {
+	Txt string // Holds the original duration text.
```
niksajakovljevic, line 27: This field name should be more descriptive. eg. `durationStr`

pkg/internal/day/duration.go
```diff
@@ -20,7 +20,13 @@ const (
...
+	T   time.Duration
```
niksajakovljevic, line 28: This as well should be more descriptive, eg `duration`

pkg/internal/day/duration.go
```diff
@@ -68,8 +75,23 @@ func handleDays(s []byte) (time.Duration, error) {
...
 
 // String returns a string value of DayDuration.
-func (d Duration) String() string {
-	return time.Duration(d).String()
+func (d *Duration) String() string {
+	return d.T.String()
+}
+
+// Text returns the original text received while parsing.
+func (d *Duration) Text() string {
```
niksajakovljevic, line 83: If fields are public we don't need function to access them

pkg/internal/day/duration.go
```diff
@@ -68,8 +75,23 @@ func handleDays(s []byte) (time.Duration, error) {
...
+	return d.Txt
+}
+
+// Duration returns the parsed duration.
+func (d *Duration) Duration() time.Duration {
+	return d.T
+}
+
+// SetDuration returns the parsed duration.
+func (d *Duration) SetDuration(t time.Duration) {
```
niksajakovljevic, line 93: Since this field is public we don't need a setter func

pkg/internal/day/duration.go
niksajakovljevic: >Doing duration.String() is not good, since then we will have schemas like ds_1h0m0s which is not tidy

Is that problematic in any way? Mostly asking b/c I want us to avoid Duration wrapper

pkg/internal/day/duration.go
niksajakovljevic, line 93: Also it feels that `Duration` should be immutable because it's derived from text input.

pkg/downsample/downsample.go
niksajakovljevic: I can hardly tell what is the difference between: `update_downsampling_state` and `set_downsampling_state`. Also there is another function named `create_or_update_downsampling`.

niksajakovljevic:
>Note: Commit [Adds support to query default downsampling views in absence of __column__](https://github.com/timescale/promscale/pull/1808/commits/7464174d04281b2e81652fa8c1c6c08b13378075) is not directly related to this PR, but it was more comfortable and faster for me to include in this PR, than wait for approval and merge and then open a following PR.

Btw it does seem that commit is gone so can't really follow through that part

pkg/downsample/downsample.go
jgpruitt: is this correct?

1. if existing config is missing in new set, disable it
2. if new config is not existing, insert
3. if new config matches existing config on schema name, update it UNLESS the interval is different, in which case, throw an error
4. no two configs can have the same interval
5. no two configs can have the same name

pkg/internal/day/duration.go
jgpruitt: can the user do `1.5d` and `36h`?

jgpruitt: I believe this will produce a desired standardized output regardless of how the user specifies the input.

here is the output for the samples given:

```
1d
1d12h
12h
12h15m
15m30s
1d12h15m30s
```

```go
package main

import (
	"fmt"
	"time"
)

func main() {
        // sample tests
	fmt.Println(String(time.Hour * 24))
	fmt.Println(String(time.Hour * 36))
	fmt.Println(String(time.Hour * 12))
	fmt.Println(String(time.Hour*12 + time.Minute*15))
	fmt.Println(String(time.Minute*15 + time.Second*30))
	fmt.Println(String(time.Hour*36 + time.Minute*15 + time.Second*30))
}

func String(d time.Duration) string {
	const day = int64(time.Hour * 24)

	remainder := int64(d)
	days := remainder / day
	remainder = remainder % day
	hours := remainder / int64(time.Hour)
	remainder = remainder % int64(time.Hour)
	mins := remainder / int64(time.Minute)
	remainder = remainder % int64(time.Minute)
	secs := remainder / int64(time.Second)

	display := ""
	if days != 0 {
		display = fmt.Sprintf("%dd", days)
	}
	if hours != 0 {
		display = fmt.Sprintf("%s%dh", display, hours)
	}
	if mins != 0 {
		display = fmt.Sprintf("%s%dm", display, mins)
	}
	if secs != 0 {
		display = fmt.Sprintf("%s%ds", display, secs)
	}
	return display
}
```

jgpruitt: We need to have a deterministic algorithm for how we name schemas. There may be situations (hopefully not) where we have to reverse engineer schema names to figure out what they were.

jgpruitt: So we could use the standard library's `time.Duration` for parsing, and use the function above (or similar) for generating the schema name.

jgpruitt: my vote: if we want to support `d` units in input, use `duration.Day` BUT use a standardize output format like the above instead of the raw user input.

jgpruitt: If we have a deterministic algorithm for the schema names, then we don't have to worry about detecting the case where two different schema names have the same interval. That's a big win IMO

pkg/downsample/downsample.go
jgpruitt: honestly, I'd write a single database function to handle this. It's much fewer lines of code and one database round-trip.

something like this:
```sql
create or replace function apply_downsample_config(config jsonb) returns void
as $proc$
declare
    _input _prom_catalog.downsample[];
    _schema_name text;
begin
    select array_agg(x) into strict _input
    from jsonb_populate_recordset(null::_prom_catalog.downsample, config) x;

    for _schema_name in
    (
        select x.schema_name
        from unnest(_input) x
        where not exists
        (
            select 1
            from _prom_catalog.downsample d
            where d.schema_name = x.schema_name
        )
    )
    loop
        execute format('create schema %I', _schema_name);
    end loop;

    -- insert or update
    insert into _prom_catalog.downsample (schema_name, resolution, retention, should_refresh)
    select x.schema_name, x.resolution, x.retention, true
    from unnest(_input) x
    on conflict (schema_name) do
    update set
        retention = excluded.retention,
        should_refresh = true
    ;

    -- disable if missing
    update _prom_catalog.downsample set should_refresh = false
    where schema_name not in 
    (
        select schema_name
        from unnest(_input)
    );
end;
$proc$ language plpgsql;
```

then you can just push the new config to the database as json. almost no logic required on the go side

```sql
select apply_downsample_config(
$$
[
    {
        "schema_name": "d_1d",
        "resolution": "1d",
        "retention": "30d"
    },
    {
        "schema_name": "d_12h",
        "resolution": "12h",
        "retention": "18d"
    },
    {
        "schema_name": "d_1h",
        "resolution": "1h",
        "retention": "15d"
    }
]
$$::jsonb
);
```

jgpruitt (changes requested):
I strongly advocate for deterministic schema names.
I'd also like to see the simplified logic for updating the downsample config table per my comments

pkg/downsample/downsample.go
jgpruitt: If we named the setting `enable_downsampling` instead of `downsample` and then named the functions `set_enable_downsampling` and `get_enable_downsampling` that would clear things up a little IMO

pkg/downsample/downsample.go
Harkishen-Singh: I am not confident for pushing down the entire config for 3 reasons:
1. The point of connector applying the configuration is that in future, if the `downsample.Config` needs more fields, we can easily change there and connector handles it. With `apply_downsample_config`, I don't see how we can do unless we make changes in the extension code.
2. This can be specific to me. I wrote SQL code so it's easy to read and understand (I am not an expert in SQL). In the current SQL code, I tried to keep everything very straightforward since for me, I find following SQL not as easy as Golang. I am thankful for the new suggestion, but I think `apply_downsample_config` is not a straightforward function. Plus, if there happens to be any changes in `downsample.Config` in future, we cannot simply push the jsonb until we filter some fields, since the `_prom_catalog.downsample` will not have the additional fields in future. I feel this is creating a strong coupling.
3. With `apply_downsample_config()`, the SQL API is no longer simple. It does more than 1 task and I am not sure if its fine.

These are my thoughts. I do not want to oppose it, as I feel I am not the right person in SQL to suggest. So if everyone is happy with `apply_downsample_config` then I can write this function.

alejandrodnm: My 2 cents. I like the approach of the DB function:

1- If we add more fields to the config we would still need to store them in the DB, right?. This is just to handle the logic of applying the whole config. I find this simpler than doing the whole code in the connector and managing the logic there. For me, it feels less error prone and doesn't seem like an out of place extension to the downsample API.

2- 
> I am not an expert in SQL

This is precisely why you should try to write it. You'll have the guidance of the awesome data-platform team. They'll review and help you, and get you a little bit closer to being a SQL expert.

If you don't feel confortable working with the JSONB because of the loose schema, maybe there are alternatives. Could we have the config be a custom type and just pass an array of those, or something, just an idea I'm not a PostgreSQL expert either 😅 

3- You could say that this function does 1 thing, it replaces the current config with the given config.

Harkishen-Singh: Sorry, my point was more towards keeping things simple and straightforward. That will make life easier for all of us. Plus, keeping things in Golang allows us to test easily.

Harkishen-Singh: There are more points than just these. If we make create/update/disable in extension, then I will have to remove `TestMetricDownsampleSync` completely (and write in SQL). Otherwise, it will be creation/updation in extension, and its tests in Go. Then, I will have to write tests in SQL and it will be a very big PR to review. Given we are not sure of this feature's impact ATM, I don't feel confident investing more time and designing more. We should probably go now with speed than accuracy.

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return err
+		}
+		if !got {
+			return fmt.Errorf("timeout: unable to take the advisory lock for syncing downsampling state")
+		}
+	}
+
+	defer func() {
+		if _, err = pgLock.Unlock(); err != nil {
```
niksajakovljevic, line 83: Do we need both `Close` and `Unlock`? I'd assume that `Close` will unlock

niksajakovljevic (approved):

jgpruitt (approved):

PR #1809, opened 2022-12-21
Upgrade to pgx v5

alejandrodnm:
Version 5 of pgx introduced several breaking changes that affected our code base. You can read more about them here:

https://github.com/jackc/pgx/blob/29ad306e47c491a0ecc52d502241539aedd636bd/CHANGELOG.md#codec-and-value-split

In summary:

- NULL Representation

> Previously, types had a Status field that could be Undefined, Null, or
> Present. This has been changed to a Valid bool field to harmonize with
> how database/sql represents NULL and to make the zero value useable.

- Codec and Value Split

> Previously, the type system combined decoding and encoding values with
> the value types...
> This concepts have been separated. A Codec only has responsibility for
> encoding and decoding values. Value types are generally defined by
> implementing an interface that a particular Codec understands (e.g.
> PointScanner and PointValuer for the PostgreSQL point type).

- Array Types

> All array types are now handled by ArrayCodec instead of using code
> generation for each new array type...

- Other Changes

> JSON and JSONB types removed. Use []byte or string directly.

There are some changes that don't seem to appear in the CHANGELOG:

- An error is returned when scanning NULL into a variable that can't handle nil values. For example, scanning NULL into an int64.

- Custom types must be register into the connection's `typeMap *pgtype.Map`, otherwise, an error will be returned when the encoding/decoding plan is not found for the custom type OID.

- Requires go 1.18 since it uses generics for some features, like arrays.

To deal with the custom types, we added the register logic in the `afterConnect` method of the pool. Every time a connection is created a new pgtype.Map is generated with all the default PostgreSQL, meaning that whenever a connection is created, we need to register our custom types with it.

## Merge requirements


alejandrodnm:
I'm going to take a day off this PR and then re-review everything with a clear mind.

pkg/pgmodel/querier/row.go
```diff
@@ -1,148 +1,40 @@
...
 
 var tPool = sync.Pool{
 	New: func() interface{} {
-		return new(pgtype.TimestamptzArray)
+		return new(model.ReusableArray[pgtype.Timestamptz])
 	},
 }
 
-// wrapper to allow DecodeBinary to reuse the existing array so that a pool is effective
-type timestamptzArrayWrapper struct {
```
alejandrodnm, line 28: I moved this to the `custom_types.go` file https://github.com/timescale/promscale/pull/1809/files#diff-d774183a1d65a22e1378babeec3d1e7aed802ebf5f33ed2c5f018d4a3af9a00bR203

niksajakovljevic:
It would be interesting to compare memory profiles while running https://github.com/timescale/promscale-benchmark/ . Like you noted before we expect this upgrade to use less memory due to improvements in PGX.

niksajakovljevic (commented):
Left few comments, but need more time to review. I'll share more during the day.

pkg/dataset/config.go
```diff
@@ -72,12 +72,12 @@ func (c *Config) Apply(conn *pgx.Conn) error {
...
 
 	queries := map[string]interface{}{
-		setDefaultMetricChunkIntervalSQL:    time.Duration(c.Metrics.ChunkInterval),
+		setDefaultMetricChunkIntervalSQL:    time.Duration(c.Metrics.ChunkInterval).String(),
 		setDefaultMetricCompressionSQL:      c.Metrics.Compression,
-		setDefaultMetricHAReleaseRefreshSQL: time.Duration(c.Metrics.HALeaseRefresh),
-		setDefaultMetricHAReleaseTimeoutSQL: time.Duration(c.Metrics.HALeaseTimeout),
-		setDefaultMetricRetentionPeriodSQL:  time.Duration(c.Metrics.RetentionPeriod),
-		setDefaultTraceRetentionPeriodSQL:   time.Duration(c.Traces.RetentionPeriod),
+		setDefaultMetricHAReleaseRefreshSQL: time.Duration(c.Metrics.HALeaseRefresh).String(),
```
niksajakovljevic, line 77: I am a bit surprised with this. How is `go` duration string working with PgSQL interval?

pkg/jaeger/store/get_operations.go
```diff
@@ -34,7 +35,7 @@ WHERE
 
 func getOperations(ctx context.Context, conn pgxconn.PgxConn, query spanstore.OperationQueryParameters) ([]spanstore.Operation, error) {
 	var (
-		pgOperationNames, pgSpanKinds pgtype.TextArray
+		pgOperationNames, pgSpanKinds pgtype.FlatArray[pgtype.Text]
```
niksajakovljevic, line 38: I wonder why we use PG array instead of `[]string`? My understanding is that PG arrays should only be used if we need to access to some of it's internals.

niksajakovljevic: The same question applies to other trace functions where we use PG array and just convert to string slice. I got a bit confused. Maybe I am missing something?

niksajakovljevic: I do understand that array was used prior to this change so it's not directly part of this refactoring, but just wanted to get understanding and maybe fix it if not needed.

pkg/dataset/config.go
jgpruitt: +1

alejandrodnm: It works because this is valid in Postgres

```
promscale@/tmp:promscale> select '8h0m0s'::interval;
+----------+
| interval |
|----------|
| 8:00:00  |
+----------+
```

But I double checked, and just the Lease values needs to be converted to string because the SQL does `$1::text`. The way pgx works is that it has the text OID registered to a Codec, that Codex expects the argument to be a `string`, `[]byte` or a TextValuer interface, since time.Duration is none of those it errors.

I'm updating this part and removing the `.String()` where it's not needed.

jgpruitt: well, that's handy

jgpruitt (approved):
I didn't exhaustively look at every line, but I don't see anything that gives me concern beyond Niksa's questions. Please address his.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/timescale/promscale
 
-go 1.17
+go 1.18
```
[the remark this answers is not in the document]
alejandrodnm: Not really, I think we are using 1.19 for building the image, this is just the minimum version required. I had to bump it because pgx uses generics. Do you want me to bump it to 1.19?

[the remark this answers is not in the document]
niksajakovljevic: If it's no effort I'd rather pick 1.19 :)

pkg/telemetry/telemetry.go
```diff
@@ -171,10 +171,10 @@ func (t *engineImpl) syncDynamicMetadata() error {
 func (t *engineImpl) syncWithMetadataTable(queryFormat string, m Metadata) error {
 	batch := t.conn.NewBatch()
 	for key, metadata := range m {
-		safe := pgutf8str.Text{}
-		if err := safe.Set(metadata); err != nil {
-			return fmt.Errorf("setting in pgutf8 safe string: %w", err)
-		}
+		safe := pgutf8str.Text{Text: pgtype.Text{
```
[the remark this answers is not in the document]
niksajakovljevic: It seems `pgutf8str.Text` only purpose is to sanitize NULL character within a string. However I   telemetry metadata values are controlled by us and will never contain NULL character so I think it's safe to just use `string`. I don't see a reason to use `pgtype.Text` either.

pkg/pgmodel/model/label_list.go
```diff
@@ -7,37 +7,23 @@ package model
...
 
-	"github.com/jackc/pgtype"
+	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/timescale/promscale/pkg/pgmodel/model/pgutf8str"
 )
 
 type LabelList struct {
-	names  *pgutf8str.TextArray
-	values *pgutf8str.TextArray
+	names  pgtype.FlatArray[pgutf8str.Text]
```
niksajakovljevic, line 15: Label names can't have NULL character so it's better to just use `pgtype.Text`. I assume this was not originally part of this change but since we are touching those code lines I think it's good to use opportunity to improve/fix things.

niksajakovljevic: Actually it seems it was `pgtype.Text` before.

pkg/pgmodel/model/custom_types.go
```diff
@@ -7,113 +7,232 @@ package model
...
-		return pgtype.NewArrayType("prom_api.label_value_array", GetCustomTypeOID(t), labelValueArrayTranscoder)
-	default:
-		panic("invalid type")
+// Wrapper to allow DecodeBinary to reuse the existing array so that a pool is
+// effective
+type ReusableArray[T any] struct {
+	pgtype.FlatArray[T]
+}
+
+func (a *ReusableArray[T]) SetDimensions(dimensions []pgtype.ArrayDimension) error {
```
niksajakovljevic, line 207: Since I haven't looked close enough into new PGX implementation is this function supposed to be called by `DecodeBinary` ? Do we need to provide `DecodeBinary` implementation?

niksajakovljevic: Also a bit confused with combining `pgtype.FlatArray` and `pgtype.ArrayDimension` ? Isn't FlatArray designed to ignore PostgreSQL dimensions?

alejandrodnm: All the arrays are decoded/encoded by an ArrayCodec. The ArrayCodec works with 2 interfaces which are [ArraySetter and ArrayGetter](https://github.com/jackc/pgx/blob/master/pgtype/array_codec.go#L14-L38) which both [Array](https://github.com/jackc/pgx/blob/master/pgtype/array.go#L400-L433) and [FlatArray](https://github.com/jackc/pgx/blob/master/pgtype/array.go#L445-L481) implement.

ArrayDimensions tells how big the array needs to be, and it uses in the SetDimensions function to allocate the underlaying Array, the difference between Array and FlatArray is that the array struct keeps a reference to the dimensions object, while FlatArray doesn't.

pkg/jaeger/store/get_operations.go
```diff
@@ -76,10 +78,13 @@ func getOperations(ctx context.Context, conn pgxconn.PgxConn, query spanstore.Op
...
 
-func textArraytoStringArr(s pgtype.TextArray) ([]string, error) {
-	var d []string
-	if err := s.AssignTo(&d); err != nil {
-		return []string{}, fmt.Errorf("assign to: %w", err)
+func textArraytoStringArr(s pgtype.FlatArray[pgtype.Text]) ([]string, error) {
+	d := make([]string, len(s))
+	for i, v := range s {
+		if !v.Valid {
+			return nil, errors.New("can't assign NULL to string")
```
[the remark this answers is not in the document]
alejandrodnm: > I wonder if we should go with string pointers? Or maybe just use empty string for NULL? I feel that erroring is maybe too strick. For example if some array values have NULLs (legacy?) it will actually break the connector.

The behaviour didn't change, the `s.AssignTo` method will return an error if there is a NULL value, because it will try to assign a nil to a string. I modify a test from the deprecated pgtype package to check this:

```go
func TestTextArrayDecodeTextNull(t *testing.T) {
	textArray := &pgtype.TextArray{}
	err := textArray.DecodeText(nil, []byte(`{abc,NULL}`))
	require.NoError(t, err)
	require.Len(t, textArray.Elements, 2)
	assert.Equal(t, pgtype.Present, textArray.Elements[0].Status)
	assert.Equal(t, pgtype.Null, textArray.Elements[1].Status)
	var a []string
	err = textArray.AssignTo(&a)
	require.Errorf(t, err, "cannot assign NULL to *string")
}
```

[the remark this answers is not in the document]
alejandrodnm: > Btw I've also noted that we have `TextArrayToSlice` function.

Yeah, both methods are almost the same. The thing is that `TextArrayToSlice` is basically replacing the [Get method from pgutf8str](https://github.com/timescale/promscale/pull/1809/files#diff-537f9ac4eefdd4668983d9efa6d91797a67abac9d26259193a155d1f94e2e80bL46-L52)  (The code on the left). It does the `revertSanitization` and I just realised that I left a question there, because the method panics, the old one also has a panic, but maybe we can handle it as an error instead.

pkg/telemetry/telemetry.go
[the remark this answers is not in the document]
alejandrodnm: Some telemetry values are taken from the OS/HOST I'd assume those won't return the NULL char.

pkg/jaeger/store/get_operations.go
niksajakovljevic: >IIRC it's due to the way data is serialised/deserialised, PGX expect one of its types because of the interfaces it implements. I'll double check this.

Any updates on this one? Thinking if we can be using `[]string`

PR #1811, opened 2022-12-21
Dictionary encoding for series cache key

niksajakovljevic:
This makes key much smaller and have positive impact on memory usage. Benchmark is showing 2X reduction in memory usage.

Before:
Screenshot 2022-12-21 at 12 13 37

Screenshot 2022-12-21 at 12 08 03

After:
Screenshot 2022-12-21 at 13 18 34

Screenshot 2022-12-21 at 13 17 33

### DIFF
--- a/pkg/pgmodel/cache/series_cache.go
+++ b/pkg/pgmodel/cache/series_cache.go
@@ -6,12 +6,11 @@ package cache
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/binary"
-	"fmt"
-	"math"
 	"sort"
 	"sync"
-	"unsafe"
+	"sync/atomic"
 
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/timescale/promscale/pkg/clockcache"
@@ -26,7 +25,7 @@ const DefaultSeriesCacheSize = 1000000
 // SeriesCache is a cache of model.Series entries.
 type SeriesCache interface {
 	Reset()
-	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)
+	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, err error)
 	Len() int
 	Cap() int
 	Evictions() uint64
@@ -34,13 +33,16 @@ type SeriesCache interface {
 
 type SeriesCacheImpl struct {
 	*ResizableCache
+	labelsDict  sync.Map
+	dictCounter atomic.Uint32
+	dictLock    sync.Mutex
 }
 
 func NewSeriesCache(config Config, sigClose <-chan struct{}) *SeriesCacheImpl {
 	return &SeriesCacheImpl{NewResizableCache(
 		clockcache.WithMetrics("series", "metric", config.SeriesCacheInitialSize),
 		config.SeriesCacheMemoryMaxBytes,
-		sigClose)}
+		sigClose), sync.Map{}, atomic.Uint32{}, sync.Mutex{}}
 }
 
 // Get the canonical version of a series if one exists.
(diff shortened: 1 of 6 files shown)


pkg/pgmodel/cache/series_cache.go
```diff
@@ -164,48 +84,59 @@ func (t *SeriesCacheImpl) GetSeriesFromLabels(ls labels.Labels) (*model.Series,
...
+			pairEnc, _ = s.labelsDict.LoadOrStore(l, s.dictCounter.Add(1))
+			s.dictLock.Unlock()
+		}
+		encodedSize := binary.PutUvarint(idEncodeBuf, uint64(pairEnc.(uint32)))
+		tmp := idEncodeBuf[:encodedSize]
+		//since length can never exceed 4 bytes this is safe
+		seriesKeyBytes.WriteByte(byte(len(tmp)))
+		seriesKeyBytes.Write(tmp)
+	}
+	return base64.StdEncoding.EncodeToString(seriesKeyBytes.Bytes())
```
cevian, line 128: why do we need this base64 encoded. Why not `string(seriesKeyBytes.Bytes())`

pkg/pgmodel/cache/series_cache.go
```diff
@@ -164,48 +84,59 @@ func (t *SeriesCacheImpl) GetSeriesFromLabels(ls labels.Labels) (*model.Series,
...
+// Bytes are then serialized to string using Base64 encoding
+func (s *SeriesCacheImpl) getSeriesKey(labelPairs []prompb.Label) string {
+	seriesKeyBytes := keyPool.Get().(*bytes.Buffer)
+	idEncodeBuf := make([]byte, binary.MaxVarintLen32)
+	ensureLabelsSorted(labelPairs)
+	for _, labelPair := range labelPairs {
+		l := labels.Label{Name: labelPair.Name, Value: labelPair.Value}
+		pairEnc, found := s.labelsDict.Load(l)
+		if !found {
+			s.dictLock.Lock()
```
cevian, line 118: why are we using both a lock AND atomic primitives. Usually you use one or the other

pkg/pgmodel/cache/series_cache.go
```diff
@@ -26,21 +25,24 @@ const DefaultSeriesCacheSize = 1000000
...
-	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, metricName string, err error)
+	GetSeriesFromProtos(labelPairs []prompb.Label) (series *model.Series, err error)
 	Len() int
 	Cap() int
 	Evictions() uint64
 }
 
 type SeriesCacheImpl struct {
 	*ResizableCache
+	labelsDict  sync.Map
```
cevian, line 36: don't we want to refcount the entries here and delete from the map if all series using a dict entry get evicted. As it is I think the dict size grows in an unbounded way

pkg/pgmodel/cache/series_cache.go
```diff
@@ -164,48 +84,59 @@ func (t *SeriesCacheImpl) GetSeriesFromLabels(ls labels.Labels) (*model.Series,
...
+// We use dictionary encoding to generate series key.
+// Each labelPair is encoded to uint32
+// We perform variable encoding on uint32 to byte and append all to byte buffer
+// Bytes are then serialized to string using Base64 encoding
+func (s *SeriesCacheImpl) getSeriesKey(labelPairs []prompb.Label) string {
+	seriesKeyBytes := keyPool.Get().(*bytes.Buffer)
+	idEncodeBuf := make([]byte, binary.MaxVarintLen32)
+	ensureLabelsSorted(labelPairs)
+	for _, labelPair := range labelPairs {
+		l := labels.Label{Name: labelPair.Name, Value: labelPair.Value}
```
arajkumar, line 115: I think it would be much efficient to use prompb.Label as is(may be prompb.Label.Name + prompb.Label.Value) instead of converting to labels.Label which is a map?

pkg/pgmodel/cache/series_cache.go
```diff
@@ -164,48 +84,59 @@ func (t *SeriesCacheImpl) GetSeriesFromLabels(ls labels.Labels) (*model.Series,
...
+	return series, nil
+}
 
-	return series, metricName, nil
+// We use dictionary encoding to generate series key.
+// Each labelPair is encoded to uint32
+// We perform variable encoding on uint32 to byte and append all to byte buffer
+// Bytes are then serialized to string using Base64 encoding
+func (s *SeriesCacheImpl) getSeriesKey(labelPairs []prompb.Label) string {
+	seriesKeyBytes := keyPool.Get().(*bytes.Buffer)
```
arajkumar, line 111: Looks like seriesKeyBytes is not put back into the keyPool.

pkg/pgmodel/model/series.go
```diff
@@ -41,14 +41,12 @@ type Series struct {
...
 
 	metricName string
-	str        string
 }
 
-func NewSeries(key string, labelPairs []prompb.Label) *Series {
+func NewSeries(labelPairs []prompb.Label) *Series {
 	series := &Series{
 		names:    make([]string, len(labelPairs)),
 		values:   make([]string, len(labelPairs)),
```
arajkumar, line 49: May be adding separate dictionary for keys & values would reduce size further and the same encoding could be used here in names and values instead of string[]?

niksajakovljevic:
Thanks for the feedback however no need to review yet since this is just an initial draft to validate the approach.

niksajakovljevic:
Project got deprecated

PR #1812, opened 2022-12-22
Track metrics per workload in vacuum engine

jgpruitt:
## Description

Adjusts vacuum engine metrics to add a `workload` label to keep separate metrics per workload.

## Merge requirements

### DIFF
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -69,7 +69,7 @@ var (
 	vacuumErrorsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: util.PromNamespace,
 		Subsystem: "vacuum",
-		Name:      "vacuum_errors_total",
+		Name:      "errors_total",
 		Help:      "Total number of errors encountered by the Promscale vacuum engine while vacuuming tables.",
 	},
 		[]string{"workload"},
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -63,7 +63,7 @@ var (
 	numberVacuumConnections = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: util.PromNamespace,
 		Subsystem: "vacuum",
-		Name:      "number_vacuum_connections",
+		Name:      "num_connections",
 		Help:      "Number of database connections currently in use by the vacuum engine. One taken up by the advisory lock, the rest by vacuum commands.",
 	})
 	vacuumErrorsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
--- a/pkg/vacuum/vacuum.go
+++ b/pkg/vacuum/vacuum.go
@@ -60,40 +60,48 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/vacuum/vacuum.go
```diff
@@ -60,40 +60,48 @@ import (
 )
 
 var (
-	tablesVacuumedTotal = prometheus.NewCounter(prometheus.CounterOpts{
+	numberVacuumConnections = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: util.PromNamespace,
 		Subsystem: "vacuum",
-		Name:      "tables_vacuumed_total",
-		Help:      "Total number of tables vacuumed by the Promscale vacuum engine.",
+		Name:      "number_vacuum_connections",
```
Harkishen-Singh, line 66: We are repeating `vacuum` used in subsystem.
```suggestion
		Name:      "num_connections",
```

pkg/vacuum/vacuum.go
```diff
@@ -60,40 +60,48 @@ import (
...
+		Help:      "Number of database connections currently in use by the vacuum engine. One taken up by the advisory lock, the rest by vacuum commands.",
 	})
-	vacuumErrorsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+	vacuumErrorsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: util.PromNamespace,
 		Subsystem: "vacuum",
 		Name:      "vacuum_errors_total",
```
Harkishen-Singh, line 72:
```suggestion
		Name:      "errors_total",
```

pkg/vacuum/vacuum.go
```diff
@@ -338,10 +353,11 @@ func (e *Engine) Run(ctx context.Context) {
 		chunks, err := e.listChunks(ctx, con, w.query)
 		if err != nil {
 			log.Error("msg", fmt.Sprintf("failed to list %s", w.name), "error", err)
+			vacuumErrorsTotal.WithLabelValues(w.workload).Inc()
 			w.stop = true
 			continue
 		}
-		tablesNeedingVacuum.Set(float64(len(chunks)))
+		tablesToVacuum.WithLabelValues(w.workload).Set(float64(len(chunks)))
```
Harkishen-Singh, line 360: If I am understanding this correctly, it seems like the only time the label value other than `locking` is `test` in the unit test. Maybe we declare `locking` out of the functional value and ue that constant everywhere?

Harkishen-Singh (changes requested):

pkg/vacuum/vacuum.go
jgpruitt: Not sure I follow. `w.workload` will either be `compressed-chunks-to-freeze` or `compressed-chunks-missing-stats` depending on whether `w` is set to `chunksToFreeze` or `chunksMissingStats`

jgpruitt: So the label will either be `locking`, `test`, `compressed-chunks-to-freeze`, or `compressed-chunks-missing-stats`

niksajakovljevic (approved):

Harkishen-Singh (approved):

PR #1814, opened 2022-12-23
Custom COPY with only one DB roundtrip

alejandrodnm:
### Add custom CopyFrom command that avoids an extra roundtrip 
 
 The CopyFrom command from pgx requires the OIDs of the data types is
going to insert when constructing the binary message that contains
the data.

To retrieve the OIDs a statement is prepared per COPY operation. In our
case the statements are very simmilar:

```
copy {tableName} ( `time`, `value`, `series_id` ) from stdin binary
```

These statements can be cached, the problem is that we will have a cache
entry per table name.

Since we always insert to the same columns and we know the OIDs of the
columns we are going to be inserting to, we can avoid the extra
roundtrip and the cache by manually passing the OIDs.

This custom copy command is basically a copy of the pgx CopyCommand that
accepts the list of OIDs.

## Merge requirements

### DIFF
--- a/pkg/pgxconn/copy_from.go
+++ b/pkg/pgxconn/copy_from.go
@@ -1,0 +1,234 @@
+// This file was copied from the PGX repository and modified. PGX is
+// licensed under MIT.
+//
+// https://github.com/jackc/pgx/blob/master/copy_from.go
+//
+// The MIT License
+//
+// Copyright (c) 2011 Dominic Tarr
+//
+// Permission is hereby granted, free of charge,
+// to any person obtaining a copy of this software and
+// associated documentation files (the "Software"), to
+// deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify,
+// merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom
+// the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice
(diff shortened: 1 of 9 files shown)


pkg/pgxconn/copy_from.go
```diff
@@ -0,0 +1,208 @@
+// This file and its contents are licensed under the Apache License 2.0.
```
niksajakovljevic, line 1: If we borrowed a code from PGX we should include their licenses here as well.

niksajakovljevic: Also worth adding a pointer/url to the original code

pkg/pgxconn/copy_from.go
```diff
@@ -0,0 +1,208 @@
...
+	"sync"
+
+	"github.com/jackc/pgio"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+var bufPool = sync.Pool{
+	New: func() any {
+		s := make([]byte, 0, 1024)
```
niksajakovljevic, line 24: Nit: Maybe increase this to 4096? One row is 8+8+4=20 bytes. IIRC one COPY batch is usually around 100-200 items.

alejandrodnm: I used the same value as PGX, but I'll increase it.

niksajakovljevic (approved):

PR #1819, opened 2023-01-09
Add support for PG15

alejandrodnm:
## Merge requirements

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -44,7 +44,7 @@ var (
 	Branch                     = ""      // Comes from -ldflags settings
 	EarliestUpgradeTestVersion = "0.3.0" // 0.3.0 earliest version an image with correct extension versions exists
 
-	PgVersionNumRange       = ">=12.x <15.x" // Corresponds to range within pg 12.0 to pg 14.99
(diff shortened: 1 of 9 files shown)


pkg/version/version.go
```diff
@@ -44,7 +44,7 @@ var (
 	Branch                     = ""      // Comes from -ldflags settings
 	EarliestUpgradeTestVersion = "0.3.0" // 0.3.0 earliest version an image with correct extension versions exists
 
-	PgVersionNumRange       = ">=12.x <15.x" // Corresponds to range within pg 12.0 to pg 14.99
+	PgVersionNumRange       = ">=12.x <16.x" // Corresponds to range within pg 12.0 to pg 14.99
```
niksajakovljevic, line 47: Nit: tweak comment as well

niksajakovljevic (approved):
Lgtm. Maybe worth changing docker image for docker compose to use PG15 image built by extension (this is to be extra safe but maybe not needed)

pkg/version/version.go
alejandrodnm: I had auto merge, I'll fix it in the release PR

PR #1826, opened 2023-01-09
Prepare for 0.17.0 release

alejandrodnm:
## Merge requirements


alejandrodnm:
> Shouldn't we tweak required extension version to be >0.8.0 ? I think we added some vacuuming stuff that requires 0.8.0 at least

That was already done here https://github.com/timescale/promscale/pull/1817/files

